PR #10, opened 2018-11-02
add a comment about C* config

yito88:
Scalar DB assumes commitlog sync `batch`


docs/getting-started.md
```diff
@@ -10,6 +10,14 @@ Scalar DB v1 is written in Java and uses Cassandra as an underlining storage imp
 * [Oracle JDK 8](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) (OpenJDK 8) or higher
 * [Casssandra](http://cassandra.apache.org/) 3.11.x (the current stable version as of writing)
     * Take a look at [this document](http://cassandra.apache.org/download/) for how to set up Cassandra.
+    * Change `commitlog_sync` from `periodic` to `batch` in `cassandra.yaml`
```
feeblefakie: Can you update it to this ? (I just think it needs reasoning)

 * Change `commitlog_sync` from `periodic` to `batch` in `cassandra.yaml` not to lose data when quorum of replica nodes go down

PR #13, opened 2018-11-13
add clock-drift tests

yito88:
Add Jepsen clock-drift tests for Scalar DB

### DIFF
--- a/jepsen/scalardb/test/scalardb/scalardb_test.clj
+++ b/jepsen/scalardb/test/scalardb/scalardb_test.clj
@@ -8,10 +8,12 @@
             [scalardb.transfer :as transfer]
             [jepsen [core :as jepsen]]))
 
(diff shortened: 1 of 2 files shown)


jepsen/scalardb/test/scalardb/scalardb_test.clj
```diff
@@ -8,34 +8,53 @@
             [scalardb.transfer :as transfer]
             [jepsen [core :as jepsen]]))
 
+(def nodes ["n1" "n2" "n3" "n4" "n5"])
```
feeblefakie: Does it only work with 5-node setting ? (or this clock-drift only work with 5-node setting ?)
Is there any way to make it work with arbitrary number of nodes ?

feeblefakie: Got it. Thank you !

PR #14, opened 2018-11-16
add integrations tests to the verification group

siyopao:


siyopao:
@feeblefakie `group` is used to group together like tasks. For example, there is a `build` group which contains, among others, the `assemble`, `build`, `clean` tasks. All the test tasks seem to be grouped in the `verification` group, so this is just adding `integrationTest` to that group.

screen shot 2018-11-26 at 16 41 16 2

feeblefakie:
@siyopao oh, ok, that's nice !

PR #27, opened 2019-02-20
If readRecordsWithRetry fails exit properly

siyopao:
Want to return an error code so the calling script knows when to stop.

### DIFF
--- a/src/test/java/com/scalar/database/verification/GetTotal.java
+++ b/src/test/java/com/scalar/database/verification/GetTotal.java
@@ -33,7 +33,14 @@ public class GetTotal {
     TransferContext context = new TransferContext(numAccounts, 0, 0, 0);
     AccountBalanceTransferHandler handler = new AccountBalanceTransferHandler(config, context);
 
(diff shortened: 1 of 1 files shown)


src/test/java/com/scalar/database/verification/GetTotal.java
```diff
@@ -33,7 +33,14 @@ public static void main(String[] args) throws IOException {
     TransferContext context = new TransferContext(numAccounts, 0, 0, 0);
     AccountBalanceTransferHandler handler = new AccountBalanceTransferHandler(config, context);
 
-    List<Result> results = handler.readRecordsWithRetry();
+    List<Result> results = null;
+    try {
+      results = handler.readRecordsWithRetry();
```
feeblefakie: When the exception is thrown and propagated to the main, doesn't this script return 1 ?

siyopao: No, the script just hangs... That does seem pretty weird doesn't it.

Issue #25, opened 2019-02-14
transaction.commit() unexpectedly succeeded without clustering keys


yito88:
A record was committed unexpectedly by the `Put` has a clustering key as a value.
This commit should fail because the Cassandra driver throws `InvalidQueryException`.

This issue happens when a record is an initial record for the primary key.
The put for insertion isn't checked if the `Put` has the correct keys in the driver.
When the insertion succeeds, the preparation phase of the transaction also succeeds.

The commit of the transaction for the record fails because of the lack of clustering keys.
But it is eventual committed because this commit phase is after the state updating.

feeblefakie:
It should be fixed with https://github.com/scalar-labs/scalardb/commit/46e3562223409688307c8693344b4b5bdab63e83 .
Please reopen the issue if you still have the issue.

PR #28, opened 2019-02-20
fix the reported bug and add some unit tests

feeblefakie:
https://github.com/scalar-labs/scalardb/issues/25

This is caused by unchecked primary key in put in storage, so this PR fixes that to fix the above issue.

### DIFF
--- a/src/main/java/com/scalar/database/storage/cassandra/TableMetadata.java
+++ b/src/main/java/com/scalar/database/storage/cassandra/TableMetadata.java
@@ -8,18 +8,10 @@ import javax.annotation.concurrent.ThreadSafe;
 
 @ThreadSafe
 public class TableMetadata {
-  private final ClusterManager manager;
-  private final String keyspace;
-  private final String tableName;
-  private final com.datastax.driver.core.TableMetadata tableMetadata;
   private final Set<String> partitionKeyNames;
   private final Set<String> clusteringColumnNames;
 
-  public TableMetadata(ClusterManager manager, String keyspace, String tableName) {
-    this.manager = manager;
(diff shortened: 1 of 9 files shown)


src/main/java/com/scalar/database/storage/cassandra/TableMetadata.java
```diff
@@ -0,0 +1,46 @@
...
+public class TableMetadata {
+  private final ClusterManager manager;
+  private final String keyspace;
+  private final String tableName;
+  private final com.datastax.driver.core.TableMetadata tableMetadata;
+  private final Set<String> partitionKeyNames;
+  private final Set<String> clusteringColumnNames;
+
+  public TableMetadata(ClusterManager manager, String keyspace, String tableName) {
+    this.manager = manager;
```
scalarindetail: It might be not necessary to use properties to preserve manager, keyspace and tableName individually since they seem only used to get Cassandra's table metadata.

feeblefakie: Good catch ! Right. Probably Passing `com.datastax.driver.core.TableMetadata` only to the constructor might be even better.
Let me work on that.

feeblefakie: @scalarindetail I fixed as below. Please take an another look !
https://github.com/scalar-labs/scalardb/pull/28/commits/42b0151d6e656efc838b15df5f1dd45150deee88

scalarindetail: LGTM

PR #29, opened 2019-03-07
a client reconnects to the cluster when many failures

yito88:
Jepsen crash tests often failed because a client couldn't get the final balances due to disconnection to the cluster.

I've added reconnection when many failures happen.
- A client has DistributedStorageService and DistributedTransactionService
- After many failures, a client tries to reconnect to the cluster
- Close all connections at the end

### DIFF
--- a/jepsen/scalardb/project.clj
+++ b/jepsen/scalardb/project.clj
@@ -7,6 +7,6 @@
                  [jepsen "0.1.10-SNAPSHOT"]
                  [cassandra "0.1.0-SNAPSHOT"]
                  [clojurewerkz/cassaforte "3.0.0-alpha2-SNAPSHOT"]
-                 [com.scalar-labs/scalardb "1.0.0-rc1" :exclusions [org.slf4j/slf4j-log4j12]]]
+                 [com.scalar-labs/scalardb "1.0.0" :exclusions [org.slf4j/slf4j-log4j12]]]
   :main scalardb.runner
   :aot [scalardb.runner])
--- a/jepsen/scalardb/src/scalardb/runner.clj
+++ b/jepsen/scalardb/src/scalardb/runner.clj
@@ -33,7 +33,7 @@
     :parse-fn #(Long/parseLong %)
     :validate [pos? "Must be positive"]]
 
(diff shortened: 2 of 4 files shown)


jepsen/scalardb/src/scalardb/runner.clj
```diff
@@ -33,7 +33,7 @@
     :parse-fn #(Long/parseLong %)
     :validate [pos? "Must be positive"]]
 
-   (jc/tarball-opt "http://www.us.apache.org/dist/cassandra/3.11.3/apache-cassandra-3.11.3-bin.tar.gz")])
+   (jc/tarball-opt "http://www.us.apache.org/dist/cassandra/3.11.4/apache-cassandra-3.11.4-bin.tar.gz")])
```
siyopao: It may be better to use `https://archive.apache.org/dist/cassandra/3.11.4/apache-cassandra-3.11.4-bin.tar.gz` which won't break if/when Cassandra is updated to 3.11.5.

jepsen/scalardb/project.clj
```diff
@@ -7,6 +7,6 @@
                  [jepsen "0.1.10-SNAPSHOT"]
                  [cassandra "0.1.0-SNAPSHOT"]
                  [clojurewerkz/cassaforte "3.0.0-alpha2-SNAPSHOT"]
-                 [com.scalar-labs/scalardb "1.0.0-rc1" :exclusions [org.slf4j/slf4j-log4j12]]]
+                 [com.scalar-labs/scalardb "1.0.0" :exclusions [org.slf4j/slf4j-log4j12]]]
```
feeblefakie: Regarding this kind of version numbers (also for C*), can they be defined outside of the programs instead of hard-coding them in the code ?

yito88: `project.clj` is like `build.gradle`.
We can change only this file to define dependent libraries.

feeblefakie: Got it ! Thanks

yito88:
@siyopao Sorry, I mistakenly dismissed your review state. Could you review it again?

PR #30, opened 2019-03-22
Fix jepsen reconnection

yito88:
- Modified reconnection to the cluster
- Skip an `op` when there is no `DistributedTransactionService`
- Modified checking the coordinator table

### DIFF
--- a/jepsen/scalardb/src/scalardb/core.clj
+++ b/jepsen/scalardb/src/scalardb/core.clj
@@ -27,7 +27,7 @@
 
 (defn exponential-backoff
   [r]
-  (Thread/sleep (* 1000 (reduce * (repeat r 2)))))
+  (Thread/sleep (reduce * 1000 (repeat r 2))))
 
 (defn create-coordinator-table!
   [session test]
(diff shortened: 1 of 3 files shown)


jepsen/scalardb/src/scalardb/core.clj
```diff
@@ -19,13 +18,17 @@
...
+(def ^:const RETRIES 8)
 (def ^:const RETRIES_FOR_RECONNECTION 3)
 
 (def ^:const COORDINATOR "coordinator")
 (def ^:const STATE_TABLE "state")
 (def ^:const VERSION "tx_version")
 
+(defn exponential-backoff
+  [r]
+  (Thread/sleep (* 1000 (reduce * (repeat r 2)))))
```
siyopao: Not that it makes any difference, but `reduce` can also take an initial value, so could be rewritten here as

```
(reduce * 1000 (repeat r 2))
```

yito88: Yes, it is better. Thanks.

PR #31, opened 2019-03-29
add javadocs

siyopao:
https://scalar-labs.atlassian.net/browse/DLT-2228

### DIFF
--- a/src/main/java/com/scalar/database/api/DistributedStorage.java
+++ b/src/main/java/com/scalar/database/api/DistributedStorage.java
@@ -6,12 +6,12 @@ import java.util.Optional;
 
 /**
  * A storage abstraction for interacting with distributed storage implementations. The user can
- * execute CRUD operations to the storage by passing {@link Operation} commands such as {@link Get},
+ * execute CRUD operations on the storage by passing {@link Operation} commands such as {@link Get},
  * {@link Scan}, {@link Put} and {@link Delete}, which specify the location of storage entries and
  * optional information.
  *
- * <p>The data model behind this abstraction is a multi-dimensional map based on key-value data
- * model. A logical record (or entry) is composed of partition key, clustering key and a set of
+ * <p>The data model behind this abstraction is a multi-dimensional map based on the key-value data
+ * model. A logical record (or entry) is composed of a partition key, a clustering key and a set of
  * values. The value is uniquely mapped by a primary key composed of partition key, clustering key
  * and value name as described in the following scheme.
  *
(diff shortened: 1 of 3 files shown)


src/main/java/com/scalar/database/api/TransactionState.java
```diff
@@ -1,6 +1,6 @@
 package com.scalar.database.api;
 
-/** */
+/** The state of a transaction */
```
yito88: I think it is better to describe in more detail because this state is used for two meanings, state of a transaction and state of a record.

siyopao: Thank you for your comment. I made a very simple change to "Flag to represent the state of a record in a transaction." I am not sure if it is worth adding anything on how the flags are used as this would be very large I think. I simply removed mention of "state of a transaction" as as far as I can tell it only represents the state of records.

src/main/java/com/scalar/database/api/TransactionState.java
```diff
@@ -1,6 +1,6 @@
 package com.scalar.database.api;
 
-/** */
+/** Flag to represent the state of a record in a transaction. */
```
siyopao:
```suggestion
/** Flag used to represent either the state of a record in a transaction or a transaction itself. */
```

src/main/java/com/scalar/database/api/TransactionState.java
yito88: I often check this state if a record is consistent as expected.
I think a user wants to know what these states mean when the user checks the coordinator table or user tables.
How about adding descriptions for each state like [Consistency](https://github.com/scalar-labs/scalardb/blob/a4e223766cd703fabf0c95d82b547988d0760550/src/main/java/com/scalar/database/api/Consistency.java#L8)?

For example;

PREPARED state is for only a record. PREPARED state of a record shows that the record isn't committed because the corresponding transaction is in progress or has failed in the commit phase.

src/main/java/com/scalar/database/api/DistributedStorage.java
```diff
@@ -21,24 +21,26 @@
...
+ * props.setProperty(DatabaseConfig.USERNAME, USERNAME);
+ * props.setProperty(DatabaseConfig.PASSWORD, PASSWORD);
  *
  * // In case of Cassandra storage implementation.
- * DistributedStorage storage = new Cassandra(config, credential);
+ * DistributedStorage storage = new Cassandra(new DatabaseConfig(props));
  *
- * // Uses NAMESPACE and TABLE by default in this storage instance.
- * storage.with(NAMESPACE, TABLE);
+ * // Uses KEYSPACE and TABLE by default in this storage instance.
```
feeblefakie: Scalar DB uses `namespace` for generalization of database/keyspace/... .

src/main/java/com/scalar/database/api/TransactionState.java
```diff
@@ -1,6 +1,6 @@
 package com.scalar.database.api;
 
-/** */
+/** Flag used to represent either the state of a record in a transaction or a transaction itself. */
```
feeblefakie: What is the relationship between the state of a record and transaction state ?
What is says is not false but I think it is not very informative and probably misleading.

PR #34, opened 2019-03-29
added javadocs

siyopao:
https://scalar-labs.atlassian.net/browse/DLT-205

This will be 1/3 for `transaction` probably.

### DIFF
--- a/src/main/java/com/scalar/database/transaction/consensuscommit/CommitHandler.java
+++ b/src/main/java/com/scalar/database/transaction/consensuscommit/CommitHandler.java
@@ -42,7 +42,7 @@ public class CommitHandler {
   }
 
   /**
-   * Commit the specified {@link Snapshot} to {@link DistributedStorage}
+   * Commits the specified {@link Snapshot} to {@link DistributedStorage}
    *
    * @param snapshot a {@link Snapshot} to be committed
    * @throws CommitException
--- a/src/main/java/com/scalar/database/transaction/consensuscommit/CommitHandler.java
+++ b/src/main/java/com/scalar/database/transaction/consensuscommit/CommitHandler.java
@@ -18,7 +18,7 @@ import javax.annotation.concurrent.ThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** */
+/** Commit snapshots to {@link DistributedStorage} */
 @ThreadSafe
 public class CommitHandler {
   private static final Logger LOGGER = LoggerFactory.getLogger(CommitHandler.class);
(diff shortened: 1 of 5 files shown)


src/main/java/com/scalar/database/transaction/consensuscommit/CommitHandler.java
```diff
@@ -18,21 +18,36 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** */
+/** Commit snapshots to {@link DistributedStorage} */
```
yito88: I think this class handles the commit process of a transaction, not just store updated values to the storage.

src/main/java/com/scalar/database/transaction/consensuscommit/CommitMutationComposer.java
```diff
@@ -19,10 +19,19 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Keep track of mutations that need to be committed to {@link
```
yito88: I think it is confusing. These mutations change the state of records to commit state.

siyopao: That is confusing! Thank you. Fixed.

src/main/java/com/scalar/database/transaction/consensuscommit/CommitHandler.java
siyopao: Thank you! Fixed.

src/main/java/com/scalar/database/transaction/consensuscommit/CommitHandler.java
```diff
@@ -18,21 +18,41 @@
...
     this.storage = checkNotNull(storage);
     this.coordinator = checkNotNull(coordinator);
     this.recovery = checkNotNull(recovery);
   }
 
+  /**
+   * Implements the commit phase of a transaction. That is, all the records that have been changed
+   * by the transaction (dirty objects) will be written to persistent storage.
+   *
+   * <p>The commit phase is performed in two stages: the prepare stage and the commit stage. If the
```
feeblefakie: No offence at all, but I think this explanation is too vague and seems not really correct.
What is the definition of prepare stage and commit stage here ? And what is the `phase` ?
The code has some methods called `prepareRecords`, `commitState` and `commitRecords` and what belongs to which stage/phase ?
In other words, what do you exactly mean by `If the commit stage finishes successfully` ?

src/main/java/com/scalar/database/transaction/consensuscommit/CommitHandler.java
```diff
@@ -18,21 +18,41 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** */
+/** Handler for the commit phase of a transaction. */
```
feeblefakie: I think here is the place to explain how a transaction work in scalar db. Please carefully read the paper, understand it and come back.

PR #35, opened 2019-03-29
Javadocs transaction2

siyopao:
https://scalar-labs.atlassian.net/browse/DLT-205

2/3 for transaction.

### DIFF
--- a/src/main/java/com/scalar/database/transaction/consensuscommit/CrudHandler.java
+++ b/src/main/java/com/scalar/database/transaction/consensuscommit/CrudHandler.java
@@ -103,7 +103,7 @@ public class CrudHandler {
   }
 
   /**
-   * Perform a {@link Put} operation on the {@link Snapshot}
+   * Performs a {@link Put} operation on the {@link Snapshot}
    *
    * @param put the {@code Put} to be performed
    */
@@ -112,7 +112,7 @@ public class CrudHandler {
   }
 
   /**
-   * Perform a {@link Delete} operation on the {@link Snapshot}
+   * Performs a {@link Delete} operation on the {@link Snapshot}
    *
    * @param delete the {@code Delete} to be performed
    */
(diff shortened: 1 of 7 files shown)


src/main/java/com/scalar/database/transaction/consensuscommit/CrudHandler.java
```diff
@@ -22,7 +22,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** */
+/** Perform CRUD operations on {@link DistributedStorage} */
```
yito88: It looks some operations don't operate on DistributedStorage.

src/main/java/com/scalar/database/transaction/consensuscommit/PrepareMutationComposer.java
```diff
@@ -21,9 +21,11 @@
 import java.util.Optional;
 import javax.annotation.concurrent.NotThreadSafe;
 
+/** Prepare mutations in order to commit to {@link com.scalar.database.api.DistributedStorage}. */
```
yito88: It is not wrong. I think it is confusing a little.
`PrepareMutation` in this class means the mutation mutates values, metadata and changes the state of record to the `PREPARED` state.

How about this?
```
Compose mutations to prepare records for a transaction
```

src/main/java/com/scalar/database/transaction/consensuscommit/CrudHandler.java
feeblefakie: I think it's misleading and not very informative. It should explain how it operates with snapshot and the storage.
Maybe, understanding snapshot isolation properly before writing a doc is good. 
Also, it's a comment for a class, please start with a noun.

src/main/java/com/scalar/database/transaction/consensuscommit/PrepareMutationComposer.java
siyopao: Thank you! I used your suggestion.

src/main/java/com/scalar/database/transaction/consensuscommit/CrudHandler.java
siyopao: Thank you for the comments. I updated based on your feedback, so please check it when you have time.

PR #42, opened 2019-05-14
Update to the latest version of Jepsen

siyopao:
Sorry, these files were updated in the directory where they were being run, but I seemed to have not updated them when copying to this repository.

In any case, I think this fixes the problem.

### DIFF
--- a/jepsen/scalardb/src/scalardb/transfer.clj
+++ b/jepsen/scalardb/src/scalardb/transfer.clj
@@ -246,7 +246,7 @@
 (defn consistency-checker
   []
   (reify checker/Checker
-    (check [this test model history opts]
+    (check [this test history opts]
       (let [read-result (->> history
                              (r/filter op/ok?)
                              (r/filter #(= :get-all (:f %)))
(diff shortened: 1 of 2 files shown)


jepsen/scalardb/src/scalardb/transfer.clj
```diff
@@ -246,7 +246,7 @@
 (defn consistency-checker
   []
   (reify checker/Checker
-    (check [this test model history opts]
+    (check [this test history opts]
```
feeblefakie: Is this supposed to be deleted ? If so, can you clarify why ?

siyopao: This `reify` thing is implementing the `Checker` interface, and in the latest Jepsen the interface has been changed to what you see here. I think it is because not all checkers need a `model`, and if one does, it can be constructed with the checker.

feeblefakie: Got it. Thanks !

PR #45, opened 2019-05-26
add deref

yito88:
To add `deref` for `:transaction` of `test` before `read-all`.

The final reading all records failed due to NullPointerException in Jepsen testing.
When no connection, the client should reconnect to the cluster before the read.
But the checking has a bug. The client tried to read records without a connection.


yito88:
@siyopao Good catch! Thanks!
I want to do refactoring for these common functions in another PR.

PR #46, opened 2019-07-03
Delete the Jepsen tests

siyopao:
https://scalar-labs.atlassian.net/browse/DLT-2532

Also removed the alias to the Jepsen folder from `mkdocs.yml` and updated the Jepsen link to point to the `scalar-jepsen` repo.


mkdocs.yml
```diff
@@ -7,7 +7,6 @@ nav:
     - Database schema: schema.md
     - Schema loader: schema-loader.md
 - Design: design.md
-- Jepsen: jepsen.md
```
feeblefakie: Let's create a new jepsen.md to point to the new repo.

siyopao: Thank you. Added.

mkdocs.yml
```diff
@@ -7,6 +7,7 @@ nav:
     - Database schema: schema.md
     - Schema loader: schema-loader.md
 - Design: design.md
+- Jepsen: https://github.com/scalar-labs/scalar-jepsen
```
feeblefakie: Does it work ?

siyopao: I don't know how to check without it being merged. In any case, according to https://www.mkdocs.org/user-guide/configuration/#nav it certainly should.

feeblefakie: It should work but can you check it with `mkdocs serve` quickly ?
mkdocs can be installed through brew.

siyopao: Works fine with mkdocs, version 1.0.4. It did not, however, with some older version of mkdocs.

feeblefakie: Hmm, ok. not sure about what version of mkdocs readthedocs uses. We'll see...

feeblefakie:
Merged and checked. It is working but it directs you to github so it might not a very good UX...

siyopao:
Should it link directly to the readme?

feeblefakie:
It shows the README so I think it's fine.

PR #47, opened 2019-07-11
start with an user-provided transaction ID

feeblefakie:
This is a scalardb part of [the story](https://scalar-labs.atlassian.net/browse/DLT-1691).

Usually we should use `start()` or `start(isolation)`, but accepting user-provided transaction id if user can be responsible for uniqueness of the transaction id.

### DIFF
--- a/src/main/java/com/scalar/database/api/DistributedTransactionManager.java
+++ b/src/main/java/com/scalar/database/api/DistributedTransactionManager.java
@@ -12,7 +12,7 @@ public interface DistributedTransactionManager {
   DistributedTransaction start();
 
   /**
-   * Starts a new transaction with the specified transaction ID. It is a users' responsibility to
+   * Starts a new transaction with the specified transaction ID. It is users' responsibility to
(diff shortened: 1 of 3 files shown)


src/main/java/com/scalar/database/api/DistributedTransactionManager.java
```diff
@@ -4,10 +4,42 @@
...
 
+  /**
+   * Starts a new transaction.
+   *
+   * @return {@link DistributedTransaction}
+   */
   DistributedTransaction start();
 
+  /**
+   * Starts a new transaction with the specified transaction ID. It is a users' responsibility to
```
siyopao: Regarding "a users' responsibility". Probably it should be "It is the user's responsibility..." or "It is users' responsibility ...". Similarly on lines 33-34.

feeblefakie: Cool. Thanks ! Will change it.

feeblefakie: Fixed

PR #53, opened 2019-11-28
[Driver4] Add ClusterManager

yito88:
For supporting Cassandra Java 4.x

- `Cluster` does not exist anymore
- Make `CqlSessionBuilder`, then build `CqlSession`

### DIFF
--- a/src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
+++ b/src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
@@ -91,7 +91,6 @@ public class ClusterManager {
     config
         .getContactPoints()
         .forEach(contactPoint -> contactPoints.add(new InetSocketAddress(contactPoint, port)));
-    System.out.println(contactPoints);
 
     CqlSessionBuilder builder =
         CqlSession.builder()
--- a/src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
+++ b/src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
@@ -1,0 +1,102 @@
+package com.scalar.database.storage.cassandra4driver;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.CqlSessionBuilder;
+import com.datastax.oss.driver.api.core.metadata.schema.TableMetadata;
+import com.google.common.annotations.VisibleForTesting;
+import com.scalar.database.config.DatabaseConfig;
+import com.scalar.database.exception.storage.ConnectionException;
+import com.scalar.database.exception.storage.StorageRuntimeException;
+import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A cluster manager
+ *
+ * @author Hiroyuki Yamada
+ */
+@ThreadSafe
+public class ClusterManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ClusterManager.class);
+  private static final int DEFAULT_CASSANDRA_PORT = 9042;
+  private final DatabaseConfig config;
+  private CqlSessionBuilder builder;
+  private CqlSession session;
+
+  /**
+   * Constructs a {@code ClusterManager} with the specified {@code Config}
+   *
+   * @param config database configuration needed to create a cluster
+   */
+  public ClusterManager(DatabaseConfig config) {
+    this.config = checkNotNull(config);
+  }
+
+  /**
+   * Returns a session. It will create a session if it is not created yet.
+   *
+   * @return a {@code Session}
+   */
+  public synchronized CqlSession getSession() {
+    if (session != null) {
+      return session;
+    }
+    try {
+      if (builder == null) {
+        builder = getBuilder();
+      }
+      // TODO: set Credentials
+      session = builder.build();
+      LOGGER.info("session to the cluster is created.");
+      return session;
+    } catch (RuntimeException e) {
+      LOGGER.error("connecting the cluster failed.", e);
+      throw new ConnectionException("connecting the cluster failed.", e);
+    }
+  }
+
+  /**
+   * Returns a {@link TableMetadata} with the specified keyspace and table name
+   *
+   * @param keyspace keyspace name
+   * @param table table name
+   * @return {@code TableMetadata}
+   */
+  @Nonnull
+  public TableMetadata getMetadata(String keyspace, String table) {
+    return session
+        .getMetadata()
+        .getKeyspace(keyspace)
+        .orElseThrow(() -> new StorageRuntimeException("no keyspace information found"))
+        .getTable(table)
+        .orElseThrow(() -> new StorageRuntimeException("no table information found"));
+  }
+
+  /** Closes the session. */
+  public void close() {
+    session.close();
+    session = null;
+  }
+
+  @VisibleForTesting
+  CqlSessionBuilder getBuilder() {
+    List<InetSocketAddress> contactPoints = new ArrayList<InetSocketAddress>();
+    int port = config.getContactPort() == 0 ? DEFAULT_CASSANDRA_PORT : config.getContactPort();
+    config
+        .getContactPoints()
+        .forEach(contactPoint -> contactPoints.add(new InetSocketAddress(contactPoint, port)));
+    System.out.println(contactPoints);
+
+    // TODO: set LoadBalancePolicy and RetryPolicy
+    return CqlSession.builder()
+        .addContactPoints(contactPoints)
+        .withLocalDatacenter("datacenter1"); // for SimpleSnitch
+  }
+}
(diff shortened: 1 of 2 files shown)


src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
```diff
@@ -0,0 +1,106 @@
...
+    int port = config.getContactPort() == 0 ? DEFAULT_CASSANDRA_PORT : config.getContactPort();
+    config
+        .getContactPoints()
+        .forEach(contactPoint -> contactPoints.add(new InetSocketAddress(contactPoint, port)));
+
+    CqlSessionBuilder builder =
+        CqlSession.builder()
+            .addContactPoints(contactPoints)
+            .withLocalDatacenter("datacenter1"); // for SimpleSnitch
+
```
feeblefakie: We didn't do anything special for retry and load balancing policy but I'm wondering if you know where those configurations are gone to.

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
```diff
@@ -0,0 +1,106 @@
...
+    return session
+        .getMetadata()
+        .getKeyspace(keyspace)
+        .orElseThrow(() -> new StorageRuntimeException("no keyspace information found"))
+        .getTable(table)
+        .orElseThrow(() -> new StorageRuntimeException("no table information found"));
+  }
+
+  /** Closes the session. */
+  public void close() {
```
feeblefakie: I feel it should also close (set null) builder since it's closing the cluster.
What do you think ?

src/test/java/com/scalar/database/storage/cassandra4driver/ClusterManagerTest.java
```diff
@@ -0,0 +1,224 @@
...
+    // Assert
+    verify(manager).getBuilder();
+    verify(builder).build();
+  }
+
+  @Test
+  public void getSession_CalledAfterBuilt_ShouldNotBuild() {
+    // Arrange
+    manager.getSession();
+    manager.close();
```
feeblefakie: I think it is testing a little different thing from what it used to. 
so, maybe we should rename the test to something like `getSession_CalledAfterClosed_ShouldBuildAgain()`.

I feel it's a bit weird to see `getBuilder()` is called only once even after the cluster is closed. (it's related to the comment about `close()` method above.)
From users' perspective, I feel the expected behavior would be calling `getBuilder` and `build()` both twice ?
 
Whether or not we need to keep testing the original behavior, I feel we might not need to do it since there is no build() method in the class.

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
```diff
@@ -0,0 +1,106 @@
...
+  public TableMetadata getMetadata(String keyspace, String table) {
+    return session
+        .getMetadata()
+        .getKeyspace(keyspace)
+        .orElseThrow(() -> new StorageRuntimeException("no keyspace information found"))
+        .getTable(table)
+        .orElseThrow(() -> new StorageRuntimeException("no table information found"));
+  }
+
+  /** Closes the session. */
```
feeblefakie: If we also set null to builder,
```suggestion
  /** Closes the cluster. */
```

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A cluster manager
+ *
+ * @author Hiroyuki Yamada
```
feeblefakie: It should be 
```suggestion
 * @author Yuji Ito, Hiroyuki Yamada
```
Or we can delete this line and delegate it to commit logs.
(Sorry all the initial codes include this author annotation since it is automatically added by the editor I use...)

src/test/java/com/scalar/database/storage/cassandra4driver/ClusterManagerTest.java
```diff
@@ -0,0 +1,224 @@
...
+    when(keyspaceMetadata.getTable(anyString())).thenReturn(Optional.of(tableMetadata));
+
+    // Act
+    TableMetadata actual = manager.getMetadata(ANY_KEYSPACE_NAME, ANY_TABLE_NAME);
+
+    // Assert
+    assertThat(actual).isEqualTo(tableMetadata);
+  }
+
+  // @Test
```
feeblefakie: What is the reason of keeping this commented out  ?
If we want to keep in with this form, it would probably be nice to add some comments about why it is here.

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
feeblefakie: Or `closes the session and builder. ``` This might sound more correct.

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
```diff
@@ -0,0 +1,106 @@
...
+    List<InetSocketAddress> contactPoints = new ArrayList<InetSocketAddress>();
+    int port = config.getContactPort() == 0 ? DEFAULT_CASSANDRA_PORT : config.getContactPort();
+    config
+        .getContactPoints()
+        .forEach(contactPoint -> contactPoints.add(new InetSocketAddress(contactPoint, port)));
+
+    CqlSessionBuilder builder =
+        CqlSession.builder()
+            .addContactPoints(contactPoints)
+            .withLocalDatacenter("datacenter1"); // for SimpleSnitch
```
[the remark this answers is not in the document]
yito88: The datacenter name must be needed.
Ref. https://github.com/datastax/java-driver/tree/4.x/manual/core/load_balancing#local-only
And, for SimpleSnitch, it locates every node on "rack1" in "datacenter1"

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
```diff
@@ -0,0 +1,106 @@
...
+   * @param config database configuration needed to create a cluster
+   */
+  public ClusterManager(DatabaseConfig config) {
+    this.config = checkNotNull(config);
+  }
+
+  /**
+   * Returns a session. It will create a session if it is not created yet.
+   *
+   * @return a {@code Session}
```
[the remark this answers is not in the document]
yito88: Fixed

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
yito88: The driver provides only one balancing policy for now.
For retry policy, `withRetryPolicy()` has removed.
It can be configured with the driver's configuration.
Ref. https://github.com/datastax/java-driver/tree/4.x/manual/core/retries#retries

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
yito88: I think it is better to close both.
Fixed.

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
yito88: Fixed

src/main/java/com/scalar/database/storage/cassandra4driver/ClusterManager.java
yito88: Thanks!

src/test/java/com/scalar/database/storage/cassandra4driver/ClusterManagerTest.java
yito88: `getMetadata()` doesn't throw an exception for now.
https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/core/session/Session.html#getMetadata--
Removed the test.

src/test/java/com/scalar/database/storage/cassandra4driver/ClusterManagerTest.java
yito88: I renamed the test name. I think this test can check `close()` (setting null).

src/test/java/com/scalar/database/storage/cassandra4driver/ClusterManagerTest.java
```diff
@@ -0,0 +1,208 @@
...
+            })
+        .doesNotThrowAnyException();
+
+    // Assert
+    verify(manager).getBuilder();
+    verify(builder).build();
+  }
+
+  @Test
+  public void getSession_ConfigWithInvalidPortGiven_ShouldThrowExceptionProperly() {
```
feeblefakie: I'm wondering why this test is checking an invalid port instead of an invalid endpoint.
The new driver won't really throw exception any more even if an invalid endpoint is given ?
Maybe `InetSocketAddress` won't throw ?

yito88: `InetSocketAddress` constractor doesn't throw `IllegalArgumentException` for an invalid endpoint.
I added a check if an endpoint can be resolved and add a test for an invalid endpoint.

feeblefakie: Looking good !

PR #55, opened 2019-12-02
[Driver4] Add statement handlers for mutation

yito88:
- Change the import paths
- Use [`Relation`](https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/querybuilder/relation/Relation.html) for where clause
- Modify `ConditionSetter` to use `BuildableQuery` and [`Condition`](https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/querybuilder/condition/Condition.html) and update it by replacing the `BuildableQuery` instance
  - Get `BuildableQuery` instance after all conditions are set
- `StatementHandler#bind()` returns [`BoundStatementBuilder`](https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/core/cql/BoundStatementBuilder.html)
  - A mutable `builder` will be set consistency level and will be built for execution
- Remove the cache of prepared statements because the driver 4.x caches them

### DIFF
--- a/src/main/java/com/scalar/database/storage/cassandra4driver/InsertStatementHandler.java
+++ b/src/main/java/com/scalar/database/storage/cassandra4driver/InsertStatementHandler.java
@@ -21,7 +21,7 @@ import javax.annotation.concurrent.ThreadSafe;
 /**
  * A handler class for insert statements
  *
- * @author Hiroyuki Yamada
+ * @author Hiroyuki Yamada, Yuji Ito
  */
 @ThreadSafe
 public class InsertStatementHandler extends MutateStatementHandler {
--- a/src/main/java/com/scalar/database/storage/cassandra4driver/InsertStatementHandler.java
+++ b/src/main/java/com/scalar/database/storage/cassandra4driver/InsertStatementHandler.java
@@ -1,0 +1,92 @@
+package com.scalar.database.storage.cassandra4driver;
+
+import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.bindMarker;
+import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.insertInto;
+
+import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.cql.BoundStatement;
+import com.datastax.oss.driver.api.core.cql.BoundStatementBuilder;
+import com.datastax.oss.driver.api.core.cql.PreparedStatement;
+import com.datastax.oss.driver.api.core.cql.ResultSet;
+import com.datastax.oss.driver.api.querybuilder.insert.Insert;
+import com.datastax.oss.driver.api.querybuilder.insert.OngoingValues;
+import com.datastax.oss.driver.api.querybuilder.term.Term;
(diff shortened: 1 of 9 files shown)


src/main/java/com/scalar/database/storage/cassandra4driver/StatementHandler.java
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Returns a {@link PreparedStatement} based on the given query string
+   *
+   * @param queryString
+   * @return a {@code PreparedStatement}
+   */
+  @Nonnull
+  protected PreparedStatement prepare(String queryString) {
+    LOGGER.debug("query to prepare : [" + queryString + "].");
+    return session.prepare(queryString);
```
feeblefakie: I'm just curious. Is there any specific reason why it does't cache prepared statements as before ?
It used to warn if it prepares the same query multiple times. It doesn't warn any more ?
The doc seems still saying this.
> Preparing the same query string twice is suboptimal and a bad practice, and will cause the driver to log a warning.

https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/core/cql/PreparedStatement.html

yito88: Thank you!
According to the upgrade guide, prepared statements are cached.
I'll check the implementation.
https://github.com/datastax/java-driver/tree/4.x/upgrade_guide#statements

feeblefakie: Thanks ! That's cool ! It should have been done in the library.

yito88: I've confirmed that the preparation process checks the cache before making a new prepared statement.

https://github.com/datastax/java-driver/blob/46817dfa60a70b0c22f926600efe39cbe7891435/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java#L54-L83

feeblefakie: Cool ! Thanks.

src/main/java/com/scalar/database/storage/cassandra4driver/InsertStatementHandler.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  @Nonnull
+  protected ResultSet execute(BoundStatement bound, Operation operation) {
+    return session.execute(bound);
+  }
+
+  private Insert prepare(Put put) {
+    OngoingValues insert = insertInto(put.forNamespace().get(), put.forTable().get());
+    Map<String, Term> values = new LinkedHashMap<>();
```
feeblefakie: Just a question to understand it better.
The reason why it uses `LinkedHashMap` is to make preparation order and bind order the same ?

yito88: Yes, it's used for the order.

PR #57, opened 2019-12-02
[Driver4] Add batch handler and handler manager

yito88:
- Change the import paths
- Use [`BatchStatementBuilder`](https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/core/cql/BatchStatementBuilder.html)

### DIFF
--- a/src/test/java/com/scalar/database/storage/cassandra4driver/BatchHandlerTest.java
+++ b/src/test/java/com/scalar/database/storage/cassandra4driver/BatchHandlerTest.java
@@ -1,0 +1,386 @@
+package com.scalar.database.storage.cassandra4driver;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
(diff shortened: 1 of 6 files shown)


src/test/java/com/scalar/database/storage/cassandra4driver/BatchHandlerTest.java
```diff
@@ -0,0 +1,386 @@
...
+  public void handle_CorrectHandlerAndConditionalOperationsGiven_ShouldExecuteProperly()
+      throws ExecutionException {
+    // Arrange
+    configureBehavior();
+    mutations = prepareConditionalPuts();
+    when(session.execute(any(Statement.class))).thenReturn(results);
+    when(results.wasApplied()).thenReturn(true);
+
+    // Act Assert
+    assertThatCode(
```
siyopao: Here, and in other places, do we really have to check that no exceptions are thrown? If an exception is thrown the the test will fail (great!), and if not, the assert part of the test will check the various assertions. Is there something that I am missing?

yito88: Thank you for the comment!
I think it is useful to check that no exception is thrown to be also compared to other unit tests because it is possible to throw various exceptions from `batch.handle()`.
For example, `handle_EmptyOperationsGiven_ShouldThrowIllegalArgumentException` should be checked to throw an exception. It is pretty sure that this test should NOT throw any exception.

PR #61, opened 2019-12-06
Rename check

feeblefakie:
I just realized (after release...) that `check` might not be a clear method name for getting the state of a transaction.


siyopao:
LGTM!

PR #63, opened 2020-01-20
update the package

feeblefakie:
Just renamed the package name except for 1 change from the getting started doc.


docs/getting-started.md
```diff
@@ -34,10 +34,10 @@ $ make
 $ cd -
 ```
 Or you can download from [maven central repository](https://mvnrepository.com/artifact/com.scalar-labs/scalardb).
-For example in Gradle, you can add the following dependency to your build.gradle.
+For example in Gradle, you can add the following dependency to your build.gradle. Please replace the `<version>` with the version you want to use.
 ```
 dependencies {
-    compile group: 'com.scalar-labs', name: 'scalardb', version: '1.0.0-rc1'
+    compile group: 'com.scalar-labs', name: 'scalardb', version: '<version>'
```
feeblefakie: Sorry, this change was intentionally included even though it's not directly related to this update.

Issue #65, opened 2020-04-20
Non-serializable execution with write skew in some cases


feeblefakie:
### Environment
- Version 2.0.0
- with SERIALIZABLE

### How to reproduce
```
CREATE TRANSACTION TABLE foo.foo (
  id TEXT PARTITIONKEY,
  sub_id TEXT CLUSTERINGKEY,
  value INT,
);
```

- Initial state: no records

- Steps
  - T1: start
  - T2: start
  - T1: Scan all records that have `foo` as a partition key, and calculate a sum of value. (name the sum A). 0 since the table is empty.
    - (this is also fine) T1: Read ("foo", "t2") and put the value into A. Put 0 if the record does not exist.
  - T2: Scan all records that have `foo` as a partition key, and calculate a sum of value (name the sum B). 0 since the table is empty.
    - (this is also fine) T2: Read ("foo", "t1") and put the value into B. Put 0 if the record does not exist.
  - T1: Add 1 to A and put the value to ("foo", "t1")
  - T2: Add 1 to B and put the value to ("foo", "t2")
  - T1: commit
  - T2: commit

- Actual states :

id | sub_id | value
-- | ------ | ------
foo |     t1 |     1
foo |     t2 |     1

The above states can NOT be produced by either executing T1 and T2 in this order or T2 and T1 in this order as follows, which means the execution is not serializable.

- T1->T2

id | sub_id | value
-- | ------ | ------
foo |     t1 |     1
foo |     t2 |     2

- T2->T1

id | sub_id | value
-- | ------ | ------
foo |     t1 |     2
foo |     t2 |     1

feeblefakie:
We implemented some workarounds to make it not happen instead of actually fixing them.
Regarding the issue around reading (by get) a non-existing record, it will probably be fixed with writing DELETED as an enhancement after we decide on our delete strategy. 
Regarding the issue around scanning non-existing records, we will keep the current workaround until we find a good alternative. If we find it, we will work on as an enhancement.
Anyways, closing the issue for now.

@Yonezawa-T2 Thank you very much for the report.

Issue #66, opened 2020-04-21
Non-serializable execution (lost update) when a recode is deleted


feeblefakie:
### Environment
- Version 2.0.0
- with SERIALIZABLE

### How to reproduce
```
CREATE TRANSACTION TABLE foo.foo (
  id TEXT PARTITIONKEY,
  value TEXT
);
```

- Initial state: 

id | value | tx_version (tx metadata) 
-- | ------ | ---
foo |     2 |  1

- Steps
  - T1: start
  - T1: Read a record that has id = foo and put the value into A. If the record does not exist, put 0. 
    - (id, value, tx_version) = ("foo", 2, 1)
  - T2: start
  - T2: Delete a record that has id = foo
  - T2: commit
  - T3: start
  - T3: Read a record that has id = foo and put the value into B. If the record does not exist, put 0. 
  - T3: Add 1 to B and put
    - (id, value, tx_version) = ("foo", 1, 1)
  - T3: commit
  - T1: Add 1 to A and put
    - (id, value, tx_version) = ("foo", 3, 2)
  - T1: commit

- Actual states :

id | value | tx_version
-- | ------ | ------
foo |    3 |     2

The above state can NOT be produced by T1, T2 and T3 in some serial order, which means the execution is not serializable.

- T1->T3->T2 and T3->T1->T2
no record

- T2->T1->T3 and T2->T3->T1

id | value | tx_version
-- | ------ | ------
foo |    2 |     2

- T1->T2->T3 and T3->T2->T1

id | value | tx_version
-- | ------ | ------
foo |   1 |    1

Yonezawa-T2:
Cheap solution: randomize initial version.
Precise solution: check `tx_id` and `tx_version` in the prepare phase.

Randomizing initial version almost fix the issue except 1/2^32 of collision.

`tx_id` is a UUIDv4 value newly generated for each transaction, so the collision is neglectable. Checking two values may impact the performance, although it might not be significant.

feeblefakie:
@Yonezawa-T2 Thanks. That sounds a bit tricky and make verification more difficult.
I feel introducing logical delete and guaranteeing monotonic version increase would probably be better from many perspectives.

feeblefakie:
It is closed since it is fixed. 
@Yonezawa-T2 Thank you very much for the report.

PR #67, opened 2020-04-22
fix write skew issue

feeblefakie:
It fixes part of https://github.com/scalar-labs/scalardb/issues/65 .

The fix should probably be called workaround since it aborts if it reads empty record. So the write skew case with empty records can not happen in Scalar DB. 
It's a little limitation in the current implementation but I think there seems nothing else we can do since we detect conflicts by using records.

### DIFF
--- a/src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTest.java
+++ b/src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTest.java
@@ -3,6 +3,7 @@ package com.scalar.db.transaction.consensuscommit;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.catchThrowable;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
@@ -13,12 +14,14 @@ import com.scalar.db.api.Delete;
 import com.scalar.db.api.DistributedStorage;
 import com.scalar.db.api.DistributedTransaction;
 import com.scalar.db.api.Get;
+import com.scalar.db.api.Isolation;
 import com.scalar.db.api.Put;
 import com.scalar.db.api.Result;
 import com.scalar.db.api.Scan;
 import com.scalar.db.api.Selection;
 import com.scalar.db.api.TransactionState;
 import com.scalar.db.exception.storage.ExecutionException;
+import com.scalar.db.exception.transaction.CommitConflictException;
 import com.scalar.db.exception.transaction.CommitException;
 import com.scalar.db.exception.transaction.CoordinatorException;
 import com.scalar.db.exception.transaction.CrudException;
@@ -29,6 +32,7 @@ import com.scalar.db.io.IntValue;
 import com.scalar.db.io.Key;
 import com.scalar.db.io.Value;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.IntStream;
@@ -1082,7 +1086,8 @@ public class ConsensusCommitIntegrationTest {
             () -> {
               transaction.commit();
             })
-        .isInstanceOf(InvalidUsageException.class);
+        .isInstanceOf(CommitException.class)
+        .hasCauseInstanceOf(InvalidUsageException.class);
   }
 
   public void commit_DeleteGivenForNonExisting_ShouldThrowInvalidUsageException()
@@ -1099,7 +1104,8 @@ public class ConsensusCommitIntegrationTest {
             () -> {
               transaction.commit();
             })
-        .isInstanceOf(InvalidUsageException.class);
+        .isInstanceOf(CommitException.class)
+        .hasCauseInstanceOf(InvalidUsageException.class);
   }
 
   public void commit_DeleteGivenForExistingAfterRead_ShouldDeleteRecord()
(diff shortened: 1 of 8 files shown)


src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTest.java
```diff
@@ -1189,6 +1195,124 @@ public void commit_NonConflictingDeletesGivenForExisting_ShouldCommitBoth()
     assertThat(another.get(gets.get(account4)).isPresent()).isFalse();
   }
 
+  public void commit_WriteSkewOnExistingRecordsWithSnapshot_ShouldProduceNonSerializableResult()
```
feeblefakie, line 1198: This is just a new test verifying if write skew actually happens in SNAPSHOT. Just added for confirmation.

src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTest.java
```diff
@@ -1189,6 +1195,124 @@ public void commit_NonConflictingDeletesGivenForExisting_ShouldCommitBoth()
...
+    transaction = manager.start();
+    result1 = transaction.get(get1_1);
+    result2 = transaction.get(get2_1);
+    // the results can not be produced by executing the transactions serially
+    assertThat(result1.get().getValue(BALANCE).get()).isEqualTo(new IntValue(BALANCE, 2));
+    assertThat(result2.get().getValue(BALANCE).get()).isEqualTo(new IntValue(BALANCE, 2));
+  }
+
+  public void
+      commit_WriteSkewOnExistingRecordsWithSerializable_OneShouldCommitTheOtherShouldThrowCommitConflictException()
```
feeblefakie, line 1239: This is also a new test verifying if write skew actually avoided in SERIALIZABLE with non-empty records. Just added for confirmation.

src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTest.java
```diff
@@ -1189,6 +1195,124 @@ public void commit_NonConflictingDeletesGivenForExisting_ShouldCommitBoth()
...
+    // Assert
+    transaction = manager.start(Isolation.SERIALIZABLE);
+    result1 = transaction.get(get1_1);
+    result2 = transaction.get(get2_1);
+    assertThat(result1.get().getValue(BALANCE).get()).isEqualTo(new IntValue(BALANCE, 1));
+    assertThat(result2.get().getValue(BALANCE).get()).isEqualTo(new IntValue(BALANCE, 2));
+    assertThat(thrown).isInstanceOf(CommitConflictException.class);
+  }
+
+  public void commit_WriteSkewOnNonExistingRecordsWithSerializable_ShouldThrowCommitException()
```
feeblefakie, line 1279: This is a test to verify the fix is working as expected.
Since there is no way to detect conflicts for empty records correctly, it will abort when it reads empty record with SERIALIZABLE. It is a false positive abort so there might be a case where a history is serializable without such abort.

src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -25,7 +26,7 @@
   private static final Logger LOGGER = LoggerFactory.getLogger(Snapshot.class);
   private final String id;
   private final Isolation isolation;
-  private final Map<Key, TransactionResult> readSet;
+  private final Map<Key, Optional<TransactionResult>> readSet;
```
feeblefakie, line 29: Now, empty record is also put into the snapshot to identify reading empty records.

src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -117,11 +122,16 @@ void toSerializable() {
                 return;
               }
 
+              if (!e.getValue().isPresent()) {
```
feeblefakie, line 125: Here is the fix.

feeblefakie:
Yonezawa-san is added to a reviewer. 
@Yonezawa-T2 Thank you very much for the report !

Yonezawa-T2:
There is a better fix: creating a record with state `DELETED` in prepare phase prevents the anomary:

- T1: start
- T2: start
- T1: Read ("foo", "t2") and put the value into A. Put 0 if the record does not exist.
- T2: Read ("foo", "t1") and put the value into B. Put 0 if the record does not exist.
- T1: Add 1 to A and put the value to ("foo", "t1")
- T2: Add 1 to B and put the value to ("foo", "t2")
- T1: commit (succeeds)
- T2: commit (fails since T2 cannot create recored ("foo", "t1") with `DELETED` state).

feeblefakie:
@Yonezawa-T2 Thanks ! Yes, that's one of my future plans and is maybe coming.
But we need to make the current delete implementation logical instead of physical which
* causes lots of critical changes in the current code (so it requires lots of verification)
* might not be acceptable by users who manage personal information
so, we will work on after the above are settled.
Logical delete would fix another issue nicely so it should be the from technical perspective.
Thank you, anyways !

Yonezawa-T2:
Prepared records with state `DELETED` are deleted physically in `commitRecords` phase.

https://github.com/scalar-labs/scalardb/blob/f2c7dd4235a7fd481dcfde78c8ef754ccaefae73/src/main/java/com/scalar/db/transaction/consensuscommit/CommitMutationComposer.java#L84-L94

yito88:
@Yonezawa-T2 Thank you for your contribution!
I'm in charge of verification for Scalar DB.

I think it's an impractical change to add the new state of a record as @feeblefakie said.
I would be glad to review and verify your proposal if you make that PR.
However, it will take a few weeks to make a big change pass our verification.

I would like to avoid this issue with the small change for now.

feeblefakie:
@Yonezawa-T2 Yes, writing DELETE would work better when thinking about only this issue. But I think introducing logical delete would be better when thinking about #65 and #66 . So, even if writing DELETE here, the code will be updated, so I want to keep this change as minimum as possible for safety.

Yonezawa-T2:
@yito88 `DELETED` state is not a new state. It is used to delete a existing record.

https://github.com/scalar-labs/scalardb/blob/f2c7dd4235a7fd481dcfde78c8ef754ccaefae73/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java#L79-L105

yito88:
@Yonezawa-T2 Sorry it's not adding the new state as you said. It's adding a new state transition.

PR #69, opened 2020-04-23
Repeatable scan

feeblefakie:
https://scalar-labs.atlassian.net/browse/DLT-5833

### DIFF
--- a/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
+++ b/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
@@ -57,34 +57,35 @@ public class CrudHandler {
   }
 
   public List<Result> scan(Scan scan) throws CrudException {
-    // NOTICE : scan needs to always look at storage first since no primary key is specified
     List<Result> results = new ArrayList<>();
-    List<TransactionResult> uncommitted = new ArrayList<>();
+    List<Snapshot.Key> keys;
 
+    keys = snapshot.get(scan);
+    if (!keys.isEmpty()) {
+      keys.forEach(key -> results.add(snapshot.get(key).get()));
+      return results;
+    }
+
+    keys = new ArrayList<>();
     for (Result r : getFromStorage(scan)) {
       TransactionResult result = new TransactionResult(r);
       if (!result.isCommitted()) {
-        uncommitted.add(result);
-        continue;
+        throw new UncommittedRecordException(result, "the record needs recovery");
       }
+
+      Snapshot.Key key =
+          getSnapshotKey(r, scan)
+              .orElseThrow(() -> new CrudRuntimeException("can't get a snapshot key"));
+
+      if (snapshot.get(key).isPresent()) {
+        LOGGER.warn("scanned records are already in snapshot. overwriting snapshot...");
+      }
+
+      snapshot.put(key, Optional.of(result));
+      keys.add(key);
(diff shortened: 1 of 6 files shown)


Yonezawa-T2:
If the first scan results in empty results, the second scan doesn't use the snapshot, so that it may return newly committed results.

feeblefakie:
@Yonezawa-T2 Good point. Thanks. I'll fix it.

src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -57,34 +57,35 @@ public CrudHandler(DistributedStorage storage, Snapshot snapshot) {
...
-        continue;
+        throw new UncommittedRecordException(result, "the record needs recovery");
+      }
+
+      Snapshot.Key key =
+          getSnapshotKey(r, scan)
+              .orElseThrow(() -> new CrudRuntimeException("can't get a snapshot key"));
+
+      if (snapshot.get(key).isPresent()) {
+        LOGGER.warn("scanned records are already in snapshot. overwriting snapshot...");
```
Yonezawa-T2, line 81: This makes `get` not repeatable.

- T1: Get some value at key `foo`.
- T2: Put updated value to key `foo` and commit.
- T1: Scan a range including key `foo`.
- T1: Get again at key `foo`.

If a record exists in the snapshot, we should use it.
```
  result = snapshot.get(key).get();
```

Yonezawa-T2: At worse, this breaks isolation, results in lost update.

Initial state:
  foo: 1 (version 1)

- T1: Get the value at key `foo`, that is `1`, and store it to a variable A.
- T2: Increment the value at key `foo` and commit (value: 2, version: 2).
- T1: Scan a range including key `foo`, updating snapshot with value 2 (version 2).
- T1: Put `A + 2` to `foo`.
- T1: commit.  This will success because the snapshot hold a entry with version 2, which match the version on the Cassandra.

The update by T2 is lost. This cannot be prevented by even `SERIALIZABLE` isolation level.

Yonezawa-T2: This is not a new bug, should be fixed in a separate PR.

feeblefakie: Oh, you are really great at this :) 
Can you come up with suggestions ?

feeblefakie: Seems like the same issue as the #66. checking tx_id would seem to fix it.

Yonezawa-T2: I prefer using the old value rather than overriding the snapshot. Inserting the below line would fix it.

```
  result = snapshot.get(key).get();
```

feeblefakie: Yes, that one is just fixed.

feeblefakie:
@Yonezawa-T2 The non-repeatable read issue when scan result is empty is fixed. Please take 
a look again.

feeblefakie:
An integration test is added.

@yito88 Please take a look when you get a chance !

PR #70, opened 2020-04-24
avoid another write skew case (could be tentative workaround)

feeblefakie:
Another workaround for a case where scan reads empty records in  #65.
This could technically be fixed by writing DELETE record but it is even tricker since clustering key might be unknown in scan.

### DIFF
--- a/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -129,7 +129,6 @@ public class Snapshot {
 
   @VisibleForTesting
   void toSerializable() {
-    // TODO: scan might cause anomalies since it might cause phantoms ?
     readSet
         .entrySet()
         .forEach(
@@ -150,6 +149,18 @@ public class Snapshot {
                       .forNamespace(key.getNamespace())
                       .forTable(key.getTable());
               writeSet.put(e.getKey(), put);
+            });
+
+    scanSet
+        .entrySet()
+        .forEach(
+            e -> {
+              // if there is a scan on empty records and a write in a transaction
+              if (e.getValue().isPresent() && e.getValue().get().isEmpty() && !writeSet.isEmpty()) {
+                throw new CommitRuntimeException(
+                    "reading empty records might cause write skew anomaly "
+                        + "so aborting the transaction for safety.");
+              }
             });
   }
 
--- a/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -155,8 +155,8 @@ public class Snapshot {
         .entrySet()
         .forEach(
             e -> {
-              // if there is a scan on empty records and a write in a transaction
-              if (e.getValue().isPresent() && e.getValue().get().isEmpty() && !writeSet.isEmpty()) {
+              // if there is a scan and a write in a transaction
+              if (e.getValue().isPresent() && !writeSet.isEmpty()) {
                 throw new CommitRuntimeException(
                     "reading empty records might cause write skew anomaly "
                         + "so aborting the transaction for safety.");
--- a/src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTest.java
+++ b/src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTest.java
@@ -1338,6 +1338,39 @@ public class ConsensusCommitIntegrationTest {
     assertThat(thrown2).isInstanceOf(CommitException.class);
   }
 
+  public void
+      commit_WriteSkewWithScanOnNonExistingRecordsWithSerializable_ShouldThrowCommitException()
+          throws CrudException {
+    // Arrange
+    // no records
+
+    // Act
+    DistributedTransaction transaction1 = manager.start(Isolation.SERIALIZABLE);
+    DistributedTransaction transaction2 = manager.start(Isolation.SERIALIZABLE);
+    List<Result> results1 = transaction1.scan(prepareScan(0, 0, 1, NAMESPACE, TABLE_1));
+    int count1 = results1.size();
+    List<Result> results2 = transaction2.scan(prepareScan(0, 0, 1, NAMESPACE, TABLE_1));
+    int count2 = results2.size();
+    Put put1 = preparePut(0, 0, NAMESPACE, TABLE_1).withValue(new IntValue(BALANCE, count1 + 1));
+    transaction1.put(put1);
+    Put put2 = preparePut(0, 1, NAMESPACE, TABLE_1).withValue(new IntValue(BALANCE, count2 + 1));
+    transaction2.put(put2);
+    Throwable thrown1 = catchThrowable(transaction1::commit);
+    Throwable thrown2 = catchThrowable(transaction2::commit);
+
+    // Assert
+    assertThat(results1).isEmpty();
+    assertThat(results2).isEmpty();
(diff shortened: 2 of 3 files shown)


src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -129,7 +129,6 @@ public void to(MutationComposer composer) {
 
   @VisibleForTesting
   void toSerializable() {
-    // TODO: scan might cause anomalies since it might cause phantoms ?
```
Yonezawa-T2, line 132: Phantoms have not gone. We should keep this comment.

Initial state:

|`key`|`sub_key`|`value`|
|-----|---------|-------|
|`foo`|`t0`|1|
|`foo`|`t3`|1|

- T1: Scan where `'t0' <= sub_key and sub_key < 't3'` and store the sum of `value` into `sum1`.
- T2: Scan where `'t1' <= sub_key and sub_key < 't4'` and store the sum of `value` into `sum2`.
- T1: Put `sum1` into key ("foo", "t1").
- T2: Put `sum2` into key ("foo", "t2").
- T1: Commit. Checks `t0` and `t1` then succeeds.
- T2: Commit. Checks `t3` and `t2` then succeeds.

Final state:

|`key`|`sub_key`|`value`|
|-----|---------|-------|
|`foo`|`t0`|1|
|`foo`|`t1`|1|
|`foo`|`t2`|1|
|`foo`|`t3`|1|

Neither T1 → T2 nor T2 → T1 results in this state.

feeblefakie: Great 👍 
I think scan and write in a transaction could cause phantom/write skew in the current implementation (without something like predicate locking, which is a little hard to implement in the current architecture), so it probably needs to be like this.
```
              if (e.getValue().isPresent() && !writeSet.isEmpty()) {
```

feeblefakie: Probably,
R U W ⊇ S => OK (Serializable) 

R: a set of keys in read set by get
W: a set of keys in write set by put
S: a set of keys in scan ( ∞ for scan on empty records because keys could be anything)

feeblefakie:
Writing a DELETED record (with some specific clustering key) in scan and write is possible to lock a partition like predicate locking but mutation will occur twice as much (actual write and DELETED write in prepare and actual commit write and delete the DELETED record in commit) every time even scan is not happening concurrently, which I don't think is acceptable. I think we need to avoid scan and write in a transaction. Thought ?

Yonezawa-T2:
To prevent phantom read, we should do one of:

- locking ranges of keys,
- checking write-sets of transactions started after the start of the current transaction,
- rescanning the range atomically in prepareing phase, or
- any others?

The first option doesn't fit into Scalar DB since we have no range locks.
The second option also doesn't fit into Scalar DB since it doesn't have reliable clocks.
The third option works but expensive:

1. Create a lock table for each data tables.  It has the partition keys of the original table.
2. When a scan operation is executed, record the scan condition.
3. In prepareing phase, lock the entire partition by creating a lock record in the lock table.
4. Scan the table again and ensure the versions are same.
5. Put `PREPARED` records (we can use unconditional put here).
6. Update the transaction state to `COMMITTED`.
7. Delete the lock record.
8. Put `COMMITTED` records (we must use conditional put here).

Locking entire partitions may impact performance. On the other hand, we can use unconditional put for each data records in prepareing phase, since we rescan and verify condition ourselves and no other transactions can update records while locking the partitions.
We still need prepareing and committing since readers don't check the lock.
We should benchmark the positive and negative impact of this.

src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
Yonezawa-T2: Counter example (initial state are same to the previous example):

- T1: Scan where `'t0' <= sub_key and sub_key < 't3'` and store the sum of `value` into `sum1`.
- T2: Scan where `'t1' <= sub_key and sub_key < 't4'` and store the sum of `value` into `sum2`.
- T1: Put `sum1` into key ("foo", "t1") **and ("foo", "t0")**.
- T2: Put `sum2` into key ("foo", "t2") **and ("foo", "t3")**.
- T1: Commit. Checks `t0` and `t1` then succeeds.
- T2: Commit. Checks `t3` and `t2` then succeeds.

feeblefakie:
@Yonezawa-T2 Thanks.
The third option is essentially pretty much similar to what I said above (locking a partition). I think that is the doable option with the current implementation, but it is obviously expensive (we could do a benchmark though ...) and I wouldn't want Scalar DB to be complex to support only this case. Instead, I think we should rather avoid making it happened and recommend users to adjust applications. (because if you can do, you would do)
So, I would like to go with the current workaround now.
We could think about other better alternatives of-course.
Thank you anyways. Your feedbacks are really helpful.

src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
feeblefakie: T1 and T2 are reading non-existing records. so S is infinite.

Yonezawa-T2:
How about this committing procedure?

1. Prepare records in the write-set (without `toSerializable`).
2. Re-scan using conditions in the scan-set.
3. Re-get records in the read-set.
4. Update the transaction state to `COMMITTED`.
5. Update records in write-set to `COMMITTED`.

Read skews, write skews, and phantom read cannot happen using this scheme, I believe.

Suppose transactions T1 and T2. Read skews cannot happen:

- T1: Read A.
- T2: Write A.
- T2: Write B.
- T2: Commit.
- T1: Read B.
- T1: Commit. Fails since A is updated.

Nor phantom reads:

- T1: Scan.
- T2: Create a record in scanned range.
- T2: Commit.
- T1: Commit. Fails since scanned range is updated.

If T2 commit between re-scanning and updating the transaction state of T1, it cannot depend on the records in the write-set of T1 since those records are prepared, so it is equivalent to sequential execution T1 → T2.

Write skews cannot happen:

- T1: Read B.
- T2: Read A.
- T1: Write A.
- T2: Write B.
- T1: Prepare A.
- T1: Re-read B.
- T2: Prepare B.
- T2: Re-read A. Fails since A is prepared.
- T1: Update the transaction state to `COMMITTED`. Success.

To make T2 to successfully re-read A, it must be before preparing A, but then, T1 fails when re-reading B.

Benefits:

- No locking of entire partitions.
- No write for read-set (better than current implementation of `SERIALIZABLE`).
- `SNAPSHOT` and `SERIALIZABLE` is almost identical, only re-scan and re-get is added.

This scheme is equivalent to implicitly holding reader locks between read and re-read, and holding writer locks between read and transaction state update.

feeblefakie:
Thank you for the proposal. I think it is promising.
It can be seen as another way of removing anti-dependencies by actually checking if data is written after read.
So, I think it is a proper way.

I wouldn't say yet that it is generally better than the current way since there are still some unknowns/concerns as follows, for example.

- It is not certain which case it is going to be faster and it is not since C* read is way slower than write due to LSM-tree.
	- Proposed: For each record, a few disk read I/Os for a record + some reads and writes in Paxos read additionally occur.
	- Current: For each record, one Paxos write that comes with more network round trips and more writes additionally occur.
- More aborts would possibly happen since prepare phase is longer by extra read(s)
	- Might be problematic when there are contended/many write-skew causing conflicting transactions, which is one of the major cases that people choose Serializable over SI.
- Not starvation-free
	- For the write skew example above, two transactions can enter prepare phase at the same time (since their write sets don't conflict) and they can be both aborted in extra read phase.
	- Might be problematic as above

But, it is simple (which is the most important criteria in Scalar DB design) and definitely worth trying benchmarking to see how it actually goes. 
Let us plan it and do it.

PR #72, opened 2020-05-11
add another serializable strategy

feeblefakie:
This PR contains updates for another serializable strategy proposed by @Yonezawa-T2  in
https://github.com/scalar-labs/scalardb/pull/70 .

There are still some unknowns as discussed in #70  but I think it is a good alternative so I actually implemented it as yet another `SerializableStrategy` and named it `Extra Read` strategy (the existing strategy is now called `Extra Write` strategy). We will do benchmarking and verification once it is merged.

### DIFF
--- a/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
+++ b/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
@@ -62,15 +62,39 @@ public class ConsensusCommitManager implements DistributedTransactionManager {
 
   @Override
   public synchronized ConsensusCommit start(Isolation isolation) {
-    String txId = UUID.randomUUID().toString();
-    return start(txId, isolation);
+    return start(isolation, SerializableStrategy.EXTRA_WRITE);
   }
 
   @Override
   public synchronized ConsensusCommit start(String txId, Isolation isolation) {
+    return start(txId, isolation, SerializableStrategy.EXTRA_WRITE);
+  }
+
+  @Override
+  public synchronized ConsensusCommit start(
+      Isolation isolation, com.scalar.db.api.SerializableStrategy strategy) {
+    String txId = UUID.randomUUID().toString();
+    return start(txId, isolation, strategy);
+  }
+
+  @Override
+  public synchronized ConsensusCommit start(com.scalar.db.api.SerializableStrategy strategy) {
+    String txId = UUID.randomUUID().toString();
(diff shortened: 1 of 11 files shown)


src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -105,7 +116,9 @@ public void put(Snapshot.Key key, Delete delete) {
   }
 
   public void to(MutationComposer composer) {
-    if ((composer instanceof PrepareMutationComposer) && isolation == Isolation.SERIALIZABLE) {
+    if ((composer instanceof PrepareMutationComposer)
+        && isolation == Isolation.SERIALIZABLE
+        && strategy == SerializableStrategy.EXTRA_WRITE) {
       toSerializable();
```
feeblefakie, line 122: This method will be refactored in another PR by having `toSerializableWithExtraWrite` and the above if statement will be in the method for consistency with `toSerializableWithExtraRead`.

src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -62,15 +62,39 @@ public ConsensusCommit start(String txId) {
 
   @Override
   public synchronized ConsensusCommit start(Isolation isolation) {
-    String txId = UUID.randomUUID().toString();
-    return start(txId, isolation);
+    return start(isolation, SerializableStrategy.EXTRA_WRITE);
```
feeblefakie, line 65: After some benchmarking and verification, we can change the default strategy based on the results.

src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -164,6 +178,71 @@ void toSerializable() {
...
+        if (transactionResult.getId().equals(id)) {
+          continue;
+        }
+        currentReadSetByScan.add(transactionResult);
+      }
+
+      for (Key key : entry.getValue().get()) {
+        if (writeSet.containsKey(key)) {
+          continue;
+        }
```
Yonezawa-T2, line 204: Should we also check delete-set?

feeblefakie: Good catch. Yes, it should be checked as well. Thank you.

feeblefakie: Fixed in cdb12d1.

PR #73, opened 2020-05-14
refactor extra write and add/update some related tests

feeblefakie:
Basically renamed `toSerializable` to `toSerializableWithExtraWrite` and updated the for loop with usual `for` instead of using lambda `forEach` for consistency with extra read strategy added in #72. Note that the logic is not changed.

Additionally, some related unit tests are added.

### DIFF
--- a/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -14,7 +14,6 @@ import com.scalar.db.api.Result;
 import com.scalar.db.api.Scan;
 import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.exception.transaction.CommitConflictException;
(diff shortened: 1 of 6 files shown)


src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -141,41 +138,35 @@ public void to(MutationComposer composer) {
...
-              }
-            });
+    for (Map.Entry<Key, Optional<TransactionResult>> entry : readSet.entrySet()) {
+      Key key = entry.getKey();
+      if (writeSet.containsKey(key) || deleteSet.containsKey(key)) {
+        continue;
+      }
+
+      if (!entry.getValue().isPresent()) {
+        throwExceptionDueToPotentialAntiDependency();
```
feeblefakie, line 153: This part will be updated in a later PR with writing a DELETED record.

PR #77, opened 2020-07-20
Delete multiple records from Cosmos DB

yito88:
https://scalar-labs.atlassian.net/browse/DLT-6574

### Example
- Delete operation
```java
    Key partitionKey = new Key(new IntValue("c1", 0));
    Key clusteringKey = new Key(new IntValue("c2", 1));
    Delete delete = new Delete(partitionKey);
    storage.delete(delete);
```

- Before

|"c1" (PartitionKey)|"c2" (ClusteringKey)|"c3" (ClusteringKey)|"c4"|
--|--|--|--
|0|0|0|100|
|0|1|0|100|
|0|1|1|100|
|0|1|2|100|
|0|2|0|100|
|0|2|1|100|
|0|2|2|100|

- After

|"c1" (PartitionKey)|"c2" (ClusteringKey)|"c3" (ClusteringKey)|"c4"|
--|--|--|--
|0|0|0|100|
|0|2|0|100|
|0|2|1|100|
|0|2|2|100|

### DIFF
--- a/src/main/java/com/scalar/db/storage/cosmos/DeleteStatementHandler.java
+++ b/src/main/java/com/scalar/db/storage/cosmos/DeleteStatementHandler.java
@@ -47,7 +47,7 @@ public class DeleteStatementHandler extends MutateStatementHandler {
 
       getContainer(mutation).deleteItem(id, partitionKey, options);
     } else {
-      // clustering key is not specified
+      // clustering key is not fully specified
       executeStoredProcedure(mutation);
     }
   }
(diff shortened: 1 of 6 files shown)


src/main/java/com/scalar/db/storage/cosmos/DeleteStatementHandler.java
```diff
@@ -40,10 +40,15 @@ private void execute(Mutation mutation) throws CosmosException {
...
-    CosmosItemRequestOptions options = new CosmosItemRequestOptions();
+    if (cosmosMutation.isPrimaryKeySpecified()) {
+      String id = cosmosMutation.getId();
+      PartitionKey partitionKey = cosmosMutation.getCosmosPartitionKey();
+      CosmosItemRequestOptions options = new CosmosItemRequestOptions();
 
-    getContainer(mutation).deleteItem(id, partitionKey, options);
+      getContainer(mutation).deleteItem(id, partitionKey, options);
+    } else {
+      // clustering key is not specified
```
feeblefakie, line 50: Just a question. Is it OK if clustering keys are partially specified ?
If it is OK, I think the comment should be updated like this.
```
// clustering key is not fully specified
```

yito88: Thank you! Partially specified clustering keys are supported.
I've updated the comment.

PR #78, opened 2020-08-04
fix request for blob data

yito88:
https://scalar-labs.atlassian.net/browse/DLT-6876

### DIFF
--- a/src/main/java/com/scalar/db/storage/cosmos/MapVisitor.java
+++ b/src/main/java/com/scalar/db/storage/cosmos/MapVisitor.java
@@ -8,7 +8,6 @@ import com.scalar.db.io.FloatValue;
 import com.scalar.db.io.IntValue;
 import com.scalar.db.io.TextValue;
 import com.scalar.db.io.ValueVisitor;
-import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Map;
 import javax.annotation.concurrent.NotThreadSafe;
@@ -91,7 +90,8 @@ public class MapVisitor implements ValueVisitor {
   }
 
   /**
-   * Sets the specified {@code BlobValue} to the map
+   * Sets the specified {@code BlobValue} to the map. Scalar DB converts the blob as a string
+   * because Cosmos DB doesn't receive the value as a byte array.
    *
    * @param value a {@code BlobValue} to be set
    */
@@ -101,8 +101,7 @@ public class MapVisitor implements ValueVisitor {
         .get()
         .ifPresent(
             b -> {
-              ByteBuffer buffer = (ByteBuffer) ByteBuffer.allocate(b.length).put(b).flip();
-              values.put(value.getName(), buffer);
+              values.put(value.getName(), new String(value.get().get()));
             });
   }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/scalar/db/storage/cosmos/MapVisitor.java
```diff
@@ -91,7 +90,8 @@ public void visit(TextValue value) {
   }
 
   /**
-   * Sets the specified {@code BlobValue} to the map
+   * Sets the specified {@code BlobValue} to the map. Scalar DB converts the blob as a string
+   * because Cosmos DB doesn't receive the value as a byte array.
```
feeblefakie, line 94:
```suggestion
   * because Cosmos DB doesn't receive a byte array.
```

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
```diff
@@ -148,7 +148,7 @@ private Value convert(Object recordValue, String name, String type)
       case "varchar":
         return new TextValue(name, (String) recordValue);
       case "blob":
-        return new BlobValue(name, (byte[]) recordValue);
+        return new BlobValue(name, ((String) recordValue).getBytes());
```
feeblefakie, line 151: I feel it should specify an encoding to make it able to work with non-ascii.
```suggestion
        return new BlobValue(name, ((String) recordValue).getBytes(StandardCharsets.UTF_8));
```

src/test/java/com/scalar/db/storage/cosmos/ResultImplTest.java
```diff
@@ -74,7 +74,8 @@ public void setUp() throws Exception {
             .put(ANY_COLUMN_NAME_4, Float.MAX_VALUE)
             .put(ANY_COLUMN_NAME_5, Double.MAX_VALUE)
             .put(ANY_COLUMN_NAME_6, "string")
-            .put(ANY_COLUMN_NAME_7, "string".getBytes())
+            // Cosmos DB converts byte[] to a string
+            .put(ANY_COLUMN_NAME_7, "blob")
```
feeblefakie, line 78: I think it doesn't work with "ブロブ".

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
```diff
@@ -148,7 +148,7 @@ private Value convert(Object recordValue, String name, String type)
       case "varchar":
         return new TextValue(name, (String) recordValue);
```
feeblefakie, line 149: This might not work as well for non-ascii.

src/main/java/com/scalar/db/storage/cosmos/MapVisitor.java
```diff
@@ -101,8 +101,7 @@ public void visit(BlobValue value) {
         .get()
         .ifPresent(
             b -> {
-              ByteBuffer buffer = (ByteBuffer) ByteBuffer.allocate(b.length).put(b).flip();
-              values.put(value.getName(), buffer);
+              values.put(value.getName(), new String(value.get().get()));
```
feeblefakie, line 104: I think this also needs to be updated.

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
yito88: Thank you! I've fixed it.

PR #79, opened 2020-08-31
Fix reading blob from Cosmos DB

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7035

I found a bug about reading blob value from Cosmos DB.
In this modification, Scalar DB requests a write of blob value, does not convert it to `String`, and Cosmos DB will convert to String by Base64 encoding.
When reading the blob value, Scalar DB decodes the value because Cosmos DB returns it as String.

I tested the blob conversion to String in Scalar DB before. It worked well when the blob was short.
This PR supports long blob values.

### DIFF
--- a/src/main/java/com/scalar/db/storage/cosmos/MapVisitor.java
+++ b/src/main/java/com/scalar/db/storage/cosmos/MapVisitor.java
@@ -8,7 +8,7 @@ import com.scalar.db.io.FloatValue;
 import com.scalar.db.io.IntValue;
 import com.scalar.db.io.TextValue;
 import com.scalar.db.io.ValueVisitor;
-import java.nio.charset.StandardCharsets;
+import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Map;
 import javax.annotation.concurrent.NotThreadSafe;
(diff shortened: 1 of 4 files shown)


src/main/java/com/scalar/db/storage/cosmos/MapVisitor.java
```diff
@@ -102,7 +101,7 @@ public void visit(BlobValue value) {
         .get()
         .ifPresent(
             b -> {
-              values.put(value.getName(), new String(value.get().get(), StandardCharsets.UTF_8));
+              values.put(value.getName(), (ByteBuffer) ByteBuffer.allocate(b.length).put(b).flip());
```
feeblefakie, line 104: Just a question.
So, encoding to base64-based string is done by Cosmos DB but decoding should be done by users?

If so, what is the trigger of the automatic encoding? Is it done in case of byte array?
I'm just curious.

feeblefakie (commented):
Overall looking good. I just left a question for clarification.

src/main/java/com/scalar/db/storage/cosmos/MapVisitor.java
yito88: A byte array is automatically encoded to Base64 string in Azure Cosmos SDK.
But Cosmos DB always returns the encoded string because Cosmos DB doesn't know the actual type when it's written.
So, users should decode it.

feeblefakie: Got it. Thank you! (It's a pretty weird spec though...)

feeblefakie (approved):
LGTM!

PR #80, opened 2020-08-31
Set null or a default value when the value doesn't exist

yito88:
Scalar DB `get()` returns a Value even if a column doesn't exist in a record.

Before this PR, no Value is returned when the column doesn't exist on Cosmos DB.
An application throws NullPointerException in this case.
For Cassandra, Datastax Cassandra driver returns the default value like 0 for `row.getInt(name)` (when the `name` column doesn't exist) and Scalar DB makes a Value with it.

https://github.com/scalar-labs/scalardb/blob/f43227f8ded7982705f2d7913e9c57af353184d9/src/main/java/com/scalar/db/storage/cassandra/ResultImpl.java#L145-L167


src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
```diff
@@ -93,69 +92,72 @@ public String toString() {
...
   }
 
   private Optional<Key> getKey(Set<String> names) {
     List<Value> list = new ArrayList<>();
     for (String name : names) {
       Value value = values.get(name);
-      if (value == null) {
-        LOGGER.warn("full key doesn't seem to be projected into the result");
-        return Optional.empty();
-      }
```
yito88, line 130: This case doesn't happen anymore because the `values` always have the corresponding entry for the`name`.
All keys' entries are put in `interpret()`.

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
```diff
@@ -93,69 +92,72 @@ public String toString() {
...
+    metadata.getClusteringKeyNames().forEach(name -> add(name, record.getClusteringKey()));
   }
 
-  private void add(String name, Object value) {
-    if (metadata.getColumns().containsKey(name)) {
-      values.put(name, convert(value, name, metadata.getColumns().get(name)));
-    } else {
-      throw new InvalidMetadataException("metadata doesn't have the specified column: " + name);
-    }
+  private void add(String name, Map<String, Object> recordValues) {
```
feeblefakie, line 117: I think it's slightly better to define the method like this (more like keeping the method as it is) since the users would imagine it's going to create a `name -> recordValues` relationship by the method.
```suggestion
  private void add(String name, Object value) {
```

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
```diff
@@ -93,69 +92,72 @@ public String toString() {
...
-
-    // This isn't actual projection...
+    Map<String, Object> recordValues = record.getValues();
     if (selection.getProjections().isEmpty()) {
-      record.getValues().forEach((name, value) -> add(name, value));
+      metadata
+          .getColumns()
+          .forEach(
+              (name, type) -> {
+                add(name, recordValues);
```
feeblefakie, line 101:
```suggestion
                add(name, recordValues.get(name));
```

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
```diff
@@ -93,69 +92,72 @@ public String toString() {
...
-      Map<String, Object> recordValues = record.getValues();
+      // This isn't actual projection...
       selection
           .getProjections()
           .forEach(
               name -> {
-                if (recordValues.containsKey(name)) {
-                  add(name, recordValues.get(name));
-                }
+                add(name, recordValues);
```
feeblefakie, line 109:
```suggestion
                add(name, recordValues.get(name));
```

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
```diff
@@ -93,69 +92,72 @@ public String toString() {
...
               });
     }
+
+    metadata.getPartitionKeyNames().forEach(name -> add(name, record.getPartitionKey()));
```
feeblefakie, line 113:
```suggestion
    metadata.getPartitionKeyNames().forEach(name -> add(name, record.getPartitionKey().get(name)));
```

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
```diff
@@ -93,69 +92,72 @@ public String toString() {
...
+    metadata.getClusteringKeyNames().forEach(name -> add(name, record.getClusteringKey()));
```
feeblefakie, line 114:
```suggestion
    metadata.getClusteringKeyNames().forEach(name -> add(name, record.getClusteringKey().get(name)));
```

feeblefakie (approved):
LGTM! Left some minor suggestions but they might be just from my preference.

src/main/java/com/scalar/db/storage/cosmos/ResultImpl.java
yito88: Thank you! I've fixed it.

feeblefakie (approved):
LGTM! Thank you!

PR #84, opened 2020-09-28
Integrate DynamoDB

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7137

This PR covers basic CRUD operations and conditional mutations of Scalar DB on Cosmos DB.

## TODO
- [x] Manage DynamoDB client
- [x] Data model conversion & Metadata
- [x] Put operation without conditions
- [x] ResultImpl
- [x] Get operation
- [x] Scan operation
- [x] Delete operation without conditions
- [x] Conditional mutations
- [x] Batch operation
- [x] Integration tests
- [x] Select storages

## Added TODO
- [x] Specify AWS credentials by the config file
- [x] Support multiple clustering keys
- [x] Use updateItem() for insertion
- [x] Unit tests
- [x] Update the schema tool for DynamoDB

## Others
- ~Only 1 clustering key can be used~
  - Dynamo DB cannot support multiple sort(clustering) keys
    - Scalar DB supports multiple clustering keys by local secondary indexes
  - ~I'll modify the schema tool to reject the multiple clustering keys for Dynamo DB~
- `delete` cannot multiple records at once with partition keys on DynamoDB
  - Need to specify the primary key (partition key + clustering(sort) key)
- ~Need to modify the schema tool to set up the metadata tables on DynamoDB~ [done]

### DIFF
--- a/src/main/java/com/scalar/db/storage/dynamo/BatchHandler.java
+++ b/src/main/java/com/scalar/db/storage/dynamo/BatchHandler.java
@@ -1,0 +1,166 @@
+package com.scalar.db.storage.dynamo;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.scalar.db.api.DeleteIfExists;
+import com.scalar.db.api.Mutation;
+import com.scalar.db.api.MutationCondition;
+import com.scalar.db.api.PutIf;
+import com.scalar.db.api.PutIfExists;
+import com.scalar.db.api.PutIfNotExists;
+import com.scalar.db.exception.storage.ExecutionException;
+import com.scalar.db.exception.storage.MultiPartitionException;
+import com.scalar.db.exception.storage.NoMutationException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.concurrent.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import software.amazon.awssdk.services.dynamodb.model.CancellationReason;
+import software.amazon.awssdk.services.dynamodb.model.Delete;
+import software.amazon.awssdk.services.dynamodb.model.DynamoDbException;
+import software.amazon.awssdk.services.dynamodb.model.Put;
+import software.amazon.awssdk.services.dynamodb.model.TransactWriteItem;
+import software.amazon.awssdk.services.dynamodb.model.TransactWriteItemsRequest;
+import software.amazon.awssdk.services.dynamodb.model.TransactionCanceledException;
+import software.amazon.awssdk.services.dynamodb.model.Update;
+
+/**
+ * A handler for a batch
+ *
+ * @author Yuji Ito
+ */
+@ThreadSafe
+public class BatchHandler {
+  static final Logger LOGGER = LoggerFactory.getLogger(BatchHandler.class);
+  private final DynamoDbClient client;
+  private final TableMetadataManager metadataManager;
+
+  /**
+   * Constructs a {@code BatchHandler} with the specified {@link DynamoDbClient} and {@link
+   * TableMetadataManager}
+   *
+   * @param client {@code DynamoDbClient} to create a statement with
+   * @param metadataManager {@code TableMetadataManager}
+   */
+  public BatchHandler(DynamoDbClient client, TableMetadataManager metadataManager) {
+    this.client = client;
+    this.metadataManager = metadataManager;
+  }
+
+  /**
+   * Execute the specified list of {@link Mutation}s in batch. All the {@link Mutation}s in the list
+   * must be for the same partition. Otherwise, it throws an {@link MultiPartitionException}.
+   *
+   * @param mutations a list of {@code Mutation}s to execute
+   * @throws NoMutationException if at least one of conditional {@code Mutation}s failed because it
+   *     didn't meet the condition
+   */
+  public void handle(List<? extends Mutation> mutations) throws ExecutionException {
+    checkNotNull(mutations);
+    if (mutations.size() < 1) {
+      throw new IllegalArgumentException("please specify at least one mutation.");
+    } else if (mutations.size() > 25) {
+      throw new IllegalArgumentException("DynamoDB cannot batch more than 25 mutations at once.");
+    }
+
+    TransactWriteItemsRequest.Builder builder = TransactWriteItemsRequest.builder();
+    List<TransactWriteItem> transactItems = new ArrayList<>();
+    mutations.forEach(m -> transactItems.add(makeWriteItem(m)));
+    builder.transactItems(transactItems);
+
+    try {
+      client.transactWriteItems(builder.build());
+    } catch (TransactionCanceledException e) {
+      for (CancellationReason reason : e.cancellationReasons()) {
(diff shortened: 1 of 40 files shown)


src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -0,0 +1,267 @@
...
+    List<Value> end = scan.getEndClusteringKey().get().get();
+    String startKeyName = start.get(start.size() - 1).getName();
+    String endKeyName = end.get(end.size() - 1).getName();
+
+    DynamoOperation dynamoOperation = new DynamoOperation(scan, metadataManager);
+
+    if (startKeyName.equals(endKeyName)) {
+      conditions.add(startKeyName + DynamoOperation.RANGE_CONDITION);
+      if (!scan.getStartInclusive() || !scan.getEndInclusive()) {
+        throw new IllegalArgumentException(
```
feeblefakie, line 171: Similar to the clustering key limitation, it probably needs to exclude entries in the adapter side not to break the abstraction.

BTW, `com.amazonaws.services.dynamodbv2` is an old library?
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ScanJavaDocumentAPI.html
In this, Scan seems to support exclusive for start, but not for end...

src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -0,0 +1,267 @@
...
+    DynamoOperation dynamoOperation = new DynamoOperation(scan, metadataManager);
+    String partitionKey = dynamoOperation.getConcatenatedPartitionKey();
+
+    bindMap.put(
+        DynamoOperation.PARTITION_KEY_ALIAS, AttributeValue.builder().s(partitionKey).build());
+
+    return bindMap;
+  }
+
+  private boolean setRangeCondition(Scan scan, List<String> conditions) {
```
feeblefakie, line 156: I think it's a bit weird to return `boolean` for `setXXX` method.

src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -0,0 +1,267 @@
...
+    return client.query(builder.build());
+  }
+
+  private void setConditions(QueryRequest.Builder builder, Scan scan) {
+    List<String> conditions = new ArrayList<>();
+
+    conditions.add(getPartitionKeyCondition());
+
+    boolean isRangeEnabled = setRangeCondition(scan, conditions);
+    if (!isRangeEnabled) {
```
feeblefakie, line 125: Sorry, I'm not really sure why `range` is specially treated in here and [below](https://github.com/scalar-labs/scalardb/pull/84/files#diff-fd1d1c3e8897b8263f236fe2b59f5ff7R132).
Could you elaborate on this?

src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -0,0 +1,267 @@
...
+import software.amazon.awssdk.services.dynamodb.model.QueryRequest;
+import software.amazon.awssdk.services.dynamodb.model.QueryResponse;
+
+/**
+ * A handler class for select statement
+ *
+ * @author Yuji Ito
+ */
+@ThreadSafe
+public class SelectStatementHandler extends StatementHandler {
```
feeblefakie, line 33: Just a question. If the secondary index is used, can this `SelectStatementHandler` be used in the same way?

feeblefakie (commented):
Looking good! Great work.

I left some comments, suggestions, and questions.
I think the biggest concern at the moment is scan since DynamoDB scan (including 1 sort key thing) is pretty limited. We probably need to deal with it in the adapter side.

src/main/java/com/scalar/db/storage/dynamo/BatchHandler.java
```diff
@@ -0,0 +1,166 @@
...
+      throw new ExecutionException(e.getMessage(), e);
+    }
+  }
+
+  private TransactWriteItem makeWriteItem(Mutation mutation) {
+    TransactWriteItem.Builder itemBuilder = TransactWriteItem.builder();
+
+    if (mutation instanceof com.scalar.db.api.Put) {
+      MutationCondition condition = mutation.getCondition().orElse(null);
+      if (condition != null && (condition instanceof PutIf || condition instanceof PutIfExists)) {
```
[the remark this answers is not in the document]
yito88: Good question! I need to modify the behavior.

In the case where the item doesn't exist with the primary key, PutItem and Update Item work the same.
When the item already exists, they are different.
- PutItem replaces the existing item with the new item
- UpdateItem updates or adds the values of the specific attributes in the existing item.

In Cosmos DB, UpsertItem API is used. Cassandra's INSERT is also upsert.
Cassandra INSERT: https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlInsert.html

To be the same behavior, we should use UpdateItem for inserting a new item.

src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+  }
+
+  private void setStartCondition(Scan scan, List<String> conditions) {
+    DynamoOperation dynamoOperation = new DynamoOperation(scan, metadataManager);
+    scan.getStartClusteringKey()
+        .ifPresent(
+            k -> {
+              if (k.get().size() > 1) {
+                throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
yito88: Thank you for your suggestion. It makes sense.

I'd like to modify `scan` operation in another PR to support multiple clustering keys.

src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
yito88: Thank you.
I'm not sure what `exclusiveStartKey` does. It seems that it doesn't suite this usage.

> To fetch the next page, you execute the scan method again by providing the primary key value of the last item in the previous page so that the scan method can return the next set of items. You provide this information in the request by using the withExclusiveStartKey method.`

> The primary key of the first item that this operation will evaluate. Use the value that was returned for LastEvaluatedKey in the previous operation.

src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
yito88: At first, I tried to use multiple conditions to specify the range like `:ck > :start_val AND :ck < :end_val`.
However, it didn't work because a condition string can include only 1 condition per key.
(If it could work, we can use `exclusive` option in DynamoDB)
That's why the range condition is specified by `a BETWEEN b AND c `.

src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
yito88: I think it can be used in the same way because we can use `client.query()` API for the secondary index too.

yito88:
@feeblefakie Thank you for your review!
I'd like to update this branch by making other branches and PRs based on this branch.
That's because I think this integration is immature for now.
After the updates, we can merge this branch into the master.

yito88:
@feeblefakie Thank you for the lots of reviews!
I'd like to merge this PR to the master.

All additional PRs have been merged for the integration.
Could you take a look at this PR again?

PR #87, opened 2020-10-06
Support multiple clustering keys

yito88:
To support multiple clustering keys with secondary indexes

### DIFF
--- a/src/main/java/com/scalar/db/storage/dynamo/ScannerImpl.java
+++ b/src/main/java/com/scalar/db/storage/dynamo/ScannerImpl.java
@@ -3,9 +3,13 @@ package com.scalar.db.storage.dynamo;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.scalar.db.api.Result;
+import com.scalar.db.api.Scan;
 import com.scalar.db.api.Scanner;
 import com.scalar.db.api.Selection;
+import com.scalar.db.exception.storage.UnsupportedTypeException;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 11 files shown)


src/main/java/com/scalar/db/storage/dynamo/ScannerImpl.java
```diff
@@ -53,4 +58,70 @@ public ScannerImpl(
...
+  private void sort() {
+    Scan scan = (Scan) selection;
+    DynamoOperation dynamoOperation = new DynamoOperation(scan, metadata);
+    if (dynamoOperation.isSingleClusteringKey()) {
+      // the ordering and the limitation already are applied in DynamoDB
+      return;
+    }
+
+    if (!scan.getOrderings().isEmpty()) {
+      Collections.sort(
```
feeblefakie, line 71: I think this part should be separated as another class like `ItemSorter` for separation of concerns and it should be tested independently to check the sorting capability intensively. (like testing all the types and complex sort orders because bugs can be easily introduced in this kind of code)

In that case, `ItemSorter` can be given in the constructor like this.
```
public ScannerImpl
    List<Map<String, AttributeValue>> items, Selection selection, TableMetadata metadata, ItemSorter sorter) {
```

src/test/java/com/scalar/db/storage/dynamo/ScannerImplTest.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import org.junit.Before;
+import org.junit.Test;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+
+public class ScannerImplTest {
```
feeblefakie, line 19: I think `ScannerImplTest` should test if `isSingleClusteringKey` check is properly working and the expected behaviors are happening for each case.
For example, if `isSingleClusteringKey` is `false` then checking if the sorting function is called.
The sorting function itself should be tested in the unit test of `ItemSorter` (a sorting class whatever the name is).

feeblefakie (commented):
Overall looking good. Great work 👍 

It's not super critical, but I left some suggestions for better design and testability.
I feel we'd better test the sorting with all the types and various sorting orders because bugs can be easily introduced in this kind of code.

src/main/java/com/scalar/db/storage/dynamo/ScannerImpl.java
yito88: Thank you! I've added `ItemSorter`. Could you take a look again?

src/test/java/com/scalar/db/storage/dynamo/ScannerImplTest.java
yito88: Thank you! I've added `ItemSorterTest`. Could you take a look again?

feeblefakie (approved):
LGTM! Thank you!

PR #90, opened 2020-10-12
Add --port option to scalar-schema tool

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-7392

Add `-P` `--port` option to the scalar-schema tool.
It is needed to replace [the old schema-loader for Cassandra](https://github.com/scalar-labs/scalardl-schema-loader-cassandra).


yito88 (approved):
LGTM! Thank you!

tools/scalar-schema/src/scalar_schema/cassandra_schema.clj
```diff
@@ -91,9 +92,9 @@
        doall))
 
 (defn operate-cassandra
-  [{:keys [schema-file host user password]
+  [{:keys [schema-file host port user password]
     :or {user "cassandra" password "cassandra"} :as options}]
```
yito88, line 96:
```suggestion
    :or {user "cassandra" password "cassandra" port 9042} :as options}]
```

tools/scalar-schema/src/scalar_schema/core.clj
```diff
@@ -8,6 +8,8 @@
   [[nil "--cassandra" "Operate for Cassandra"]
    [nil "--cosmos" "Operate for Cosmos DB"]
    ["-h" "--host DB_HOST" "Address of Cassandra like IP or URI address of your Cosmos DB account"]
+   ["-P" "--port DB_PORT" "Port of Cassandra. This option is ignored when Cosmos DB."
+    :default 9042 :parse-fn #(Integer/parseInt %)]
```
yito88, line 12: I think this port number might be used for other data stores in the future.
So, the default value should be set in the `cassandra_schema.clj` as I've suggested.

yito88 (changes requested):
I added a suggestion.

tools/scalar-schema/README.md
```diff
@@ -22,7 +22,7 @@ $ java -jar target/scalar-schema.jar --cosmos -h <YOUR_ACCOUNT_URI> -p <YOUR_ACC
 
 ```console
 # For Cassandra
-$ java -jar target/scalar-schema.jar --cassandra -h <CASSANDRA_IP> -u <CASSNDRA_USER> -p <CASSANDRA_PASSWORD> -f schema.json [-n <NETWORK_STRATEGY> -R <REPLICATION_FACTOR>]
+$ java -jar target/scalar-schema.jar --cassandra -h <CASSANDRA_IP> -P <CASSANDRA_PORT> -u <CASSNDRA_USER> -p <CASSANDRA_PASSWORD> -f schema.json [-n <NETWORK_STRATEGY> -R <REPLICATION_FACTOR>]
```
feeblefakie, line 25: Based on the suggestions from Yuji, `-P` should be optional?

feeblefakie (commented):
Overall looking good! Thanks.
Left 1 comment.

tools/scalar-schema/src/scalar_schema/core.clj
ymorimo: Should `:default 9042` here be removed?

tools/scalar-schema/README.md
ymorimo: Yes, and probably `-u` and `-p` could also be optional since they have default values (`cassandra`/`cassandra`)? What do you think? @yito88

tools/scalar-schema/src/scalar_schema/core.clj
yito88: Yes, I think so. Each data store's schema process couldn't set each default value if the default value would be set here.

tools/scalar-schema/README.md
yito88: Yes. These default values are set in `cassandra_schema.clj` when we don't specify them by the options.

ymorimo: Thanks! Updated in 6d24892. (Brackets are changed to surround each optional param.)

yito88 (approved):
LGTM 👍

feeblefakie (approved):
LGTM!

Issue #89, opened 2020-10-10
Typecast issue exists in scalardb with cosmos db. 


scalar-boney:
**Describe the bug**
Typecast issue exists in scalardb while reading data from Cosmos DB using scalardb

**To Reproduce**
1. Setup [QA application with cosmos db](https://github.com/scalar-labs/getting-started-with-scalardb)
2. Login to QA application
3. Insert some questions using the `Submit a Question` button. 
4. See some error `An error occurred while looking up the question` after creating the question.

**Expected behavior**
Read data from cosmos db using scalardb without any error.

**Screenshots**
Typecast issue exists while reading data from cosmos db.

**Backend Details**
 - Scalardb: 2.2.0
 - Storage: Cosmos DB

**Additional context**
Issue related logs
```
2020-10-08 23:12:59.355 ERROR 8243 --- [0.1-8090-exec-3] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long] with root cause

java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
        at com.scalar.db.storage.cosmos.ResultImpl.convert(ResultImpl.java:142) ~[scalardb-2.2.0.jar:na]
        at com.scalar.db.storage.cosmos.ResultImpl.add(ResultImpl.java:120) ~[scalardb-2.2.0.jar:na]
        at com.scalar.db.storage.cosmos.ResultImpl.lambda$interpret$0(ResultImpl.java:101) ~[scalardb-2.2.0.jar:na]
        at com.google.common.collect.ImmutableSortedMap.forEach(ImmutableSortedMap.java:588) ~[guava-24.1-jre.jar:na]
        at java.util.Collections$UnmodifiableMap.forEach(Collections.java:1505) ~[na:1.8.0_152-ea]
        at com.scalar.db.storage.cosmos.ResultImpl.interpret(ResultImpl.java:99) ~[scalardb-2.2.0.jar:na]
        at com.scalar.db.storage.cosmos.ResultImpl.<init>(ResultImpl.java:44) ~[scalardb-2.2.0.jar:na]
        at com.scalar.db.storage.cosmos.ScannerIterator.next(ScannerIterator.java:34) ~[scalardb-2.2.0.jar:na]
[...]

PR #96, opened 2020-10-13
Fix the cast for numbers on Cosmos DB

yito88:
#89

## Issue
Scalar DB failed to cast to a long value on Cosmos DB.

```
Execution error (ClassCastException) at com.scalar.db.storage.cosmos.ResultImpl/convert (ResultImpl.java:142).
java.lang.Integer cannot be cast to java.lang.Long
```

## Cause
A value read from Cosmos DB is a Number. Some values like zero were cast to Integer.

## Fix
Cast a number via Number


feeblefakie (approved):
LGTM!  Thank you!

PR #97, opened 2020-10-20
Replace putItem with updateItem for upsert

yito88:
To upsert items

According to this [disscussion](https://github.com/scalar-labs/scalardb/pull/84#discussion_r498612446), we should be able to overwrite the existing item like C*.
Dynamo DB client's `putItem()` cannot overwrite it.
So, I replaced it with `updateItem()`.


feeblefakie (approved):
LGTM!

PR #99, opened 2020-10-22
fix RU type for DynamoDB

yito88:
To specify the RU for DynamoDB.

### Issue
- RU couldn't set for DynamoDB
```
Exception in thread "main" java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
```

### Cause
- The given type wasn't correct for `readCapacityUnits()` and `writeCapacityUnits()`

### Fix
- Cast to Long


feeblefakie (approved):
LGTM!

PR #101, opened 2020-10-22
fix metadata

yito88:
### Issue
- When no clustering key exists, NullPointerException was thrown.
```
Execution error (NullPointerException) at com.scalar.db.storage.dynamo.TableMetadata/convert (TableMetadata.java:61).
```

### Cause
- The metadata couldn't be converted from the values on the metadata table because the value was null when no clustering key

### Fix
- When no clustering key, the name set of clustering keys is set as an empty set.


feeblefakie (commented):
Looks good!
Can you write some unit tests to cover the case?

yito88:
Thank you for the suggestion!
I added some tests.

feeblefakie (approved):
LGTM! Thank you!

PR #102, opened 2020-10-22
Replace a put request for batch

yito88:
This PR is almost the same as #97 to replace put request with update request.


feeblefakie (approved):
LGTM!

PR #103, opened 2020-10-22
Simplify the condition

yito88:
We can simplify the condition for IfNotExists and IfExists because we have `concatenatedClusteringKey` attribute after supporting multiple clustering keys.


feeblefakie (approved):
LGTM!

PR #106, opened 2020-10-29
Alias for update expression

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7500
To use column names reserved by DynamoDB like `date`, `user`.

Set aliases of columns and the name map to the update request.


feeblefakie (approved):
LGTM!

PR #107, opened 2020-10-29
Fix the schema tool to modify the secondary index table

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7404
https://scalar-labs.atlassian.net/browse/DLT-7499

### Issue
- Schema tool failed to create a secondary index which had more than 20 non-key attributes
```
'localSecondaryIndexes.1.member.projection.nonKeyAttributes' failed to satisfy constraint: Member must have length less than or equal to 20
```

### Cause
- The API `nonKeyAttributes()` couldn't receive more than 20 attributes

### Fix
- Use `ProjectionType/ALL` instead of `ProjectionType/INCLUDE` to set all attributes


ymorimo (approved):
It worked! LGTM.

feeblefakie (approved):
LGTM! Nice!

PR #108, opened 2020-11-02
fix for a null value

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7534

### Issue
- Put with a null value failed
```
Caused by: com.scalar.db.exception.storage.ExecutionException: Invalid UpdateExpression: An expression attribute value used in expression is not defined; ...
```

### Cause
- A null string(bytes) could not be defined as an AttributeValue
```
# non null string
:val0=AttributeValue(S=foo, SS=[], NS=[], BS=[], M={}, L=[])
# null string
:val4=AttributeValue(SS=[], NS=[], BS=[], M={}, L=[])
```

### Fix
- A null string(bytes) is set as a null type value
```
:val4=AttributeValue(SS=[], NS=[], BS=[], M={}, L=[], NUL=true)
```

### DIFF
--- a/src/main/java/com/scalar/db/storage/dynamo/ResultImpl.java
+++ b/src/main/java/com/scalar/db/storage/dynamo/ResultImpl.java
@@ -17,7 +17,6 @@ import com.scalar.db.io.IntValue;
 import com.scalar.db.io.Key;
 import com.scalar.db.io.TextValue;
 import com.scalar.db.io.Value;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


src/main/java/com/scalar/db/storage/dynamo/ResultImpl.java
```diff
@@ -131,22 +130,23 @@ private Value convert(AttributeValue itemValue, String name, String type)
       throws UnsupportedTypeException {
     // When itemValue is NULL, the value will be the default value.
     // It is the same behavior as the datastax C* driver
+    boolean isNull = itemValue == null || (itemValue.nul() != null && itemValue.nul());
```
feeblefakie, line 133: Just a question.
Is there any case where `itemValue.nul()` is null?

feeblefakie (approved):
LGTM! 
(left a question but it's non blocking)

src/main/java/com/scalar/db/storage/dynamo/ResultImpl.java
yito88: Yes, `AttributeValue#nul()` returns `Boolean`. It returns null when the itemValue has a non-null value.

PR #109, opened 2020-11-05
use aliases of columns for batch

yito88:
Similar to #106
Sorry, I forgot to add the aliases of columns for a bach operation.

https://scalar-labs.atlassian.net/browse/DLT-7500
To use column names reserved by DynamoDB like `date`, `user`.

Set aliases of columns and the name map to the update request for a batch operation.


feeblefakie (approved):
LGTM!

PR #112, opened 2020-11-10
Refactor the schema tool

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7617

Added a protocol (like an interface) to reduce redundant processes


tools/scalar-schema/project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject scalar-schema "2.2.2"
+(defproject scalar-schema "2.3.0"
```
yito88, line 1: I guess the next version will be 2.3.0.

feeblefakie (approved):
LGTM!

PR #113, opened 2020-11-10
Update scalar-schema tool to exit with status 1 if errors in CLI options

ymorimo:
In case of an error in the scalar-schema CLI options, it should fail with an exit status other than `0`.

It is needed to notify tools like Ansible and Terraform of the result of the scalar-schema tool.
For example, if the argument for the `--region` option is missing, it should fail with a status other than `0` to tell Terraform to abort.
The following case should not complete successfully.

```
module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_schema[0] (remote-exec): ERROR: ["Missing required argument for \"--region REGION\""]
...
module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_schema[0]: Creation complete after 8s [id=5751361347562494241]
```

https://scalar-labs.atlassian.net/browse/DLT-7415

### DIFF
--- a/tools/scalar-schema/src/scalar_schema/core.clj
+++ b/tools/scalar-schema/src/scalar_schema/core.clj
@@ -30,11 +30,13 @@
   (let [{:keys [options summary errors]
          {:keys [cassandra cosmos dynamo help]} :options}
         (parse-opts args cli-options)]
-    (if (or help errors)
-      (do (when (not help)
-            (println (str "ERROR: " errors)))
-          (println summary))
-      (do
-        (when cassandra (cassandra-schema/operate-cassandra options))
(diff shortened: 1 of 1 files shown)


feeblefakie:
@ymorimo Seems like it's got a conflict after Yuji's refactoring. Please fix 🙇

yito88:
@ymorimo @feeblefakie Sorry, my refactoring wasn't verified well. It works on only DynamoDB. That's because I verified it with the old binary... 🙇 I'll test and fix it. I guess it is better for you to fix this PR after that.

yito88 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #115, opened 2020-11-12
Fix clustering key order

yito88:
Fixed the ordering because the clustering keys were reordered by Set


feeblefakie (approved):
Super quick! Thank you!

PR #116, opened 2020-11-13
Support auto-scaling for DynamoDB

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7579

To register ScalableTarget and set ScalingPocily to each table.
ScalableTarget specifies the minimum and the maximum Capacity Unit.
ScalingPolicy specifies the interval and the usage rate to scale in/out the CU.

Ref. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/AutoScaling.html

### DIFF
--- a/tools/scalar-schema/README.md
+++ b/tools/scalar-schema/README.md
@@ -38,21 +38,20 @@ $ java -jar target/scalar-schema-standalone-<version>.jar --cassandra -h <CASSAN
   - If `-p <CASSANDRA_PASSWORD>` is not supplied, it defaults to `cassandra`.
   - `<NETWORK_STRATEGY>` should be `SimpleStrategy` or `NetworkTopologyStrategy`
 
-### Delete all tables
+### Delete tables
 ```console
 # For Cosmos DB
-$ java -jar target/scalar-schema-standalone-<version>.jar --cosmos -h <YOUR_ACCOUNT_URI> -p <YOUR_ACCOUNT_PASSWORD> -D
+$ java -jar target/scalar-schema-standalone-<version>.jar --cosmos -h <YOUR_ACCOUNT_URI> -p <YOUR_ACCOUNT_PASSWORD> -f schema.json -D
 ```
 
 ```console
 # For DynamoDB
 $ java -jar target/scalar-schema.jar --dynamo -u <AWS_ACCESS_KEY_ID> -p <AWS_ACCESS_SECRET_KEY> --region <REGION> -f schema.json -D
 ```
-  - For DynamoDB, only the tables which are included in the schema file will be deleted.
 
 ```console
 # For Cassandra
-$ java -jar target/scalar-schema-standalone-<version>.jar --cassandra -h <CASSANDRA_IP> [-P <CASSANDRA_PORT>] [-u <CASSNDRA_USER>] [-p <CASSANDRA_PASSWORD>] -D
+$ java -jar target/scalar-schema-standalone-<version>.jar --cassandra -h <CASSANDRA_IP> [-P <CASSANDRA_PORT>] [-u <CASSNDRA_USER>] [-p <CASSANDRA_PASSWORD>] -f schema.json -D
 ```
 
 ### Show help
--- a/tools/scalar-schema/README.md
+++ b/tools/scalar-schema/README.md
@@ -103,13 +103,13 @@ $ java -jar target/scalar-schema-standalone-<version>.jar --help
 - `compaction-strategy` should be `STCS`, `LCS` or `TWCS`. This is ignored in Cosmos DB and DynamoDB.
 - This `ru` value is set for all tables on this database even if `-r BASE_RESOURCE_UNIT` is set when Cosmos DB and DynamoDB. `ru` is ignored in Cassandra.
 
-## Performance tuning
-We can tune the throughput of Cosmos DB and DynamoDB by `-r` option or `ru` parameter for each table. For Cassandra, the tool cannot tune it since the performance depends on the resources of the node.
+## Scaling Performance
 
 ### RU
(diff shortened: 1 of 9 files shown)


feeblefakie (commented):
Sorry, I'm not really sure about this yet, but is it something configurable right?
Also, do we need to update the doc of schema tool?

yito88:
Thank you for the comment.
This change makes the tool enable auto-scaling of DynamoDB.
A user doesn't need to do anything for that.
I've updated the README.

It might be better to be able to choose to disable the auto-scaling.
What do you think?

yito88:
I'll modify to be able to choose to disable the auto-scaling.
For example, it's useful when benchmarking.
And, DynamoDB takes some time to scale in/out. Some users might want to receive unexpected requests suddenly.

feeblefakie:
@yito88 I think it's better if it can be disabled because some customers might care about constant cost due to a budget constraint.

tools/scalar-schema/README.md
```diff
@@ -38,21 +38,20 @@ $ java -jar target/scalar-schema-standalone-<version>.jar --cassandra -h <CASSAN
   - If `-p <CASSANDRA_PASSWORD>` is not supplied, it defaults to `cassandra`.
   - `<NETWORK_STRATEGY>` should be `SimpleStrategy` or `NetworkTopologyStrategy`
 
-### Delete all tables
+### Delete tables
 ```console
 # For Cosmos DB
-$ java -jar target/scalar-schema-standalone-<version>.jar --cosmos -h <YOUR_ACCOUNT_URI> -p <YOUR_ACCOUNT_PASSWORD> -D
+$ java -jar target/scalar-schema-standalone-<version>.jar --cosmos -h <YOUR_ACCOUNT_URI> -p <YOUR_ACCOUNT_PASSWORD> -f schema.json -D
```
yito88, line 44: Now we need the file to delete tables

tools/scalar-schema/sample_schema/sample_schema.json
```diff
@@ -13,7 +13,7 @@
       "c4": "INT",
       "c5": "BOOLEAN"
     },
-    "ru": 123,
+    "ru": 5000,
```
yito88, line 16: It's just for testing to create tables on Cosmos DB.

tools/scalar-schema/src/scalar_schema/cosmos.clj
```diff
@@ -85,14 +85,14 @@
...
+    (create-database client common/METADATA_DATABASE 400 true))
   (when-not (container-exists? client common/METADATA_DATABASE common/METADATA_TABLE)
     (create-container client common/METADATA_DATABASE common/METADATA_TABLE))
   (let [metadata (doto (TableMetadata.)
                    (.setId (common/get-fullname (:database schema)
                                                 (:table schema)))
-                   (.setPartitionKeyNames (:partition-key schema))
-                   (.setClusteringKeyNames (:clustering-key schema))
+                   (.setPartitionKeyNames (set (:partition-key schema)))
+                   (.setClusteringKeyNames (set (:clustering-key schema)))
```
yito88, line 95: Sorry for the mess. It is for fixing a bug for Cosmos DB.

tools/scalar-schema/README.md
```diff
@@ -103,3 +102,14 @@ $ java -jar target/scalar-schema-standalone-<version>.jar --help
 ```
 - `compaction-strategy` should be `STCS`, `LCS` or `TWCS`. This is ignored in Cosmos DB and DynamoDB.
 - This `ru` value is set for all tables on this database even if `-r BASE_RESOURCE_UNIT` is set when Cosmos DB and DynamoDB. `ru` is ignored in Cassandra.
+
+## Performance tuning
```
feeblefakie, line 106: I updated the descriptions based on the original ones. 
(First, I updated one by one but it got messy so I created a new doc)
It might be wrong so please check 🙇 
```
## Scaling Performance

### RU

You can scale the throughput of Cosmos DB and DynamoDB by specifying `-r` option (which applies to all the tables) or `ru` parameter for each table. Those configurations are ignored in Cassandra. The default values are `400` for Cosmos DB and `10` for DynamoDB respectively, which are set with `-r` option.

Note that the schema tool abstracts [Request Unit](https://docs.microsoft.com/azure/cosmos-db/request-units) of Cosmos DB and [Capacity Unit](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadWriteCapacityMode.html#HowItWorks.ProvisionedThroughput.Manual) of DynamoDB with `RU`.
So, please set an appropriate value depending on the database implementations. Please also note that the schema tool sets the same value to both Read Capacity Unit and Write Capacity Unit for DynamoDB.

### Auto-scaling

By default, the schema tool enables auto-scaling of RU for all tables:  RU is scaled in or out between 10% and 100% of a specified RU depending on a workload. For example, if you specify `-r 10000`, RU of each table is scaled in or out between 1000 and 10000.  Note that auto-scaling of Cosmos DB is enabled only when you set more than or equal to 4000 RU.
```

Questions.
1. The default values are specified with `-r` implicitly? (so, for example, `400` is applied to each table in Cosmos DB?)

feeblefakie (approved):
LGTM!
Sorry for late.
I was struggling with suggestions. I ended up suggesting a revised section.
It might be wrong so please check.

tools/scalar-schema/README.md
yito88: Thank you for the suggestion!
I've replaced it with your suggestion.

> The default values are specified with -r implicitly? (so, for example, 400 is applied to each table in Cosmos DB?)
The default values are set implicitly when you don't add `-r` option.

feeblefakie (approved):

PR #118, opened 2020-11-24
partial update for Cosmos DB

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7425

### Issue
- Non-updated columns are removed when partial updating
  - before updating

    | partition key | clustering key | c1 | c2 |
    --|--|--|--
    | 0 | 1 | "inserted" |  null |

  - after updating the only c2, c1's value is removed

    | partition key | clustering key | c1 | c2 |
    --|--|--|--
    | 0 | 1 | null |  "new" |

### Cause
- [Cosmos DB doesn't support partial update](https://feedback.azure.com/forums/263030-azure-cosmos-db/suggestions/6693091-be-able-to-do-partial-updates-on-document#%7Btoggle_previous_statuses%7D)

### Fix
- Use the stored procedure to read the existing record and update it
  - Insert a new record when the record doesn't exist

### DIFF
--- a/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosIntegrationTest.java
+++ b/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosIntegrationTest.java
@@ -340,34 +340,6 @@ public class CosmosIntegrationTest {
   }
 
   @Test
-  public void put_SinglePutGiven_ShouldStoreProperly() throws ExecutionException {
-    // Arrange
-    int pKey = 0;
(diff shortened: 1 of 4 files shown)


feeblefakie (approved):
LGTM!
I'm just wondering how it can be switched back to the current way once the upsert feature is released. Just reverting the PR might be good enough?

yito88:
@feeblefakie Yes, it's enough to revert this PR if partial updating would be supported.

PR #123, opened 2020-11-26
Add namespace prefix

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7606

Add the specified prefix to each namespace

This PR requires #124 to create tables with the prefix.

### DIFF
--- a/src/main/java/com/scalar/db/api/Operation.java
+++ b/src/main/java/com/scalar/db/api/Operation.java
@@ -19,6 +19,7 @@ import javax.annotation.concurrent.NotThreadSafe;
 public abstract class Operation {
   private final Key partitionKey;
   private final Optional<Key> clusteringKey;
+  private Optional<String> namespacePrefix;
   private Optional<String> namespace;
   private Optional<String> tableName;
   private Consistency consistency;
@@ -26,9 +27,20 @@ public abstract class Operation {
   public Operation(Key partitionKey, Key clusteringKey) {
     this.partitionKey = checkNotNull(partitionKey);
     this.clusteringKey = Optional.ofNullable(clusteringKey);
+    namespacePrefix = Optional.empty();
     namespace = Optional.empty();
     tableName = Optional.empty();
     consistency = Consistency.SEQUENTIAL;
+  }
+
+  /**
+   * Returns the namespace prefix for this operation
+   *
+   * @return an {@code Optional} with the returned namespace prefix
+   */
+  @Nonnull
+  public Optional<String> forNamespacePrefix() {
+    return namespacePrefix;
   }
 
   /**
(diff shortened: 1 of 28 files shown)


src/main/java/com/scalar/db/api/Operation.java
```diff
@@ -51,6 +79,45 @@ public Operation(Key partitionKey, Key clusteringKey) {
...
+   *
+   * @return an {@code Optional} with the returned the full table name
+   */
+  @Nonnull
+  public Optional<String> forFullTableName() {
+    if (!namespace.isPresent() || !tableName.isPresent()) {
+      return Optional.empty();
+    }
+
+    if (namespacePrefix.isPresent()) {
```
feeblefakie, line 93: I think it is a bit redundant. How about this?
```    
    StringBuilder builder = new StringBuilder();
    if (namespacePrefix.isPresent()) {
      builder.append(namespacePrefix.get()));
    }
    builder.append(namespace.get());
    builder.append(".");
    builder.append(tableName.get()); 
    return Optional.of(builder.toString());
```

src/main/java/com/scalar/db/api/Operation.java
```diff
@@ -41,6 +53,22 @@ public Operation(Key partitionKey, Key clusteringKey) {
...
   }
 
+  /**
+   * Returns the namespace with the prefix for this operation
+   *
+   * @return an {@code Optional} with the returned namespace with the prefix
+   */
+  @Nonnull
+  public Optional<String> forFullNamespace() {
+    if (namespace.isPresent() && namespacePrefix.isPresent()) {
```
feeblefakie, line 63: I think it should at least warn if this method is called when `namespace` is not present but `namespacePrefix` is present.
What do you think?

feeblefakie (commented):
Overall looking good! Left two suggestions.

src/main/java/com/scalar/db/api/Operation.java
yito88: Thank you! The suggestion looks better. Modifled.

src/main/java/com/scalar/db/api/Operation.java
yito88: I've added warning meesages when the `namespace` is not present.

src/main/java/com/scalar/db/api/Operation.java
```diff
@@ -41,6 +57,26 @@ public Operation(Key partitionKey, Key clusteringKey) {
...
 
+  /**
+   * Returns the namespace with the prefix for this operation
+   *
+   * @return an {@code Optional} with the returned namespace with the prefix
+   */
+  @Nonnull
+  public Optional<String> forFullNamespace() {
+    if (!namespace.isPresent()) {
+      LOGGER.warn("namespace isn't specified");
```
feeblefakie, line 68:
```suggestion
      LOGGER.warn("namespace is not set. So, you might get an unexpected return value.");
```

feeblefakie (dismissed):
LGTM! Thank you!

PR #124, opened 2020-11-30
Modify the schema tool to add prefix

yito88:
https://scalar-labs.atlassian.net/browse/DLT-7606

Modify the schema tool to add the specified prefix to each namespace with `--prefix` option

### DIFF
--- a/tools/scalar-schema/src/scalar_schema/core.clj
+++ b/tools/scalar-schema/src/scalar_schema/core.clj
@@ -22,6 +22,7 @@
     "The network topology strategy. SimpleStrategy or NetworkTopologyStrategy. This options is ignored when Cosmos DB and DynamoDB."]
    ["-D" "--delete-all" "All database will be deleted, if this is enabled."]
    [nil "--region REGION" "Region where the tool creates tables for DynamoDB"]
+   [nil "--prefix DATABASE_PREFIX" "Database prefix. The prefix will be added to all database."]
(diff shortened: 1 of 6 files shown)


tools/scalar-schema/src/scalar_schema/core.clj
```diff
@@ -22,6 +22,7 @@
     "The network topology strategy. SimpleStrategy or NetworkTopologyStrategy. This options is ignored when Cosmos DB and DynamoDB."]
    ["-D" "--delete-all" "All database will be deleted, if this is enabled."]
    [nil "--region REGION" "Region where the tool creates tables for DynamoDB"]
+   [nil "--prefix DATABASE_PREFIX" "Database prefix. The prefix will be added to all database."]
```
feeblefakie, line 25: I think it should be consistent with Scalar DB configuration?
```suggestion
   [nil "--prefix NAMESPACE_PREFIX" "Namespace prefix. The prefix is added to all the namespaces."]
```

feeblefakie (commented):
Overall looking good! Left 1 suggestion.

tools/scalar-schema/src/scalar_schema/core.clj
yito88: Thanks!

feeblefakie (approved):
LGTM! Thanks.

Maybe, we need to use `namespace` consistently in the program at some later point.

PR #133, opened 2020-12-18
Add JDBC adapter of Scalar DB 

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-7724
https://scalar-labs.atlassian.net/browse/DLT-7725


yito88 (commented):
Overall looks good! Thank you for the big implementation!

I left some comments. Also, I think we basically avoid `null` by using `Optional`. For example, we check if `clusteringKey` exists in an operation by `clusiteringKey.isPresent()`, not using `!= null`. Did you have a specific reason?

I'll review them again for the details.

src/main/java/com/scalar/db/storage/jdbc/JDBCService.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  public boolean put(
+      Put put, Connection connection, @Nullable String defaultSchema, @Nullable String defaultTable)
+      throws SQLException {
+    Table table = getTable(put, defaultSchema, defaultTable);
+    @Nullable Key clusteringKey = put.getClusteringKey().orElse(null);
+    @Nullable MutationCondition condition = put.getCondition().orElse(null);
+
+    operationChecker.checkPut(
```
feeblefakie, line 135: How to check seems checker's responsibility so I think it would be better it is determined and implemented by the check method. By doing this, the users of the checker don't need to care what to pass to check any more.

```suggestion
    operationChecker.check(table, delete);
```

src/main/java/com/scalar/db/storage/jdbc/JDBC.java
```diff
@@ -0,0 +1,187 @@
...
+
+  @Override
+  public Scanner scan(Scan scan) throws ExecutionException {
+    Connection connection = null;
+    try {
+      connection = dataSource.getConnection();
+      return jdbcService.scan(scan, connection, defaultSchema, defaultTable);
+    } catch (SQLException e) {
+      try {
+        if (connection != null) {
```
feeblefakie, line 109: If there is a chance of `null` for connection`,  I think it should be checked when created. 
https://github.com/scalar-labs/scalardb/pull/133/files#diff-269c4117791939509f752c12912fd795785d698c30c82d92b05a54bbc29c62f5R105

src/main/java/com/scalar/db/storage/jdbc/JDBCService.java
```diff
@@ -0,0 +1,213 @@
...
+        scan.getLimit(),
+        scan.getOrderings());
+
+    SelectQuery selectQuery =
+        queryBuilder
+            .select(scan.getProjections())
+            .from(table)
+            .where(
+                scan.getPartitionKey(),
+                startClusteringKey,
```
feeblefakie, line 105: I feel, passing optional as it is seems much simpler than getting nullable key and doing null check in the method.
https://github.com/scalar-labs/scalardb/pull/133/files#diff-8810b9d8331d1df08af5790eacab666ecde1ee1523a8f6f1db3e893987dc0ebfR73

Is there any particular reason of this?

src/main/java/com/scalar/db/storage/jdbc/query/UpdateQuery.java
```diff
@@ -0,0 +1,131 @@
...
+        @Nullable List<ConditionalExpression> otherConditions) {
+      this.partitionKey = partitionKey;
+      this.clusteringKey = clusteringKey;
+      this.otherConditions = otherConditions;
+      return this;
+    }
+
+    public UpdateQuery build() {
+      if (partitionKey == null || values == null) {
+        throw new IllegalStateException("partitionKey or values is null.");
```
feeblefakie, line 52: You meant `IllegalArgumentException`?

src/main/java/com/scalar/db/storage/jdbc/query/UpdateQuery.java
```diff
@@ -0,0 +1,131 @@
...
+    table = builder.table;
+    partitionKey = builder.partitionKey;
+    clusteringKey = builder.clusteringKey;
+    values = builder.values;
+    otherConditions = builder.otherConditions;
+  }
+
+  @Override
+  protected String sql() {
+    return "UPDATE " + table + " SET " + makeSetSQLString() + " WHERE " + makeConditionSQLString();
```
feeblefakie, line 74: Guave's Strings.Joiner might be useful here to handle spaces nicely. (Also, it's a minor thing but, I'm sure it is more efficient than simple string concatenation, which is supposed to be not efficient)
```suggestion
    return Joiner.join("UPDATE", table, "SET", makeSetSQLString(), "WHERE", makeConditionSQLString());
```
https://github.com/google/guava/wiki/StringsExplained#joiner

feeblefakie (changes requested):
Great step forward!!! Thank you!
I took a first glance and I left some comments and suggestions. Please take a look.
Since it is pretty big, I would like to take a few more looks.

brfrn169 (commented):
@yito88 Thank you for taking a look at this!

Just left some comments for your comments.

> Also, I think we basically avoid null by using Optional. For example, we check if clusteringKey exists in an operation by clusiteringKey.isPresent(), not using != null. Did you have a specific reason?

Yeah.. I think it's an annoying problem in Java. IntelliJ says as a warning that Optional should be used for a return type only. If following this rule, we can't use Optional as an argument of a method and a field of a class. So I deiced to use Null pattern instead of using Optional and use the Nullable annotation. But not sure it's good or not..

src/main/java/com/scalar/db/storage/jdbc/JDBCUtils.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  public static BasicDataSource initDataSource(DatabaseConfig config, boolean transactional) {
+    BasicDataSource dataSource = new BasicDataSource();
+    dataSource.setUrl(config.getContactPoints().get(0));
+    dataSource.setUsername(config.getUsername());
+    dataSource.setPassword(config.getPassword());
+
+    if (!transactional) {
+      dataSource.setDefaultAutoCommit(false);
```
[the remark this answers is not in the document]
brfrn169: This is a bug. Thank you for pointing it out!

src/main/java/com/scalar/db/storage/jdbc/query/AbstractQuery.java
```diff
@@ -0,0 +1,28 @@
...
+  public PreparedStatement prepareAndBind(Connection connection) throws SQLException {
+    PreparedStatement preparedStatement = prepare(connection);
+    bind(preparedStatement);
+    return preparedStatement;
+  }
+
+  protected abstract String sql();
+
+  protected PreparedStatement prepare(Connection connection) throws SQLException {
+    return connection.prepareStatement(sql());
```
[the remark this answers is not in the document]
brfrn169: As discussed online, we don't need to cache the PrepareStatement objects in JDBC unlike Cassandra.

brfrn169 (commented):

Just left some comments for your reviews. I will do some modifications for your and Ito-san's reviews and push a new commit. Thanks.

src/main/java/com/scalar/db/storage/jdbc/JDBC.java
feeblefakie: Sorry, I misunderstood the code. Based on the chat discussion, it's not needed to be handled.

feeblefakie:
@brfrn169 
Regarding the client-side caching for prepared statements, it is actually recommended (thus not done in JDBC) or done in JDBC driver (implicitly or explicitly) in most databases we support.

MySQL: https://vladmihalcea.com/how-does-the-mysql-jdbc-driver-handle-prepared-statements/
PostgreSQL: https://jdbc.postgresql.org/documentation/head/server-prepare.html
Oracle: https://docs.oracle.com/cd/E11882_01/java.112/e16548/stmtcach.htm

But, it seems to require a database-specific configuration or our own implementation to deal with it.
The simplest and database-agnostic solution might be just pooling multiple PreparedStatement Objects per specific statement and take it from the pool when executing it.
Probably it doesn't have to be implemented now, but we might need to do something for this in another PR.

brfrn169:
@feeblefakie Thank you for the investigation! I didn't know we need to cache the PreparedStatement objects at the client side honestly. And looks using the prepared-statement pooling feature in DBCP2 is a good idea. I think we can make it configurable in the JdbcDatabaseConfig in this JDBC adapter (and it looks like disabling it by default is okay as it's disabled by default in DBCP2).

src/main/java/com/scalar/db/storage/jdbc/JDBCService.java
brfrn169: As discussed, I will update the code to use Optional as a parameter and a class field.

src/main/java/com/scalar/db/storage/jdbc/ConditionalUpdater.java
```diff
@@ -0,0 +1,156 @@
...
+
+  private void throwSQLExceptionIfOccurred() throws SQLException {
+    if (sqlException != null) {
+      throw sqlException;
+    }
+  }
+
+  @Override
+  public void visit(PutIf condition) {
+    UpdateQuery updateQuery =
```
feeblefakie, line 74: I think if all the `visit`just  keep  `AbstractQuery` instead of preparing and excuting,
all the redundant `prepareAndBind`, `executeUpdate`, and the exception handling can be unified except `PutIfNotExists` case.

src/main/java/com/scalar/db/storage/jdbc/JDBC.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  @Override
+  public Optional<String> getTable() {
+    return Optional.ofNullable(defaultTable);
+  }
+
+  @Override
+  public Optional<Result> get(Get get) throws ExecutionException {
+    try (Connection connection = dataSource.getConnection()) {
```
feeblefakie, line 94: I think the following part is a bit redundant and is shown several times, so it would be great if it can be unified.

```
    try (Connection connection = dataSource.getConnection()) {
       // ...
    } catch (SQLException e) {
      throw new ExecutionException("An error occurred", e);
    }
```

I have't fully verified it, but how about having a method like this?

```
  @Override
  public Optional<Result> get(Get get) throws ExecutionException {
    ThrowableFunction<Get, Optional<Result>> f =
        o -> jdbcService.get(o, connection, defaultSchema, defaultTable);
    processOperation(f);
  }

 private void processOperation(ThrowableFunction f) {
    try (Connection connection = dataSource.getConnection()) {
      return f.apply();
    } catch (SQLException e) {
      throw new ExecutionException("An error occurred", e);
    }
  }

// it can be separately defined.
@FunctionalInterface
public interface ThrowableFunction<T, R> {
  R apply(T t) throws Exception;
}
```

src/main/java/com/scalar/db/storage/jdbc/JDBCService.java
```diff
@@ -0,0 +1,213 @@
...
+      List<? extends Mutation> mutations,
+      Connection connection,
+      @Nullable String defaultSchema,
+      @Nullable String defaultTable)
+      throws SQLException {
+    operationChecker.checkMutate(mutations);
+
+    for (Mutation mutation : mutations) {
+      if (mutation instanceof Put) {
+        if (!put((Put) mutation, connection, defaultSchema, defaultTable)) {
```
feeblefakie, line 202: I think `put` and `delete` methods can use the same method name like `mutate` and use overloading so that this part can be simplified without `instanceof` and casting.

feeblefakie (commented):
Left additional suggestions.

src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -0,0 +1,188 @@
...
+  public Optional<String> getTable() {
+    return tableName;
+  }
+
+  @Override
+  public Optional<Result> get(Get get) throws ExecutionException {
+    try (Connection connection = dataSource.getConnection()) {
+      return jdbcService.get(get, connection, namespace, tableName);
+    } catch (SQLException e) {
+      throw new ExecutionException("An error occurred", e);
```
feeblefakie, line 98: I think the message should be a bit more specific because `An error occurred` is obvious and not very informational.
I left a similar comment on the JdbcTransaction code, so please take a look!

src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -0,0 +1,188 @@
...
+        connection = dataSource.getConnection();
+        connection.setAutoCommit(false);
+        if (!jdbcService.mutate(mutations, connection, namespace, tableName)) {
+          connection.rollback();
+          throw new NoMutationException("no mutation was applied");
+        } else {
+          connection.commit();
+        }
+      } catch (SQLException e) {
+        if (connection != null) {
```
feeblefakie, line 166: IMO, instead of having nested `try`, it would be better to `try` smaller range and throw with a specific message. What do you think?

```
      try {
        if (connection != null) {
          connection.rollback();
        }
      } catch (SQLException e) {
        throw new ExecutionException("rollback failed.", e);
      }
```

src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -0,0 +1,188 @@
...
+          connection.rollback();
+        }
+        throw new ExecutionException("An error occurred", e);
+      } finally {
+        if (connection != null) {
```
feeblefakie, line 171: The same for this.
BTW, should it throw an exception if `close` throws an exception?
If the processing is succeeded (except for closing the connection), it might be misleading to throw an exception because a user would assume the processing is failed. 
In such a case, giving a warning message might be the way it is.

src/main/java/com/scalar/db/storage/jdbc/ConditionalPutExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+        queryBuilder
+            .insertInto(put.forFullTableName().get())
+            .values(put.getPartitionKey(), put.getClusteringKey(), put.getValues())
+            .build();
+    try (PreparedStatement preparedStatement = insertQuery.prepareAndBind(connection)) {
+      preparedStatement.executeUpdate();
+      result = true;
+    } catch (SQLException e) {
+      if (e.getSQLState().equals("23000") || e.getSQLState().equals("23505")) {
+        // The duplicate key error
```
feeblefakie, line 101: In this case, should set `result = true`?
If so, I feel we might need a test case for that.

feeblefakie (commented):
Sorry for leaving comments gradually, I left several comments, questions, and suggestions.
PTAL when you get a chance.

src/main/java/com/scalar/db/storage/jdbc/ConditionalPutExecutor.java
feeblefakie: This is my misunderstanding. Sorry!

brfrn169: It looks like the name `result` is a bit confusing. I will change it. Thanks!

src/main/java/com/scalar/db/storage/jdbc/query/InsertQuery.java
```diff
@@ -0,0 +1,93 @@
...
+
+public class InsertQuery extends AbstractQuery {
+
+  public static class Builder {
+    private final String fullTableName;
+    private Key partitionKey;
+    private Optional<Key> clusteringKey;
+    private Map<String, Value> values;
+
+    Builder(String fullTableName) {
```
feeblefakie, line 22: Just a question. 
Some `XXXQuery` accepts `fullTableName` in its builder's constructor, but some accept it with its builder's method (like SelectQuery).
Do you have any particular reason for that?

src/main/java/com/scalar/db/storage/jdbc/query/InsertQuery.java
```diff
@@ -0,0 +1,93 @@
...
+      this.values = values;
+      return this;
+    }
+
+    public InsertQuery build() {
+      return new InsertQuery(this);
+    }
+  }
+
+  private final String fullTableName;
```
feeblefakie, line 39: It might be just a style thing (not sure if google's style guide defines something for this style), but is it common to instance variables in the middle if some static inner class comes above?

feeblefakie (commented):
Overall looking good!
Left some relatively minor comments.

src/main/java/com/scalar/db/storage/jdbc/query/InsertQuery.java
brfrn169: The reason is that only `SelectQuery` accepts projections as a first parameter while other `XXXQuery` accepts a full table name as a first parameter. 

For example:
```
// SelectQuery
queryBuilder.select(<projections>)
  .from(<the full table name>)
  .where(...)
  .build()

// InsertQuery
queryBuilder.insertInto(<a full table name>)
  .values(...)
  .build()

// UpdateQuery
queryBuilder.update(<a full table name>)
  .set(...)
  .where(...)
  .build()

// DeleteQuery
queryBuilder.deleteFrom(<a full table name>)
  .where(...)
  .build()

// UpsertQuery
queryBuilder.upsertInto(<a full table name>)
  .values(...)
  .build()
```

As you can see, only `SelectQuery` doesn't accept a full table name as a first parameter, but the others accept a full table name as a fist parameter.

src/main/java/com/scalar/db/storage/jdbc/query/InsertQuery.java
brfrn169: Not sure it's common or not actually, but my policy was that static fields and methods were placed first in a class. Let me check google's style guide.

feeblefakie: As we discussed, it would be great if we can follow [this](https://github.com/scalar-labs/scalar/blob/master/common/src/main/java/com/scalar/dl/ledger/asset/AssetProof.java) for consistency.

feeblefakie (approved):
LGTM! 👍 👍 👍 
Left 1 minor suggestion but it does not block it from being merged.

yito88 (dismissed):
LGTM!

src/main/java/com/scalar/db/service/TransactionModule.java
```diff
@@ -20,8 +20,21 @@ public TransactionModule(DatabaseConfig config) {
 
   @Override
   protected void configure() {
-    bind(DistributedTransactionManager.class).to(ConsensusCommitManager.class).in(Singleton.class);
     bind(DistributedStorage.class).to(config.getStorageClass()).in(Singleton.class);
+
+    if (config.getStorageClass() == JdbcDatabase.class) {
```
feeblefakie, line 25: For more readability, how about having a method like this?

```
  ...
  if (useJdbcTransaction) {
    bind(DistributedTransactionManager.class).to(JdbcTransactionManager.class).in(Singleton.class);
  }
  ...

private boolean useJdbcTransaction() {
  if (config.getStorageClass() == JdbcDatabase.class) {
    JdbcDatabaseConfig jdbcDatabaseConfig = provideJdbcDatabaseConfig();
    if (jdbcDatabaseConfig
        .getTransactionManagerType()
        .equals(JdbcDatabaseConfig.TRANSACTION_MANAGER_TYPE_JDBC)) {
      return true;
    }
  }
  return false;
}
```

feeblefakie (commented):
Thank you for the updates. 
Left some comments on the commits after the first approval.
I think we can create a new branch out of this branch and create a PR (being merged to add-jdbc-adapter branch) to for ease of review.

src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -175,11 +176,17 @@ public void commit() throws CommitException, UnknownTransactionStatusException {
       } catch (SQLException e) {
         LOGGER.warn("failed to close the connection", e);
       }
+      isCommitCalled = true;
     }
   }
 
   @Override
   public void abort() throws AbortException {
+    if (isCommitCalled) {
```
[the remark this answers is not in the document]
brfrn169: When I checked the document of DBCP,  I thought `connection.isClosed` didn't work. But I rechecked the source code and ran a simple test code, and it looks like it works. So I will modify it to use `connection.isClosed`. Thanks.

src/main/java/com/scalar/db/storage/jdbc/query/QueryUtils.java
```diff
@@ -24,4 +25,21 @@ public static String getOperatorString(ConditionalExpression.Operator operator)
         throw new AssertionError("invalid operator: " + operator);
     }
   }
+
+  public static String enclose(String name, RdbEngine rdbEngine) {
```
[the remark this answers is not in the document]
brfrn169: As discussed, we will keep to use `enclose` for this.

feeblefakie (approved):
LGTM! Thank you!

PR #134, opened 2020-12-23
Some scanners in CrudHandler and Snapshot are not closed

brfrn169:
While adding the JDBC adapter, I found some scanners in CrudHandler and Snapshot are not closed. We should close them.


feeblefakie (approved):
LGTM! Thank you!

yito88 (approved):
LGTM!

PR #136, opened 2021-01-18
Check operations if they are given with correct keys

feeblefakie:
https://scalar-labs.atlassian.net/browse/DLT-3333

There is no check in the current C* adapter if given operations have proper keys or not.
So, I fixed it in this PR as well since it is pretty related to the bug.


yito88 (approved):
LGTM!

brfrn169 (approved):
LGTM!

PR #138, opened 2021-01-19
Fix Snapshot to handle deleteSet properly when a delete operation is mixed with other operations.

feeblefakie:
https://scalar-labs.atlassian.net/browse/DLT-7979

### DIFF
--- a/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -93,20 +93,24 @@ public class Snapshot {
   }
 
   public void put(Snapshot.Key key, Put put) {
+    deleteSet.remove(key);
     writeSet.put(key, put);
   }
 
   public void put(Snapshot.Key key, Delete delete) {
+    writeSet.remove(key);
     deleteSet.put(key, delete);
   }
 
   public boolean containsKey(Snapshot.Key key) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -93,20 +93,24 @@ public void put(Scan scan, Optional<List<Key>> keys) {
...
   public boolean containsKey(Snapshot.Key key) {
-    return writeSet.containsKey(key) || readSet.containsKey(key);
+    return writeSet.containsKey(key) || deleteSet.containsKey(key) || readSet.containsKey(key);
   }
 
   public Optional<TransactionResult> get(Snapshot.Key key) {
     if (writeSet.containsKey(key)) {
       throw new CrudRuntimeException("reading already written data is not allowed");
+    } else if (deleteSet.containsKey(key)) {
+      return Optional.empty();
```
brfrn169, line 113: Why we allow reading already deleted data while we don't allow reading already written data?

brfrn169: Ah okay. I understood. Thank you for the clear explanation! I'll approve this.

brfrn169 (approved):

feeblefakie:
@Yonezawa-T2 I'm mering it but PTAL when you get a chance since I might miss something.

PR #140, opened 2021-01-21
[Scalar-Schema] Set Cassandra secondary index

yito88:
https://scalar-labs.atlassian.net/browse/DLT-8021

This PR is for Cassandra

For Cosmos DB: #141 
For DynamoDB: #142

PR #141, opened 2021-01-21
[Scalar-Schema] Set Cosmos secondary index

yito88:
https://scalar-labs.atlassian.net/browse/DLT-8021

This PR is for Cosmos DB

For Cassandra: #140 
For DynamoDB: #142 

Please merge after #140

### DIFF
--- a/tools/scalar-schema/java/src/CosmosTableMetadata.java
+++ b/tools/scalar-schema/java/src/CosmosTableMetadata.java
@@ -17,6 +17,7 @@ public class CosmosTableMetadata {
   private String id;
   private SortedSet<String> partitionKeyNames;
   private SortedSet<String> clusteringKeyNames;
+  private SortedSet<String> secondaryIndexNames;
   private SortedMap<String, String> columns;
   private List<String> keyNames;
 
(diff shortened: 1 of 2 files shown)


feeblefakie (approved):
LGTM!

tools/scalar-schema/java/src/CosmosTableMetadata.java
```diff
@@ -17,6 +17,7 @@
   private String id;
```
brfrn169, line 17: Just out of my curiosity. Why do we have CosmosTableMetadata in this schema tool project?

brfrn169 (approved):
Left one question, but it's just out of my curiosity. I'll approve this.

tools/scalar-schema/java/src/CosmosTableMetadata.java
yito88: The class is used for metadata insertion.
Cosmos DB SDK serializes/deserializes the data according to this class's setter and getter.

This tool gives the instance to the Upsert API.
https://github.com/scalar-labs/scalardb/blob/0f9cd909fd94ac04e46575860d0110ba6d57dab7/tools/scalar-schema/src/scalar_schema/cosmos.clj#L96

PR #149, opened 2021-01-25
Fix pip install failure in CI

feeblefakie:
It came to fail suddenly. By explicitly specifying the version of `get-pip`, it seems working fine.


yito88 (approved):
LGTM

Issue #150, opened 2021-01-25
Scan after put doesn't return the written value, nor raise an error for EXTRA_READ


Yonezawa-T2:
**Describe the bug**

With `EXTRA_READ` strategy, scanning values after putting values doesn't return the written value, nor results in an error.

**To Reproduce**

Initial state: empty

1. Begin a transaction.
1. Read a value at `key05` (returns empty).
2. Put a value `1` at `key05`.
3. Scan from `key01` to `key10`.
4. Commit the transaction.

**Expected behavior**

The scan returns a value `1`, or the commit results in an error.

**Actual behavior**

The scan returns empty result and the commit succeeds.

**Additional context**

Linux and Cassandra 3 on Docker.

**Possible fixes**

- Checking the write set and the delete set when scanning a range and returning the written value. Since the write set contains partial records, the written value must be merged into scanned value, if exists.
- Raising an error on commit or scan if the write set is not empty, like `EXTRA_WRITE` strategy, or if keys in the write set overlap the scanned range.

feeblefakie:
@Yonezawa-T2 Thank you!
I updated the code to make it throw an exception if a scan range overlaps with a write-set.

feeblefakie:
Fixed in #151.

PR #151, opened 2021-02-04
Fix snapshot scan bug

feeblefakie:
https://github.com/scalar-labs/scalardb/issues/150

### DIFF
--- a/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -118,6 +118,9 @@ public class Snapshot {
   }
 
   public Optional<List<Key>> get(Scan scan) {
+    if (isWriteSetOverlappedWith(scan)) {
+      throw new CrudRuntimeException("reading already written data is not allowed");
+    }
     if (scanSet.containsKey(scan)) {
       return scanSet.get(scan);
     }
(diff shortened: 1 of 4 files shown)


Yonezawa-T2 (approved):
LGTM

yito88 (approved):
LGTM!

src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -145,6 +148,61 @@ public void to(MutationComposer composer) throws CommitConflictException {
             });
   }
 
+  private boolean isWriteSetOverlappedWith(Scan scan) {
```
brfrn169, line 151: I was wondering if we need to consider `limit` and `projections` of the scan operation here. But, I found we clear `projections` at the following line, so it looks like we don't need to consider `projections`:
https://github.com/scalar-labs/scalardb/blob/master/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java#L92

I have two questions:
- Why do we clear `projections` in scan/get operations in `ConsensusCommit`?
- Do we need to consider `limit` in `isWriteSetOverlappedWith`?

Thanks.

feeblefakie: Thank you for the questions!

> Why do we clear projections in scan/get operations in ConsensusCommit?

This is because ConsensusCommit stores its metadata in the columns of a record and needs to force applications to get all the columns to properly handle transactions.
Even if it doesn't use such metadata columns, as long as conflicts are detected on the basis of records (not columns), projections shouldn't be taken into consideration.

> Do we need to consider limit in isWriteSetOverlappedWith?

I don't think so.
First, `limit` cannot be utilized here because there is no way to identify what records would be returned by the limit.
Even if we know what records would be returned by the `limit`, 
the same issue still exists if the records and write set don't overlap.
Essentially, conflict detection for scan should be done with a specified range (like predicate locking) instead of returned records otherwise it would cause phantom.

But, I might have missed your points, so please feel free to ask more questions and concerns.

brfrn169: Thank you!

brfrn169 (approved):
LGTM! Thank you!

PR #153, opened 2021-02-10
Divide the integration test into Cassandra and JDBC

brfrn169:
After this change, we can run the integration tests for Cassandra and JDBC as follows:

Cassandra:
```
gradle integrationTestCassandra
```

JDBC (MySQL):
```
gradle integrationTestJdbc -Dscalardb.jdbc.url=jdbc:mysql://localhost:3306/ -Dscalardb.jdbc.username=root -Dscalardb.jdbc.password=mysql
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -95,7 +95,6 @@ task integrationTestCassandra(type: Test) {
     testClassesDirs = sourceSets.integrationTestCassandra.output.classesDirs
     classpath = sourceSets.integrationTestCassandra.runtimeClasspath
     outputs.upToDateWhen { false }  // ensures integration tests are run every time when called
-    shouldRunAfter test
 }
 
 task integrationTestJdbc(type: Test) {
@@ -104,7 +103,6 @@ task integrationTestJdbc(type: Test) {
     testClassesDirs = sourceSets.integrationTestJdbc.output.classesDirs
     classpath = sourceSets.integrationTestJdbc.runtimeClasspath
     outputs.upToDateWhen { false }  // ensures integration tests are run every time when called
-    shouldRunAfter test
     options {
         systemProperties(System.getProperties())
     }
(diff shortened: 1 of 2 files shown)


yito88 (approved):
LGTM! Thanks 👍

build.gradle
```diff
@@ -9,19 +9,38 @@ repositories {
 }
 
 sourceSets {
-    integrationTest {
+    integrationTestCassandra {
         java {
             compileClasspath += main.output + test.output
             runtimeClasspath += main.output + test.output
             srcDir file('src/integration-test/java')
+            exclude '**/com/scalar/db/storage/cosmos/*.java'
```
feeblefakie, line 17: I'm sure you have tried it but I'm wondering if there is any way to just include what to test or exclude except what to test. Let's chat.

feeblefakie (commented):
Overall looking good! Left a question.

build.gradle
brfrn169: As discussed, we can use `include` here which is better than `exclude`. I will push a new commit for it.

build.gradle
```diff
@@ -71,21 +89,30 @@ task testJar(type: Jar) {
...
     shouldRunAfter test
+}
 
+task integrationTestJdbc(type: Test) {
+    description = 'Runs the integration tests for a JDBC database.'
+    group = 'verification'
+    testClassesDirs = sourceSets.integrationTestJdbc.output.classesDirs
+    classpath = sourceSets.integrationTestJdbc.runtimeClasspath
+    outputs.upToDateWhen { false }  // ensures integration tests are run every time when called
+    shouldRunAfter test
```
feeblefakie, line 107: Just wondering if this is needed. (since it is only executed on-demand basis, it's not needed?)

feeblefakie (approved):
LGMT! Thank you!
I just left one question but it's not blocking it from being merged.

PR #154, opened 2021-02-12
Fix to set RU for DynamoDB

yito88:
https://scalar-labs.atlassian.net/browse/DLT-8333

When auto-scaling is enabled for DynamoDB, the RU of each table wasn't applied.


feeblefakie (approved):
LGTM!

PR #159, opened 2021-02-19
Enable dynamo backup

yito88:
https://scalar-labs.atlassian.net/browse/DLT-8207

Enabled the continuous backup (for Point Time In Recovery)

Note:
This doesn't enable the backup of the `scalardb.metadata` table.
The restored metadata would be useless because the user table name should be changed.
(DynamoDB cannot restore data to the existing table.)
So, we have to remake the metadata table with this tool with the new prefix after restoring other tables.

### DIFF
--- a/tools/scalar-schema/src/scalar_schema/core.clj
+++ b/tools/scalar-schema/src/scalar_schema/core.clj
@@ -24,6 +24,7 @@
    [nil "--region REGION" "Region where the tool creates tables for DynamoDB"]
    [nil "--prefix NAMESPACE_PREFIX" "Namespace prefix. The prefix is added to all the namespaces."]
    [nil "--no-scaling" "Disable auto-scaling"]
+   [nil "--no-backup" "Disable continuous backup"]
    [nil "--help"]])
 
 (defn -main [& args]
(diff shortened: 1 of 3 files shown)


brfrn169 (approved):
LGTM!

tools/scalar-schema/src/scalar_schema/core.clj
```diff
@@ -24,6 +24,7 @@
    [nil "--region REGION" "Region where the tool creates tables for DynamoDB"]
    [nil "--prefix NAMESPACE_PREFIX" "Namespace prefix. The prefix is added to all the namespaces."]
    [nil "--no-scaling" "Disable auto-scaling"]
+   [nil "--no-backup" "Disable continuous backup"]
```
feeblefakie, line 27: Just a question. This option can only be specified for DynamoDB at the moment and would be ignored if specified for other databases?

feeblefakie (approved):
LGTM! 
Left a clarifying question.

tools/scalar-schema/src/scalar_schema/core.clj
yito88: @feeblefakie Thank you for the question. Right, it will be ignored for other databases.

PR #160, opened 2021-02-19
 Label correction for Cosmos DB credential

scalar-boney:
Minor label correction for Cosmos DB credentials for better understanding

Ref:- https://scalar-labs.atlassian.net/browse/DLT-8371

### DIFF
--- a/docs/getting-started-with-scalardb-on-cosmosdb.md
+++ b/docs/getting-started-with-scalardb-on-cosmosdb.md
@@ -44,10 +44,11 @@ scalar.db.contact_points=<COSMOS_DB_ACCOUNT_URI>
 
 # Credential information to access the database
 scalar.db.username=
-scalar.db.password=<COSMOS_DB_PRIMARY_KEY>
+scalar.db.password=<COSMOS_DB_KEY>
 
 # Storage implementation. Either cassandra or cosmos can be set. Default storage is cassandra.
 scalar.db.storage=cosmos
 ```
+Note that you can use a primary key or a secondary key for `<COSMOS_DB_KEY>`.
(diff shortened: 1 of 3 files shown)


docs/getting-started-with-scalardb-on-cosmosdb.md
```diff
@@ -37,14 +37,14 @@ The **scalardb.properties** (getting-started/scalardb.properties) file holds the
...
-scalar.db.contact_points=<YOUR_ACCOUNT_URI>
+scalar.db.contact_points=<COSMOS_DB_ACCOUNT_URI>
 
 # Port number for all the contact points. Default port number for each database is used if empty.
 #scalar.db.contact_port=
 
 # Credential information to access the database
 scalar.db.username=
-scalar.db.password=<YOUR_ACCOUNT_PASSWORD>
+scalar.db.password=<COSMOS_DB_PRIMARY_KEY>
```
yito88, line 47: It might be too specific because not only the primary key but also the secondary key can be used.

feeblefakie: I think Yuji's suggestion is reasonable, so why don't you apply it and leave something like this as a note before `Please follow ...` sentence.

```
Note that you can use a primary key or a secondary key for `<COSMOS_DB_KEY>`.
```

scalar-boney: @yito88 @feeblefakie san,
Thank you for your suggestions.
Document updated accordingly.

feeblefakie (approved):
LGTM! Thanks.

PR #161, opened 2021-02-22
Remove prefix from indexes of DynamoDB

yito88:
https://scalar-labs.atlassian.net/browse/DLT-8207

For DyanmoDB recovery, the prefix for index tables is removed.

Scalar DB requires the same name about index tables after restoring the table. However, the name of the restored tables should be different from that of the source table to restore data. This requirement can be satisfied by adding the prefix. Also, the name of an index table couldn't be modified after restoring the table on DyanmoDB.
That's why I don't want to add a prefix to the index.

### DIFF
--- a/src/main/java/com/scalar/db/api/Operation.java
+++ b/src/main/java/com/scalar/db/api/Operation.java
@@ -94,13 +94,28 @@ public abstract class Operation {
    */
   @Nonnull
   public Optional<String> forFullTableName() {
+    return forFullTableName(true);
+  }
+
+  /**
+   * Returns the full table name without the prefix with the namespace for this operation
+   *
+   * @return an {@code Optional} with the returned the full table name
+   */
+  @Nonnull
+  public Optional<String> forFullTableNameWithoutPrefix() {
+    return forFullTableName(false);
+  }
+
+  @Nonnull
+  private Optional<String> forFullTableName(boolean withPrefix) {
     if (!namespace.isPresent() || !tableName.isPresent()) {
       LOGGER.warn("namespace or table name isn't specified");
       return Optional.empty();
     }
 
     StringBuilder builder = new StringBuilder();
-    if (namespacePrefix.isPresent()) {
+    if (withPrefix && namespacePrefix.isPresent()) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/scalar/db/api/Operation.java
```diff
@@ -94,13 +94,28 @@ public Operation(Key partitionKey, Key clusteringKey) {
...
+    return forFullTableName(true);
+  }
+
+  /**
+   * Returns the full table name without the prefix with the namespace for this operation
+   *
+   * @return an {@code Optional} with the returned the full table name
+   */
+  @Nonnull
+  public Optional<String> forFullTableNameWithoutPrefix() {
```
feeblefakie, line 106:
```suggestion
  public Optional<String> forUnprefixedFullTableName() {
```
This sounds a bit better to me.

src/main/java/com/scalar/db/api/Operation.java
```diff
@@ -94,13 +94,28 @@ public Operation(Key partitionKey, Key clusteringKey) {
    */
   @Nonnull
   public Optional<String> forFullTableName() {
+    return forFullTableName(true);
+  }
+
+  /**
+   * Returns the full table name without the prefix with the namespace for this operation
```
feeblefakie, line 101:
```suggestion
   * Returns the full table name without the unprefixed namespace for this operation
```

feeblefakie (commented):
Looking good! Thank you!
Left minor naming suggestion.

brfrn169:
@yito88 So an index name is a table-local name? I mean, there is no conflict with the same index names between the original table and the restored table?

src/main/java/com/scalar/db/api/Operation.java
yito88: Thanks! Fixed.

yito88:
@brfrn169 Good question! An index table is bound to the table. There is no conflict with that of the source table.

brfrn169 (approved):
@yito88 Thank you! LGTM!

feeblefakie (approved):
LGTM! Thank you!

PR #165, opened 2021-02-24
Fix not to return unexpected results

feeblefakie:
Since DynamoDB seems not able to scan with an exclusive range like `a < x < b`,  I updated it to throw an exception instead of giving the warning message.


yito88 (approved):
LGTM! Thank you

brfrn169 (approved):
LGTM!

PR #169, opened 2021-03-03
Unify the operation check logic of the adapters

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-8128


src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
```diff
@@ -0,0 +1,281 @@
...
+  private static void checkClusteringKeys(Scan scan, TableMetadata metadata) {
+    scan.getStartClusteringKey()
+        .ifPresent(startClusteringKey -> checkStartClusteringKey(scan, metadata));
+
+    scan.getEndClusteringKey().ifPresent(endClusteringKey -> checkEndClusteringKey(scan, metadata));
+
+    if (scan.getStartClusteringKey().isPresent() && scan.getEndClusteringKey().isPresent()) {
+      Key startClusteringKey = scan.getStartClusteringKey().get();
+      Key endClusteringKey = scan.getEndClusteringKey().get();
+      Supplier<String> message =
```
feeblefakie, line 94: Just a question.
Is there any benefit of doing this instead of just doing `String message = " ..."`?

src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -100,15 +101,17 @@ private String orderBySqlString() {
     List<Scan.Ordering> orderingList = new ArrayList<>(orderings);
 
     Boolean reverse = null;
-    for (int i = 0; i < tableMetadata.getClusteringKeys().size(); i++) {
+    Iterator<String> iterator = tableMetadata.getClusteringKeyNames().iterator();
```
feeblefakie, line 104: Just a question.
Is there any particular reason for using `iterator` instead of `for`?

src/test/java/com/scalar/db/storage/jdbc/JdbcServiceTest.java
```diff
@@ -65,10 +69,19 @@
...
 
+  private MockedStatic<OperationChecker> mockedStatic;
+
   @Before
   public void setUp() throws Exception {
     MockitoAnnotations.initMocks(this);
-    jdbcService = new JdbcService(operationChecker, queryBuilder, Optional.empty());
+    jdbcService = new JdbcService(tableMetadataManager, queryBuilder, Optional.empty());
+
+    mockedStatic = Mockito.mockStatic(OperationChecker.class);
```
feeblefakie, line 79: Oh, Mockito now supports mocking for static methods!

feeblefakie (commented):
Overall looking good! Thank you for the big refactoring.
Since it's a big PR, let me start off with some questions and check a few times more.

src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
brfrn169: Here, I wanted to do a lazy evaluation for the plus (+) operation in building the message. If just doing `String message = "..."`, the plus operation is executed even if `message` is not used with IllegalArgumentException. That's why I use `Supplier` here.

src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
brfrn169: I'm not sure why I did it... Maybe I was tired... I will fix it with `for`. Thanks!

src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
feeblefakie: Good to know! Thanks!

src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
```diff
@@ -0,0 +1,281 @@
...
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+public final class OperationChecker {
+  private OperationChecker() {}
+
+  public static void check(Get get, TableMetadata metadata) {
```
feeblefakie, line 26: Is there any particular reason for making the methods static?
I think it passes around `TableMetadata` to a pretty deep level so it would be a bit simpler to make it have a constructor and manage an instance variable. What do you think?
(also it's a bit more consistent with other checkers like ColumnChecker and ConditionChecker)

src/main/java/com/scalar/db/storage/cassandra/CassandraTableMetadata.java
```diff
@@ -28,11 +37,57 @@ public CassandraTableMetadata(com.datastax.driver.core.TableMetadata tableMetada
...
                 .map(IndexMetadata::getTarget)
                 .collect(Collectors.toSet()));
+    columnDataTypes =
+        ImmutableMap.copyOf(
+            tableMetadata.getColumns().stream()
+                .collect(
+                    Collectors.toMap(
+                        ColumnMetadata::getName, c -> convertDataType(c.getType().getName()))));
+
+    Map<String, Scan.Ordering.Order> tmp = new HashMap<>();
```
feeblefakie, line 52:
```suggestion
    Map<String, Scan.Ordering.Order> tmpClusteringOrder = new HashMap<>();
```

feeblefakie: It seems hard to avoid creating `tmp`, right?
I've thought about it but couldn't get a better solution.

feeblefakie (commented):
Left minor suggestions and comments.

src/main/java/com/scalar/db/storage/cassandra/CassandraTableMetadata.java
brfrn169: We can use `ImmutableMap.builder()` here, but I think it's essentially the same. I will refactor it a bit to make it more readable.

src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
brfrn169: Honestly, I don't have any big reason for that. I made the methods static because the original check methods in the `Utility` class were static. Okay, let's make it have a constructor and manage an instance variable.

feeblefakie (approved):
LGTM! Thank you!

PR #170, opened 2021-03-05
Specify a driver class corresponding to the JDBC URL to the dataSource

brfrn169:
We need to specify a driver class corresponding to the specified JDBC URL to the dataSource in order to avoid the "No suitable driver" error when ServiceLoader in java.sql.DriverManager doesn't work (ex. we dynamically load a driver class from a fatJar).

I'm not sure this is the best solution, but I think it's a reasonable one. If you have any better idea for it, please let me know.

### DIFF
--- a/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
+++ b/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
@@ -34,7 +34,7 @@ public final class JdbcUtils {
     BasicDataSource dataSource = new BasicDataSource();
 
     /*
-     * We need to specify a driver class corresponding to the JDBC URL to the dataSource in order
+     * We need to set the driver class of an underlining database to the dataSource in order
      * to avoid the "No suitable driver" error when ServiceLoader in java.sql.DriverManager doesn't
      * work (e.g., when we dynamically load a driver class from a fatJar).
      */
--- a/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
+++ b/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
@@ -1,8 +1,12 @@
 package com.scalar.db.storage.jdbc;
 
+import com.microsoft.sqlserver.jdbc.SQLServerDriver;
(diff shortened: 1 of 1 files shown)


src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
```diff
@@ -26,8 +30,17 @@ public static BasicDataSource initDataSource(JdbcDatabaseConfig config) {
...
 
   public static BasicDataSource initDataSource(JdbcDatabaseConfig config, boolean transactional) {
+    String jdbcUrl = config.getContactPoints().get(0);
     BasicDataSource dataSource = new BasicDataSource();
-    dataSource.setUrl(config.getContactPoints().get(0));
+
+    /*
+     * We need to specify a driver class corresponding to the JDBC URL to the dataSource in order
+     * to avoid the "No suitable driver" error when ServiceLoader in java.sql.DriverManager doesn't
+     * work (ex. we dynamically load a driver class from a fatJar).
```
feeblefakie, line 39:
```suggestion
     * work (e.g., when we dynamically load a driver class from a fatJar).
```

src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
```diff
@@ -26,8 +30,17 @@ public static BasicDataSource initDataSource(JdbcDatabaseConfig config) {
   }
 
   public static BasicDataSource initDataSource(JdbcDatabaseConfig config, boolean transactional) {
+    String jdbcUrl = config.getContactPoints().get(0);
     BasicDataSource dataSource = new BasicDataSource();
-    dataSource.setUrl(config.getContactPoints().get(0));
+
+    /*
+     * We need to specify a driver class corresponding to the JDBC URL to the dataSource in order
```
feeblefakie, line 37:
```suggestion
     * We need to set the driver class of an underlining database to the dataSource in order
```

feeblefakie (dismissed):
LGTM! Thank you!
(Left minor English suggestions)

feeblefakie (approved):
LGTM! Thank you!

PR #173, opened 2021-03-10
Wrong index name when a namespace prefix is not specified in the sche…

brfrn169:
…ma tool for DynamoDB

If a namespace prefix is not specified, the original schema tool for DynamoDB removes the first `_` from the table name that's in part of the index name.

For example, let's see the table name is `ns.table_1_a` and the indexed column name is `col1`. In this case, the index name will wrongly be `ns.table1_a.global_index.col1`, the correct one is `ns.table_1_a.global_index.col1` though.


feeblefakie (dismissed):
LGTM!

tools/scalar-schema/src/scalar_schema/dynamo/common.clj
```diff
@@ -18,8 +18,9 @@
 
 (defn- get-table-name-without-prefix
   [{:keys [prefix] :as schema}]
-  (str/replace-first (get-table-name schema)
-                     (java.util.regex.Pattern/compile (str prefix \_)) ""))
+  (let [table-name (get-table-name schema)]
+    (if prefix (str/replace-first table-name (java.util.regex.Pattern/compile (str prefix \_)) "")
+        table-name)))
```
yito88, line 23: I think it can be more readable.
It is better to avoid making lines longer than 80 characters according to the [clojure-style-guide](https://github.com/bbatsov/clojure-style-guide).
```suggestion
    (if prefix
      (str/replace-first table-name
                         (java.util.regex.Pattern/compile (str prefix \_)) "")
      table-name)))
```

yito88 (approved):
Thank you! Looks good.
It is not a blocking comment.

feeblefakie (approved):

PR #174, opened 2021-03-12
Add a new storage implementation for multi-storage

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-8328
https://scalar-labs.atlassian.net/browse/DLT-8452


src/main/java/com/scalar/db/storage/multistorage/MultiStorageDatabaseConfig.java
```diff
@@ -0,0 +1,117 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Map;
+import java.util.Properties;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class MultiStorageDatabaseConfig {
```
feeblefakie, line 15:
```suggestion
public class MultiStorageConfig {
```
MultiStorage should accept MultiStorageConfig since JdbcDatabase accepts JdbcDatabaseConfig?

src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
```diff
@@ -0,0 +1,153 @@
...
+
+  private final Map<String, DistributedStorage> storageMap;
+  private final DistributedStorage defaultStorage;
+
+  private Optional<String> namespace;
+  private Optional<String> tableName;
+
+  @Inject
+  public MultiStorage(MultiStorageDatabaseConfig config) {
+    Map<String, DistributedStorage> databaseStorageMap = new HashMap<>();
```
feeblefakie, line 44: I think `databaseStorageMap` is a bit confusing and not necessarily correct since it's a map of <alias-of-storage, storage-object>?
So should it be something like `aliasStorageMap` or `nameStorageMap`?

.circleci/config.yml
```diff
@@ -269,3 +270,66 @@ jobs:
...
+      - run:
+          name: Save Gradle integration test reports
+          command: |
+            mkdir -p /tmp/gradle_integration_test_reports
+            cp -a build/reports/tests/integrationTestMultiStorage /tmp/gradle_integration_test_reports/
+          when: always
+
+      - store_artifacts:
+          path: /tmp/gradle_integration_test_reports
+          destination: gradle_integration_test_reports
```
feeblefakie, line 335: I think a return code should be added just in case.

feeblefakie (commented):
Overall looking great!!!
Left some minor comments mostly about naming.
(Sorry, I'm picky about names)

src/main/java/com/scalar/db/storage/multistorage/MultiStorageConfig.java
```diff
@@ -0,0 +1,116 @@
...
+    load();
+  }
+
+  public Properties getProperties() {
+    return props;
+  }
+
+  private void load() {
+    String storage = props.getProperty(DatabaseConfig.STORAGE);
+    if (storage == null || !storage.equals("multi-storage")) {
```
feeblefakie, line 49: I think it would be probably better to make "multi-storage" a constant since it's used in several places.

src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import javax.annotation.concurrent.ThreadSafe;
+import javax.inject.Inject;
+
+/**
+ * A storage implementation with multi-storage for {@link DistributedStorage}.
+ *
+ * <p>This storage implementation holds multiple storage instances and have mapping a table name to
```
feeblefakie, line 31:
```suggestion
 * <p>This storage implementation holds multiple storage instances. It chooses a storage instance on the basis of the specified configuration and a given operation. 
```

feeblefakie: I feel it's OK but it can be extended by giving examples if it's too blur.
Also, the following lines can be deleted if the above suggestion is accepted.

feeblefakie (approved):
LGTM! Thank you!
(but left minor suggestions.)

feeblefakie (approved):
LGTM! Thank you!

PR #175, opened 2021-03-14
Fix the broken TLA+ link

y-taka-23:
ReadTheDocs doesn't resolve links that are outside of the `docs` dir correctly.


feeblefakie (approved):
LGTM! Thank you!

PR #176, opened 2021-03-22
Extend the ConsensusCommit integration tests to cover multi-table transactions

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-8704

Currently, the ConsensusCommit integration tests are performed only for single table transactions. From the test coverage perspective, we should perform the integration tests for multi-table transactions. In this story, we will extend the existing integration tests to cover multi-table transactions.


feeblefakie (approved):
LGTM! It's a great enhancement!

Issue #86, opened 2020-10-02
Add missing table name to this exception message `StorageRuntimeException: no table information found`


Torch3333:
**Is your feature request related to a problem? Please describe.**
Let's state that this is a program executing a Put request targeting a Cassandra table named `FooTable` using the **transaction** mode.

When Scalar DB fails to execute the `Put` because `FooTable` was not created in Cassandra, the stack trace will look like this.
```
no table information found
com.scalar.db.exception.storage.StorageRuntimeException: no table information found
	at com.scalar.db.storage.cassandra.ClusterManager.getMetadata(ClusterManager.java:83)
	at com.scalar.db.storage.cassandra.Cassandra.getTableMetadata(Cassandra.java:185)
	at com.scalar.db.storage.cassandra.Cassandra.get(Cassandra.java:79)
	at com.scalar.db.transaction.consensuscommit.RollbackMutationComposer.getLatestResult(RollbackMutationComposer.java:145)
	at com.scalar.db.transaction.consensuscommit.RollbackMutationComposer.add(RollbackMutationComposer.java:60)
	at com.scalar.db.transaction.consensuscommit.Snapshot.lambda$to$0(Snapshot.java:126)
	at java.util.concurrent.ConcurrentHashMap$EntrySetView.forEach(ConcurrentHashMap.java:4795)
	at com.scalar.db.transaction.consensuscommit.Snapshot.to(Snapshot.java:122)
	at com.scalar.db.transaction.consensuscommit.RecoveryHandler.rollback(RecoveryHandler.java:58)
	at com.scalar.db.transaction.consensuscommit.CommitHandler.commit(CommitHandler.java:44)
	at com.scalar.db.transaction.consensuscommit.ConsensusCommit.commit(ConsensusCommit.java:121)
        at com.scalar.ist.api.ScalarDBTest.put(ScalarDBTest.java:71)
...
```
Since the stracktrace is originating from the faulty user code, I tend to think that `FooTable` is missing which is correct in that case.

But, in the case the `FooTable` was created but the  `coordinator.state` table is missing. The exception will look like this :
```
no table information found
com.scalar.db.exception.storage.StorageRuntimeException: no table information found
	at com.scalar.db.storage.cassandra.ClusterManager.getMetadata(ClusterManager.java:83)
	at com.scalar.db.storage.cassandra.Cassandra.getTableMetadata(Cassandra.java:185)
[...]

PR #178, opened 2021-03-29
Unify the behavior between the adapters when a specified table in a o…

brfrn169:
…peration is not found

https://scalar-labs.atlassian.net/browse/DLT-8742

Currently, the behavior when a specified table in an operation is not found is different between the adapters in Scalar DB. We should unify it.

This will fix the following, as well:
https://github.com/scalar-labs/scalardb/issues/86


feeblefakie (approved):
LGTM! Thank you!

PR #184, opened 2021-04-13
Unify the behavior with -D (deleting all tables) in Schema Tool

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-8878

Currently, Schema Tool for JDBC doesn’t drop schemas and the metadata table/schema, which is different behavior from the others. We should unify it in this PR.


feeblefakie (approved):
LGTM!  
But, I feel this PR includes some unrelated changes (e.g., naming and format changes).
If it's easy to do, can you separate them into 2 PRs (one for changes for -D and the other for refactoring)?

feeblefakie (approved):
LGTM! Thank you!

PR #191, opened 2021-04-16
Use "FETCH FIRST n ROWS ONLY" instead of ROWNUM for LIMIT for Oracle

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-8778


feeblefakie (approved):
LGTM! Thank you!

PR #192, opened 2021-04-19
Add a validation that checks not to use BOOLEAN type for a clustering…

brfrn169:
… key or a secondary index in Schema Tool for DynamoDB


feeblefakie (approved):
LGTM! Thank you!

PR #196, opened 2021-05-06
Fix warnings of raw use of parameterized class for the Value class

brfrn169:
To fix warnings of raw use of parameterized class for the Value class, I added wildcards to the declarations of the Value class.


feeblefakie (approved):
LGTM! Thank you!

PR #198, opened 2021-05-13
Add utility methods for Value/Put and a builder class for Key

brfrn169:
I added utility methods for Value/Put and a builder class for Key, which I believe will make Scalar DB easy to use.

### DIFF
--- a/src/main/java/com/scalar/db/api/Put.java
+++ b/src/main/java/com/scalar/db/api/Put.java
@@ -2,7 +2,14 @@ package com.scalar.db.api;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableMap;
+import com.scalar.db.io.BigIntValue;
+import com.scalar.db.io.BlobValue;
+import com.scalar.db.io.BooleanValue;
+import com.scalar.db.io.DoubleValue;
+import com.scalar.db.io.FloatValue;
+import com.scalar.db.io.IntValue;
 import com.scalar.db.io.Key;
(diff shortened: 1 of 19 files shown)


src/main/java/com/scalar/db/api/Put.java
```diff
@@ -51,6 +58,90 @@ public Put withValue(Value<?> value) {
     return this;
   }
 
+  /**
+   * Adds the specified BigInt value to the list of put values.
```
feeblefakie, line 62: Just a minor question.
It seems these `withValue` methods are ordered by the types (?)
Is it common to do that?
(I just thought I would do with boolean -> int -> bigint -> float -> double -> text -> bytes or something, but I don't know that order is. )

feeblefakie (approved):
LGTM! Great improvement. Thank you.
Left one very minor question.

src/main/java/com/scalar/db/api/Put.java
brfrn169: Thank you for reviewing this!

Actually, I wasn't concerned about the method order very much. And it looks like the Google style plugin doesn't reorder the methods. Maybe I chose the alphabetical order (BigInt -> Blob -> Boolean -> Double -> Float -> Int -> Text). Should I reorder the methods?

And if we need to care about the method order, do we need to add this rule to our code rule? Or Google style already has this rule?

feeblefakie (approved):
LGTM! Thank you!

PR #199, opened 2021-05-17
Refactor the code around metadata

brfrn169:
I wanted to add a `DistributedStorageAdmin` interface that gets table metadata and creates/drops tables. But, before that, we need to do some refactoring for the code around metadata. This PR is for this.

In this change, I removed the existing `TableMetadata` interface and added the new `TableMetadata` class with the builder class. Because of this, we can remove `CassandraTableMetadata` and `DynamoTableMetadata` and `JdbcTableMetadata` (`CosmosTableMetadata` is still needed though).

After it's merged, I want to add `DistributedStorageAdmin` like the following:
```Java
public interface DistributedStorageAdmin {
  TableMetadata getTableMetadata(String namespace, String table);
  void createTable(String namespace, String table, TableMetadata metadata, Map<String, String> options);
  void dropTable(String namespace, String table);
}
```

Please take a look at this!


src/main/java/com/scalar/db/storage/cosmos/CosmosTableMetadata.java
```diff
@@ -1,34 +1,24 @@
...
 import java.util.Set;
-import java.util.stream.Collectors;
 
 /**
  * A metadata class for a table of Scalar DB to know the type of each column
  *
  * @author Yuji Ito
  */
-public class CosmosTableMetadata implements TableMetadata {
+public class CosmosTableMetadata {
```
feeblefakie, line 16: Just a question.  Why this cannot be unified to `TableMetadata`?

src/main/java/com/scalar/db/api/TableMetadata.java
```diff
@@ -0,0 +1,172 @@
...
+import com.scalar.db.storage.common.util.ImmutableLinkedHashSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+/** Represents the table metadata */
```
feeblefakie, line 16: It's not in the google style guide, but I think it would be a bit better and consistent if a comment for a class starts with a noun.

feeblefakie (commented):
Overall looking good! Thank you.
Left 1 question and 1 minor comment.

src/main/java/com/scalar/db/storage/cosmos/CosmosTableMetadata.java
brfrn169: Thank you for reviewing this!

We still need `CosmosTableMetadata` because we need POJO (that means a class with some fields and their getters and setters maybe) to serialize and deserialize data in Cosmos DB.

For example, the following code writes the metadata into Cosmos DB with `CosmosTableMetadata`:
https://github.com/scalar-labs/scalardb/blob/2ff0ff36163158fd17627d476fcca7bd9c4bbba8/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosIntegrationTest.java#L85-L100

And the following code retrieves the metadata from Cosmos DB with `CosmosTableMetadata`:
https://github.com/scalar-labs/scalardb/blob/2ff0ff36163158fd17627d476fcca7bd9c4bbba8/src/main/java/com/scalar/db/storage/cosmos/CosmosTableMetadataManager.java#L49-L51

As the new `TableMetadata` is not POJO, we can't use it for that.

feeblefakie: Thank you! I see.
Since `CosmosTableMetadata` is only used in `CosmosTableMetadataManager`, can it be an inner class of the manager class?

brfrn169: `CosmosTableMetadata` is used only in `CosmosTableMetadataManager ` in the main code, but it's actually used in the integration test code as well. And we might use this class another part when we introduce `DistributedStorageAdmin`. So I want to keep it an outer class for now. If we realize it doesn't need to be an outer class in the future, we can do that at that time. Thanks.

feeblefakie: OK, I understand.

feeblefakie (approved):
LGTM! Thank you!

feeblefakie:
@yito88 I merged it but since it's a bit big PR so please take a look when you get a chance.

PR #200, opened 2021-05-17
Use com.google.inject.Inject instead of javax.inject.Inject

brfrn169:
I found that we are using the `javax.inject.Inject` annotations in some parts. In this change, I fixed it to use the `com.google.inject.Inject` annotations.


feeblefakie (approved):
LGTM!
Just a question. Was it working?

brfrn169:
Yes. Looks like It was working.

PR #202, opened 2021-05-19
Refactor the code around ScannerIterator

brfrn169:
- Throw NoSuchElementException when ScannerIterator doesn't have next element in Cassandra, Cosmos DB and DynamoDB
- Cache an iterator object in Scanner

### DIFF
--- a/src/main/java/com/scalar/db/storage/cassandra/ScannerImpl.java
+++ b/src/main/java/com/scalar/db/storage/cassandra/ScannerImpl.java
@@ -18,6 +18,8 @@ public final class ScannerImpl implements Scanner {
   private final ResultSet resultSet;
   private final CassandraTableMetadata metadata;
 
+  private ScannerIterator scannerIterator;
+
   public ScannerImpl(ResultSet resultSet, CassandraTableMetadata metadata) {
     this.resultSet = checkNotNull(resultSet);
     this.metadata = checkNotNull(metadata);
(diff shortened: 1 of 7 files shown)


src/main/java/com/scalar/db/storage/cassandra/ScannerImpl.java
```diff
@@ -43,7 +45,10 @@ public ScannerImpl(ResultSet resultSet, CassandraTableMetadata metadata) {
 
   @Override
   public Iterator<Result> iterator() {
-    return new ScannerIterator(resultSet, metadata);
+    if (scannerIterator == null) {
+      scannerIterator = new ScannerIterator(resultSet, metadata);
```
feeblefakie, line 49: I'm not fully sure but is it OK to cache an iterator?
If a user gets an iterator with `iterator()` and call `next()` several times and the call `iterator()` again,
doesn't the user expect a new iterator that hasn't moved the cursor?

feeblefakie (commented):
Thank you!
I'm not sure if it is OK to cache an iterator so left a question.

src/main/java/com/scalar/db/storage/cassandra/ScannerImpl.java
brfrn169: I think users expect that `iterator()` always returns an iterator whose cursor is reset to the first row as the general collections like ArrayList do. And currently, it looks like the behaviors between the adapters are different. Let's discuss it tomorrow.

brfrn169:
I will create another PR for these changes. Thanks.

PR #204, opened 2021-05-24
Introduce DistributedStorageAdmin that is an administrative interface…

brfrn169:
… for distributed storage implementations

I added DistributedStorageAdmin that is an administrative interface for distributed storage implementations. createTable/dropTable/truncateTable have not been implemented yet for all the adapters. We can implement them in another PR.

Please take a look!


src/main/java/com/scalar/db/service/AdminModule.java
```diff
@@ -0,0 +1,40 @@
+package com.scalar.db.service;
```
feeblefakie, line 1: I'm wondering if we need another module for Admin that is identical to StorageModule.
Is there any problem updating StorageModule just by adding the line?
```
bind(DistributedStorageAdmin.class).to(config.getAdminClass()).in(Singleton.class);
```

feeblefakie (commented):
Overall looking good! Nice improvement!
I left one question.

src/main/java/com/scalar/db/service/AdminModule.java
brfrn169: Thank you for reviewing this!

> I'm wondering if we need another module for Admin that is identical to StorageModule.

Yes, it's better!  I will fix it in that way. Thanks.

feeblefakie (approved):
LGTM!

feeblefakie:
@yito88 PTAL when you get a chance!

PR #205, opened 2021-05-27
Remove MultiPartitionException

brfrn169:
I don't think we need `MultiPartitionException`. We can use `IllegalArgumentException` instead.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #206, opened 2021-05-29
Add some more useful methods to Key.Builder

brfrn169:
I added two methods (add() and addAll()) to the `Key.Builder` class. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #207, opened 2021-05-29
Support namespace mappings in MultiStorage

brfrn169:
Currently, we only have a table mapping support in MultiStorage as follows:
```
# Define table mappings from table names to storages. The format is “<table name>:<storage name>,...”
scalar.db.multi_storage.table_mapping=user.ORDER:cassandra,user.CUSTOMER:mysql,coordinator.state:cassandra
```

This PR adds a namespace mapping support where we can define namespace mappings as follows:
```
# Define namespace mappings from namespace names to storages. The format is “<namespace name>:<storage name>,...”
scalar.db.multi_storage.namespace_mapping=user:cassandra,coordinator:mysql
```

Please take a look!

### DIFF
--- a/src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
+++ b/src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
@@ -29,7 +29,7 @@ import javax.annotation.concurrent.ThreadSafe;
  * A storage implementation with multi-storage for {@link DistributedStorage}.
  *
  * <p>This storage implementation holds multiple storage instances. It chooses a storage instance on
- * the basis of the specified configuration and a given operation. If conflicting mapping between a
+ * the basis of the specified configuration and a given operation. If there is a conflict between a
  * table mapping and a namespace mapping, it prefers the table mapping because table mappings are
  * more specific than namespace mappings.
  *
--- a/src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
+++ b/src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
@@ -29,7 +29,9 @@ import javax.annotation.concurrent.ThreadSafe;
  * A storage implementation with multi-storage for {@link DistributedStorage}.
  *
  * <p>This storage implementation holds multiple storage instances. It chooses a storage instance on
- * the basis of the specified configuration and a given operation.
+ * the basis of the specified configuration and a given operation. If conflicting mapping between a
+ * table mapping and a namespace mapping, it prefers the table mapping because table mappings are
+ * more specific than namespace mappings.
  *
  * @author Toshihiro Suzuki
  */
--- a/src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
+++ b/src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
@@ -2,8 +2,7 @@ package com.scalar.db.storage.multistorage;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableMap.Builder;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Guice;
 import com.google.inject.Inject;
 import com.google.inject.Injector;
@@ -19,6 +18,7 @@ import com.scalar.db.api.Scanner;
 import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.service.StorageModule;
 import com.scalar.db.storage.common.util.Utility;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


src/main/java/com/scalar/db/storage/multistorage/MultiStorageConfig.java
```diff
@@ -88,17 +91,34 @@ private void loadTableStorageMapping() {
...
 
     String tableMapping = props.getProperty(TABLE_MAPPING);
-    for (String tableAndStorage : tableMapping.split(",")) {
-      String[] s = tableAndStorage.split(":");
-      String table = s[0];
-      String storage = s[1];
-
-      checkIfStorageExists(storage);
-      builder.put(table, storage);
+    if (tableMapping != null) {
```
feeblefakie, line 94: It is a minor style thing, but I feel the following is slightly better since less indenting.
```
if (tableMapping == null) {
  return null;
}
for (String tableAndStorage : tableMapping.split(",")) {
  // ...
}
```

src/main/java/com/scalar/db/storage/multistorage/MultiStorageConfig.java
```diff
@@ -88,17 +91,34 @@ private void loadTableStorageMapping() {
...
+      }
     }
     tableStorageMap = builder.build();
   }
 
+  private void loadNamespaceStorageMapping() {
+    Builder<String, String> builder = ImmutableMap.builder();
+
+    String namespaceMapping = props.getProperty(NAMESPACE_MAPPING);
+    if (namespaceMapping != null) {
```
feeblefakie, line 110: Ditto.

src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
```diff
@@ -141,16 +164,18 @@ public void mutate(List<? extends Mutation> mutations) throws ExecutionException
...
-    DistributedStorage storage = storageMap.get(fullTaleName);
-    if (storage == null) {
-      return defaultStorage;
+    DistributedStorage storage = tableStorageMap.get(fullTaleName);
+    if (storage != null) {
+      return storage;
     }
-    return storage;
+    String namespace = operation.forNamespace().get();
+    storage = namespaceStorageMap.get(namespace);
```
feeblefakie, line 172: So, if conflicting mapping like below is given, it prefers namespace mapping, right? Or we shouldn't use both mapping format together?
```
scalar.db.multi_storage.table_mapping=user.ORDER:cassandra,user.CUSTOMER:mysql,coordinator.state:cassandra
scalar.db.multi_storage.namespace_mapping=user:cassandra
```

Anyway, I think it would be great if we note it somewhere.

feeblefakie (commented):
Left a minor question and comments.

src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
brfrn169: Thank you for reviewing this!

If conflicting mapping between a table mapping and a namespace mapping, it prefers the table mapping because table mappings are more specific than namespace mappings.

Sure, I will add it to the Javadoc. Thanks.

feeblefakie: Ah, sorry, I meant table mapping is preferred. Thank you!

src/main/java/com/scalar/db/storage/multistorage/MultiStorage.java
```diff
@@ -29,43 +29,68 @@
  * A storage implementation with multi-storage for {@link DistributedStorage}.
  *
  * <p>This storage implementation holds multiple storage instances. It chooses a storage instance on
- * the basis of the specified configuration and a given operation.
+ * the basis of the specified configuration and a given operation. If conflicting mapping between a
+ * table mapping and a namespace mapping, it prefers the table mapping because table mappings are
```
feeblefakie, line 33:
```suggestion
 * the basis of the specified configuration and a given operation. If there is a conflict between a
 * table mapping and a namespace mapping, it prefers the table mapping because table mappings are
```

feeblefakie: We might need to reformat after the change.

src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
```diff
@@ -1,42 +1,76 @@
...
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import javax.annotation.concurrent.ThreadSafe;
 
+/**
+ * An implementation with multi-storage for {@link DistributedStorageAdmin}.
+ *
+ * <p>This implementation holds multiple DistributedStorageAdmin instances. It chooses a instance on
+ * the basis of the specified configuration and a given operation. If conflicting between a table
```
feeblefakie, line 20:
```suggestion
 * the basis of the specified configuration and a given operation. If there is a conflict between a table
```

feeblefakie (approved):
LGTM! Thanks!
I left minor suggestions on the English comments.

feeblefakie (approved):
LGMT!

PR #208, opened 2021-05-30
Should close PreparedStatement and ResultSet in scan operations of Jd…

brfrn169:
…bcTransaction

I found we don't close PreparedStatement and ResultSet in scan operations of JdbcTransaction. This PR will fix this issue.

Please take a look!

### DIFF
--- a/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
+++ b/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
@@ -18,6 +18,7 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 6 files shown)


src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
```diff
@@ -108,6 +109,40 @@ public Scanner scan(
         resultSet);
   }
 
+  public List<Result> scan(
+      Scan scan, Connection connection, Optional<String> namespace, Optional<String> tableName)
+      throws SQLException {
+    Utility.setTargetToIfNot(scan, namespacePrefix, namespace, tableName);
```
feeblefakie, line 115: I think `getScanner` and `scan` are pretty redundant, so should `scan` use `getScanner` (maybe it's not possible)  or should both delegate the common procedures to the same method?

feeblefakie (commented):
Looking good! Left 1 clarifying question.

src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
brfrn169: Thank you for reviewing this! 

Sure, I will try that.

feeblefakie (approved):
LGTM!

PR #209, opened 2021-05-30
Support the endpoint override feature for Dynamo

brfrn169:
Supporting the endpoint override feature for Dynamo is useful for tests.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #210, opened 2021-06-01
Add gPRC layer for DistributedStorage and DistributedStorageAdmin

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-9013

Please take a look!


src/main/java/com/scalar/db/server/DistributedStorageService.java
```diff
@@ -0,0 +1,294 @@
...
+  private static void execute(Callable<Void> callable, StreamObserver<?> responseObserver) {
+    try {
+      callable.call();
+    } catch (IllegalArgumentException | IllegalStateException e) {
+      LOGGER.error("an invalid argument error happened during the execution", e);
+      responseObserver.onError(
+          Status.INVALID_ARGUMENT.withDescription(e.getMessage()).asRuntimeException());
+    } catch (NoMutationException e) {
+      responseObserver.onError(
+          Status.FAILED_PRECONDITION.withDescription(e.getMessage()).asRuntimeException());
```
feeblefakie, line 262: For storage, it might be good enough to use gRPC status but I think we will need scalardb-specific error status for transaction. So, I'm wondering if we should have scalardb-specific error statuses for storage for consistency.
Maybe, it's something that is worth discussing.

src/main/java/com/scalar/db/api/PutIf.java
```diff
@@ -26,7 +26,18 @@
   public PutIf(ConditionalExpression... expressions) {
     checkNotNull(expressions);
     this.expressions = new ArrayList<>(expressions.length);
-    Arrays.stream(expressions).forEach(v -> this.expressions.add(v));
+    this.expressions.addAll(Arrays.asList(expressions));
```
feeblefakie, line 29: This change is something we can do in a small PR and merge it before this?

src/main/java/com/scalar/db/storage/rpc/GrpcStorage.java
```diff
@@ -0,0 +1,207 @@
+package com.scalar.db.storage.rpc;
```
feeblefakie, line 1: This is more like a client SDK?
I'm wondering if it shouldn't be under `storage`.

feeblefakie (commented):
Great work! Overall looking good at first glance.
I left some comments and questions that came up with so far, bu let me check it a few more times.

src/main/java/com/scalar/db/storage/rpc/GrpcStorage.java
brfrn169: As discussed, we will keep it for now.

src/main/java/com/scalar/db/api/PutIf.java
brfrn169: I have created a PR for it:
https://github.com/scalar-labs/scalardb/pull/211

Please take a look! After it merged, I will push a commit to revert those changes.

feeblefakie (approved):

feeblefakie:
@yito88 PTAL when you get a chance!

PR #213, opened 2021-06-04
Some refactoring for JdbcTransaction

brfrn169:
What I did here is as follows:
- Add a pseudo transaction ID `txId` to `JdbcTransaction`
- Make `JdbcTransactionManager` always return `UNKNOWN`

Please take a look!


feeblefakie (approved):
LGTM! 
But, I'm not sure why this is needed. It's for grpc server?

brfrn169:
Thank you for reviewing and merging this!

Yes, it's for the grpc server. I'm thinking about using the `txId` in the grpc server. That's why I needed this change.

PR #215, opened 2021-06-09
Refactor the integration tests for transaction

brfrn169:
This is a very minor change. Please take a look!

### DIFF
--- a/src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTestBase.java
+++ b/src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTestBase.java
@@ -13,7 +13,6 @@ import static org.mockito.Mockito.verify;
 import com.scalar.db.api.Consistency;
 import com.scalar.db.api.Delete;
 import com.scalar.db.api.DistributedStorage;
-import com.scalar.db.api.DistributedTransaction;
 import com.scalar.db.api.Get;
 import com.scalar.db.api.Isolation;
 import com.scalar.db.api.Put;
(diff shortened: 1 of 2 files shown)


src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTestBase.java
```diff
@@ -841,14 +837,14 @@ public void get_GetGivenForDeletedWhenCoordinatorStateNotExistAndExpired_ShouldA
   public void getAndScan_CommitHappenedInBetween_ShouldReadRepeatably()
       throws CrudException, CommitException, UnknownTransactionStatusException {
     // Arrange
-    DistributedTransaction transaction = manager.start();
+    ConsensusCommit transaction = manager.start();
```
feeblefakie, line 840: Just a question. Is there any reason of using  `ConsensusCommit` instead of `DistributedTransaction`?
(usually we should use `DistributedTransaction`, right?)

feeblefakie (commented):
Thank you!
The code looks good but I don't know why we use `ConsensusCommit` and `JdbcTransaction` instead of `DistributedTransaction`.

src/integration-test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTestBase.java
brfrn169: Thank you for reviewing this!

The reason of those changes is that the integration tests are actually just for `ConsensusCommit`, but not for general `DistributedTransaction` (I found that when I tried to use it for gRPC transaction but I couldn't). To highlight that, I made those changes, but they are not very important. And yes, usually we should use DistributedTransaction.

feeblefakie (approved):
LGTM! (from mobile)

PR #216, opened 2021-06-10
Add gPRC layer for DistributedTransaction

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-9013

Please take a look!


src/main/java/com/scalar/db/server/DistributedTransactionService.java
```diff
@@ -0,0 +1,292 @@
...
+      LOGGER.error("an invalid argument error happened during the execution", e);
+      responseObserver.onError(
+          Status.INVALID_ARGUMENT.withDescription(e.getMessage()).asRuntimeException());
+    } catch (CrudConflictException | CommitConflictException e) {
+      responseObserver.onError(
+          Status.FAILED_PRECONDITION.withDescription(e.getMessage()).asRuntimeException());
+    } catch (UnknownTransactionStatusException e) {
+      LOGGER.error("an unknown transaction error happened during the execution", e);
+      responseObserver.onError(Status.UNKNOWN.withDescription(e.getMessage()).asRuntimeException());
+    } catch (Throwable e) {
```
feeblefakie, line 256: I'm wondering Errors should be caught here.
I've never caught Errors but should it rather stop the server or something if an Error occurs?

src/main/java/com/scalar/db/server/DistributedTransactionService.java
```diff
@@ -0,0 +1,292 @@
...
+                  .build());
+          responseObserver.onCompleted();
+        },
+        responseObserver);
+  }
+
+  private DistributedTransaction getTransaction(String transactionId) throws TransactionException {
+    ActiveTransaction activeTransaction = activeTransactions.get(transactionId);
+    if (activeTransaction == null) {
+      throw new TransactionException(
```
feeblefakie, line 236: A question and a comment.
I think this case happens mostly in a case where different connections are used in a transaction, which only happens when a connection is closed for some reason after the transaction is started.
First of all, is this right?

If so, should the exception be treated separately?
I feel it's not an `INTERNAL` error.
What do you think?

feeblefakie (commented):
Overall looking good! Great work!
Left some questions and comments. PTAL.

src/main/java/com/scalar/db/server/DistributedTransactionService.java
brfrn169: I think this case could happen due to the following reasons:
1. A client specifies a wrong transaction ID
2. A client specifies a transaction ID that associates with a transaction that's already closed 
3. A transaction associated with the specified transaction ID is already expired by the `transactionExpirationThread`

An `INVALID_ARGUMENT` error suits the 1st and 2nd cases, but I think they are rare cases. And an `INTERNAL` error suits the 3rd case, and I think it most likely happens. So I decided to throw an `INTERNAL` error for this case. What do you think?

src/main/java/com/scalar/db/server/DistributedTransactionService.java
brfrn169: I catch `Throwable` here is because whenever any `Throwable` is thrown, I thought we should call `responseObserver.onError`, otherwise the client will get stuck. And for OutOfMemoryError, we have the `-XX:OnOutOfMemoryError` JVM flag, and we can kill the server with it. What do you think?

src/main/java/com/scalar/db/server/DistributedTransactionService.java
```diff
@@ -0,0 +1,292 @@
...
+    execute(
+        () -> {
+          DistributedTransaction transaction = getTransaction(request.getTransactionId());
+
+          List<Mutation> mutations = new ArrayList<>(request.getMutationsCount());
+          for (com.scalar.db.rpc.Mutation mutation : request.getMutationsList()) {
+            mutations.add(ProtoUtil.toMutation(mutation));
+          }
+
+          if (mutations.size() == 1) {
```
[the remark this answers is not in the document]
brfrn169: Yes, I thought the overhead of creating an unnecessary list is negligible, too. My concern here is that we are not sure the performance (or overhead) difference between the single put/delete and the multiple put/delete/mutate, which depends on the storage implementation. For example, one storage implementation has an optimization for single put/delete, and the single put/delete might be faster than the multiple put/delete/mutate when we execute one mutation. That's why I have this special case for one mutation. What do you think?

brfrn169 (commented):
@feeblefakie Thank you for reviewing this!

I left some responses for your reviews. Please check them. Thanks.

src/main/java/com/scalar/db/server/DistributedTransactionService.java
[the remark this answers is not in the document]
feeblefakie: From the perspective of separation of concerns, I think an upper layer shouldn't care about the implementation or the optimization of the underlining layer if it's passing the same information, and the underlining layer should do proper optimization based on the given information.
So, in this case, the underlining layer should check if a given list contains only single item if it has an optimization for single item operation. (but the current code might not be handling such things properly and should probably be fixed)
Thought? (maybe we should discuss?)

feeblefakie (commented):
Thank you for the replies!
I left comments so PTAL when you get a chance.

feeblefakie (approved):
LGTM! Thank you for the discussion!

PR #217, opened 2021-06-11
Upgrade the gradle build to 7

brfrn169:
Since the default Gradle version in the openjdk image in Circle CI was upgraded to 7, it's a good time to upgrade our Gradle build to 7, as well.

Note that the Cosmos DB integration test is still failing. I will resolve that in the next PR after it's merged.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #219, opened 2021-06-14
Add duplicatesStrategy to installDist

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-9205

This issue is created by upgrading Gradle version to 7 in #217. 

It looks like we need to add `duplicatesStrategy` to the `installDist` task to fix the issue described in the above Jira. 
Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #220, opened 2021-06-14
Some refactoring for the gRPC services

brfrn169:
- Use `Empty.getDefaultInstance()` instead of `Empty.newBuilder().build()`
- Rename Throwable `e` to `t`
- Use the `mutate(List)` method for all mutation cases in DistributedStorageService and DistributedTransactionService

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #222, opened 2021-06-21
Reorganize the CI config

brfrn169:
I did some refactoring for the CI config. Note that I removed the short verification here because we conduct the periodic verification everyday.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #226, opened 2021-06-29
Use bidirectional streaming RPCs in DistributedTransactionService

brfrn169:
I modified DistributedTransactionService to use bidirectional streaming RPCs as discussed offline. This code passes the integration test but I haven't added unit tests for it yet. Before that, I want to have some discussion based on it. Thanks.


brfrn169:
As discussed, I will continue to work on this. Closing for now.

PR #228, opened 2021-07-01
Use bidirectional streaming RPCs in DistributedTransactionService

brfrn169:
This PR changes `DistributedTransactionService` to use bidirectional streaming RPCs in gRPC.

After it's merged, I will work on changing `DistributedStorageService` to use bidirectional streaming RPCs for scan operations, as well.

Please take a look!


core/src/main/java/com/scalar/db/transaction/rpc/GrpcTransactionManager.java
```diff
@@ -88,24 +90,16 @@ public void withTable(String tableName) {
...
   @Override
   public GrpcTransaction start() throws TransactionException {
-    return execute(
-        () -> {
-          StartTransactionResponse response =
-              stub.start(StartTransactionRequest.newBuilder().build());
-          return new GrpcTransaction(
-              response.getTransactionId(), stub, metadataManager, namespace, tableName);
-        });
+    GrpcTransactionService service = new GrpcTransactionService(stub, metadataManager);
```
feeblefakie, line 93: I'm wondering if it should create `GrpcTransactionService` object in every start.
If `GrpcTransactionManager` object singleton is shared (, which is most of the cases,) it seems not really effective?
Or, should `GrpcTransactionManager` be created for each transaction?

Ah, maybe, it's created for managing `finished` and the latch?

feeblefakie (commented):
Thank you! It's way more clean than I assumed! Great work 🎉 
It looks pretty good but I want to check it again to understand it better.
Left one clarifying question.

core/src/main/java/com/scalar/db/transaction/rpc/GrpcTransactionManager.java
brfrn169: Thank you for reviewing this!

> Ah, probably, it's created for managing finished and the latch separately?

Yes, `GrpcTransactionService` indicates things like a transaction session (assigned to a gRPC bidirectional stream), so we need to instantiate it per transaction. Maybe, the name (GrpcTransactionService) is not good. I will try to rename it to a better one. Thanks!

core/src/main/java/com/scalar/db/transaction/rpc/GrpcTransactionBidirectionalStream.java
```diff
@@ -0,0 +1,316 @@
...
+import java.util.Optional;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+@NotThreadSafe
+public class GrpcTransactionBidirectionalStream implements StreamObserver<TransactionResponse> {
```
feeblefakie, line 41:
```suggestion
public class GrpcTransactionOnBidirectionalStream implements StreamObserver<TransactionResponse> {
```
maybe?

feeblefakie (dismissed):
LGTM! Thank you!
Left a minor naming suggestion.

PR #229, opened 2021-07-02
Update compile commade new gradle version

MilanCorreggio:
the `compile` configurations introduced by the Java plugin have been deprecated since Gradle 4.10 (Aug 27, 2018), and were finally removed in Gradle 7.0 (Apr 9, 2021).
The aforementioned configurations should be replaced by `implementation`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,6 +22,7 @@ subprojects {
     }
 
     dependencies {
+        implementation group: 'com.scalar-labs', name: 'scalardb', version: '3.0.0'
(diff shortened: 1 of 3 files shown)


brfrn169 (dismissed):
LGTM! Thank you!

brfrn169:
We also use the old configurations in the getting-started sample code:
https://github.com/scalar-labs/scalardb/blob/master/docs/getting-started/build.gradle

We might need to fix it in another PR.

feeblefakie (approved):
LGTM! Thanks!

PR #233, opened 2021-07-02
Fix the getting started project

brfrn169:
It seems like the getting started project doesn't work after converting this project to multi-project. I realized that while reviewing #229. This PR will fix this issue. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #236, opened 2021-07-05
Reorganize the protocol buffer messages for DistributedTransaction

brfrn169:
I'm very sorry that we need to modify the protocol buffer messages right after they were merged. I think the code after this change is much better for readability. You can take a look at only `rpc/src/main/proto/scalardb.proto`. 
https://github.com/scalar-labs/scalardb/pull/236/files#diff-5656c45db5f1aec6f6fecf69830df05debbc544247e197cc4f3dd52e3aaf8f3a

Please take a look!


feeblefakie (approved):
Looks less redundant and good!
Thank you!

PR #238, opened 2021-07-06
Make Scalar DB transactions be able to abort eagerly

brfrn169:
https://scalar-labs.atlassian.net/browse/DLT-9267

Please take a look!


core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransactionManager.java
```diff
@@ -140,6 +140,11 @@ public TransactionState getState(String txId) {
     return TransactionState.UNKNOWN; // always returns UNKNOWN
   }
 
+  @Override
+  public TransactionState abort(String txId) {
+    return TransactionState.UNKNOWN; // always returns UNKNOWN
```
feeblefakie, line 145: I rethought this method and `getState` method.
I now think it's better to throw an exception since the application cannot rely on the APIs.
(I think, If UNKNOWN is returned, users would expect if it is returning something else sometimes or eventually)
Thought?

feeblefakie (commented):
Thank you!
The code looks good.
I'm just wondering if we should return UNKNOWN for unsupported operations.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransactionManager.java
brfrn169: Thank you for reviewing this!

Yes, I agree that the methods should throw an exception (`UnsupportedOperationException`) in the JDBC transaction. I will fix it in that way. Thanks!

feeblefakie (dismissed):
LGTM! Thank you!

feeblefakie:
@brfrn169 Sorry, please rebuild the rpc to resolve the conflict.

brfrn169:
@feeblefakie I rebased this to fix the conflicts. Please take a look at it!

PR #239, opened 2021-07-08
Add transaction manager config

brfrn169:
To support transactions through the Scalar DB server, I added a new transaction manager property (`scalar.db.transaction_manager`). We can set `consensus-commit` (default) or `jdbc` or `grpc` to this property. When we set `grpc`, we can perform transactions through the Scalar DB server. And if we set `jdbc`, we also need to set `jdbc` to `scalar.db.storage`.

I also removed the `scalar.db.jdbc.transaction_manager_type` property in this PR because we can replace it with the new property. So this is an incompatible change.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
+++ b/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
@@ -6,6 +6,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.common.base.Strings;
 import com.scalar.db.api.DistributedStorage;
 import com.scalar.db.api.DistributedStorageAdmin;
+import com.scalar.db.api.DistributedTransactionManager;
 import com.scalar.db.api.Isolation;
 import com.scalar.db.storage.cassandra.Cassandra;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/scalar/db/config/DatabaseConfig.java
```diff
@@ -132,6 +139,35 @@ protected void load() {
       namespacePrefix = Optional.empty();
     }
 
+    if (Strings.isNullOrEmpty(props.getProperty(TRANSACTION_MANAGER))) {
+      transactionManagerClass = ConsensusCommitManager.class;
+    } else {
```
feeblefakie, line 144: Very minor suggestion. I won't argue strongly.
```
transactionManagerClass = ConsensusCommitManager.class;
if (!Strings.isNullOrEmpty(props.getProperty(TRANSACTION_MANAGER))) {
```

feeblefakie (approved):
LGTM! Thank you!
Left one minor suggestion which I won't argue for not changing.

PR #241, opened 2021-07-13
Revisit the code around onCompleted() in bidirectional streaming of gRPC

brfrn169:
This is a refactoring thing for the code around onCompleted() in bidirectional streaming of gRPC.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/rpc/GrpcScanOnBidirectionalStream.java
+++ b/core/src/main/java/com/scalar/db/storage/rpc/GrpcScanOnBidirectionalStream.java
@@ -12,7 +12,6 @@ import com.scalar.db.util.retry.ServiceTemporaryUnavailableException;
 import io.grpc.Status.Code;
 import io.grpc.StatusRuntimeException;
 import io.grpc.stub.StreamObserver;
-import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -71,9 +70,10 @@ public class GrpcScanOnBidirectionalStream implements StreamObserver<ScanRespons
     }
   }
 
-  private List<Result> getResults(ScanResponse response) {
+  private List<Result> getResults(ScanResponse response) throws ExecutionException {
     if (!response.getHasMoreResults()) {
-      hasMoreResults.set(false);
+      // We can early close the scanner as it has no more results
+      closeScanner();
     }
     return response.getResultList().stream()
         .map(r -> ProtoUtil.toResult(r, metadata))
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/scalar/db/server/DistributedStorageService.java
```diff
@@ -206,8 +207,10 @@ public void onError(Throwable t) {
```
feeblefakie, line 207: (Sorry, it cannot comment on the code, but it's for the `onError` method)
Not really sure but is it OK not to call `responseObserver.onError(t)` when onError is called?

server/src/main/java/com/scalar/db/server/DistributedStorageService.java
```diff
@@ -206,8 +207,10 @@ public void onError(Throwable t) {
 
     @Override
     public void onCompleted() {
-      responseObserver.onCompleted();
-      cleanUp();
+      if (!cleanedUp.get()) {
```
feeblefakie, line 210: I'm not sure but, is there any case where `cleanedUp` is set to true but `onCompleted` is not called?
If there is such a case, `onCompleted` won't be called properly?

core/src/main/java/com/scalar/db/storage/rpc/GrpcScanOnBidirectionalStream.java
```diff
@@ -141,15 +141,15 @@ private void throwIfErrorForNext(ResponseOrError responseOrError) throws Executi
...
 
-  public void closeScanner() throws IOException {
+  public void closeScanner() throws ExecutionException {
     try {
       if (!hasMoreResults.get()) {
         return;
       }
-      requestObserver.onCompleted();
       hasMoreResults.set(false);
+      requestObserver.onCompleted();
```
feeblefakie, line 150: Just a question.
When users call `closeScanner` explicitly, this `onCompleted` will be called twice?
Is it OK to call `onCompleted` multiple times? (If OK, then I think there is no problem)

feeblefakie (commented):
Overall looking good. Left several clarifying questions.

server/src/main/java/com/scalar/db/server/DistributedStorageService.java
brfrn169: The server side proactively cleans up only when the scanner has no more results. So when a user calls `closeScanner` before the server cleans up, the case will happen.

brfrn169 (commented):
@feeblefakie Thank you for taking a loot at this!

I left some replies. Please check them when you have time!

feeblefakie (commented):
Left additional questions. Sorry, I'm not familiar with the streaming...

feeblefakie (approved):
LGTM! Thank you!

PR #244, opened 2021-07-14
Support graceful shutdown

brfrn169:
I tested this in my env and it looks like it worked properly.

Please take a look!


server/src/main/java/com/scalar/db/server/ScalarDbServer.java
```diff
@@ -39,8 +41,8 @@ public ScalarDbServer(Properties properties) {
 
   @Override
   public Integer call() throws Exception {
-    start();
     addShutdownHook();
+    start();
```
feeblefakie, line 45: Ah, it's better to add a shutdown hook before start. 👍

feeblefakie (approved):
LGTM! Thank you!

PR #246, opened 2021-07-15
Support Prometheus in Scalar DB Server

brfrn169:
To support Prometheus, I used `simpleclient_dropwizard` (https://github.com/prometheus/client_java/tree/master/simpleclient_dropwizard) published by Prometheus officially instead of `jmx_exporter` (https://github.com/prometheus/jmx_exporter) that's also published by Prometheus officially.

This change will start an HTTP endpoint for Prometheus (port: 8080, path: /metrics).

Please take a look!

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -35,6 +35,9 @@ dependencies {
     }
     implementation group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.2.2'
     implementation group: 'io.dropwizard.metrics', name: 'metrics-jmx', version: '4.2.2'
+    implementation group: 'io.prometheus', name: 'simpleclient_dropwizard', version: '0.11.0'
+    implementation group: 'io.prometheus', name: 'simpleclient_servlet', version: '0.11.0'
+    implementation group: 'org.eclipse.jetty', name: 'jetty-servlet', version: '9.4.43.v20210629'
     testImplementation project(':core').sourceSets.integrationTest.output
 }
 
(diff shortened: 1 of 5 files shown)


server/build.gradle
```diff
@@ -35,6 +35,9 @@ dependencies {
     }
     implementation group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.2.2'
     implementation group: 'io.dropwizard.metrics', name: 'metrics-jmx', version: '4.2.2'
+    implementation group: 'io.prometheus', name: 'simpleclient_dropwizard', version: '0.11.0'
+    implementation group: 'io.prometheus', name: 'simpleclient_servlet', version: '0.11.0'
+    implementation group: 'org.eclipse.jetty', name: 'jetty-servlet', version: '9.4.43.v20210629'
```
feeblefakie, line 40: jetty-servlet seems to have newer versions but any particular reason for using the specified version?

feeblefakie (commented):
It's nice and clean. Looking good!
Left a comment and a question.

server/build.gradle
brfrn169: Thank you for taking a look at this!

According to the following doc, Jetty `9.4.x` is for Java 8, and `10.0.x `and `11.0.x` is for Java 11:
https://www.eclipse.org/jetty/

As Scalar DB supports Java 8 at the moment, I decided to use `9.4.43.v20210629` that's the latest version in `9.4.x`.

feeblefakie (approved):
LGTM! Thank you!

PR #247, opened 2021-07-16
Refactoring the code around DatabaseConfig

brfrn169:
What I did is as follows:
- Rename `DynamoDatabaseConfig` to `DynamoConfig`
- Rename `JdbcDatabaseConfig` to `JdbcConfig`
- Add a check if 'grpc' storage is set when 'grpc' transaction manager is set

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
+++ b/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
@@ -156,6 +156,14 @@ public class DatabaseConfig {
           transactionManagerClass = JdbcTransactionManager.class;
           break;
         case "grpc":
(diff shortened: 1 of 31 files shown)


core/src/main/java/com/scalar/db/config/DatabaseConfig.java
```diff
@@ -156,6 +156,14 @@ protected void load() {
           transactionManagerClass = JdbcTransactionManager.class;
           break;
         case "grpc":
+          if (storageClass != GrpcStorage.class) {
```
feeblefakie, line 159: 👍

feeblefakie (approved):
LGTM! Thank you!

PR #248, opened 2021-07-17
Support deadline in Scalar DB Server clients

brfrn169:
Please take a look!


core/src/main/java/com/scalar/db/transaction/rpc/GrpcTransactionOnBidirectionalStream.java
```diff
@@ -64,12 +79,22 @@ public void onError(Throwable t) {
...
+    requestStream.onCompleted();
   }
 
   private ResponseOrError sendRequest(TransactionRequest request) {
-    requestObserver.onNext(request);
-    return Uninterruptibles.takeUninterruptibly(queue);
+    requestStream.onNext(request);
+
+    ResponseOrError responseOrError =
+        Utility.pollUninterruptibly(
```
feeblefakie, line 89: Just a clarifying question.
The queue used here is empty or has one entry always?
If so, the reason for using the queue is to use `poll` to wait for data easily?

feeblefakie (approved):
LGTM! Thank you!
Left a non-blocking clarifying question.

PR #253, opened 2021-07-20
Add documentation of Scalar DB Server

brfrn169:
I added very simple documentation of the Scalar DB Server. We need to extend it in the future.

Please take a look!

### DIFF
--- a/docs/scalardb-server.md
+++ b/docs/scalardb-server.md
@@ -2,11 +2,20 @@
 
 Scalar DB server is a gRPC server that implements Scalar DB interface. 
 With Scalar DB server, you can use Scalar DB features from multiple programming languages that are supported by gRPC.
+
+Currently, we provide only a Java client officially, and we will support other language clients officially in the future.
+Of course, you can generate language-specific client stubs by yourself.
+However, note that it might not be straightforward to implement a client since it's using a bidirectional streaming RPC in gRPC, and you need to be familiar with it.
+
 This document explains how to install and use Scalar DB server.
 
-## Usage
+## Install prerequisites
 
-### Install
+Scalar DB server is written in Java. So the following software is required to run it.
+
+* [Oracle JDK 8](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) (OpenJDK 8) or higher
+
+## Install Scalar DB server
 
 We have Docker images in [our repository](https://github.com/orgs/scalar-labs/packages) and zip archives of Scalar DB server available in [releases](https://github.com/scalar-labs/scalardb/releases).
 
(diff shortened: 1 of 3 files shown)


docs/scalardb-server.md
```diff
@@ -0,0 +1,38 @@
+# Scalar DB server
+
+This document explains the Scalar DB server that is a gRPC interface of Scalar DB.
```
feeblefakie, line 3:
```suggestion
Scalar DB server is a gRPC server that implements Scalar DB interface. 
With Scalar DB server, you can use Scalar DB features from multiple programming languages that are supported by gRPC.
This document explains how to install and use Scalar DB server.
```

docs/scalardb-server.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+For zip archives, you can start Scalar DB server with the following commands:
+
+```
+unzip server.zip
+cd server
+bin/scalardb-server --config <your configration file path>
+```
+
```
feeblefakie, line 35: I think we need something like how to write a client section including how to generate language-specific client stubs. We can basically refer to the official docs but we should note what proto file to use. (and give some examples?)

feeblefakie (commented):
Looking good! 
Left some suggestions.

docs/scalardb-server.md
brfrn169: As we use the gRPC bidirectional stream in the Scalar DB server, I think it isn't straightforward to implement client code. So, I didn't think we should mention that the Scalar DB server can support multiple programming languages via gRPC. What do you think?

docs/scalardb-server.md
brfrn169: As mentioned above, I didn't think we should mention that we support multiple programming languages in the Scalar DB server.

brfrn169 (commented):

Just left some replies to your comments. Please check them when you have time!

docs/scalardb-server.md
thongdk8: I think we can mention that for other language support, the client sample will be provided in the future ( in x.x.x release for example)?

feeblefakie: I think we should mention but we should note that it is difficult since it's using the bi-directional stream of gRPC and you need to be familiar with it to properly implement a client. We can also mention that we will provide sample code or official client SDKs in the future.

Otherwise, I think users/readers would think what the server is for.

brfrn169: Thank you for the suggestions! I will revise it.

docs/scalardb-server.md
```diff
@@ -0,0 +1,102 @@
...
+```
+unzip scalardb-server-<version>.zip
+cd scalardb-server-<version>
+bin/scalardb-server --config <your property file path>
+```
+
+## Usage of the Java client of Scalar DB server
+
+You can use the Java client of Scalar DB server almost the same way as other storages/databases.
```
feeblefakie, line 83:
```suggestion
You can use the Java client of Scalar DB server in almost the same way as other storages/databases.
```

docs/scalardb-server.md
```diff
@@ -0,0 +1,102 @@
+# Scalar DB server
+
+Scalar DB server is a gRPC server that implements Scalar DB interface. 
+With Scalar DB server, you can use Scalar DB features from multiple programming languages that are supported by gRPC.
+
+Currently, we provide only a Java client officially, and we will support other language clients officially in the future.
+Of course, you can generate language-specific client stubs by yourself.
+However, note that it might not be straightforward to implement a client since it's using a bidirectional streaming RPC in gRPC, and you need to be familiar with it.
```
feeblefakie, line 8:
```suggestion
However, note that it is not necessarily straightforward to implement a client since it's using a bidirectional streaming RPC in gRPC, and you need to be familiar with it.
```

feeblefakie (approved):
LGTM! Thank you!

PR #258, opened 2021-07-23
Some more release related changes

brfrn169:
I realized that the `maven` plugin is removed from gradle 7.0, and we need to use the `maven-publish` plugin instead. This PR will change to use  the `maven-publish` plugin. And after this fix, tests jar and test-sources jar won't be published.

Please take a look!


build.gradle
```diff
@@ -54,24 +55,4 @@ subprojects {
...
-        classifier = 'sources'
-        from sourceSets.main.allSource
-    }
-
-    task javadocJar(type: Jar) {
-        classifier = 'javadoc'
-        from javadoc
-    }
-
-    task testJar(type: Jar) {
```
feeblefakie, line 68: I'm just curious. Any reason you decided not to publish `test` and `testSources`?

tools/scalar-schema/project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject scalar-schema "3.0.0"
+(defproject scalar-schema "4.0.0-SNAPSHOT"
```
feeblefakie, line 1: Any reason for `4.0.0-SNAPSHOT`?

build.gradle
```diff
@@ -39,8 +32,16 @@ subprojects {
...
     targetCompatibility = 1.8
 
+    java {
+        withJavadocJar()
+        withSourcesJar()
+    }
+
     javadoc {
         options.addStringOption("notimestamp", "com.scalar.db")
+        if(JavaVersion.current().isJava9Compatible()) {
```
feeblefakie, line 42:
```suggestion
        if (JavaVersion.current().isJava9Compatible()) {
```

server/build.gradle
```diff
@@ -38,7 +38,11 @@ dependencies {
     implementation group: 'io.prometheus', name: 'simpleclient_dropwizard', version: '0.11.0'
     implementation group: 'io.prometheus', name: 'simpleclient_servlet', version: '0.11.0'
     implementation group: 'org.eclipse.jetty', name: 'jetty-servlet', version: '9.4.43.v20210629'
-    testImplementation project(':core').sourceSets.integrationTest.output
+    testImplementation group: 'junit', name: 'junit', version: '4.12'
+    testImplementation group: 'org.assertj', name: 'assertj-core', version: '3.9.1'
+    testImplementation group: 'org.mockito', name: 'mockito-core', version: '2.16.0'
+    testImplementation group: 'org.mockito', name: 'mockito-inline', version: '2.16.0'
+    integrationTestScalarDbServerImplementation project(':core').sourceSets.integrationTest.output
```
feeblefakie, line 45: Just a question for my future update for the ledger project.
Any reason for using `integrationTestScalarDbServerImplementation` instead of `testImplementation`?

feeblefakie (commented):
Looking good! Thank you!
Left some questions and minor comments.

build.gradle
brfrn169: Including `test` and `testSources` jars to a release artifact seems to be a bit tricky in the `maven-publish` plugin. And I rethought why we needed to publish them, and I thought we didn't need to do that, at least for now.

What do you think?

tools/scalar-schema/project.clj
brfrn169: This is because the `master` branch is for the next release, `4.0.0`.

Similarly, I changed the version to `3.0.2-SNAPSHOT` in the `3.0` branch, and I'm going to change the versions to `3.2.0-SNAPSHOT` and `3.1.1-SNAPSHOT` in the `3` and `3.1` branches, respectively.

I think the version should be the next release version in the branch + `-SNAPSHOT`.

brfrn169 (commented):

I just left some replies on your review comments. Please check them when you have time! Thanks.

tools/scalar-schema/project.clj
feeblefakie: Oh, I didn't know that's the way it is. Thank you!

build.gradle
feeblefakie: I think it's fine not to have them since we don't really use them.
Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #259, opened 2021-07-28
Implement addition and deletion of metadata table for jdbc metadata manager

Torch3333:
This implements the methods to add and delete table metadata of the JdbcMetadataManager

### DIFF
--- a/core/src/main/java/com/scalar/db/api/TableMetadata.java
+++ b/core/src/main/java/com/scalar/db/api/TableMetadata.java
@@ -40,6 +40,15 @@ public class TableMetadata {
   }
 
   /**
+   * Creates a new builder instance
+   *
+   * @return a new builder instance
+   */
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  /**
    * Returns the column names
    *
    * @return an {@code LinkedHashSet} of column names
@@ -95,13 +104,43 @@ public class TableMetadata {
     return secondaryIndexNames;
   }
 
-  /**
-   * Creates a new builder instance
-   *
-   * @return a new builder instance
-   */
-  public static Builder newBuilder() {
-    return new Builder();
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    TableMetadata metadata = (TableMetadata) o;
+    if (!Objects.equals(columnNames, metadata.columnNames)) {
+      return false;
+    }
+    if (!Objects.equals(columnDataTypes, metadata.columnDataTypes)) {
+      return false;
+    }
+    if (!Objects.equals(partitionKeyNames, metadata.partitionKeyNames)) {
+      return false;
+    }
+    if (!Objects.equals(clusteringKeyNames, metadata.clusteringKeyNames)) {
+      return false;
+    }
+    if (!Objects.equals(clusteringOrders, metadata.clusteringOrders)) {
+      return false;
+    }
+    return Objects.equals(secondaryIndexNames, metadata.secondaryIndexNames);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(
+        columnNames,
+        columnDataTypes,
+        partitionKeyNames,
+        clusteringKeyNames,
+        clusteringOrders,
+        secondaryIndexNames);
   }
 
   /** A builder class that creates a TableMetadata instance */
@@ -168,45 +207,5 @@ public class TableMetadata {
       return new TableMetadata(
           columns, partitionKeyNames, clusteringKeyNames, clusteringOrders, secondaryIndexNames);
     }
-  }
-
-  @Override
-  public boolean equals(Object o) {
-    if (this == o) {
-      return true;
-    }
-    if (o == null || getClass() != o.getClass()) {
-      return false;
-    }
-
-    TableMetadata metadata = (TableMetadata) o;
-
-    if (!Objects.equals(columnNames, metadata.columnNames)) {
-      return false;
-    }
-    if (!Objects.equals(columnDataTypes, metadata.columnDataTypes)) {
-      return false;
-    }
-    if (!Objects.equals(partitionKeyNames, metadata.partitionKeyNames)) {
-      return false;
-    }
-    if (!Objects.equals(clusteringKeyNames, metadata.clusteringKeyNames)) {
-      return false;
-    }
-    if (!Objects.equals(clusteringOrders, metadata.clusteringOrders)) {
-      return false;
-    }
-    return Objects.equals(secondaryIndexNames, metadata.secondaryIndexNames);
-  }
-
-  @Override
-  public int hashCode() {
-    int result = columnNames != null ? columnNames.hashCode() : 0;
-    result = 31 * result + (columnDataTypes != null ? columnDataTypes.hashCode() : 0);
-    result = 31 * result + (partitionKeyNames != null ? partitionKeyNames.hashCode() : 0);
-    result = 31 * result + (clusteringKeyNames != null ? clusteringKeyNames.hashCode() : 0);
-    result = 31 * result + (clusteringOrders != null ? clusteringOrders.hashCode() : 0);
-    result = 31 * result + (secondaryIndexNames != null ? secondaryIndexNames.hashCode() : 0);
-    return result;
   }
 }
--- a/core/src/main/java/com/scalar/db/api/TableMetadata.java
+++ b/core/src/main/java/com/scalar/db/api/TableMetadata.java
@@ -114,22 +114,12 @@ public class TableMetadata {
     }
 
     TableMetadata metadata = (TableMetadata) o;
-    if (!Objects.equals(columnNames, metadata.columnNames)) {
-      return false;
-    }
-    if (!Objects.equals(columnDataTypes, metadata.columnDataTypes)) {
-      return false;
-    }
-    if (!Objects.equals(partitionKeyNames, metadata.partitionKeyNames)) {
-      return false;
-    }
-    if (!Objects.equals(clusteringKeyNames, metadata.clusteringKeyNames)) {
-      return false;
-    }
-    if (!Objects.equals(clusteringOrders, metadata.clusteringOrders)) {
-      return false;
-    }
-    return Objects.equals(secondaryIndexNames, metadata.secondaryIndexNames);
+    return Objects.equals(columnNames, metadata.columnNames)
+        && Objects.equals(columnDataTypes, metadata.columnDataTypes)
+        && Objects.equals(partitionKeyNames, metadata.partitionKeyNames)
+        && Objects.equals(clusteringKeyNames, metadata.clusteringKeyNames)
+        && Objects.equals(clusteringOrders, metadata.clusteringOrders)
+        && Objects.equals(secondaryIndexNames, metadata.secondaryIndexNames);
   }
 
   @Override
--- a/core/src/main/java/com/scalar/db/api/TableMetadata.java
+++ b/core/src/main/java/com/scalar/db/api/TableMetadata.java
@@ -169,4 +169,44 @@ public class TableMetadata {
           columns, partitionKeyNames, clusteringKeyNames, clusteringOrders, secondaryIndexNames);
     }
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    TableMetadata metadata = (TableMetadata) o;
+
+    if (!Objects.equals(columnNames, metadata.columnNames)) {
+      return false;
+    }
+    if (!Objects.equals(columnDataTypes, metadata.columnDataTypes)) {
+      return false;
+    }
+    if (!Objects.equals(partitionKeyNames, metadata.partitionKeyNames)) {
+      return false;
+    }
+    if (!Objects.equals(clusteringKeyNames, metadata.clusteringKeyNames)) {
+      return false;
+    }
+    if (!Objects.equals(clusteringOrders, metadata.clusteringOrders)) {
+      return false;
+    }
+    return Objects.equals(secondaryIndexNames, metadata.secondaryIndexNames);
+  }
+
+  @Override
+  public int hashCode() {
+    int result = columnNames != null ? columnNames.hashCode() : 0;
+    result = 31 * result + (columnDataTypes != null ? columnDataTypes.hashCode() : 0);
+    result = 31 * result + (partitionKeyNames != null ? partitionKeyNames.hashCode() : 0);
+    result = 31 * result + (clusteringKeyNames != null ? clusteringKeyNames.hashCode() : 0);
+    result = 31 * result + (clusteringOrders != null ? clusteringOrders.hashCode() : 0);
+    result = 31 * result + (secondaryIndexNames != null ? secondaryIndexNames.hashCode() : 0);
+    return result;
+  }
 }
--- a/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
@@ -29,6 +29,11 @@ public interface TableMetadataManager {
    */
   void deleteTableMetadata(String namespace, String table);
 
-  /** Add the given table metadata */
+  /**
+   * Add the given table metadata
+   * @param namespace a namespace
+   * @param table a table
+   * @param metadata a table metadata
+   */
   void addTableMetadata(String namespace, String table, TableMetadata metadata);
 }
--- a/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
@@ -20,4 +20,17 @@ public interface TableMetadataManager {
    * @return a table metadata. null if the table is not found.
    */
   TableMetadata getTableMetadata(String namespace, String table);
+
+  /**
+   * Delete the given table metadata
+   *
+   * @param namespace a namespace
+   * @param table a table metadata
+   */
+  void deleteTableMetadata(String namespace, String table);
+
+  /**
+   * Add the given table metadata
+   */
+  void addTableMetadata(TableMetadata metadata);
 }
--- a/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
@@ -29,8 +29,6 @@ public interface TableMetadataManager {
    */
   void deleteTableMetadata(String namespace, String table);
 
-  /**
-   * Add the given table metadata
-   */
+  /** Add the given table metadata */
   void addTableMetadata(String namespace, String table, TableMetadata metadata);
 }
--- a/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
@@ -31,6 +31,7 @@ public interface TableMetadataManager {
 
   /**
    * Add the given table metadata
+   *
    * @param namespace a namespace
    * @param table a table
    * @param metadata a table metadata
--- a/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
@@ -32,5 +32,5 @@ public interface TableMetadataManager {
   /**
    * Add the given table metadata
    */
-  void addTableMetadata(TableMetadata metadata);
+  void addTableMetadata(String namespace, String table, TableMetadata metadata);
 }
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
@@ -126,7 +126,7 @@ public class JdbcTableMetadataManager implements TableMetadataManager {
         + ", "
         + enclose(INDEXED)
         + " FROM "
-        + encloseFullTableNames(getMetadataSchema(), TABLE)
+        + encloseFullTableName(getMetadataSchema(), TABLE)
         + " WHERE "
         + enclose(FULL_TABLE_NAME)
         + " = ? ORDER BY "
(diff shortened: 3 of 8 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -44,40 +60,34 @@ public JdbcTableMetadataManager(
...
-      RdbEngine rdbEngine)
-      throws SQLException {
+  private Optional<TableMetadata> load(String fullTableName) throws SQLException {
     TableMetadata.Builder builder = TableMetadata.newBuilder();
     boolean tableExists = false;
 
     try (Connection connection = dataSource.getConnection();
         PreparedStatement preparedStatement =
-            connection.prepareStatement(getSelectColumnsStatement(schemaPrefix, rdbEngine))) {
+            connection.prepareStatement(getSelectColumnsStatement(rdbEngine))) {
```
brfrn169, line 74:
```suggestion
            connection.prepareStatement(getSelectColumnsStatement())) {
```

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -105,26 +115,213 @@ public JdbcTableMetadataManager(
     return Optional.of(builder.build());
   }
 
-  private String getSelectColumnsStatement(Optional<String> schemaPrefix, RdbEngine rdbEngine) {
+  private String getSelectColumnsStatement(RdbEngine rdbEngine) {
```
brfrn169, line 118:
```suggestion
  private String getSelectColumnsStatement() {
```

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -105,26 +115,213 @@ public JdbcTableMetadataManager(
...
+  }
+
+  private void execute(Connection connection, String sql) throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      stmt.execute(sql);
+    }
+  }
+
+  private String getMetadataSchema() {
+    return schemaPrefix.orElse("") + JdbcTableMetadataManager.SCHEMA;
```
brfrn169, line 273:
```suggestion
    return schemaPrefix.orElse("") + SCHEMA;
```

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -147,4 +344,20 @@ public TableMetadata getTableMetadata(String fullTableName) {
...
+  public void deleteTableMetadata(String namespace, String table) {
+    try (Connection connection = dataSource.getConnection()) {
+      execute(connection, getDeleteStatement(namespace, table));
+    } catch (SQLException e) {
+      if (e.getMessage().contains("Unknown table") || e.getMessage().contains("does not exist")) {
+        return;
+      }
+      throw new StorageRuntimeException(
+          String.format(
+              "deleting the %s table metadata failed", schemaPrefix + namespace + "." + table),
```
brfrn169, line 358:
```suggestion
              "deleting the %s table metadata failed", schemaPrefix.orElse("") + namespace + "." + table),
```

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -105,26 +115,213 @@ public JdbcTableMetadataManager(
...
+      case ORACLE:
+      case SQL_SERVER:
+        return value ? "1" : "0";
+      default:
+        return value ? "true" : "false";
+    }
+  }
+
+  private String getDeleteStatement(String namespace, String tableName) {
+    String fullTableName = schemaPrefix.orElse("") + namespace + "." + tableName;
```
brfrn169, line 172: Maybe `schema` is better than `namespace` for consistency. Also `table` is better than `tableName`:
```suggestion
  private String getDeleteStatement(String schema, String table) {
    String fullTableName = schemaPrefix.orElse("") + schema + "." + table;
```

core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
```diff
@@ -20,4 +20,15 @@
...
+
+  /**
+   * Delete the given table metadata
+   *
+   * @param namespace a namespace
+   * @param table a table metadata
+   */
+  void deleteTableMetadata(String namespace, String table);
+
+  /** Add the given table metadata */
```
brfrn169, line 32: Can you please add `@param` for this method?

core/src/test/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManagerTest.java
```diff
@@ -0,0 +1,539 @@
...
+  }
+
+  public ResultSet mockResultSet(List<ResultSetMocker.Row> rows) throws SQLException {
+    ResultSet resultSet = mock(ResultSet.class);
+    // Everytime the ResultSet.next() method will be called, the ResultSet.getXXX methods call be
+    // mocked to return the current row data
+    doAnswer(new ResultSetMocker(rows)).when(resultSet).next();
+    return resultSet;
+  }
+  // Utility class used to mock ResultSet
```
brfrn169, line 492:
```suggestion

  // Utility class used to mock ResultSet
```

core/src/test/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManagerTest.java
```diff
@@ -0,0 +1,539 @@
...
+  }
+
+  @Test
+  public void deleteMetadata_forOracleWithExistingTable_ShouldDeleteMetadata() throws SQLException {
+    deleteMetadata_ForXWithExistingTable_ShouldDeleteMetadata(
+        RdbEngine.ORACLE, "DELETE FROM `scalardb`.`metadata` WHERE `full_table_name` = 'ns1.t2'");
+  }
+
+  private void deleteMetadata_ForXWithExistingTable_ShouldDeleteMetadata(
+      RdbEngine rdbEngine, String deleteQuery) throws SQLException {
```
brfrn169, line 448: It looks like this `rdbEngine` is never used. It that okay?

core/src/main/java/com/scalar/db/storage/rpc/GrpcTableMetadataManager.java
```diff
@@ -105,4 +105,16 @@ public int hashCode() {
       return Objects.hash(namespace, table);
     }
   }
+
+  @Override
+  public void deleteTableMetadata(String namespace, String table) {
+    // TODO To implement
```
brfrn169, line 111: Since we don't expose `deleteTableMetadata` and `addTableMetadata` gRPC endpoints, `GrpcTableMetadataManager` will support read only operations. So we don't need this `TODO`.
```suggestion
```

core/src/main/java/com/scalar/db/storage/rpc/GrpcTableMetadataManager.java
```diff
@@ -105,4 +105,16 @@ public int hashCode() {
...
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public void addTableMetadata(String namespace, String table, TableMetadata metadata) {
+    // TODO To implement
```
brfrn169, line 117: Ditto.
```suggestion
```

brfrn169 (commented):
Overall LGTM! Thank you!

I just lest some comments and suggestions. Please check them when you have time!

core/src/main/java/com/scalar/db/api/TableMetadata.java
```diff
@@ -169,4 +169,44 @@ public TableMetadata build() {
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    TableMetadata metadata = (TableMetadata) o;
+
+    if (!Objects.equals(columnNames, metadata.columnNames)) {
```
feeblefakie, line 184: Maybe, we can simplify this comparison using ComparisonChain of guava.
https://github.com/google/guava/wiki/CommonObjectUtilitiesExplained#comparecompareto

core/src/main/java/com/scalar/db/api/TableMetadata.java
```diff
@@ -169,4 +169,44 @@ public TableMetadata build() {
...
+    }
+    if (!Objects.equals(clusteringOrders, metadata.clusteringOrders)) {
+      return false;
+    }
+    return Objects.equals(secondaryIndexNames, metadata.secondaryIndexNames);
+  }
+
+  @Override
+  public int hashCode() {
+    int result = columnNames != null ? columnNames.hashCode() : 0;
```
feeblefakie, line 204: I think we should use Objects.hash() or guava's Objects.hashCode() instead of re-implement the hashing.
https://github.com/google/guava/wiki/CommonObjectUtilitiesExplained#hashcode

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -105,26 +115,213 @@ public JdbcTableMetadataManager(
...
+      try {
+        createMetadataSchema(connection);
+        execute(connection, createTableQuery);
+      } catch (SQLException e) {
+        connection.rollback();
+        throw e;
+      }
+      connection.commit();
+    } catch (SQLException e) {
+      if (e.getMessage().contains("database exists")
```
feeblefakie, line 243: Not sure but checking the actual messages looks implementation-specific.
Do MySQL and PostgreSQL return the same message as long as we use the same JDBC?

feeblefakie (commented):
Thank you! 
Took a glance and left some comments.
Please take a look!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
Torch3333: Indeed, I will update it.

core/src/test/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManagerTest.java
Torch3333: Oh thank you. Indeed it is not good as test will always be run for the same rdbengine which is not the purpose.

core/src/main/java/com/scalar/db/api/TableMetadata.java
Torch3333: I was not aware of this, It's a nice utility method! 
Though all the class attributes are objects that do not implement the `Comparable` interface I am not sure it's possible to use in this case
```
  private final LinkedHashSet<String> columnNames;
  private final Map<String, DataType> columnDataTypes;
  private final LinkedHashSet<String> partitionKeyNames;
  private final LinkedHashSet<String> clusteringKeyNames;
  private final Map<String, Order> clusteringOrders;
  private final Set<String> secondaryIndexNames;
  ```

Comment (feeblefakie) [reply core/src/main/java/com/scalar/db/api/TableMetadata.java:184]: Or we can simply write as follows so that we can remove `return false` statements?
```
return (Objects.equals(columnNames, metadata.columnNames) &&
    Objects.equals(columnDataTypes, metadata.columnDataTypes) &&
    ...);
```

File: core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java (hunk 13 above)
Comment (Torch3333) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: Indeed, the error message is implementation specific. 

> Do MySQL and PostgreSQL return the same message as long as we use the same JDBC?

It seems it is the case but I would have to do more extensive test. 

Otherwise I think we could either : 
- Option 1 : Rely on [SqlException.getErrorCode()](https://docs.oracle.com/javase/jp/8/docs/api/java/sql/SQLException.html#getErrorCode--) which returns a vendor specific error code. I checked will Mysql, Postgresql, SqlServer and Oracle. Each of them throw a custom error when trying to create a duplicate schema/table. I suppose such error code are stable through the different driver versions.
- Option 2 : Mysql and Postgresql supports the `IF EXISTS` keyword that we can use when creating schema and table. In that case no exception would be thrown for the duplication. 
But Oracle and Sql server do not have `IF EXISTS` keyword so we can rely on checking error code for only these two. If error code checking is not satisfying, we can run some select query towards system database to verify if such user or table already exist, which seems a bit hacky. 

What do you think ?

File: core/src/main/java/com/scalar/db/api/TableMetadata.java (hunk 11 above)
Comment (Torch3333) [reply core/src/main/java/com/scalar/db/api/TableMetadata.java:184]: Sure, let me update it.

File: core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java (hunk 13 above)
Comment (feeblefakie) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: If the exception object has a proper message even though they are vendor-specific or different, I think we can always throw an exception as you do now instead of doing special care depending on the messages.
(So, the if block can be simply removed)

Comment (Torch3333) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: Maybe there was a misunderstanding but remove the exception suppression will break the current design.
 Each time the `TableMetadataManager.addTableMetadata` method is called, it tries to create the `scalardb` schema and `scalardb.metadata` table and then add the given sample_schema.sample_table table metadata. If we remove the exception suppression block in case the `scalardb.metadata` has already been created, we won't be able to run `TableMetadataManager.addTableMetadata` more than once.

Comment (feeblefakie) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: Sorry, I'm saying this without fully understanding the goal. 
OK, if the logic is required, I think we should implement the logic for each supported database like MySQL, PostgreSQL, Oracle, and SQLServer since JDBC doesn't abstract the if-exists logic.
I think the current JDBC adapter implements PutIfNotExists so you can do it in the same way at least logically.

Comment (feeblefakie) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: Ah, OK, it's for creating schema. What I was saying in the last sentence is about inserting with IF NOT EXISTS, so it's a different thing. Please ignore the last sentecne.

Comment (Torch3333) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: Alright. 👍 
To be sure I understood correctly, I will update as follows : 
- For **Mysql** & **Postgresql**: use `IF EXISTS` statement in the sql request
- For Oracle & SqlServer : execute a `select` query to check shema/table existence and execute  a `create` if the schema/table does not exist.
- If an exception is thrown, do not suppress it.

Comment (feeblefakie) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: The approach sounds good to me.

@brfrn169 (sorry for the interrupt) What do you think?

Comment (brfrn169) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: The approach @Torch3333 mentioned is fine with me. 

Actually, the logic in the schema tool checks the JDBC implementation-specific message, which is the same as the current one. That is because I didn't want to have different logics (queries) between MySQL/PostgreSQL and Oracle/SQLServer, which I thought was a bit hacky. However, now that there are similar hacky logics in other parts, so we no longer have to worry about that anymore. 😅

Comment (feeblefakie) [reply core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java:243]: @brfrn169 Thank you! (Sorry for the interruption!)

File: core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java [hunk 14]
```diff
@@ -105,26 +113,291 @@ public JdbcTableMetadataManager(
...
+                getMetadataSchema(), TABLE);
+        // If the metadata table do not exists, the create table statement will be executed
+        createTableIfNotExistsStatement =
+            "IF NOT EXISTS (" + selectTableQuery + ") " + createTableStatement;
+        break;
+      default:
+        throw new UnsupportedOperationException("rdb engine not supported");
+    }
+    execute(connection, createTableIfNotExistsStatement);
+  }
```
Torch3333, line 275: For the table creation

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -105,26 +113,291 @@ public JdbcTableMetadataManager(
...
+                + " IF (userExistCount = 0) THEN EXECUTE IMMEDIATE "
+                + createUserStatement
+                + "\n"
+                + "END IF;\n"
+                + "END;");
+        execute(
+            connection, "ALTER USER " + enclose(getMetadataSchema()) + " quota unlimited on USERS");
+        break;
+    }
+  }
```
Torch3333, line 340: For the schema creation

Torch3333 (commented):
Thank you for your comments ! Please have a look.

I updated all the queries to create the metadata schema and table only if it exists. 
With Postgresql and Mysql, it's quite straightforward thanks to the `IF EXISTS` keyword that is allowed but with Oracle and SqlServer, it's quite complicated.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -105,26 +113,291 @@ public JdbcTableMetadataManager(
...
+            + String.format("%s %s NOT NULL,", enclose(INDEXED), getBooleanType())
+            + String.format("%s INTEGER NOT NULL,", enclose(ORDINAL_POSITION))
+            + String.format(
+                "PRIMARY KEY (%s, %s))", enclose(FULL_TABLE_NAME), enclose(COLUMN_NAME));
+    String createTableIfNotExistsStatement;
+    switch (rdbEngine) {
+      case ORACLE:
+        // Save into the tableExistCount variable if the metadata table
+        // exists (1=true, 0=false) by querying a system table
+        String selectTableExistStatement =
```
feeblefakie, line 231: I'm curious what would happen if you create a table that exists in Oracle and SQLServer.
If it's throwing a certain exception for duplicate table creation,  I think simply ignoring the exception might be better rather than select and create in a stored procedure.
I'm sure you have tried it and came to this but asking just in case.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -105,26 +113,291 @@ public JdbcTableMetadataManager(
...
+      execute(connection, insertStatement);
+    } catch (SQLException e) {
+      connection.rollback();
+      throw e;
+    }
+  }
+
+  private void createTableIfNotExists(Connection connection) throws SQLException {
+    String createTableStatement =
+        String.format("CREATE TABLE %s(", encloseFullTableNames(getMetadataSchema(), TABLE))
```
feeblefakie, line 216: It's a very minor comment.
I feel just calling one `String.format` makes the query more readable.
Or we can use do simple String concatenation or Guava's Joiner to remove redundant `String.format`.
https://github.com/google/guava/wiki/StringsExplained#joiner

feeblefakie (commented):
Thank you! Left a question and a commet.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
Torch3333: All right. I updated it using simple string concatenation.

Torch3333:
I updated the schema/table creation as we discussed this morning. Please have a look.

feeblefakie (dismissed):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -147,4 +413,29 @@ public TableMetadata getTableMetadata(String fullTableName) {
...
+          e);
+    }
+    tableMetadataCache.put(schemaPrefix.orElse("") + namespace + "." + table, Optional.empty());
+  }
+
+  private String enclose(String name) {
+    return QueryUtils.enclose(name, rdbEngine);
+  }
+
+  private String encloseFullTableNames(String schema, String table) {
```
brfrn169, line 438:
```suggestion
  private String encloseFullTableName(String schema, String table) {
```

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
```diff
@@ -31,11 +33,23 @@
 public class JdbcTableMetadataManager implements TableMetadataManager {
```
brfrn169, line 33: According to the spotbugs report, we need to suppress or fix the warning:
https://4146-151700458-gh.circle-artifacts.com/0/gradle_spotbugs_reports_for_core/spotbugs/main.html

core/src/test/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManagerTest.java
```diff
@@ -0,0 +1,533 @@
...
+import java.util.List;
+import java.util.Optional;
+import javax.sql.DataSource;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.mockito.invocation.InvocationOnMock;
+
+public class JdbcTableMetadataManagerTest {
```
brfrn169, line 30: It looks like the spotbugs report shows some false positives:
https://4146-151700458-gh.circle-artifacts.com/0/gradle_spotbugs_reports_for_core/spotbugs/test.html

I think we can suppress the warnings with `@SuppressFBWarnings("OBL_UNSATISFIED_OBLIGATION")`.

brfrn169 (commented):
After merging the master branch into this branch, the CI shows some spotbugs warnings that are basically false positives. Can you please handle them? Other than that, LGTM!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
Torch3333: Good eye, thanks!

core/src/test/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManagerTest.java
Torch3333: Indeed, since the objects are mocked, it's a false positive

core/src/main/java/com/scalar/db/storage/jdbc/JdbcTableMetadataManager.java
Torch3333: I think this one is a false positive to as even though is it dynamically generated. No external input is used to build the query.

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):

PR #263, opened 2021-08-03
Introduce Error Prone plugin

brfrn169:
This PR introduces the Error Prone plugin: https://errorprone.info/

I also fixed the warnings that the plugin showed in this PR.

Please take a look!

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/storage/jdbc/test/TestEnv.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/jdbc/test/TestEnv.java
@@ -1,5 +1,6 @@
 package com.scalar.db.storage.jdbc.test;
 
+import com.google.common.collect.ImmutableMap;
 import com.scalar.db.api.Scan;
 import com.scalar.db.api.TableMetadata;
 import com.scalar.db.config.DatabaseConfig;
@@ -16,7 +17,6 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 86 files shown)


core/src/main/java/com/scalar/db/io/Value.java
```diff
@@ -1,13 +1,14 @@
...
 import java.util.Optional;
+import javax.annotation.Nonnull;
 
 /**
  * An abstraction for storage entry's value
  *
  * @author Hiroyuki Yamada
  */
-public interface Value<T> extends Comparable<T> {
+public interface Value<T> extends Comparable<Value<T>> {
```
brfrn169, line 11: I think this is the biggest change. Please take a look at this carefully.

brfrn169:
I rebased this to fix the conflict.

core/src/main/java/com/scalar/db/io/Value.java
feeblefakie: I'm not fully sure, but should we implement `Comparable` in each sub-class so that we can have, for example, `compareTo(BigIntValue)` in BigIntValue and we can avoid runtime casting causing ClassCastException?

feeblefakie (commented):
Left one question.

core/src/main/java/com/scalar/db/io/BlobValue.java
```diff
@@ -50,18 +47,15 @@ public BlobValue(byte[] value) {
...
-  /**
-   * Returns the content of this {@code Value}
-   *
-   * @return an {@code Optional} of the content of this {@code Value}
-   */
+  @Override
+  @Nonnull
   public Optional<byte[]> get() {
-    return bytes.map(b -> Arrays.copyOf(b.array(), b.limit()));
+    return value;
```
feeblefakie, line 53: Does this break the immutability? (value can be updated by the outside of the class?)

feeblefakie (commented):
Left one additional question.

core/src/main/java/com/scalar/db/io/Value.java
brfrn169: Thank you for taking a look at this!

Regarding this, Error Prone showed the following warning:
https://errorprone.info/bugpattern/ComparableType

feeblefakie: I meant, I think it seems better to remove the `extends Comparable<Value<T>>` from Value class and add type-specific `implements Comparable` to each subclass like `implements Comparable<BitIntValue>` in BigIntValue class from users' perspective.

brfrn169: Ah, that can be a good idea. Let me try. Thanks.

brfrn169:
@feeblefakie @Torch3333 Thank you for the discussion! I just fixed `compareTo` methods of the `Value` implementations as discussed. Please take a look!

core/src/main/java/com/scalar/db/service/TransactionService.java
```diff
@@ -57,34 +54,40 @@ public DistributedTransaction start(String txId) throws TransactionException {
     return manager.start(txId);
   }
 
+  @Deprecated
```
Torch3333, line 57: If I'm not wrong, this is a public interface so adding a comment would be very helpful to the user to know what he should be using instead.
https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/deprecation/deprecation.html
For example : 
```
/**
 * @deprecated  As of release 1.3, replaced by {@link #getPreferredSize()}
 */
 ```

**Review by Torch3333 — COMMENTED** [review]
I am sorry for the late review. Thank you for the massive corrections. 
I left 2 comments, please take a look.

File: core/src/main/java/com/scalar/db/io/BlobValue.java [hunk 4]
```diff
@@ -110,29 +112,35 @@ public boolean equals(Object o) {
...
+    }
+    return !value.isPresent() && !other.value.isPresent();
   }
 
   @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
         .add("name", name)
-        .add("value", bytes.toString())
+        .add("value", value.map(Arrays::toString).toString())
```
feeblefakie, line 128:
```suggestion
        .add("value", value.map(Arrays::toString).get())
```

feeblefakie (commented):
Left one minor suggestion. Other than that, LGTM!

core/src/integration-test/java/com/scalar/db/storage/jdbc/test/TestEnv.java
```diff
@@ -175,6 +158,7 @@ private String namespacePrefix() {
     return config.getNamespacePrefix().orElse("");
   }
 
+  @SuppressFBWarnings("OBL_UNSATISFIED_OBLIGATION")
```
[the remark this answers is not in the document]
brfrn169: Yes, right. I hope that it's a rare case where we need to use those annotations.

core/src/main/java/com/scalar/db/io/BlobValue.java
brfrn169: I think `get()` throws `NoSuchElementException` if value is empty. So should we keep `toString()`?

core/src/integration-test/java/com/scalar/db/storage/jdbc/test/TestEnv.java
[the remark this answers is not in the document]
Torch3333: All right. Yeah I hope to :)

Torch3333 (approved):
Thanks, LGTM

core/src/main/java/com/scalar/db/io/BlobValue.java
feeblefakie: Ah, OK. Then, let's keep it.

feeblefakie (approved):
LGTM! Thank you!

PR #266, opened 2021-08-05
add scalar-schema dockerfile

ypeckstadt:
I added this to be able to do the following

https://docs.google.com/document/d/1FepHOSeKz7PxkHGo5xSI6Pa4FJT61ua4_0KnzsFvUAo/edit#heading=h.5yqkuawp7xby

### DIFF
--- a/tools/scalar-schema/Dockerfile
+++ b/tools/scalar-schema/Dockerfile
@@ -1,0 +1,19 @@
+FROM clojure:openjdk-11-lein-slim-buster AS BUILD
+COPY . /code
+WORKDIR /code
+RUN lein uberjar
+
+FROM openjdk:11-jre-slim
+WORKDIR /app
+COPY --from=BUILD /code/target/scalar-schema-standalone-*-SNAPSHOT.jar ./app.jar
+
+# ENV variables
+ENV SCALAR_DB_STORAGE cassandra
+ENV SCALAR_DB_HOST localhost
+ENV SCALAR_DB_USERNAME user
+ENV SCALAR_DB_PASSWORD pass
+ENV SCALAR_DB_SCHEMA schema.json
+ENV SCALAR_DB_REPLICATION_FACTOR 1
+
+# Run command
+ENTRYPOINT java -jar app.jar --$SCALAR_DB_STORAGE -h $SCALAR_DB_HOST -u $SCALAR_DB_USERNAME -p $SCALAR_DB_PASSWORD -f $SCALAR_DB_SCHEMA -R $SCALAR_DB_REPLICATION_FACTOR
--- a/tools/scalar-schema/Dockerfile
+++ b/tools/scalar-schema/Dockerfile
@@ -1,19 +1,28 @@
-FROM clojure:openjdk-11-lein-slim-buster AS BUILD
+FROM clojure:openjdk-8-lein-slim-buster AS BUILD
 COPY . /code
 WORKDIR /code
 RUN lein uberjar
 
-FROM openjdk:11-jre-slim
+FROM openjdk:8u275-jre-slim
+
+# Fix CVE-2021-3520, CVE-2021-33560, CVE-2021-20231, CVE-2021-20232, CVE-2020-24659, sCVE-2021-20305 and CVE-2021-23840
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    liblz4-1=1.8.3-1+deb10u1 \
+    libgcrypt20=1.8.4-5+deb10u1 \
+    libgnutls30=3.6.7-4+deb10u7 \
+    libhogweed4=3.4.1-1+deb10u1 \
+    libnettle6=3.4.1-1+deb10u1 \
+    libssl1.1=1.1.1d-0+deb10u6 \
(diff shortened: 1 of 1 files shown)


tools/scalar-schema/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+# ENV variables
+ENV SCALAR_DB_STORAGE cassandra
+ENV SCALAR_DB_HOST localhost
+ENV SCALAR_DB_USERNAME user
+ENV SCALAR_DB_PASSWORD pass
+ENV SCALAR_DB_SCHEMA schema.json
+ENV SCALAR_DB_REPLICATION_FACTOR 1
+
+# Run command
+ENTRYPOINT java -jar app.jar --$SCALAR_DB_STORAGE -h $SCALAR_DB_HOST -u $SCALAR_DB_USERNAME -p $SCALAR_DB_PASSWORD -f $SCALAR_DB_SCHEMA -R $SCALAR_DB_REPLICATION_FACTOR
```
brfrn169, line 19: It looks like it supports only the Cassandra options. Don't we support the other adapters (Cosmos DB, DynamoDB, JDBC)?

brfrn169 (commented):
Thank you! Just left a couple of comments. Please check them when you have time!

tools/scalar-schema/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM clojure:openjdk-11-lein-slim-buster AS BUILD
+COPY . /code
+WORKDIR /code
+RUN lein uberjar
```
feeblefakie, line 4: I think we should use the released version instead of building it itself like `scalardl-schema-loader` because the main is development branch.
https://github.com/scalar-labs/scalardl-schema-loader/blob/master/Dockerfile

tools/scalar-schema/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+FROM openjdk:11-jre-slim
```
feeblefakie, line 6: Please use openjdk 8. (maybe 8u302?)

tools/scalar-schema/Dockerfile
feeblefakie, line 19: Please use`ENTRYPOINT ["java", "-jar", ...]` format since it is recommended.
https://docs.docker.com/engine/reference/builder/#entrypoint

With the current form, I think SIGTERM won't be sent properly to the java process. (it is sent to a shell process that triggers the java instead), which makes graceful shutdown not working as expected.

ypeckstadt: Actually, it is supported by the `--$SCALAR_DB_STORAGE` part. However, there are a few options I missed that need to be passed on to make it fully work. Like region, Cassandra port, etc

I noticed that the scalardl-schema-loader does not bother with ENV variables and just used the commands option for Docker. 

Like this

```
 command:
      - "--cassandra"
      - "-h"
      - "cassandra"
      - "-R"
      - "1"
```

I guess maybe it's easier to keep it like this.

feeblefakie (commented):
Thank you!
Left some comments and suggestions.

Please also check vulnerabilities with trivy as follows.
```trivy image --no-progress  --format  table --exit-code  1 --ignore-unfixed --vuln-type  os,library --severity  CRITICAL,HIGH     ghcr.io/scalar-labs/scalar-schema:xxx```
https://github.com/aquasecurity/trivy

We upgrade some libraries depending on vulnerabilities it finds.
https://github.com/scalar-labs/scalardl-schema-loader/blob/master/Dockerfile#L4

File: tools/scalar-schema/Dockerfile (hunk 2 above)
Comment (ypeckstadt) [reply tools/scalar-schema/Dockerfile:4]: I actually first wanted to push this PR into the 3.1 branch for this exact reason. The idea is to build and publish the image based on the Dockerfile in a branch of a released version, not from master.

Comment (feeblefakie) [reply tools/scalar-schema/Dockerfile:4]: Ah, it's in the same repo as scalardb, so that it should be fine.

File: tools/scalar-schema/Dockerfile [hunk 4]
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /app
+COPY --from=BUILD /code/target/scalar-schema-standalone-*-SNAPSHOT.jar ./app.jar
```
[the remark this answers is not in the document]
ypeckstadt: Update the code based on the recommendation.

tools/scalar-schema/Dockerfile
ypeckstadt: I have removed the usage of ENV variables and just use commands (when called from docker compose).  So now everything is supported.

ypeckstadt: I have updated the code. Thanks for the recommendation. Didnt know about that.

tools/scalar-schema/Dockerfile
ypeckstadt: I have updated to code and for now just switched to the same one used as in the scalardl schema loader.

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #276, opened 2021-08-17
Introduce factory classes to instantiate storage and transaction

brfrn169:
I think we can introduce factory classes to instantiate storage and transaction objects rather than expose the Guice dependency.

For example:

The original way to get a `DistributedStorage` instance, which exposes the Guice dependency:
```
Injector injector = Guice.createInjector(new StorageModule(new DatabaseConfig(props)));
StorageService storage = injector.getInstance(StorageService.class);
```

The new way to get a `DistributedStorage` instance:
```
StorageFactory factory = new StorageFactory(databaseConfig);
DistributedStorage storage = factory.getStorage();
```

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedStorage.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedStorage.java
@@ -31,8 +31,8 @@ import java.util.Optional;
  * readability.)
  *
  * <pre>{@code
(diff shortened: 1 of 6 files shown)


Torch3333 (approved):
Good idea, Thank you.

feeblefakie (approved):
Nice! LGTM! Thank you.

PR #282, opened 2021-08-23
Update scalar db compatibility matrix

scalar-boney:
Update Scalar DB compatibility matrix
* Scalar DB 3.1.0
* Oracle database support


brfrn169 (approved):
It's nice to confirm that Scalar DB on Oracle works fine! Thank you!

PR #283, opened 2021-08-23
Update the Getting Started

brfrn169:
- Use `StorageFactory` and `TransactionFactory` in the Getting Started
- Use the `Key` constructors for a single value key in the Getting Started
- Use the new `Put.withValue()` in the Getting Started
- Use the new `Value.getAsXXX()` in the Getting Started

Please take a look!


Torch3333 (approved):
LGTM, thank you.

PR #285, opened 2021-08-23
Add some attributes to Oracle tables for performance

brfrn169:
I added three attributes to Oracle tables for performance:
- ROWDEPENDENCIES
- INITRANS
- MAXTRANS

We have already confirmed that the performance was improved after adding these attributes.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
@@ -264,7 +264,21 @@ public class JdbcDatabaseAdmin implements DistributedStorageAdmin {
                 .map(this::enclose)
                 .collect(Collectors.joining(","))
             + "))";
+    if (rdbEngine == RdbEngine.ORACLE) {
+      // For Oracle Database, add ROWDEPENDENCIES to the table to improve the performance
+      createTableStatement += " ROWDEPENDENCIES";
+    }
     execute(connection, createTableStatement);
+
+    if (rdbEngine == RdbEngine.ORACLE) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
```diff
@@ -264,7 +264,21 @@ private void createTableInternal(
...
+
+    if (rdbEngine == RdbEngine.ORACLE) {
+      // For Oracle Database, set INITRANS to 3 and MAXTRANS to 255 for the table to improve the
+      // performance
+      String alterTableStatement =
+          "ALTER TABLE "
+              + enclosedFullTableName(getFullNamespaceName(schemaPrefix, schema), table, rdbEngine)
+              + " INITRANS 3 MAXTRANS 255";
+      execute(connection, alterTableStatement);
+    }
```
Torch3333, line 281: Should we add the same properties to the metadata table created by the `JdbcTableMetadataManager` ?

Torch3333 (approved):
Thank you. LGTM but I left a question.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
brfrn169: Thank you for reviewing this!

The properties are for the performance of the parallel transactional execution. So I don't think we need the properties to the metadata.

Torch3333: All right!

PR #287, opened 2021-08-24
Reduce IntelliJ warnings

brfrn169:
This PR is to reduce the warnings that IntelliJ shows. There are a lot of changes in this PR but they are basically automatic fixes by IntelliJ, so you don't need to look into it very carefully 😅 

Please take a look when you get a chance!


Torch3333 (approved):
LGTM, thank you!

PR #291, opened 2021-08-28
Introduce GateKeeper that manages the front gate of the server that processes requests

brfrn169:
This PR introduces the `GateKeeper` interface that manages the front gate of the Scalar DB server that processes requests. And I added the following two implementations:

- LockFreeGateKeeper: the implementation that's not using any locking
- SynchronizedGateKeeper: the implementation that uses lock (Java synchronized)

Scalar DB server uses the `GateKeeper` interface to pause/unpause requests for implementing [the Scalar Admin interface](https://github.com/scalar-labs/scalar-admin).

For now, the default is `LockFreeGateKeeper`, but I will conduct a performance test to compare the two implementations and determine which one we should make the default.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM. Very interesting, thank you!

PR #294, opened 2021-09-01
Separate consensus commit related config from DatabaseConfig

brfrn169:
I moved the consensus commit related config from `DatabaseConfig` to `ConsensusCommitConfig` that I added in this PR. After this change, we will add new consensus commit related config to `ConsensusCommitConfig`.

Please take a look!


feeblefakie (approved):
It's a good idea. LGTM. Thank you!

PR #301, opened 2021-09-07
Implementation of Dynamo Admin module

thongdk8:
This PR contains the implementation of the Dynamo Admin module and tests for it.


core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,41 +1,130 @@
...
+  private final ImmutableSet<String> SECONDARY_INDEX_SCALING_TYPE =
+      ImmutableSet.<String>builder().add("index-read").add("index-write").build();
+  private final ImmutableMap<String, ScalableDimension> SCALABLE_DIMENSION_MAPPING =
+      ImmutableMap.<String, ScalableDimension>builder()
+          .put("read", ScalableDimension.DYNAMODB_TABLE_READ_CAPACITY_UNITS)
+          .put("write", ScalableDimension.DYNAMODB_TABLE_WRITE_CAPACITY_UNITS)
+          .put("index-read", ScalableDimension.DYNAMODB_INDEX_READ_CAPACITY_UNITS)
+          .put("index-write", ScalableDimension.DYNAMODB_INDEX_READ_CAPACITY_UNITS)
+          .build();
+  private ApplicationAutoScalingClient applicationAutoScalingClient;
```
brfrn169, line 102:
```suggestion
  private static final Logger LOGGER = LoggerFactory.getLogger(DynamoAdmin.class);
  private static final String PARTITION_KEY = "concatenatedPartitionKey";
  private static final String CLUSTERING_KEY = "concatenatedClusteringKey";
  private static final String GLOBAL_INDEX_NAME_PREFIX = "global_index";

  private static final int COOL_TIME_SEC = 60;
  private static final double TARGET_USAGE_RATE = 70.0;

  public static final String NO_SCALING = "no-scaling";
  public static final String NO_BACKUP = "no-backup";
  public static final String REQUEST_UNIT = "ru";
  public static final Boolean DEFAULT_NO_SCALING = false;
  public static final Boolean DEFAULT_NO_BACKUP = false;
  public static final long DEFAULT_RU = 10;

  private static final int DELETE_BATCH_SIZE = 100;
  private static final ImmutableMap<DataType, ScalarAttributeType> DATATYPE_MAPPING =
      ImmutableMap.<DataType, ScalarAttributeType>builder()
          .put(DataType.INT, ScalarAttributeType.N)
          .put(DataType.BIGINT, ScalarAttributeType.N)
          .put(DataType.FLOAT, ScalarAttributeType.N)
          .put(DataType.DOUBLE, ScalarAttributeType.N)
          .put(DataType.TEXT, ScalarAttributeType.S)
          .put(DataType.BLOB, ScalarAttributeType.B)
          .build();
  private static final ImmutableSet<String> TABLE_SCALING_TYPE =
      ImmutableSet.<String>builder().add("read").add("write").build();
  private static final ImmutableSet<String> SECONDARY_INDEX_SCALING_TYPE =
      ImmutableSet.<String>builder().add("index-read").add("index-write").build();
  private static final ImmutableMap<String, ScalableDimension> SCALABLE_DIMENSION_MAPPING =
      ImmutableMap.<String, ScalableDimension>builder()
          .put("read", ScalableDimension.DYNAMODB_TABLE_READ_CAPACITY_UNITS)
          .put("write", ScalableDimension.DYNAMODB_TABLE_WRITE_CAPACITY_UNITS)
          .put("index-read", ScalableDimension.DYNAMODB_INDEX_READ_CAPACITY_UNITS)
          .put("index-write", ScalableDimension.DYNAMODB_INDEX_READ_CAPACITY_UNITS)
          .build();

  private final DynamoDbClient client;
  private final ApplicationAutoScalingClient applicationAutoScalingClient;
  private final Optional<String> namespacePrefix;
  private final DynamoTableMetadataManager metadataManager;
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -47,32 +136,189 @@ public DynamoAdmin(DynamoConfig config) {
...
+    this.namespacePrefix = namespacePrefix;
+  }
+
   @Override
-  public void createNamespace(String namespace, Map<String, String> options)
-      throws ExecutionException {
-    throw new UnsupportedOperationException("implement later");
+  public void createNamespace(String namespace, Map<String, String> options) {
+    LOGGER.info(
+        "In Dynamo DB storage, namespace will be added to table name as prefix along with dot separator.");
```
brfrn169, line 154: I don't think we need logging here. Can you please just add a comment?

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -47,32 +136,189 @@ public DynamoAdmin(DynamoConfig config) {
...
+        GlobalSecondaryIndex.Builder globalSecondaryIndexBuilder =
+            GlobalSecondaryIndex.builder()
+                .indexName(getGlobalIndexName(namespace, table, secondaryIndex))
+                .keySchema(
+                    KeySchemaElement.builder()
+                        .attributeName(secondaryIndex)
+                        .keyType(KeyType.HASH)
+                        .build())
+                .projection(Projection.builder().projectionType(ProjectionType.ALL).build());
+        Long ru = DEFAULT_RU;
```
brfrn169, line 216:
```suggestion
        long ru = DEFAULT_RU;
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -86,18 +332,283 @@ public TableMetadata getTableMetadata(String namespace, String table) throws Exe
...
+        .pointInTimeRecoverySpecification(buildPointInTimeRecoverySpecification())
+        .build();
+  }
+
+  private void controlAutoScaling(
+      String namespace,
+      String table,
+      boolean noScaling,
+      Set<String> secondaryIndexes,
+      Optional<Long> ru)
```
brfrn169, line 399: We have the default value of this, so I don't think we need to make it `Optional`.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -47,32 +136,189 @@ public DynamoAdmin(DynamoConfig config) {
...
+    requestBuilder.tableName(Utility.getFullTableName(namespacePrefix, namespace, table));
+
+    // scaling control
+    boolean noScaling = DEFAULT_NO_SCALING;
+    if (options.containsKey(NO_SCALING)) {
+      if (Boolean.parseBoolean(options.get(NO_SCALING))) {
+        noScaling = true;
+      }
+    }
+    if (!noScaling) {
```
brfrn169, line 244: Looks like this check is done twice here and in the `controlAutoScaling` method.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -51,7 +72,7 @@ public TableMetadata getTableMetadata(Operation operation) {
 
   @Override
   public TableMetadata getTableMetadata(String namespace, String table) {
-    String fullName = namespace + "." + table;
+    String fullName = Utility.getFullTableName(namespacePrefix, namespace, table);
```
brfrn169, line 75: Maybe we can statically import `Utility.getFullTableName` to make the code simpler and shorter.

brfrn169 (commented):
Thank you! I did the first review for this and left some comments/suggestions. Please check them when you have time!

I will review it again later. Thanks.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Apply in [2f5e335](https://github.com/scalar-labs/scalardb/pull/301/commits/2f5e3350aab24ce146539e84efef99fc063d8769)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -127,19 +153,174 @@ private DataType convertDataType(String columnType) {
...
+        }
+      }
+    } catch (DynamoDbException e) {
+      throw new StorageRuntimeException("retrieving the table names failed", e);
+    }
+    return tableSet;
+  }
+
+  private void createMetadataTableIfNotExist() throws StorageRuntimeException {
+    if (!metadataTableExists()) {
```
brfrn169, line 266: We can do `early return` here to reduce nest:
```suggestion
    if (metadataTableExists()) {
      return;
    }
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -127,19 +153,174 @@ private DataType convertDataType(String columnType) {
...
+
+      try {
+        client.createTable(requestBuilder.build());
+      } catch (DynamoDbException e) {
+        throw new StorageRuntimeException("creating meta data table failed");
+      }
+
+      while (true) {
+        try {
+          Thread.sleep(CREATING_WAITING_TIME);
```
brfrn169, line 293:
```suggestion
        Uninterruptibles.sleepUninterruptibly(CREATING_WAITING_TIME, TimeUnit.MILLISECONDS);
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -127,19 +153,174 @@ private DataType convertDataType(String columnType) {
 
   @Override
   public void deleteTableMetadata(String namespace, String table) {
-    // TODO To implement
-    throw new UnsupportedOperationException();
+
+    HashMap<String, AttributeValue> keyToGet = new HashMap<>();
```
brfrn169, line 157: Maybe `keyToDelete`?
```suggestion
    HashMap<String, AttributeValue> keyToDelete = new HashMap<>();
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -127,19 +153,174 @@ private DataType convertDataType(String columnType) {
...
+          throw new StorageRuntimeException("getting table description failed", e);
+        }
+      }
+    }
+  }
+
+  private boolean metadataTableExists() throws StorageRuntimeException {
+    boolean tableExist = false;
+    try {
+      ListTablesResponse listTablesResponse = client.listTables();
```
brfrn169, line 312: Maybe it's better to use `client.describeTable()` and handle `ResourceNotFoundException` when the table is not found?

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -86,18 +388,175 @@ public TableMetadata getTableMetadata(String namespace, String table) throws Exe
...
+      return metadataManager.getTableNames(namespace);
+    } catch (RuntimeException e) {
+      throw new ExecutionException("getting list of tables failed");
+    }
   }
 
   @Override
   public boolean namespaceExists(String namespace) throws ExecutionException {
-    throw new UnsupportedOperationException("implement later");
+    boolean nameSpaceExists = false;
```
brfrn169, line 400:
```suggestion
    boolean namespaceExists = false;
```

brfrn169 (commented):
Thank you! I took a look at this again and left some comments/suggestions. Please take a look!

Maybe, I think I will take a look at it once again. Thanks.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
thongdk8: Thanks for the suggestion. Apply in [83b7020](https://github.com/scalar-labs/scalardb/pull/301/commits/83b7020c4783675821630ff67827a70b29f5692d). Other suggestions were applied as well.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -86,18 +394,175 @@ public TableMetadata getTableMetadata(String namespace, String table) throws Exe
...
+        + "/index/"
+        + getGlobalIndexName(namespace, table, globalIndex);
+  }
+
+  private TargetTrackingScalingPolicyConfiguration getScalingPolicyConfiguration(String type) {
+    return TargetTrackingScalingPolicyConfiguration.builder()
+        .predefinedMetricSpecification(
+            PredefinedMetricSpecification.builder()
+                .predefinedMetricType(
+                    type.contains("read")
```
feeblefakie, line 540: This `type` always be "read" or "write". No "index-read" is coming here?

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,386 @@
...
+          .put(DataType.BIGINT, ScalarAttributeType.N)
+          .put(DataType.FLOAT, ScalarAttributeType.N)
+          .put(DataType.DOUBLE, ScalarAttributeType.N)
+          .put(DataType.TEXT, ScalarAttributeType.S)
+          .put(DataType.BLOB, ScalarAttributeType.B)
+          .build();
+  private static final ImmutableSet<String> TABLE_SCALING_TYPE =
+      ImmutableSet.<String>builder().add("read").add("write").build();
+  private static final ImmutableSet<String> SECONDARY_INDEX_SCALING_TYPE =
+      ImmutableSet.<String>builder().add("index-read").add("index-write").build();
```
feeblefakie, line 97: Can you make "read", "write", "index-read", and "index-write" constant variables?

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,386 @@
...
+  private static final ImmutableMap<DataType, ScalarAttributeType> DATATYPE_MAPPING =
+      ImmutableMap.<DataType, ScalarAttributeType>builder()
+          .put(DataType.INT, ScalarAttributeType.N)
+          .put(DataType.BIGINT, ScalarAttributeType.N)
+          .put(DataType.FLOAT, ScalarAttributeType.N)
+          .put(DataType.DOUBLE, ScalarAttributeType.N)
+          .put(DataType.TEXT, ScalarAttributeType.S)
+          .put(DataType.BLOB, ScalarAttributeType.B)
+          .build();
+  private static final ImmutableSet<String> TABLE_SCALING_TYPE =
```
feeblefakie, line 94: The variable name doesn't sound like a set.
Can you make it `TABLE_SCALING_TYPE_SET`?

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,386 @@
...
+      ImmutableSet.<String>builder().add("read").add("write").build();
+  private static final ImmutableSet<String> SECONDARY_INDEX_SCALING_TYPE =
```
feeblefakie, line 96: Ditto.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,386 @@
...
+      ImmutableSet.<String>builder().add("index-read").add("index-write").build();
+  private static final ImmutableMap<String, ScalableDimension> SCALABLE_DIMENSION_MAPPING =
```
feeblefakie, line 98: I think MAP is more consistent with SET. (if we use SET as I suggested)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,386 @@
...
+            .attributeType(ScalarAttributeType.S)
+            .build());
+    if (!metadata.getClusteringKeyNames().isEmpty()) {
+      columnsToAttributeDefinitions.add(
+          AttributeDefinition.builder()
+              .attributeName(CLUSTERING_KEY)
+              .attributeType(ScalarAttributeType.S)
+              .build());
+      for (String column : metadata.getClusteringKeyNames()) {
+        if (metadata.getColumnDataType(column) == DataType.BOOLEAN) {
```
feeblefakie, line 184: The code block in the for loop looks identical to the one below for the secondary index.
Can you unify it?

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,386 @@
...
+          throw new ExecutionException(
+              "BOOLEAN type is not supported for a clustering key or a secondary index in DynamoDB");
```
feeblefakie, line 186: This limitation is still there? I did a quick search but I can't find it.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,386 @@
...
-  public void createNamespace(String namespace, Map<String, String> options)
-      throws ExecutionException {
-    throw new UnsupportedOperationException("implement later");
+  public void createNamespace(String namespace, Map<String, String> options) {
+    // In Dynamo DB storage, namespace will be added to table name as prefix along with dot
+    // separator.
   }
 
   @Override
   public void createTable(
```
feeblefakie, line 166: I think this method is too big.
The code seems to have several logical units (for each component to build), so can you make them turn to methods and call them to make this method slim?

feeblefakie (commented):
Thank you!
Let me start off by leaving comments on the superficial parts of the code.
I will look into it later.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: It uses `index-read` here. Thanks for your suggestions. I change to use `Map`, applied in [166ec54](https://github.com/scalar-labs/scalardb/pull/301/commits/166ec546eb44ccd9bdc33932e43e2deabfb79cda), other suggestions also were applied as well

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: It still is a limitation. There is no attribute type like `ScalarAttributeType..BOOL` came from SDK.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,330 @@
...
 
   @Override
   public void dropTable(String namespace, String table) throws ExecutionException {
-    throw new UnsupportedOperationException("implement later");
+    DeleteTableRequest request =
+        DeleteTableRequest.builder()
+            .tableName(getFullTableName(namespacePrefix, namespace, table))
+            .build();
+    try {
+      client.deleteTable(request);
```
brfrn169, line 275: Looks like the original schema loader disables auto scaling before dropping the table?
https://github.com/scalar-labs/scalardb/blob/98fc5103dd10d2a2fb60ea18f888da443d362668/tools/scalar-schema/src/scalar_schema/dynamo.clj#L228-L229

brfrn169 (commented):
Thank you! Just left some comments. Please take a look when you have time!

I found one different behavior from the original schema loader. Just in case, can you please re-check if there are any other different behaviors in this new implementation? And if you find them, please correct them. Thank you.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Thanks for pointing it out. I update the code to support this [4af745c](https://github.com/scalar-labs/scalardb/pull/301/commits/4af745cac2700265b8116b2a25ce0a98dfdbea9e)
Since it's an option that comes from CLI, and we don't have optional params for dropping table so for now dropping table will try to disable auto-scaling first even auto-scaling not existed already and drop table later.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -127,19 +156,174 @@ private DataType convertDataType(String columnType) {
 
   @Override
   public void deleteTableMetadata(String namespace, String table) {
-    // TODO To implement
-    throw new UnsupportedOperationException();
+
```
brfrn169, line 159:
```suggestion
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
brfrn169: Thank you for handle this!

> Since it's an option that comes from CLI, and we don't have optional params for dropping table so for now dropping table will try to disable auto-scaling first even auto-scaling not existed already and drop table later.

So, when dropping a table created with `no-scaling` option, it looks like it shows a warning message, correct? If it's correct, I don't think we should show the message because dropping a table created with `no-scaling` option is a usual way, and the warning message is confusing.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -86,18 +343,331 @@ public TableMetadata getTableMetadata(String namespace, String table) throws Exe
...
+    return getFullTableName(namespacePrefix, namespace, tableName)
+        + "."
+        + GLOBAL_INDEX_NAME_PREFIX
+        + "."
+        + keyName;
+  }
+
   @Override
   public void close() {
     client.close();
```
brfrn169, line 673:
```suggestion
    client.close();
    applicationAutoScalingClient.close();
```

brfrn169 (commented):
Thank you! I left one question and some minor suggestions. Please check them when you have time!

BTW, did you re-check if there are any other different behaviors from the original schema loader in this new implementation? And was it good?

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: The warning message is only shown when the exception is not `ObjectNotFoundException`, so in case the table is created with `no-scaling` option then the dropping table will not show any warning message.

thongdk8:
> BTW, did you re-check if there are any other different behaviors from the original schema loader in this new implementation? And was it good?

Thank you. I did check the behaviors, they are fine.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,335 @@
...
+  private static final int COOL_TIME_SEC = 60;
+  private static final double TARGET_USAGE_RATE = 70.0;
+  private static final int DELETE_BATCH_SIZE = 100;
+
+  private static final String NO_SCALING = "no-scaling";
+  private static final String NO_BACKUP = "no-backup";
+  private static final String REQUEST_UNIT = "ru";
+  private static final Boolean DEFAULT_NO_SCALING = false;
+  private static final Boolean DEFAULT_NO_BACKUP = false;
+  private static final long DEFAULT_RU = 10;
```
brfrn169, line 87: Please make those `public`:
```suggestion
  public static final String NO_SCALING = "no-scaling";
  public static final String NO_BACKUP = "no-backup";
  public static final String REQUEST_UNIT = "ru";
  public static final Boolean DEFAULT_NO_SCALING = false;
  public static final Boolean DEFAULT_NO_BACKUP = false;
  public static final long DEFAULT_RU = 10;
```

brfrn169 (commented):
LGTM except what I commented. Thank you!

brfrn169 (dismissed):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
feeblefakie: @thongdk8 Somehow it is resolved without checked but was it difficult or it was unintentionally resolved by another change?

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -86,20 +343,334 @@ public TableMetadata getTableMetadata(String namespace, String table) throws Exe
...
+      try {
+        applicationAutoScalingClient.deleteScalingPolicy(deleteScalingPolicyRequest);
+      } catch (ApplicationAutoScalingException e) {
+        if (!(e instanceof ObjectNotFoundException)) {
+          LOGGER.warn(
+              "Delete scaling policy " + deleteScalingPolicyRequest.policyName() + " failed. " + e);
+        }
+      }
+    }
+    for (DeregisterScalableTargetRequest deregisterScalableTargetRequest :
```
feeblefakie, line 571: Please put an empty line above.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -127,19 +156,173 @@ private DataType convertDataType(String columnType) {
 
   @Override
   public void deleteTableMetadata(String namespace, String table) {
-    // TODO To implement
-    throw new UnsupportedOperationException();
+    HashMap<String, AttributeValue> keyToDelete = new HashMap<>();
```
feeblefakie, line 159:
```suggestion
    Map<String, AttributeValue> keyToDelete = new HashMap<>();
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -127,19 +156,173 @@ private DataType convertDataType(String columnType) {
...
+            .build());
+    requestBuilder.tableName(getFullTableName(namespacePrefix, METADATA_NAMESPACE, METADATA_TABLE));
+
+    try {
+      client.createTable(requestBuilder.build());
+    } catch (DynamoDbException e) {
+      throw new StorageRuntimeException("creating meta data table failed");
+    }
+
+    while (true) {
```
feeblefakie, line 295: This block is redundant with the block in Admin.
(CREATING_WAITING_TIME is also redundantly defined)
It would be great if the redundancy is removed.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,335 @@
...
+import software.amazon.awssdk.services.dynamodb.model.TableStatus;
+import software.amazon.awssdk.services.dynamodb.model.UpdateContinuousBackupsRequest;
 
 @ThreadSafe
 public class DynamoAdmin implements DistributedStorageAdmin {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DynamoAdmin.class);
+  private static final String PARTITION_KEY = "concatenatedPartitionKey";
+  private static final String CLUSTERING_KEY = "concatenatedClusteringKey";
+  private static final String GLOBAL_INDEX_NAME_PREFIX = "global_index";
+  private static final int CREATING_WAITING_TIME = 3000;
```
feeblefakie, line 77: The variable names are not consistent.
It is is `ms` then should it end with `_MS` or something to be consistent with `_SEC`?

I would suggest to have the same time unit and adjust for each case like below.
```
CREATE_WAIT_DURATION_MILLIS
COOL_DOWN_DURATION_MILLIS
```

feeblefakie (commented):
Left additional comments and suggestions.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Thank you. I did split the method to make it slimmer in [166ec54](https://github.com/scalar-labs/scalardb/commit/166ec546eb44ccd9bdc33932e43e2deabfb79cda).
This was marked as resolved because I did apply it, I mentioned [here](https://github.com/scalar-labs/scalardb/pull/301#discussion_r705922779). All the suggestions that were marked as resolved were applied.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Apply in [391f30b](https://github.com/scalar-labs/scalardb/commit/391f30b54026c24a765bd56b672570feca195ed3)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
thongdk8: [391f30b](https://github.com/scalar-labs/scalardb/commit/391f30b54026c24a765bd56b672570feca195ed3)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
thongdk8: Yes, indeed. I created a static function for waiting table creation. Applied in [391f30b](https://github.com/scalar-labs/scalardb/commit/391f30b54026c24a765bd56b672570feca195ed3)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Cause the cooldown time is accepted in seconds so I change variables' name with postfix `_SECS`. Apply in [391f30b](https://github.com/scalar-labs/scalardb/commit/391f30b54026c24a765bd56b672570feca195ed3)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,325 @@
...
+  private static final int COOL_DOWN_DURATION_SECS = 60;
+  private static final double TARGET_USAGE_RATE = 70.0;
+  private static final int DELETE_BATCH_SIZE = 100;
+
+  public static final String NO_SCALING = "no-scaling";
+  public static final String NO_BACKUP = "no-backup";
+  public static final String REQUEST_UNIT = "ru";
+  public static final Boolean DEFAULT_NO_SCALING = false;
+  public static final Boolean DEFAULT_NO_BACKUP = false;
+  public static final long DEFAULT_RU = 10;
```
brfrn169, line 87:
```suggestion
  public static final String DEFAULT_NO_SCALING = "false";
  public static final String DEFAULT_NO_BACKUP = "false";
  public static final String DEFAULT_RU = "10";
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,325 @@
...
+    buildPrimaryKey(requestBuilder, metadata);
+
+    // build local indexes that corresponding to clustering keys
+    buildLocalIndexes(namespace, table, requestBuilder, metadata);
+
+    // build secondary indexes
+    long ru = DEFAULT_RU;
+    if (options.containsKey(REQUEST_UNIT)) {
+      ru = Long.parseLong(options.get(REQUEST_UNIT));
+    }
```
brfrn169, line 215:
```suggestion
    long ru = Long.parseLong(options.getOrDefault(REQUEST_UNIT, DEFAULT_RU));
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,325 @@
...
+      throw new ExecutionException("getting table description failed", e);
+    }
+
+    // scaling control
+    boolean noScaling = DEFAULT_NO_SCALING;
+    if (options.containsKey(NO_SCALING)) {
+      if (Boolean.parseBoolean(options.get(NO_SCALING))) {
+        noScaling = true;
+      }
+    }
```
brfrn169, line 245:
```suggestion
    boolean noScaling = Boolean.parseBoolean(options.getOrDefault(NO_SCALING, DEFAULT_NO_SCALING));
```

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,325 @@
...
+      enableAutoScaling(namespace, table, metadata.getSecondaryIndexNames(), ru);
+    }
+
+    // backup control
+    boolean noBackup = DEFAULT_NO_BACKUP;
+    if (options.containsKey(NO_BACKUP)) {
+      if (Boolean.parseBoolean(options.get(NO_BACKUP))) {
+        noBackup = true;
+      }
+    }
```
brfrn169, line 256:
```suggestion
    boolean noBackup = Boolean.parseBoolean(options.getOrDefault(NO_BACKUP, DEFAULT_NO_BACKUP));
```

brfrn169 (commented):
@thongdk8 Sorry, I missed one thing. Just left some suggestions. Please check them!

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Apply in [19938a0](https://github.com/scalar-labs/scalardb/pull/301/commits/19938a073c0e7ed103d9d226c6f128cbaa6cfdff)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Apply in [19938a0](https://github.com/scalar-labs/scalardb/pull/301/commits/19938a073c0e7ed103d9d226c6f128cbaa6cfdff)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Apply in [19938a0](https://github.com/scalar-labs/scalardb/pull/301/commits/19938a073c0e7ed103d9d226c6f128cbaa6cfdff)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1,78 +1,312 @@
...
+  private static final int COOL_DOWN_DURATION_SECS = 60;
+  private static final double TARGET_USAGE_RATE = 70.0;
+  private static final int DELETE_BATCH_SIZE = 100;
+
+  public static final String NO_SCALING = "no-scaling";
+  public static final String NO_BACKUP = "no-backup";
+  public static final String REQUEST_UNIT = "ru";
+  public static final String DEFAULT_NO_SCALING = "false";
+  public static final String DEFAULT_NO_BACKUP = "false";
+  public static final String DEFAULT_RU = "10";
```
brfrn169, line 87: I'm very sorry, one more thing. Can you please rename it to `DEFAULT_REQUEST_UNIT`:
```suggestion
  public static final String DEFAULT_REQUEST_UNIT = "10";
```

thongdk8: Apply in [1d82dc8](https://github.com/scalar-labs/scalardb/commit/1d82dc8225ed711c766f30acad9e4764e081d40e)

brfrn169 (dismissed):
LGTM! Thank you for your patience.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
feeblefakie: I meant, please reply back to all the comments. Otherwise, it's pretty hard to identify since there are many comments.

thongdk8: I see. I'll do it from now on. Sorry about that.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -86,20 +320,347 @@ public TableMetadata getTableMetadata(String namespace, String table) throws Exe
...
+      String resourceID, String type) {
+    return DeleteScalingPolicyRequest.builder()
+        .serviceNamespace(ServiceNamespace.DYNAMODB)
+        .resourceId(resourceID)
+        .scalableDimension(SCALABLE_DIMENSION_MAP.get(type))
+        .policyName(getPolicyName(resourceID, type))
+        .build();
+  }
+
+  protected static void waitForTableCreation(DynamoDbClient client, String tableFullName) {
```
feeblefakie, line 606: Can it be package private instead of `protected`?
(Any reason for protected?)

thongdk8: It was used in `DynamoTableMetadataManager` so I made it protected.

feeblefakie: So, it can be package-private, right? 
I've checked the code and `DynamoTableMetadataManager` n the same package is the only one that uses this method except for this Admin class, so please make it package-private.
Restricted scope is usually better.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -1,22 +1,44 @@
 package com.scalar.db.storage.dynamo;
 
+import static com.scalar.db.storage.dynamo.DynamoAdmin.waitForTableCreation;
```
feeblefakie, line 3: Let's not use static import for this.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
```diff
@@ -25,11 +47,14 @@
  */
```
feeblefakie, line 47: It's basically rewritten, so please put your name on it!

feeblefakie (commented):
Looking good except for the comments I additionally made.
Thank you!

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
thongdk8: Thank you. I removed the `protected` identifier [90037fd]( https://github.com/scalar-labs/scalardb/pull/301/commits/90037fdc8400d01e7387d17c33f4069dc9c02c9a)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
thongdk8: Updated in [90037fd]( https://github.com/scalar-labs/scalardb/pull/301/commits/90037fdc8400d01e7387d17c33f4069dc9c02c9a)

core/src/main/java/com/scalar/db/storage/dynamo/DynamoTableMetadataManager.java
thongdk8: Updated in [90037fd]( https://github.com/scalar-labs/scalardb/pull/301/commits/90037fdc8400d01e7387d17c33f4069dc9c02c9a)

feeblefakie (approved):
LGTM! Thank you!

PR #302, opened 2021-09-08
Support Two-phase Commit Transactions in Scalar DB server

brfrn169:
This PR adds Two-phase Commit Transactions support in Scalar DB server. Please take a look!


server/src/main/java/com/scalar/db/server/service/ServerModule.java
```diff
@@ -22,7 +25,11 @@
   public ServerModule(ServerConfig config, DatabaseConfig databaseConfig) {
     this.config = config;
     storageFactory = new StorageFactory(databaseConfig);
-    transactionFactory = new TransactionFactory(databaseConfig);
+
+    // For two-phase consensus commit in Scalar DB server, set MANAGE_ACTIVE_TRANSACTIONS to false
```
feeblefakie, line 29: I think it would be better to explain the reason for doing this in addition to what it is doing.
```suggestion
    // For two-phase consensus commit in Scalar DB server, set MANAGE_ACTIVE_TRANSACTIONS to false because Scalar DB server takes care of active transactions management
```
Is it correct?

server/src/main/java/com/scalar/db/server/ScalarDbServer.java
```diff
@@ -58,16 +59,23 @@ public void start() throws IOException {
...
             .addService(injector.getInstance(AdminService.class))
             .addService(new HealthService())
-            .addService(ProtoReflectionService.newInstance())
-            .build()
-            .start();
+            .addService(ProtoReflectionService.newInstance());
+
+    // Two-phase commit for JDBC is not supported for now
+    String transactionManager = properties.getProperty(DatabaseConfig.TRANSACTION_MANAGER);
+    if (Strings.isNullOrEmpty(transactionManager) || !transactionManager.equals("jdbc")) {
```
feeblefakie, line 74: What would happen if `jdbc` is specified for the transaction manager?
I think it should throw an exception but does it?

rpc/src/main/proto/scalardb.proto
```diff
@@ -339,3 +339,92 @@ message AbortRequest {
...
 }
+
+// TwoPhaseCommitTransaction
+
+service TwoPhaseCommitTransaction {
+  rpc TwoPhaseCommitTransaction(stream TwoPhaseCommitTransactionRequest) returns (stream TwoPhaseCommitTransactionResponse) {
+  }
+  rpc GetState(GetTransactionStateRequest) returns (GetTransactionStateResponse) {
+  }
+  rpc Abort(AbortRequest) returns (AbortResponse) {
```
feeblefakie, line 350: I'm sorry I didn't notice it last time, but what is the reason for having `Abort` rpc separately from `TwoPhaseCommitTransaction`?

core/src/main/java/com/scalar/db/transaction/rpc/GrpcTwoPhaseCommitTransactionOnBidirectionalStream.java
```diff
@@ -0,0 +1,397 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+public class GrpcTwoPhaseCommitTransactionOnBidirectionalStream
```
feeblefakie, line 52: NotThreadSafe?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
```diff
@@ -17,6 +17,13 @@
 
   private SerializableStrategy strategy;
 
+  // for two-phase consensus commit
+  public static final String TWO_PHASE_CONSENSUS_COMMIT_PREFIX = PREFIX + "2pcc.";
+  public static final String MANAGE_ACTIVE_TRANSACTIONS =
+      TWO_PHASE_CONSENSUS_COMMIT_PREFIX + "manage_active_transactions";
```
feeblefakie, line 23:
```suggestion
      TWO_PHASE_CONSENSUS_COMMIT_PREFIX + "active_transactions_management.enabled";
```
`.enabled` is common for boolean? (not fully sure but I used it for Scalar DL)

feeblefakie (commented):
I left minor suggestions and comments.

server/src/main/java/com/scalar/db/server/ScalarDbServer.java
brfrn169: In the current implementation, `TwoPhaseCommitTransactionService` doesn't start.

rpc/src/main/proto/scalardb.proto
brfrn169: I think that's the same reason why we have the `Abort` method in `DistributedTransaction`:
https://github.com/scalar-labs/scalardb/pull/238

brfrn169 (commented):
@feeblefakie Thank you for taking look at this!

I left some replies for some of your comments and I will push the fixes based on your review shortly. Thanks.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
```diff
@@ -17,6 +17,13 @@
 
   private SerializableStrategy strategy;
 
+  // for two-phase consensus commit
+  public static final String TWO_PHASE_CONSENSUS_COMMIT_PREFIX = PREFIX + "2pcc.";
+  public static final String ACTIVE_TRANSACTIONS_MANAGEMENT_ENABLED =
+      TWO_PHASE_CONSENSUS_COMMIT_PREFIX + "active_transactions_management_enabled";
```
feeblefakie, line 23: Sorry, I meant `.enabled` like this.
https://spark.apache.org/docs/latest/security.html#encryption

For this, I don't think it is very meaningful since no other options are available for `active_transactions_management` but I think this rule is useful when we have some other options like `active_transactions_management.xxx`.
What do you think?

server/src/main/java/com/scalar/db/server/ScalarDbServer.java
feeblefakie: Just a confirmation. It doesn't start and throws some exception?

feeblefakie (commented):
Thank you!
Left an additional comment and question.

server/src/main/java/com/scalar/db/server/ScalarDbServer.java
brfrn169: It doesn't throw any exception when `jdbc` is specified for the transaction manager. Other services (DistributedStorageService, DistributedStorageAdminService, DistributedTransactionService) start in that case.

feeblefakie (approved):
LGTM! Thank you! (can be merged after the warning message is added)

feeblefakie (approved):

PR #308, opened 2021-09-15
Upgrade the Error Prone plugin

brfrn169:
This PR upgrades the Error Prone plugin. Actually, It is a trivial thing but my intelliJ shows warning for it. Please take a look!


feeblefakie (approved):
LGTM!

PR #312, opened 2021-09-18
Add utility class for transactional tables

brfrn169:
This PR adds a utility class that converts a table metadata to a transactional one. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
@@ -10,11 +10,13 @@ import com.scalar.db.api.Get;
 import com.scalar.db.api.Put;
 import com.scalar.db.api.PutIfNotExists;
 import com.scalar.db.api.Result;
+import com.scalar.db.api.TableMetadata;
 import com.scalar.db.api.TransactionState;
 import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.exception.storage.NoMutationException;
 import com.scalar.db.exception.transaction.CoordinatorException;
 import com.scalar.db.exception.transaction.RequiredValueMissingException;
+import com.scalar.db.io.DataType;
 import com.scalar.db.io.Key;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -28,6 +30,15 @@
 public class Coordinator {
   public static final String NAMESPACE = "coordinator";
   public static final String TABLE = "state";
+  public static final TableMetadata TABLE_METADATA =
+      TableMetadata.newBuilder()
+          .addColumn(Attribute.ID, DataType.TEXT)
+          .addColumn(Attribute.STATE, DataType.INT)
+          .addColumn(Attribute.CREATED_AT, DataType.BIGINT)
+          .addColumn(Attribute.METADATA, DataType.TEXT)
```
brfrn169, line 38: @feeblefakie Do we need this `METADATA` column for the Coordinator table, correct?

feeblefakie: We don't use it currently, so we don't need it.
But, I added this for schema flexibility by storing a serialized JSON string or something.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTable.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private TransactionalTable() {}
+
+  /**
+   * Converts a table metadata to a transactional one
+   *
+   * @param tableMetadata a table metadata to be converted
+   * @return a transactional table metadata
+   */
+  public static TableMetadata convertToTransactionalTable(TableMetadata tableMetadata) {
```
feeblefakie, line 31: Should `TableMetadata` have a `transactional` boolean flag or something to make it safer even if the same tablemetadata object is applied to this method multiple times? (it would add `before_before_column1` or something)
It is not gonna happen probably, but I just feel it might be safer and it might be useful if we can  easily check a given tablemetadata object is transactional or not. (instead of checking if there is a `before_` column)
What do you think?

feeblefakie (commented):
Thank you!
Overall looking good! Left one minor comment/suggestion.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
brfrn169: Okay, so do we need to remove the `METADATA` related things in another PR?

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTable.java
brfrn169: The transactional table things are `Consensus Commit` implementation specific things, so I don't think `TableMetadata` should have a `transactional` boolean flag or something like that.

I'm wondering if we really need to take care of the situation where the same `TableMetadata` object is applied to the method multiple times. Or is there any situation we need to check if a `TableMetadata` is transactional or not? If we need to do that, I think we need to check if a `TableMetadata` has transaction meta columns or not.

brfrn169: I added a check if a TableMetadata has transaction meta columns or not in 5e2f7d9d77496a74e74a6bfed6ef051980986be8. Please take a look!

brfrn169:
I came up with another idea for this, so closing it and create another PR. Thanks.

PR #314, opened 2021-09-23
Add ConsensusCommitAdmin that manages transactional tables and coordinator table

brfrn169:
This PR adds `ConsensusCommitAdmin` that creates transactional tables and a coordinator table. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #317, opened 2021-09-28
Refactor XXXAdmin

brfrn169:
What I did in this PR is:
- Unify `XXXAdmin` and `XXXTableMetadataManager` because the difference of the two roles are getting ambiguous
- Make the role of `TableMetadataManager` only for caching table metadata
- Remove `ConnectionException` and `StorageRuntimeException`

Please take a look!


core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
```diff
@@ -173,7 +173,9 @@ public void mutate(List<? extends Mutation> mutations) throws ExecutionException
 
     Utility.setTargetToIfNot(mutations, namespacePrefix, namespace, tableName);
     operationChecker.check(mutations);
-    mutations.forEach(operationChecker::check);
+    for (Mutation mutation : mutations) {
```
feeblefakie, line 176: Any particular reason for not using `forEach`?

feeblefakie (dismissed):
LGTM! Thank you! It looks better 👍 
I left one minor comment but it doesn't block it from being merged.

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
brfrn169: Thank you for taking at this!

After this change, `operationChecker::check` throws `ExecutionException`.  `forEach` doesn't handle Exception well, I changed this way.

PR #321, opened 2021-09-29
Add Table metadata cache expiration configuration

brfrn169:
This PR allows us to set Table metadata cache expiration configuration with the `scalar.db.table_metadata.cache_expiration_time_secs` property. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
@@ -10,6 +10,7 @@ import com.scalar.db.config.DatabaseConfig;
 import com.scalar.db.exception.storage.ExecutionException;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 
 /** A class that manages and caches table metadata */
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
```diff
@@ -18,16 +19,19 @@
   private final LoadingCache<TableKey, Optional<TableMetadata>> tableMetadataCache;
 
   public TableMetadataManager(DistributedStorageAdmin admin, DatabaseConfig config) {
+    CacheBuilder<Object, Object> builder = CacheBuilder.newBuilder();
+    long tableMetadataCacheExpirationTimeSecs = config.getTableMetadataCacheExpirationTimeSecs();
+    if (tableMetadataCacheExpirationTimeSecs > 0) {
+      builder.expireAfterWrite(tableMetadataCacheExpirationTimeSecs, TimeUnit.SECONDS);
```
feeblefakie, line 25: I'm not really sure about the use case of specifying the expiration time. Can you clarify?
From users' perspective, it seems like it should be cached forever or it should be replaced right after it needs to 
 be.

feeblefakie (commented):
Thank you!
I left one comment since I'm not sure about the use case.

core/src/main/java/com/scalar/db/storage/common/TableMetadataManager.java
brfrn169: The use case is when the table metadata change happens. As we don't have `alter table` feature for now, `create table` and `delete table` are only the use cases. Anyway, I think we will need to handle `alter table` cases in the future, so this cache expiration feature will become more important.

feeblefakie (approved):
LGTM! Thank you!
But the CIs are failing. Can you take a look?

brfrn169:
They were successful after rerunning the CIs. I'm merging this.

PR #322, opened 2021-09-29
Make table metadata database name configurable in Cosmos DB adapter

brfrn169:
Currently, the table metadata database name in the Cosmos DB adapter is fixed `scalardb`. In some cases, we need to change it. This PR allows us to do that with the `scalar.db.cosmos.table_metadata.database` property. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cosmos/CosmosConfig.java
+++ b/core/src/main/java/com/scalar/db/storage/cosmos/CosmosConfig.java
@@ -1,0 +1,52 @@
+package com.scalar.db.storage.cosmos;
+
+import com.google.common.base.Strings;
+import com.scalar.db.config.DatabaseConfig;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Optional;
+import java.util.Properties;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+@SuppressFBWarnings("JCIP_FIELD_ISNT_FINAL_IN_IMMUTABLE_CLASS")
+public class CosmosConfig extends DatabaseConfig {
+
+  public static final String PREFIX = DatabaseConfig.PREFIX + "cosmos.";
+  public static final String TABLE_METADATA_DATABASE = PREFIX + "table_metadata.database";
+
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/scalar/db/storage/cosmos/CosmosConfig.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Optional;
+import java.util.Properties;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+@SuppressFBWarnings("JCIP_FIELD_ISNT_FINAL_IN_IMMUTABLE_CLASS")
+public class CosmosConfig extends DatabaseConfig {
+
+  public static final String PREFIX = DatabaseConfig.PREFIX + "cosmos.";
```
feeblefakie, line 18: I'm wondering if the metadata configuration should be done per database implementation.
Is it separated to make different configurations for each database when using multi-storage?
(e.g., cosmos uses `cosmos_metadata` and dynamo uses `dynamo_metadata`?)

feeblefakie (commented):
Thank you!
Left some clarifying questions. PTAL!

core/src/main/java/com/scalar/db/storage/cosmos/CosmosConfig.java
```diff
@@ -0,0 +1,52 @@
...
+  public static final String TABLE_METADATA_DATABASE = PREFIX + "table_metadata.database";
```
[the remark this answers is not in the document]
brfrn169: In the Cosmos DB context, I think it should be `database`. Similarly, it's `schema` in the JDBC context. Regarding DynamoDB, it doesn't have `namespace` level concept, so I just use `namespace` wording.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosConfig.java
brfrn169: Yes, the metadata configuration is storage implementation specific. Actually, Cassandra doesn't have the metadata table. And maybe even when using multi-storage, I think we should have separate configurations because users might want to have different metadata names for each storage.

brfrn169 (commented):
@feeblefakie Thank you for taking at this! I left some replies for your comments. Please take a look at them!

feeblefakie (approved):
LGTM! Thank you!

PR #323, opened 2021-09-29
Make table metadata namespace name configurable in DynamoDB adapter

brfrn169:
Currently, the table metadata namespace name in the DynamoDB adapter is fixed `scalardb`. This PR allows us to change it with the `scalar.db.dynamo.table_metadata.namespace` property. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #324, opened 2021-09-29
Make table metadata schema name configurable in JDBC adapter

brfrn169:
Currently, the table metadata schema name in the JDBC adapter is fixed `scalardb`. This PR allows us to change it with the `scalar.db.jdbc.table_metadata.schema` property. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #327, opened 2021-09-30
Correct comparator of item sorter in Dynamo DB adapter

thongdk8:
Current implementation using byte buffer comparator so it makes the result when scanning with clustering key of blob values not correctly. This PR made for correct the comparator. Please take a look. Thank you.


brfrn169 (dismissed):
LGTM! Good catch! Thank you!

brfrn169:
@thongdk8 Sorry, I forgot one thing. Can you please add a unit test for this? We basically need a unit test to make sure the bug is really fixed in this kind of case.

thongdk8:
@brfrn169 We had a unit test for this [here](https://github.com/scalar-labs/scalardb/blob/9619343a7192034002964dd69c10bd48f2a581ef/core/src/test/java/com/scalar/db/storage/dynamo/ItemSorterTest.java#L264-L285) but it seems that it did not cover well. I made some changes in this [commit](https://github.com/scalar-labs/scalardb/pull/327/commits/07f45ecc6710e5ca857ca4f584679213eb6cd2af), after changing this test case failed with the previous comparator and passed with the new corrected one.

core/src/test/java/com/scalar/db/storage/dynamo/ItemSorterTest.java
```diff
@@ -264,10 +266,27 @@ public void sort_ItemsGivenWithDescOrderingText_ShouldSortProperly() {
   @Test
   public void sort_ItemsGivenWithDescOrderingBlob_ShouldSortProperly() {
     // Arrange
+    Random random = new Random(777);
```
feeblefakie, line 269: I'm just wondering if there is any benefit of using Random with a fixed seed here.
I feel it's better to use fixed hard-coded byte arrays so that we (people) can see what is the expected result.
What do you think?

feeblefakie (commented):
Nice work! Thank you.
Left 1 comment. PTAL:bow:

core/src/test/java/com/scalar/db/storage/dynamo/ItemSorterTest.java
thongdk8: I see. I had been thought about using a loop for the test with many values, so I used Random. Actually, I see some libraries still using Random for testing, so I thought it's common. For example from [guava](https://github.com/google/guava/blob/8075df7ffd63b4b96cd0bdfdc2dde71d08f672c9/guava-tests/test/com/google/common/primitives/UnsignedBytesTest.java#L288-L306)
I made to use fixed hard-coded byte arrays in this [commit](https://github.com/scalar-labs/scalardb/pull/327/commits/0b85f8712e036c51b4f94a71e169b296892f494f). Thank you for your suggestion.

feeblefakie: It makes more sense if we test many cases using loop and random but the test was only comparing two byte-arrays, that's why I asked.

feeblefakie (approved):
LGTM! Thank you!

core/src/test/java/com/scalar/db/storage/dynamo/ItemSorterTest.java
thongdk8: That's right. So should we random and for loop here or keep it like the new commit I made?

brfrn169 (approved):
LGTM! Thank you!

core/src/test/java/com/scalar/db/storage/dynamo/ItemSorterTest.java
feeblefakie: I think the current code is fine and consistent with other tests.
If you feel it's not tested well, we can add more tests with loop and random. (please add instead of update in such a case)

brfrn169: @thongdk8 I will merge this. If you want add more tests with loop and random, please create a new PR for that. Thanks.

thongdk8: Got it. Thank you.

PR #330, opened 2021-10-03
Add ifExists support to dropNamespace() and dropTable() in DistributedStorageAdmin

brfrn169:
I think `ifExists` support for `dropNamespace()`/`dropTable()` in `DistributedStorageAdmin` is useful for many use cases. This PR adds it. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #320, opened 2021-09-29
Make Coordinator namespace name configurable

brfrn169:
Currently, the table metadata database name is fixed `coordinator`. This PR makes it configurable with the `scalar.db.consensus_commit.coordinator.namespace` property. Please take a look!

PR #331, opened 2021-10-04
Make Coordinator backward compatible

brfrn169:
#320 changed the signature of the Coordinator's constructor, which breaks backward compatibility. This PR will add a constructor that has the original signature again and fix the issue. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #333, opened 2021-10-04
Remove namespace prefix

brfrn169:
This PR removes the namespace prefix things. Note that I don't remove the namespace prefix things from the schema loader because it will be replaced with a new schema tool soon. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #337, opened 2021-10-06
FloatValue should be mapped to double type in MySQL

brfrn169:
I found `FloatValue` should be mapped to `double` type in MySQL as mentioned in the following:
https://stackoverflow.com/questions/46247940/why-is-mysql-float-different-than-java-float

This PR fixes the issue. Note that I didn't fix the issue in the schema loader as it will be removed and replaced with the new one soon. Please take a look!


thongdk8 (approved):
LGTM! Thank you!

PR #344, opened 2021-10-15
GraphQL schema generation from database tables

ymorimo:
The components for generating GraphQL schema from the Scalar DB table metadata.

There are two major classes in this PR:

* `TableGraphQLModel` holds the entire GraphQL types that are needed to represent the database columns and operations for a single database table. It should be instantiated for each table.
* `GraphQLFactory` is the factory for the [`grapqhql-java`](https://www.graphql-java.com/documentation/v17/)'s `GraphQL` object. The `GraphQL` object will serve as the interface for executing queries sent by the user.

At this point, the generated GraphQL object cannot run any query or mutation since no `DataFetcher`s (resolvers) have been associated with each type yet. However, the schema can be printed as shown below:

```java
System.out.println(new SchemaPrinter().print(graphql.getGraphQLSchema()));
```

For the GraphQL feature, the subsequent PRs are going to be against the `graphql` branch.


feeblefakie (commented):
I left minor comments.
I will take a relook again later.

graphql/src/main/java/com/scalar/db/graphql/GraphQLFactory.java
```diff
@@ -0,0 +1,139 @@
...
+import graphql.Scalars;
+import graphql.schema.GraphQLCodeRegistry;
+import graphql.schema.GraphQLFieldDefinition;
+import graphql.schema.GraphQLObjectType;
+import graphql.schema.GraphQLSchema;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public class GraphQLFactory {
```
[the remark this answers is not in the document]
ymorimo: Thanks! Renamed in c507662.

graphql/src/main/java/com/scalar/db/graphql/schema/TableGraphQlModel.java
```diff
@@ -0,0 +1,436 @@
...
+    this.tableMetadata = Objects.requireNonNull(tableMetadata);
+
+    this.transactionEnabled = tableMetadata.getColumnNames().contains(ID);
+    this.fieldNames =
+        tableMetadata.getColumnNames().stream()
+            .filter(
+                colName ->
+                    !CONSENSUSCOMMIT_ATTRIBUTES.contains(colName)
+                        && !colName.startsWith(BEFORE_PREFIX))
+            .collect(toList());
```
brfrn169, line 82: Also we can use `ConsensusCommitUtils.removeTransactionalMetaColumns()` here:
```suggestion
    this.fieldNames =
        new ArrayList<>(
            ConsensusCommitUtils.removeTransactionalMetaColumns(tableMetadata).getColumnNames());
```

brfrn169 (commented):
Thank you! I left several comments and suggestions. Please take a look when you have time!

graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
```diff
@@ -0,0 +1,139 @@
...
+
+    public Builder table(String namespace, String table) {
+      namespaces.add(namespace);
+      tables.add(table);
+      return this;
+    }
+
+    public GraphQlFactory build() throws ExecutionException {
+      if (storageFactory == null) {
+        throw new IllegalArgumentException("Need to specify storageFactory");
```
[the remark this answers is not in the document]
feeblefakie: Ah, `IllegalStateException` is better here. It used to be that but I made a wrong suggestion. @ymorimo Sorry!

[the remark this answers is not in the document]
ymorimo: Okay, I'll revert the change. Thanks!

graphql/src/main/java/com/scalar/db/graphql/schema/TableGraphQLModel.java
```diff
@@ -0,0 +1,436 @@
...
+import graphql.schema.GraphQLObjectType;
+import graphql.schema.GraphQLScalarType;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+public class TableGraphQLModel {
+  private static final Set<String> CONSENSUSCOMMIT_ATTRIBUTES =
```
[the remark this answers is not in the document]
ymorimo: This constant has been removed thanks to `ConsensusCommitutils`. 🙏

graphql/src/main/java/com/scalar/db/graphql/schema/TableGraphQlModel.java
ymorimo: Updated in 389d3c6. Now I'm thinking that it would be better to use a `LinkedHashSet` to hold the `fieldNames` rather than using a list. Do you think it's good to change the type to `LinkedHashSet`?

graphql/src/main/java/com/scalar/db/graphql/schema/SchemaUtils.java
```diff
@@ -0,0 +1,28 @@
...
+      case BOOLEAN:
+        return Scalars.GraphQLBoolean;
+      case INT:
+        return Scalars.GraphQLInt;
+      case FLOAT:
+      case DOUBLE:
+        return Scalars.GraphQLFloat;
+      case TEXT:
+        return Scalars.GraphQLString;
+      case BIGINT:
```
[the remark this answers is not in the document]
ymorimo: Great suggestion! I am checking to see if we can do that.

graphql/src/main/java/com/scalar/db/graphql/schema/TableGraphQlModel.java
brfrn169: Yes, I think we can change the type to `LinkedHashSet`.

graphql/src/main/java/com/scalar/db/graphql/schema/SchemaUtils.java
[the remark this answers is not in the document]
ymorimo: I have confirmed that GraphQL's `Float` type can represent Scalar's `BIGINT` values, so I'd like to map `BIGINT` to `Scalars.GraphQLFloat`.
But maybe we should define a custom `BigInt` scalar type that describes the exact Scalar's `BIGINT` spec. https://www.graphql-java.com/documentation/scalars
I want to add that type in another PR as it does not seem that difficult.

graphql/src/test/java/com/scalar/db/graphql/GraphQlFactoryTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+  @Mock private StorageFactory storageFactory;
+  @Mock private TransactionFactory transactionFactory;
+  @Mock private DistributedStorage storage;
+  @Mock private DistributedStorageAdmin storageAdmin;
+  @Mock private DistributedTransactionManager transactionManager;
+
+  @Before
+  public void setUp() {
+    initMocks(this);
```
brfrn169, line 43: We introduced Mockito 3 in #348, and `MockitoAnnotations.initMock()` is deprecated and replaced with `MockitoAnnotations.openMocks()` in Mockito 3. Can you please use `MockitoAnnotations.openMocks()`?

brfrn169 (commented):
Left a minor comment. Other than that, LGTM! Thank you!

ymorimo:
@brfrn169 Thank you for the review, but I updated some significant parts of the schema, so can you review them again?

* Changed response types of Put and Delete operations:
     it's not reasonable to return entire fields of the table records for Put and Delete, so I created a response type called `table_name_MutationResult`, which contains the `applied` flag and the primary key. The updated part is like the following.
    ```
    type Mutation {
      table_1_delete(delete: [table_1_DeleteInput!]!): [table_1_MutationResult!]!
      table_1_put(put: [table_1_PutInput!]!): [table_1_MutationResult!]!
    }
    type table_1_MutationResult {
      applied: Boolean!
      key: table_1_KeyOutput!
    }
    ```
* The `consistency` field was missing in the input objects for Put and Delete operations.

graphql/src/main/java/com/scalar/db/graphql/schema/CommonSchema.java
```diff
@@ -0,0 +1,112 @@
...
+public class CommonSchema {
+  public static Set<GraphQLNamedInputType> createCommonGraphQLTypes() {
+    GraphQLInputObjectType conditionalExpressionInputObject =
+        newInputObject()
+            .name("ConditionalExpression")
+            .field(newInputObjectField().name("name").type(nonNull(Scalars.GraphQLString)))
+            .field(newInputObjectField().name("intValue").type(Scalars.GraphQLInt))
+            .field(newInputObjectField().name("floatValue").type(Scalars.GraphQLFloat))
+            .field(newInputObjectField().name("stringValue").type(Scalars.GraphQLString))
+            .field(newInputObjectField().name("booleanValue").type(Scalars.GraphQLBoolean))
```
brfrn169, line 30: Do we need to add `bigIntValue` here?

graphql/src/main/java/com/scalar/db/graphql/schema/TableGraphQlModel.java
```diff
@@ -0,0 +1,447 @@
...
+  }
+
+  private GraphQLInputObjectType createClusteringKeyInputObjectType() {
+    return newInputObject()
+        .name(objectType.getName() + "_ClusteringKey")
+        .field(newInputObjectField().name("name").type(nonNull(clusteringKeyNameEnum)))
+        .field(newInputObjectField().name("intValue").type(Scalars.GraphQLInt))
+        .field(newInputObjectField().name("floatValue").type(Scalars.GraphQLFloat))
+        .field(newInputObjectField().name("stringValue").type(Scalars.GraphQLString))
+        .field(newInputObjectField().name("booleanValue").type(Scalars.GraphQLBoolean))
```
brfrn169, line 197: Ditto.

brfrn169 (commented):
Thank you! LGTM overall. I just left some comments. Please take a look when you have time!

graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
```diff
@@ -0,0 +1,139 @@
...
+import graphql.Scalars;
+import graphql.schema.GraphQLCodeRegistry;
+import graphql.schema.GraphQLFieldDefinition;
+import graphql.schema.GraphQLObjectType;
+import graphql.schema.GraphQLSchema;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public class GraphQlFactory {
```
feeblefakie, line 22: This class looks immutable so can you add @Immutable to the class? If not, please add an appropriate annotation like @ThreadSafe or @ThreadNotSafe.
We don't use any tools that utilize the annotations but it's helpful for users and developers to grasp what a class is designed for.

Also, for other classes, it would be nice if we have the annotations.

graphql/src/main/java/com/scalar/db/graphql/schema/TableGraphQlModel.java
```diff
@@ -0,0 +1,447 @@
...
+        .field(
+            newFieldDefinition()
+                .name(objectType.getName())
+                .type(nonNull(list(nonNull(objectType)))))
+        .build();
+  }
+
+  private GraphQLFieldDefinition createQueryScanField() {
+    return newFieldDefinition()
+        .name(objectType.getName() + "_scan")
```
feeblefakie, line 245: Can you clarify the naming rules?
For this,  it is not capitalized (`_scan`) but for others, it's like `_ScanPayload`.
I see `_put` and `_delete` so there seem some rules behind but I'm not sure and it looks inconsistent.

feeblefakie (approved):
LGTM! Thank you!
Left minor comments and a question.

graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
ymorimo: Thank you! Added `@Immutable` to this class and the `TableGraphQlModel` class in 6271071.
For other classes, I'm not fully sure but I don’t think it is needed because they are helper classes.

graphql/src/main/java/com/scalar/db/graphql/schema/TableGraphQlModel.java
ymorimo: * All generated names for a table are prefixed with the table name. (The object type name (`objectType.getName()`) is the table name itself.)
* “Types” such as `_ScanPayload`, `_GetInput`, or `_Key` are in pascal case.
* “Fields” in the `Query` and `Mutation` GraphQL type (`_get`, `_scan`, `_put`, and `_delete`) are in lowercase.

graphql/src/main/java/com/scalar/db/graphql/schema/CommonSchema.java
ymorimo: Thank you!
I added the `bigIntValue` with the type `GraphQLFloat` (double precision IEEE 754) for this PR. But I’m going to implement a custom GraphQL scalar type that exactly matches the range of `BigIntValue` (-2^53 - 2^53) in another PR.

brfrn169 (approved):
LGTM! Thank you!

graphql/src/main/java/com/scalar/db/graphql/schema/TableGraphQlModel.java
feeblefakie: Thank you!
Using different case styles for `types` and `fields` is a common practice?
It would be great if we have good reasons for those rules (ideally described in the code) since it looks inconsistent at first glance.

ymorimo: > Using different case styles for types and fields is a common practice?

For example, here is Apollo's recommendation. I think this convention is very common.
https://www.apollographql.com/docs/apollo-server/schema/schema/#naming-conventions

The difference in our app is that we simply prefix all types and field names with the `<table_name>` + `_`.

PR #360, opened 2021-10-31
Use MockitoAnnotations.openMocks() instead of MockitoAnnotations.initMocks()

brfrn169:
We introduced Mockito 3 in #348. And `MockitoAnnotations.initMock()` is deprecated and replaced with `MockitoAnnotations.openMocks()` in Mockito 3. This PR replaces `MockitoAnnotations.initMock()` with `MockitoAnnotations.openMocks()`. Please take a look!

### DIFF
--- a/core/src/test/java/com/scalar/db/storage/cassandra/BatchComposerTest.java
+++ b/core/src/test/java/com/scalar/db/storage/cassandra/BatchComposerTest.java
@@ -34,8 +34,8 @@ public class BatchComposerTest {
   @Mock private Delete del;
 
   @Before
-  public void setUp() {
-    MockitoAnnotations.initMocks(this);
+  public void setUp() throws Exception {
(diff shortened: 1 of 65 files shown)


feeblefakie (approved):
LGTM! Thank you!

The doc says `openMocks(this).close()` is equivalent to `initMocks`  but I wonder how they can be equivalent.

brfrn169:
> The doc says openMocks(this).close() is equivalent to initMocks but I wonder how they can be equivalent.

@feeblefakie It looks like unless we use static-mocks or custom MockMakers, we can call the `close` method immediately.
```
The close method should however only be called after completed usage of testClass. If using static-mocks or custom org.mockito.plugins.MockMakers, using this method might cause misbehavior of mocks injected into the test class.
```

thongdk8 (approved):
LGTM! Thank you!

PR #361, opened 2021-10-31
Change the default values of JDBC connection pool configurations

brfrn169:
The current default values of JDBC connection pool configurations might be too low for most cases. This PR changes it to new values. I'm not sure the new values are reasonable. I also want to discuss it here as well. 

the configurations `maxTotal`, `maxIdle` and `minIdle` are explained in the following document:
https://commons.apache.org/proper/commons-dbcp/configuration.html

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
@@ -28,9 +28,9 @@ public class JdbcConfig extends DatabaseConfig {
       PREFIX + "prepared_statements_pool.max_open";
   public static final String TABLE_METADATA_SCHEMA = PREFIX + "table_metadata.schema";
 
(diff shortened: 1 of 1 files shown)


feeblefakie (approved):
LGTM! 
The numbers look fine to me, but we can adjust the numbers based on the number of cores available in the node?
There seems a recommendation in Oracle.
https://docs.oracle.com/en/database/oracle/oracle-database/21/jjucp/optimizing-real-world-performance.html#GUID-BC09F045-5D80-4AF5-93F5-FEF0531E0E1D

brfrn169:

> The numbers look fine to me, but we can adjust the numbers based on the number of cores available in the node?

I think yes, we can do that. For now, we will have a bit more aggressive configurations for it rather than that.

thongdk8 (approved):
LGTM! Thank you!

PR #367, opened 2021-11-09
Add ConsensusCommitUtils

brfrn169:
This PR adds `ConsensusCommitUtils` that has the following utility methods:
- buildTransactionalTableMetadata(TableMetadata tableMetadata)
- isTransactionalTableMetadata(TableMetadata tableMetadata)
- removeTransactionalMetaColumns(TableMetadata tableMetadata)

Please take a look!


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
```diff
@@ -201,4 +218,63 @@ private TableMetadata convertToTransactionalTable(TableMetadata tableMetadata) {
...
+   * @return whether the table metadata is transactional
+   */
+  public static boolean isTransactionalTableMetadata(TableMetadata tableMetadata) {
+    for (String column : TRANSACTION_META_COLUMNS.keySet()) {
+      if (!tableMetadata.getColumnNames().contains(column)) {
+        return false;
+      }
+    }
+
+    List<String> nonPrimaryKeyColumns = getNonPrimaryKeyColumns(tableMetadata);
```
feeblefakie, line 235: I assume this block is handling some corner cases but I'm not sure why this block is required.
Can you elaborate on that?
Maybe we need a comment (with some examples) to explain why this block is needed.

feeblefakie (commented):
I left one question about what I'm not certain of.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
brfrn169: Thank you for taking a look at this PR!

Sorry I didn't get your question. Which block are you pointing out?

feeblefakie: Sorry, I meant this block.
```
    for (String nonPrimaryKeyColumn : nonPrimaryKeyColumns) {
      if (TRANSACTION_META_COLUMNS.containsKey(nonPrimaryKeyColumn)) {
        continue;
      }
      if (!tableMetadata.getColumnNames().contains(Attribute.BEFORE_PREFIX + nonPrimaryKeyColumn)
          && (nonPrimaryKeyColumn.length() < Attribute.BEFORE_PREFIX.length()
              || !tableMetadata
                  .getColumnNames()
                  .contains(nonPrimaryKeyColumn.substring(Attribute.BEFORE_PREFIX.length())))) {
        return false;
      }
    }
    return true;
```

brfrn169: I just added some comments in the last commit. Please take a look! Thanks.

feeblefakie: Thank you!
Ah, it is checking if all the columns are conformed to the rule.
So, even if a table has all the meta columns, if it doesn't have a `before_` prefix column for one non-primary key column, then this method returns `false`.
Such a case doesn't really happen if users use the schema-loader, but it looks very much like it is intended for transactional but mistakenly configured with a manual operation or something.
So, should it rather throw an exception if it has meta columns but it doesn't comply with the before rule?
(so I feel the method looks more like `isValid` rather than `isTransactional`.)

brfrn169: The `isTransactionalTableMetadata` method checks if specified table metadata is transactional or not. So I don't think we should throw an exception even if it's invalid. Rather than that, I think we should return false.

Actually, my first idea was to consider specified table metadata as transactional if it has any transactional columns. But in that case, even if a user accidentally uses the same column name as any of the transactional meta columns', we will consider it as transactional. That's why I decided to check all the transactional meta columns including the before prefix column. What do you think?

brfrn169:
I moved the utility methods to `ConsensusCommitUtils`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
feeblefakie: > But in that case, even if a user accidentally uses the same column name as any of the transactional meta columns', we will consider it as transactional.

Similarly, if a user accidentally forgets to add `before_` to a column, it will return `false` even though the schema looks almost transactional, which makes me feel we might need another method like `isValid` to avoid such mistakes.
(For example, if a table metadata has meta columns but doesn't comply with all the rules, `isTransactional` returns true, but `isValid` returns false)

Anyway, I think it's all about the assumptions and use cases that I couldn't get from the code, so I just wondered.
But, if it is OK, I'm fine with it.
Thank you for the discussion!

feeblefakie (approved):
LGTM! Thank you!
(I left a comment because I don't fully get the assumptions, but the code looks good if the spec is based on the assumptions.)

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
brfrn169: Yes, it looks like our assumptions are different. I assume that if a table doesn't have all the transactional meta columns, it's not transactional. But you assume that if a table has any transactional meta columns, it's transactional, but if the table doesn't have all the transactional meta columns, it's not valid.

Maybe the difference comes from whether accepting the situation where a user uses the transactional meta column names for non-transactional tables.
Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #369, opened 2021-11-11
Add KeyBytesEncoder that converts a key to bytes while preserving the sort order

brfrn169:
This PR adds `KeyBytesEncoder` that converts a key to bytes while preserving the sort order. It's going to be used for the DynamoDB sort key. Please take a look!


core/src/main/java/com/scalar/db/storage/dynamo/bytes/DoubleBytesEncoder.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Override
+  public int encodedLength(DoubleValue value, Order order) {
+    return 8;
+  }
+
+  @Override
+  public void encode(DoubleValue value, Order order, ByteBuffer dst) {
+    long l = Double.doubleToLongBits(value.getAsDouble());
+    l ^= ((l >> (Long.SIZE - 1)) | Long.MIN_VALUE);
```
feeblefakie, line 21: Could you explain what it does as a comment?

feeblefakie (approved):
Looking great! Thank you!
I left one question/request but it doesn't block it from being merged.

core/src/main/java/com/scalar/db/storage/dynamo/bytes/BytesUtils.java
```diff
@@ -0,0 +1,24 @@
...
+   *
+   * @param b a byte
+   * @param order an order
+   * @return a masked byte if DESC order, otherwise the specified byte without any processing
+   */
+  public static byte mask(byte b, Order order) {
+    if (order == Order.ASC) {
+      return b;
+    } else {
+      return (byte) (b ^ MASK);
```
thongdk8, line 21: Is the xor operation performed here for reversing the order of `b`? Could you add a comment here or in the function description to explain what it does?

thongdk8 (dismissed):
LGTM! Thank you!
Just left a question, but it doesn't block the PR from being merged

PR #370, opened 2021-11-11
Add null or empty check for partition key and clustering key

brfrn169:
Since we don't accept null or empty partition key/clustering key, this PR adds null or empty check for partition key and clustering key. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/common/checker/ColumnChecker.java
+++ b/core/src/main/java/com/scalar/db/storage/common/checker/ColumnChecker.java
@@ -16,11 +16,13 @@ import javax.annotation.concurrent.NotThreadSafe;
 @NotThreadSafe
 class ColumnChecker implements ValueVisitor {
   private final TableMetadata tableMetadata;
+  private final boolean notNullOrEmpty;
   private String name;
   private boolean isValid;
 
-  public ColumnChecker(TableMetadata tableMetadata) {
+  public ColumnChecker(TableMetadata tableMetadata, boolean notNullOrEmpty) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/scalar/db/storage/common/checker/ColumnChecker.java
```diff
@@ -16,11 +16,13 @@
 @NotThreadSafe
 class ColumnChecker implements ValueVisitor {
   private final TableMetadata tableMetadata;
+  private final boolean notNullOrEmpty;
   private String name;
   private boolean isValid;
 
-  public ColumnChecker(TableMetadata tableMetadata) {
+  public ColumnChecker(TableMetadata tableMetadata, boolean notNullOrEmpty) {
```
feeblefakie, line 23:
```suggestion
  public ColumnChecker(TableMetadata tableMetadata, boolean neitherNullNorEmpty) {
```
since it can be read as `not null` or `empty`.

feeblefakie (commented):
Looking good! Thank you!
Left one wording suggestion.

core/src/main/java/com/scalar/db/storage/common/checker/ColumnChecker.java
brfrn169: Thank you for taking a look at this! I modified it in the latest commit. Thanks.

feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #373, opened 2021-11-13
Remove UnsupportedTypeException

brfrn169:
Please take a look!


core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -280,7 +281,7 @@ private DataType fromCassandraDataType(
       case BLOB:
         return DataType.BLOB;
       default:
-        throw new UnsupportedTypeException(
+        throw new ExecutionException(
```
feeblefakie, line 284: Any particular reason for throwing a checked exception here?
Looks like it's a programs' error or user's error so should they be an assertion error or unchecked exception?

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -307,7 +308,7 @@ private DataType fromCassandraDataType(
       case BLOB:
         return com.datastax.driver.core.DataType.blob();
       default:
-        throw new UnsupportedOperationException(String.format("%s is not yet supported", dataType));
+        throw new AssertionError();
```
feeblefakie, line 311: `dataType` is mostly given from users?
If so, should it be an unchecked exception?

feeblefakie (commented):
Left some questions for clarification.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
feeblefakie: Thank you for the discussion! I understand the point so let's leave it as it is.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
feeblefakie: I understand the point. Thank you!

feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #374, opened 2021-11-16
Allow empty value for partition key and clustering key

brfrn169:
Sorry, I didn't have to add an empty check for partition key and clustering key in #370. I needed to add only null check there. This PR reverts the empty check. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #377, opened 2021-11-18
Fix hashCode() in XXXValue and Key classes

brfrn169:
I think we need to use `name` in `hashCode()` as well as `equals()` in `XXXValue` and `Key` classes. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #378, opened 2021-11-18
Add clustering order support to JDBC adapter

brfrn169:
For the clustering order support for the JDBC adapter, we need to add the following:
- For MySQL and SQL Server, we need to add the clustering orders in the primary key definition as follows:
```
CREATE TABLE `test_ns.test_table` (`id1` INT, `id2` INT, `col1` INT, PRIMARY KEY (`id1` ASC, `id2` DESC));
```
- For PostgreSQL and Oracle database, we need to add a unique index for the clustering orders:
```
CREATE UNIQUE INDEX "test_ns.test_table_clustering_order_idx" ON "test_ns.test_table" ("id1" ASC, "id2" DESC);
```

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
@@ -11,6 +11,7 @@ import com.google.inject.Inject;
 import com.scalar.db.api.DistributedStorageAdmin;
 import com.scalar.db.api.Scan;
 import com.scalar.db.api.Scan.Ordering;
+import com.scalar.db.api.Scan.Ordering.Order;
 import com.scalar.db.api.TableMetadata;
 import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.io.DataType;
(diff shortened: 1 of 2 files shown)


thongdk8 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
```diff
@@ -666,15 +667,31 @@ private void createTableInternal(
...
-            + Stream.concat(
-                    metadata.getPartitionKeyNames().stream(),
-                    metadata.getClusteringKeyNames().stream())
-                .map(this::enclose)
-                .collect(Collectors.joining(","))
-            + "))";
+    if (hasDescClusteringOrder
+        && (rdbEngine == RdbEngine.MYSQL || rdbEngine == RdbEngine.SQL_SERVER)) {
+      // For MySQL and SQL Server, specify the clustering orders in the primary key definition
+      createTableStatement +=
```
feeblefakie, line 677: Wouldn't it be better to specify the orders whether or not `hasDescClusteringOrder` is true so that we can avoid th the redundant code?
(if there is a good reason for not specifying the orders, then it's fine to leave it as it is)

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
```diff
@@ -690,6 +707,24 @@ private void createTableInternal(
               + " INITRANS 3 MAXTRANS 255";
       execute(connection, alterTableStatement);
     }
+
+    if (hasDescClusteringOrder
+        && (rdbEngine == RdbEngine.POSTGRESQL || rdbEngine == RdbEngine.ORACLE)) {
+      // For PostgreSQL and Oracle, create a unique index for the clustering orders
+      String createUniqueIndexStatement =
+          "CREATE UNIQUE INDEX "
```
feeblefakie, line 715: Just a question.
So, for PostgreSQL and Oracle, key ordering is not really supported and we need to use a secondary index to use desc ordering?

feeblefakie (commented):
Looking good! Thank you!
Left a comment and a question. PTAL!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
brfrn169: Yes, that's correct.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabaseAdmin.java
brfrn169: We can remove the `hasDescClusteringOrder` condition, but we still need the `rdbEngine == RdbEngine.MYSQL || rdbEngine == RdbEngine.SQL_SERVER` condition since we can specify `ASC` or `DESC` in the primary key definition only for MySQL and SQL Server. So I don't think we can avoid the redundant code. What do you think?

brfrn169 (commented):

I just left replies for your questions. Please take a look when you have time! Thanks.

feeblefakie (approved):
LGTM! Thank you!

PR #381, opened 2021-11-22
Disallow empty values for Partition key and Clustering key

brfrn169:
I'm very sorry for the confusion. After all, I think we should disallow empty values for Partition key and Clustering key. This PR makes the change for it. Please take a look!


thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!
(I'm a bit confused so let's chat about this at some point!)

PR #388, opened 2021-11-27
Add partition key integration tests

brfrn169:
This PR adds partition key related integration tests. The integration tests check if all data types for partition keys are handled correctly. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
+++ b/core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
@@ -8,9 +8,8 @@ import com.scalar.db.io.FloatValue;
 import com.scalar.db.io.IntValue;
 import com.scalar.db.io.TextValue;
 import com.scalar.db.io.ValueVisitor;
-import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Base64;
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
```diff
@@ -99,14 +98,9 @@ public void visit(TextValue value) {
...
     value
         .get()
-        .ifPresent(
-            b -> {
-              // TODO Can we really convert bytes to string directly?
-              //      Maybe need to use the BASE64 encoding?
-              ByteBuffer buffer = (ByteBuffer) ByteBuffer.allocate(b.length).put(b).flip();
-              values.add(new String(buffer.array(), StandardCharsets.UTF_8));
-            });
+        .ifPresent(b -> values.add(Base64.getUrlEncoder().withoutPadding().encodeToString(b)));
```
brfrn169, line 104: We need to use Base64 encoding to handle BLOB type for partition key.

feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

brfrn169:
The cause of the integration test failure is the tests taking a long time. Although the integration test is failing, I will merge this PR. And the next PR will handle the failure. Thanks.

PR #389, opened 2021-11-28
Add secondary index integration test

brfrn169:
This PR adds secondary index related integration tests. The integration tests check if all data types for secondary indexes are handled correctly. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
+++ b/core/src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
@@ -36,7 +36,7 @@ public class OperationChecker {
     checkProjections(get, metadata);
 
     if (Utility.isSecondaryIndexSpecified(get, metadata)) {
-      if (!new ColumnChecker(metadata, true).check(get.getPartitionKey().get().get(0))) {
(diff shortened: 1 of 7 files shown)


thongdk8 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
```diff
@@ -36,7 +36,7 @@ public void check(Get get) throws ExecutionException {
     checkProjections(get, metadata);
 
     if (Utility.isSecondaryIndexSpecified(get, metadata)) {
-      if (!new ColumnChecker(metadata, true).check(get.getPartitionKey().get().get(0))) {
+      if (!new ColumnChecker(metadata, false).check(get.getPartitionKey().get().get(0))) {
```
feeblefakie, line 39: Was it a bug?

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
brfrn169: Yes, I found we needed to accept empty values here, that's a very rare case though.

PR #397, opened 2021-12-01
Reduce integration test concurrency

brfrn169:
It looks like the current integration test concurrency is too aggressive, causing the integration test failures sometimes. This PR reduces the integration test concurrency. Please take a look!


thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #398, opened 2021-12-02
Add GraphQL custom scalar types Float32 and BigInt

ymorimo:
This PR adds two custom scalar types: `Float32` and `BigInt`.

The `Float32` scalar type represents Scalar DB's `FloatValue` values. [GraphQL's built-in `Float`](https://spec.graphql.org/June2018/#sec-Float) represents signed double-precision fractional values, so we need another single-precision type to represent `FloatValue`.
[Astra DB GraphQL API also has the `Float32` type](https://docs.datastax.com/en/astra/docs/graphql-using-schema-first.html#_data_types) that represents float values in CQL.

The `BigInt` scalar type represents Scalar DB's `BigIntValue` values. It restricts the value range from 2^-53 to 2^53 as specified in Scalar DB's `BigIntValue`.


graphql/src/test/java/com/scalar/db/graphql/schema/BigIntCoercingTest.java
```diff
@@ -0,0 +1,140 @@
...
+    Long actual = BigIntCoercing.INSTANCE.serialize(value);
+
+    // Assert
+    assertThat(actual).isEqualTo(expected);
+  }
+
+  @Test
+  public void serialize_LongOutOfRangeGiven_ShouldThrowException() {
+    // Arrange
+    Long value = BigIntValue.MAX_VALUE + 1;
```
feeblefakie, line 31: Can you also add a test that checks the other side (MIN_VALUE -1) of the range as well?

feeblefakie (commented):
Left one minor comment.

graphql/src/test/java/com/scalar/db/graphql/schema/BigIntCoercingTest.java
ymorimo: Thank you! Added some tests in fd06d00.

graphql/src/main/java/com/scalar/db/graphql/schema/BigIntCoercing.java
```diff
@@ -0,0 +1,72 @@
...
+import java.math.BigInteger;
+
+/** Handles type conversions for Scalar DB {@link BigIntValue}. */
+class BigIntCoercing implements Coercing<Long, Long> {
+  static final BigIntCoercing INSTANCE = new BigIntCoercing();
+
+  private BigIntCoercing() {}
+
+  @Override
+  public Long serialize(Object dataFetcherResult) throws CoercingSerializeException {
```
brfrn169, line 17: My Intellij shows a warning for it:
```suggestion
  public Long serialize(@Nonnull Object dataFetcherResult) throws CoercingSerializeException {
```

graphql/src/main/java/com/scalar/db/graphql/schema/BigIntCoercing.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+    Long value = (Long) dataFetcherResult;
+    if (!isInBigIntValueRange(value)) {
+      throw new CoercingSerializeException("Out of range for BigIntValue");
+    }
+    return value;
+  }
+
+  @Override
+  public Long parseValue(Object input) throws CoercingParseValueException {
```
brfrn169, line 29: Ditto.
```suggestion
  @Nonnull
  @Override
  public Long parseValue(@Nonnull Object input) throws CoercingParseValueException {
```

graphql/src/main/java/com/scalar/db/graphql/schema/BigIntCoercing.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+    if (isInBigIntValueRange(value)) {
+      return value;
+    } else {
+      throw new CoercingParseValueException("Out of range for BigIntValue");
+    }
+  }
+
+  @Override
+  public Long parseLiteral(Object input) throws CoercingParseLiteralException {
```
brfrn169, line 52: Ditto.
```suggestion
  @Nonnull
  @Override
  public Long parseLiteral(@Nonnull Object input) throws CoercingParseLiteralException {
```

graphql/src/main/java/com/scalar/db/graphql/schema/FloatCoercing.java
```diff
@@ -0,0 +1,46 @@
...
+import java.math.BigInteger;
+
+/** Handles type conversions for Scalar DB {@link com.scalar.db.io.FloatValue}. */
+class FloatCoercing implements Coercing<Float, Float> {
+  static FloatCoercing INSTANCE = new FloatCoercing();
+
+  private FloatCoercing() {}
+
+  @Override
+  public Float serialize(Object dataFetcherResult) throws CoercingSerializeException {
```
brfrn169, line 17: Ditto.
```suggestion
  public Float serialize(@Nonnull Object dataFetcherResult) throws CoercingSerializeException {
```

graphql/src/main/java/com/scalar/db/graphql/schema/FloatCoercing.java
```diff
@@ -0,0 +1,46 @@
...
+    if (dataFetcherResult instanceof Float) {
+      return (Float) dataFetcherResult;
+    } else {
+      throw new CoercingSerializeException("Expected a long value");
+    }
+  }
+
+  @Override
+  public Float parseValue(Object input) throws CoercingParseValueException {
```
brfrn169, line 26: Ditto.
```suggestion
  @Nonnull
  @Override
  public Float parseValue(@Nonnull Object input) throws CoercingParseValueException {
```

graphql/src/main/java/com/scalar/db/graphql/schema/FloatCoercing.java
```diff
@@ -0,0 +1,46 @@
...
+    if (input instanceof Number) {
+      return ((Number) input).floatValue();
+    } else {
+      throw new CoercingParseValueException("Expected an integer or float value");
+    }
+  }
+
+  @Override
+  public Float parseLiteral(Object input) throws CoercingParseLiteralException {
```
brfrn169, line 35: Ditto.
```suggestion
  @Nonnull
  @Override
  public Float parseLiteral(@Nonnull Object input) throws CoercingParseLiteralException {
```

brfrn169 (commented):
Thank you! Left several very minor suggestions. Other than that, LGTM! Thanks.

graphql/src/main/java/com/scalar/db/graphql/schema/BigIntCoercing.java
ymorimo: Thank you!
I didn't add the annotation to our classes since the interface methods are annotated with `org.jetbrains.annotations.NotNull` and it depends on the JetBrains package. I didn't know that I could remove that warning by adding `javax.annotation.Nonnull`. Is it okay to mix different kinds of nullability annotations?

brfrn169: Actually, we don't have any rule for it, but I usually use `javax.annotation.*` annotations in that case. @feeblefakie What do you think?

It looks like the author of this comment prefers `javax.annotation`, too:
https://stackoverflow.com/a/42695253/4746611

ymorimo: Since we don't want to add a dependency to `org.jetbrains` package, `javax.annotation` seems the way to go. I added it in 4a602d5 anyway. Please leave a comment if there are any problems.

feeblefakie: I think, so far, we only use javax.annotation, so let's use `javax.annotation.Nonnull`.

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #404, opened 2021-12-03
Move Isolation to the consensuscommit package

brfrn169:
This PR moves the `Isolation` class to the `consensuscommit` package. Also, the property name is changed to `scalar.db.consensus_commit.isolation_level`. But we can still use the old property name `scalar.db.isolation_level` for backward compatibility.

### DIFF
--- a/core/src/main/java/com/scalar/db/api/Isolation.java
+++ b/core/src/main/java/com/scalar/db/api/Isolation.java
@@ -1,5 +1,7 @@
 package com.scalar.db.api;
 
+/** @deprecated As of release 3.5.0. Will be removed in release 4.0.0. */
+@Deprecated
 public enum Isolation {
   SNAPSHOT,
   SERIALIZABLE,
--- a/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
+++ b/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
@@ -8,7 +8,6 @@ import static com.scalar.db.config.ConfigUtils.getString;
 import com.scalar.db.api.DistributedStorage;
 import com.scalar.db.api.DistributedStorageAdmin;
 import com.scalar.db.api.DistributedTransactionManager;
-import com.scalar.db.api.Isolation;
 import com.scalar.db.api.TwoPhaseCommitTransactionManager;
 import com.scalar.db.storage.cassandra.Cassandra;
 import com.scalar.db.storage.cassandra.CassandraAdmin;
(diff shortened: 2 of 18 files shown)


core/src/main/java/com/scalar/db/api/Isolation.java
```diff
@@ -1,5 +1,7 @@
 package com.scalar.db.api;
 
+/** @deprecated As of release 3.5.0. Will be removed in release 4.0.0. */
+@Deprecated
 public enum Isolation {
```
brfrn169, line 5: We don't remove this yet for backward compatibility.

core/src/main/java/com/scalar/db/config/DatabaseConfig.java
```diff
@@ -62,7 +60,6 @@
   public static final String PASSWORD = PREFIX + "password";
   public static final String STORAGE = PREFIX + "storage";
   public static final String TRANSACTION_MANAGER = PREFIX + "transaction_manager";
-  public static final String ISOLATION_LEVEL = PREFIX + "isolation_level";
```
brfrn169, line 65: If we use `DatabaseConfig.ISOLATION_LEVEL`, we need to change it to `ConsensusCommitConfig.ISOLATION_LEVEL`. It's backward incompatibility but I think it's okay.

core/src/main/java/com/scalar/db/config/DatabaseConfig.java
```diff
@@ -204,10 +196,6 @@ public int getContactPort() {
     return transactionManagerClass;
   }
 
-  public Isolation getIsolation() {
```
brfrn169, line 207: Ditto. We need to change `DatabaseConfig.getIsolation()` to `ConsensusCommitConfig.getIsolation()`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
```diff
@@ -31,27 +40,43 @@
...
   }
 
-  @Override
   protected void load() {
-    String transactionManager = getProperties().getProperty(DatabaseConfig.TRANSACTION_MANAGER);
-    if (transactionManager != null && !transactionManager.equals("consensus-commit")) {
-      throw new IllegalArgumentException(
-          DatabaseConfig.TRANSACTION_MANAGER + " should be 'consensus-commit'");
+    if (getProperties().containsValue("scalar.db.isolation_level")) {
+      LOGGER.warn(
```
brfrn169, line 64: If the old property is used, show a warning message.

feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #406, opened 2021-12-03
Add mergeServiceFiles for shadowJar in Schema Loader

brfrn169:
I faced the similar issue to the issue mentioned in the following article when using Schema Loader with Scalar DB server:
https://stackoverflow.com/questions/55484043/how-to-fix-could-not-find-policy-pick-first-with-google-tts-java-client

As this article says, after adding `mergeServiceFiles` for `shadowJar` in `build.gradle` for `schema-loader`, the issue is fixed.

Please take a look!


thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #407, opened 2021-12-04
Throw RetriableExecutionException when TransactionConflict happens in the Dynamo adapter

brfrn169:
In DynamoDB, `TransactionConflict` happens when we are writing/updating/deleting an item and there is an ongoing transaction for the same item. In that case, we can basically retry that operation, so I think we should throw `RetriableExecutionException`. This PR adds that logic. Please take a look!

Referencies:
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html
https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/dynamodb/model/TransactionConflictException.html
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/model/TransactionCanceledException.html


feeblefakie (approved):
LGTM! Thank you!
It's better to add a unit test to cover the case?

thongdk8 (dismissed):
LGTM! Thank you!

PR #408, opened 2021-12-05
Throw RetriableExecutionException in mutate operations when conflicts happen in the JDBC adapter

brfrn169:
Since a mutate operation in the JDBC adapter executes multiple put/delete operations in a transaction, conflicts can happen. This PR makes it throw `RetriableExecutionException` in that case. Please take a look!


feeblefakie (approved):
LGTM! Thank you!
Maybe, it's better to add some unit tests to cover the case?

thongdk8 (approved):
LGTM! Thank you!

PR #409, opened 2021-12-05
Handle RetriableExecutionException in ConsensusCommit

brfrn169:
Since we can retry when we catch `RetriableExecutionException` when preparing records in ConsensusCommit, we should throw `CommitConflictException` in that case. This PR adds that logic. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

thongdk8 (dismissed):
LGTM! Thank you!

PR #410, opened 2021-12-06
Result.getClusteringKey() should return empty when there is no clustering key

brfrn169:
Currently, `Result.getClusteringKey()` returns a `Key` object with no `Value` objects when there is no clustering key, but it should return `Optional.empty()` in that case. This PR fixes this issue. Please take a look!


thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #415, opened 2021-12-07
Fix get tables in Dynamo adapter when actual number of tables exceed the maximum in a response.

thongdk8:
In Dynamo DB, a response of getting the table list is paginated to a maximum of 100, so we have to handle it correctly. This PR is made to fix it. PTAL! Thank you!


brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #420, opened 2021-12-08
Use the internal JRE Docker image to avoid CVE fixing commands

supl:
This PR uses a self-maintained JRE Docker image to avoid adding CVE patches through Dockerfile.

The Docker file is maintained in 
https://github.com/scalar-labs/docker/blob/main/jre/8/Dockerfile


feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

brfrn169 (approved):
LGTM! Thank you!

PR #424, opened 2021-12-09
GraphQL schema change of Mutation type

ymorimo:
This PR changes the schema generation as follows, for example, for a table named `table_1`.

Before:

```graphql
type Mutation {
  abort: Boolean
  commit: Boolean
  table_1_delete(delete: [table_1_DeleteInput!]!): table_1_DeletePayload
  table_1_put(put: [table_1_PutInput!]!): table_1_PutPayload
}
```

After:

```graphql
type Mutation {
  abort: Boolean!
  commit: Boolean!
  table_1_bulkDelete(delete: [table_1_DeleteInput!]!): Boolean!
  table_1_bulkPut(put: [table_1_PutInput!]!): Boolean!
  table_1_delete(delete: table_1_DeleteInput!): Boolean!
  table_1_mutate(delete: [table_1_DeleteInput!], put: [table_1_PutInput!]): Boolean!
  table_1_put(put: table_1_PutInput!): Boolean!
}
```

1. Make the type of `commit` and `abort` non-null.
2. Add `put` and `delete` mutations that accept a single input object. The previous `put` and `delete` have been renamed to `bulkPut` and `bulkDelete`. (The naming of bulk- is following Stargate. It generates operations like `insertbooks()` and `bulkInsertbooks()` if we have a `books` table.)

    The `put` operation should translate and pass the argument to Scalar DB's `put(Put)` method, and `bulkPut` operation should do that to `put(List<Put>)`. The same applies to `delete` and `bulkDelete`.
    Previously we made a decision that the put operation receives a list of `PutInput` because I misunderstood that it is not allowed to run multiple mutations with the same name in a single execution. Actually, it's possible to send the same operation multiple times using aliases like the example below.

    ```graphql
    mutation put2mountains {
      mt_fuji: mountain_put(put: { key: { name: "Mt. Fuji" }, values: { elevation: 3776 }})
      takaosan: mountain_put(put: { key: { name: "Takaosan" }, values: { elevation: 599 }})
    }
    ```

    Additionally `table_1_mutate()` field that accepts both lists of `put` and `delete` is added to the `Mutation` type. It should map to Scalar DB's `mutate` operation.

3. Remove the result type of mutations `<table_name>_MutationResult`, since the mutations in Scalar DB API are void methods. The response type used to have an `applied` boolean field, but it doesn't make sense for `put(List<Put>)` and `delete(List<Delete>)` since the mutations in the list either all succeed or all fail.

### DIFF
--- a/graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
+++ b/graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
@@ -60,7 +60,8 @@ public class GraphQlFactory {
           .field(tableModel.getMutationPutField())
           .field(tableModel.getMutationBulkPutField())
           .field(tableModel.getMutationDeleteField())
-          .field(tableModel.getMutationBulkDeleteField());
+          .field(tableModel.getMutationBulkDeleteField())
+          .field(tableModel.getMutationMutateField());
     }
     builder
         .field(
--- a/graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
+++ b/graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
@@ -55,8 +55,8 @@ public class GraphQlFactory {
   private GraphQLObjectType createMutationObjectType() {
     GraphQLObjectType.Builder builder = GraphQLObjectType.newObject().name("Mutation");
     for (TableGraphQlModel tableModel : tableModels) {
-      builder.field(tableModel.getMutationPutField());
-      builder.field(tableModel.getMutationDeleteField());
+      builder.field(tableModel.getMutationBulkPutField());
+      builder.field(tableModel.getMutationBulkDeleteField());
     }
     builder.field(
         GraphQLFieldDefinition.newFieldDefinition().name("commit").type(Scalars.GraphQLBoolean));
--- a/graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
+++ b/graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
@@ -55,8 +55,11 @@ public class GraphQlFactory {
   private GraphQLObjectType createMutationObjectType() {
     GraphQLObjectType.Builder builder = GraphQLObjectType.newObject().name("Mutation");
     for (TableGraphQlModel tableModel : tableModels) {
-      builder.field(tableModel.getMutationBulkPutField());
-      builder.field(tableModel.getMutationBulkDeleteField());
+      builder
+          .field(tableModel.getMutationPutField())
+          .field(tableModel.getMutationBulkPutField())
+          .field(tableModel.getMutationDeleteField())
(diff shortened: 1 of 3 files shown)


ymorimo:
Maybe we can create `mutate(put: [PutInput!], delete: [DeleteInput!])` as well. With this method, `bulkPut()` and `bulkDelete()` might not be needed since it can accept both mutations.

brfrn169:
@ymorimo Thank you!

> Maybe we can create `mutate(put: [PutInput!], delete: [DeleteInput!])` as well. 

Yes, that's a good idea!

> With this method, `bulkPut()` and `bulkDelete()` might not be needed since it can accept both mutations.

I think we can keep `bulkPut()` and `bulkDelete()` since Scalar DB API also has those methods (the method names are different though).

ymorimo:
@brfrn169 Thank you for the comment. `mutate(put: [PutInput!], delete: [DeleteInput!])` is added in 4704b4c. The description of this PR is also updated.

brfrn169 (approved):
LGTM! Thank you!

ymorimo:
Now that #414 has been merged, I will also update the data fetchers to match the changed schema in this PR.

graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
```diff
@@ -55,13 +56,22 @@ private GraphQLObjectType createQueryObjectType() {
...
+          .field(tableModel.getMutationMutateField());
     }
-    builder.field(
-        GraphQLFieldDefinition.newFieldDefinition().name("commit").type(Scalars.GraphQLBoolean));
-    builder.field(
-        GraphQLFieldDefinition.newFieldDefinition().name("abort").type(Scalars.GraphQLBoolean));
+    builder
+        .field(
+            GraphQLFieldDefinition.newFieldDefinition()
+                .name("commit")
```
feeblefakie, line 69: It's more like a question about the specification.
Although even with only queries, we need to commit/abort but Is it OK to have `commit` and `abort` in `Mutation`? 
Are they also included in `Query` as well?

feeblefakie (approved):
LGTM! Thank you!
Just left a question but it doesn't block it from being merged.

graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
ymorimo: As we discussed on Friday, I'm moving the commit operation to the directive argument in #427. It will be possible to send commits with query operations as well.

ymorimo:
> Now that #414 has been merged, I will also update the data fetchers to match the changed schema in this PR.

I said this, but for ease of review, I would like to submit a separate PR of the new data fetchers for the schema change of this PR.
So can you please merge this PR? The `graphql` branch will be temporarily inconsistent, though.
@brfrn169 @feeblefakie

PR #427, opened 2021-12-10
Add transaction handling with GraphQL

ymorimo:
This PR adds `TransactionInstrumentation` for handling Scalar DB transactions. The "instrumentation" is a mechanism provided by the graphql-java library that allows us to inject code to change the runtime behavior.

The `TransactionInstrumentation` does the following.

1. It reads the `@transaction` directive from the operation, then starts a new transaction or gets an existing one.
2. If the `commit: true` argument is found in the directive, it will commit the transaction after all queries or mutations are finished. It is done via the `onCompleted` method of `InstrumentationContext`.
3. Before returning the result, it adds the transaction ID to the `extensions` field of the result.

This PR also adds a data fetcher for `abort` mutation.


ymorimo:
I will make this PR ready once #435 is merged.

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
```diff
@@ -0,0 +1,145 @@
...
+        directive.getArgument(Constants.TRANSACTION_DIRECTIVE_COMMIT_ARGUMENT_NAME);
+    boolean isCommitTrue =
+        commitArg != null
+            && commitArg.getValue() != null
+            && ((BooleanValue) commitArg.getValue()).isValue();
+    if (txIdArg == null) {
+      // start a new transaction
+      try {
+        DistributedTransaction transaction = transactionManager.start();
+        transactionMap.put(transaction.getId(), transaction);
```
brfrn169, line 71: When do we delete completed transaction objects committed or aborted from this `transactionMap`?

Also I think we need to handle the situation where a client is down while executing a transaction. In this case, the client doesn't call either commit or abort the transaction. So we need to remove transactions in `transactionMap` for a long time.
Maybe, we can use `ActiveExpiringMap` for that case:
https://github.com/scalar-labs/scalardb/blob/df1ba719bd55bb15ca7554762ec27ee90417f258/core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
```diff
@@ -0,0 +1,145 @@
...
+    public TransactionCommitInstrumentationContext(DistributedTransaction transaction) {
+      this.transaction = transaction;
+    }
+
+    @Override
+    public void onDispatched(CompletableFuture<ExecutionResult> result) {}
+
+    @Override
+    public void onCompleted(ExecutionResult result, Throwable t) {
+      if (t != null) {
```
brfrn169, line 135: It looks like the transaction is not committed if `t` is not null, even though specifying `commit=true` in the transaction directive. In this case, we need to call `abort` explicitly, right?

brfrn169 (commented):
Thank you! Looking good overall. Left several questions. Please take a look when you have time!

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
ymorimo: Yes, the transaction needs to be removed after a certain timeout if the client is down. `ActiveExpiringMap` sounds like a perfect option. Thank you for the great suggestion!
Also, I carelessly deleted a line that deletes transaction objects committed or aborted. I will update the code and tests.

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
ymorimo: Fixed in the latest commit. Thank you!

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
ymorimo: Now an `ActiveExpiringMap` is used to keep transactions. Please take a look when you have time.

ymorimo:
If any `TransactionException` is thrown from Scalar DB operations, the GraphQL frontend will report the exception name in the `"errors"` -> `"extensions"` -> `"exception"` key in the response. In this case, the `"data"` key will not be present.

```json
{
  "errors": [
    {
      "message": "Scalar DB CommitConflictException happened in the transaction c0ba4da7-8b22-4bdd-b506-080441ea20ff: preparing record exists",
      "locations": [
        {
          "line": 50,
          "column": 23
        }
      ],
      "extensions": {
        "exception": "CommitConflictException",
        "classification": "ExecutionAborted"
      }
    }
  ]
}
```

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
```diff
@@ -0,0 +1,190 @@
...
+                  LOGGER.info("committing the transaction {}", transaction.getId());
+                  transaction.commit();
+                } catch (CommitException | UnknownTransactionStatusException e) {
+                  LOGGER.warn(
+                      "an error happened when committing the transaction {}",
+                      transaction.getId(),
+                      e);
+                  graphQLContext.put(
+                      CONTEXT_TRANSACTION_ERROR_KEY,
+                      new ScalarDbTransactionError(
```
brfrn169, line 137: It looks like it only handles commit related exceptions, correct? How about `CrudException` or `CrudConflictException`?

And one more question is how about error report for storage operations (`ExecutionException`)?

brfrn169 (commented):
Thank you! I left a question. Please take a look when you have time!

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
ymorimo: Right, `CrudException` and `ExecutionException` are thrown from the data fetcher layer, so I'll check if they are handled properly and work on it in a separate pull request if necessary. The `TransactionInstrumentation` is only responsible for committing or aborting transactions.

ymorimo: @brfrn169 I added PR #467 for reporting `CrudException` and `ExecutionException`. Please take a look! 🙇‍♂️

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
```diff
@@ -0,0 +1,191 @@
...
+                            "The specified transaction %s is not found."
+                                + " It might have been expired",
+                            txId);
+                    LOGGER.warn(message);
+                    return new AbortExecutionException(message);
+                  });
+    }
+    graphQLContext.put(Constants.CONTEXT_TRANSACTION_KEY, transaction);
+
+    boolean isCommitTrue =
```
feeblefakie, line 111: It's a minor thing but I feel the variable name sounds a bit weird.
```suggestion
    boolean isCommitSpecified = 
```

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
```diff
@@ -0,0 +1,191 @@
...
+        commitArg != null
+            && commitArg.getValue() != null
+            && ((BooleanValue) commitArg.getValue()).isValue();
+    if (isCommitTrue) {
+      return SimpleInstrumentationContext.whenCompleted(
+          (executionResult, throwable) -> {
+            graphQLContext.delete(Constants.CONTEXT_TRANSACTION_KEY);
```
feeblefakie, line 118: For these two lines, it is OK to delete and remove here before committing or aborting?
(Is it OK even if it crashes after remove is called?)

feeblefakie (commented):
Left a question and a comment.

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
```diff
@@ -0,0 +1,191 @@
...
+                    String message =
+                        String.format(
+                            "The specified transaction %s is not found."
+                                + " It might have been expired",
+                            txId);
+                    LOGGER.warn(message);
+                    return new AbortExecutionException(message);
+                  });
+    }
+    graphQLContext.put(Constants.CONTEXT_TRANSACTION_KEY, transaction);
```
brfrn169, line 109: I have a question about `GraphQLContext`. What is the scope of the objects in `GraphQLContext`? Those objects are deleted automatically for each request? Or they are kept until we delete them (or disconnect the connection)?

brfrn169 (commented):
Left a question. Please take a look when you have time! Thanks.

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
ymorimo: I don't think the behavior of `GraphQLContext` is well documented, but the scope of the `GraphQLContext` itself is per request, so the scope of objects in it is the same.

As the JavaDoc of [`DataFetchingEnvironment`](https://github.com/graphql-java/graphql-java/blob/v17.2/src/main/java/graphql/schema/DataFetchingEnvironment.java#L101) says, the `GraphQLContext` object is set up when `GraphQL#execute` is run. The [`ExecutionInput` initially creates](https://github.com/graphql-java/graphql-java/blob/v17.2/src/main/java/graphql/ExecutionInput.java#L208) and holds the context. An `ExecutionInput` is created per execution (= per request), so the `GraphQLContext` itself should be created and freed per request.
And the `GraphQLContext` that the `TransactionInstrumentation` gets is the same object as that one.

brfrn169: Thank you for checking that!

So, if the objects in `GraphQLContext` are deleted automatically per request, it seems like we don't need to remove the object explicitly?

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
ymorimo: I think it is OK. The transaction object is removed in the beginning here because it is assumed that once the user sends a `commit` or `abort` request, the transaction should be finished and not be able to be continued in the next request.
In the normal case, either `commit` or `abort` is executed in this block. If the runtime crashes before committing or aborting, the transaction object is being not accessible, then it should eventually fail. (Is my understanding correct?)

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
ymorimo: Yes, the objects could be removed explicitly in the `instrumentExecutionResult()` which is the final instrumentation, but I don't think it's necessary.

brfrn169 (commented):
Left several minor suggestions. Other than that, LGTM. Please take a look when you have time!

graphql/src/main/java/com/scalar/db/graphql/datafetcher/TransactionInstrumentation.java
```diff
@@ -0,0 +1,191 @@
...
+    Directive directive = transactionDirectives.get(0); // @transaction is not repeatable
+    Argument txIdArg = directive.getArgument(Constants.TRANSACTION_DIRECTIVE_TX_ID_ARGUMENT_NAME);
+    Argument commitArg =
+        directive.getArgument(Constants.TRANSACTION_DIRECTIVE_COMMIT_ARGUMENT_NAME);
+
+    DistributedTransaction transaction;
+    if (txIdArg == null) {
+      // start a new transaction
+      try {
+        transaction = transactionManager.start();
```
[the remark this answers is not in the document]
ymorimo: Applied the suggestions in d2f72a8 and 03805f2. Thank you!

ymorimo:
Added the logger to `MutationAbortDataFetcher` also since loggers have been added to other classes in the merged base `graphql` branch.

graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
```diff
@@ -119,6 +118,7 @@ private GraphQLCodeRegistry createGraphQLCodeRegistry(
             new QueryScanDataFetcher(storage, helper));
       }
     }
+    builder.dataFetcher(coordinates(mutationObjectType, "abort"), new MutationAbortDataFetcher());
```
brfrn169, line 121: Don't we need to do this:
```suggestion
    if (transactionManager != null) {
      builder.dataFetcher(coordinates(mutationObjectType, "abort"), new MutationAbortDataFetcher());
    }
```

brfrn169 (commented):
LGTM other than what I commented. Please take a look!

graphql/src/main/java/com/scalar/db/graphql/GraphQlFactory.java
ymorimo: Good point! I missed that. Applied the change and arranged tests for this point.

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #429, opened 2021-12-10
Upgrade log4j version

brfrn169:
We need to upgrade the log4j version because it has a vulnerability explained in the following article:
https://www.lunasec.io/docs/blog/log4j-zero-day/

In Scalar DB, only Scalar DB Server uses log4j for now.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #430, opened 2021-12-11
Remove TransactionRuntimeException

brfrn169:
This is a refactoring thing. Please take a look!


thongdk8 (dismissed):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -185,15 +184,15 @@ public int hashCode() {
     private void checkNotMissingRequired(Result result) {
       if (!result.getValue(Attribute.ID).isPresent()
           || !result.getValue(Attribute.ID).get().getAsString().isPresent()) {
-        throw new RequiredValueMissingException("id is missing in the coordinator state");
+        throw new IllegalArgumentException("id is missing in the coordinator state");
```
feeblefakie, line 187: I'm not sure if `IllegalArgumentException` is appropriate here. It is more like a software's bug, so something different should be used?

feeblefakie (commented):
Overall looking good! Left one question.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
brfrn169: Thank you for taking a look at this!

Yes, that's true. We should not use `IllegalArgumentException` here. Maybe we can simply use `CoordinatorException` here. I'll fix it. Thanks.

feeblefakie (approved):
LGTM! Thank you!

PR #431, opened 2021-12-12
Use different datasource for table metadata in JDBC adapter

brfrn169:
Currently, since using the same datasource (connection pool) in Crud (Get/Scan/Put/Delete/Mutate) operations and the table metadata operation, something like a deadlock could happen (it's a very rare case, though). This PR makes the table metadata operation use a different datasource (connection pool) from Crud operations to avoid the deadlock. Please take a look!


thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #434, opened 2021-12-14
Move CoordinatorException to the consensuscommit package

brfrn169:
Since `CoordinatorException` is Consensus Commit specific, we should move it to the `consensuscommit` package. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorException.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorException.java
@@ -1,6 +1,6 @@
-package com.scalar.db.exception.transaction;
+package com.scalar.db.transaction.consensuscommit;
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorException.java
```diff
@@ -1,6 +1,6 @@
-package com.scalar.db.exception.transaction;
+package com.scalar.db.transaction.consensuscommit;
 
-public class CoordinatorException extends TransactionException {
+public class CoordinatorException extends Exception {
```
brfrn169, line 3: I also changed `CoordinatorException` to extend from `TransactionException` to `Exception`.

thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #441, opened 2021-12-24
Handling namespaces/tables/columns with reserved keywords in Cassandra and JDBC adapters

thongdk8:
Please take a look! Thank you!


core/src/test/java/com/scalar/db/storage/cassandra/CassandraAdminTest.java
```diff
@@ -58,8 +58,8 @@ public void setUp() throws Exception {
   @Test
   public void getTableMetadata_ClusterManagerShouldBeCalledProperly() throws ExecutionException {
     // Arrange
-    String namespace = "ns";
-    String table = "table";
+    String namespace = "sample_ns";
```
brfrn169, line 61: Just a question. Why do we need rename it?

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -58,7 +59,7 @@ public void createTable(
   @Override
   public void createNamespace(String namespace, Map<String, String> options)
       throws ExecutionException {
-    CreateKeyspace query = SchemaBuilder.createKeyspace(namespace);
+    CreateKeyspace query = SchemaBuilder.createKeyspace(Metadata.quoteIfNecessary(namespace));
```
brfrn169, line 62: Maybe it's better to use static import.

brfrn169 (commented):
Thank you! Left several comments. Please take a look!

core/src/test/java/com/scalar/db/storage/cassandra/CassandraAdminTest.java
thongdk8: The `table` is a reserved keyword, if we keep using it then the expected query has to be changed as well. So I renamed it.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
thongdk8: I changed to use static import [commit](https://github.com/scalar-labs/scalardb/pull/441/commits/2e093d1dcbb93f2e1387f73fcb1c4349d52af300)

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -242,7 +260,9 @@ void createSecondaryIndex(String fullKeyspace, String table, Set<String> seconda
     for (String index : secondaryIndexNames) {
       String indexName = String.format("%s_%s_%s", table, INDEX_NAME_PREFIX, index);
       SchemaStatement createIndex =
-          SchemaBuilder.createIndex(indexName).onTable(fullKeyspace, table).andColumn(index);
+          SchemaBuilder.createIndex(indexName)
```
brfrn169, line 263: Don't we need to quote this index name?

core/src/main/java/com/scalar/db/storage/cassandra/DeleteStatementHandler.java
```diff
@@ -56,7 +57,9 @@ protected ResultSet execute(BoundStatement bound, Operation operation) {
 
   private com.datastax.driver.core.querybuilder.Delete prepare(Delete del) {
     com.datastax.driver.core.querybuilder.Delete delete =
-        QueryBuilder.delete().from(del.forNamespace().get(), del.forTable().get());
+        QueryBuilder.delete()
+            .from(
+                quoteIfNecessary(del.forNamespace().get()), quoteIfNecessary(del.forTable().get()));
     com.datastax.driver.core.querybuilder.Delete.Where where = delete.where();
 
     del.getPartitionKey().forEach(v -> where.and(QueryBuilder.eq(v.getName(), bindMarker())));
```
brfrn169, line 65: Don't we need to quote `v.getName()`?

core/src/main/java/com/scalar/db/storage/cassandra/InsertStatementHandler.java
```diff
@@ -58,7 +59,9 @@ protected ResultSet execute(BoundStatement bound, Operation operation) {
   }
 
   private Insert prepare(Put put) {
-    Insert insert = insertInto(put.forNamespace().get(), put.forTable().get());
+    Insert insert =
+        insertInto(
+            quoteIfNecessary(put.forNamespace().get()), quoteIfNecessary(put.forTable().get()));
 
     put.getPartitionKey().forEach(v -> insert.value(v.getName(), bindMarker()));
     put.getClusteringKey().ifPresent(k -> k.forEach(v -> insert.value(v.getName(), bindMarker())));
```
brfrn169, line 67: Ditto.

brfrn169 (commented):
Just left several comments. I'm wondering if we need to quote column names as well. If yes, I think this PR misses a lot of column name quotations. We need unit tests to check if all namespaces/tables/columns with reserved keywords are quoted.

core/src/test/java/com/scalar/db/storage/cassandra/CassandraAdminTest.java
brfrn169: So we need to add other unit tests to check if all namespaces/tables/columns with reserved keywords are quoted.

thongdk8: I added the unit tests and the integration test as well for reserved keywords. Please take a look again.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
thongdk8: We don't need to quote the index name I think.

core/src/main/java/com/scalar/db/storage/cassandra/DeleteStatementHandler.java
thongdk8: Thanks for pointing it out. We have to quote all the columns in case it is a reserved keyword

core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraWithReservedKeywordsIntegrationTest.java
```diff
@@ -0,0 +1,25 @@
+package com.scalar.db.storage.cassandra;
+
+import com.scalar.db.config.DatabaseConfig;
+import com.scalar.db.storage.StorageIntegrationTestBase;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+@SuppressFBWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
+public class CassandraWithReservedKeywordsIntegrationTest extends StorageIntegrationTestBase {
```
brfrn169, line 8: I think we should add new integration tests for reserved keywords rather than extending `StorageIntegrationTestBase` because we don't need to run all the tests in `StorageIntegrationTestBase` to check if reserved keywords work or not. We should have minimum integration tests to check it.

Also, I think we should run the integration tests for reserved keywords not only for the Cassandra adaptors but also all the adaptors. So can you please add tests for all the adaptors? Thanks.

brfrn169 (commented):
Thank you! I left a comment for it. Please take a look!

core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraWithReservedKeywordsIntegrationTest.java
thongdk8: I fixed the reserved keyword issue and added tests for other storage adapters as well. Please take a look again when you have time.

core/src/integration-test/java/com/scalar/db/storage/StorageWithReservedKeywordIntegrationTestBase.java
```diff
@@ -0,0 +1,531 @@
...
+@SuppressFBWarnings({"ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD", "MS_CANNOT_BE_FINAL"})
+public abstract class StorageWithReservedKeywordIntegrationTestBase {
+
+  protected static String NAMESPACE = "integration_testing";
+  protected static String TABLE = "test_table";
+  protected static String COL_NAME1 = "c1";
+  protected static String COL_NAME2 = "c2";
+  protected static String COL_NAME3 = "c3";
+  protected static String COL_NAME4 = "c4";
+  protected static String COL_NAME5 = "c5";
```
brfrn169, line 52: It looks like these don't need to be `static`. And if these namespace and table and column names are different for each adaptor, I think we can add abstract methods as follows and implement them for each adaptor's test:
```java
  protected abstract String getNamespace();

  protected abstract String getTable();

  protected abstract String getColumnName1();

  protected abstract String getColumnName2();

  protected abstract String getColumnName3();

  protected abstract String getColumnName4();

  protected abstract String getColumnName5();
```

brfrn169 (commented):
Left several comments. Please take a look!

core/src/integration-test/java/com/scalar/db/storage/StorageWithReservedKeywordIntegrationTestBase.java
thongdk8: Thanks for the suggestion. I changed to use abstract methods [commit](https://github.com/scalar-labs/scalardb/pull/441/commits/0216d485c07c6ab2c4576fb11edd1ce99fced0b6).
I'm not sure but when don't use `static` the column name values become null after running the first test, so I kept using `static` for them.

core/src/main/java/com/scalar/db/storage/cassandra/ClusterManager.java
```diff
@@ -58,11 +60,11 @@ public Session getSession() {
    * @return {@code TableMetadata}
    */
   public TableMetadata getMetadata(String keyspace, String table) {
-    KeyspaceMetadata metadata = cluster.getMetadata().getKeyspace(keyspace);
-    if (metadata == null || metadata.getTable(table) == null) {
+    KeyspaceMetadata metadata = cluster.getMetadata().getKeyspace(quoteIfNecessary(keyspace));
+    if (metadata == null || metadata.getTable(quoteIfNecessary(table)) == null) {
```
brfrn169, line 64:
```suggestion
    if (metadata == null) {
```

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -320,6 +320,10 @@ public void truncateTable(String namespace, String table) throws ExecutionExcept
     }
   }
 
+  public static String quoteKeyword(String keyword) {
```
brfrn169, line 323: Please create a util class like `CosmosUtils` and move this method to it.

core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java
```diff
@@ -1,5 +1,7 @@
 package com.scalar.db.storage.cosmos;
```
brfrn169, line 1: Do we need to quote the following:
https://github.com/scalar-labs/scalardb/blob/fix/cassandra-admin-reserved-keywords/core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java#L137
https://github.com/scalar-labs/scalardb/blob/fix/cassandra-admin-reserved-keywords/core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java#L166
https://github.com/scalar-labs/scalardb/blob/fix/cassandra-admin-reserved-keywords/core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java#L200

If yes, we need to add tests for them to detect these misses.

core/src/main/java/com/scalar/db/storage/dynamo/BatchHandler.java
```diff
@@ -135,10 +136,14 @@ private Update makeUpdate(com.scalar.db.api.Put put, TableMetadata tableMetadata
...
         .updateExpression(expression)
         .conditionExpression(condition)
-        .expressionAttributeNames(columnMap)
-        .expressionAttributeValues(bindMap)
-        .build();
+        .expressionAttributeValues(bindMap);
 
+    Map<String, String> conditionAttributeNameMap = dynamoMutation.getConditionAttributeNameMap();
+    if (!conditionAttributeNameMap.isEmpty()) {
+      expressionAttributeNameMap.putAll(dynamoMutation.getConditionAttributeNameMap());
```
brfrn169, line 143:
```suggestion
      expressionAttributeNameMap.putAll(conditionAttributeNameMap);
```

core/src/main/java/com/scalar/db/storage/dynamo/DeleteStatementHandler.java
```diff
@@ -75,6 +75,11 @@ private void delete(Delete delete, TableMetadata tableMetadata) {
       builder.conditionExpression(condition);
     }
 
+    Map<String, String> conditionAttributeNameMap = dynamoMutation.getConditionAttributeNameMap();
+    if (!conditionAttributeNameMap.isEmpty()) {
+      builder.expressionAttributeNames(dynamoMutation.getConditionAttributeNameMap());
```
brfrn169, line 80:
```suggestion
      builder.expressionAttributeNames(conditionAttributeNameMap);
```

brfrn169 (commented):
I checked Cassandra and Cosmos and left several comments. And please separate this PR into multiple PRs for each adaptor since it's a little bit big PR? I think one PR for Cassandra and JDBC is fine.

core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java
thongdk8: Thank for pointing it out. We need to quote them. I fixed it and added tests for them as well.

brfrn169:
@thongdk8
> And please separate this PR into multiple PRs for each adaptor since it's a little bit big PR? I think one PR for Cassandra and JDBC is fine.

How about this?

thongdk8:
> How about this?
@brfrn169 Ah sorry, I split this PR to #461 and #462 and this one. There is a duplicate in the integration test base of reserved keywords, so better to review others after one of these is merged.

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
```diff
@@ -1,5 +1,6 @@
 package com.scalar.db.storage.cassandra;
```
brfrn169, line 1: It looks like we also need to quote column name for ordering?
https://github.com/scalar-labs/scalardb/blob/c6dc9766f2cc55cc92d8cb039a4c50fd42c1897f/core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java#L264
https://github.com/scalar-labs/scalardb/blob/c6dc9766f2cc55cc92d8cb039a4c50fd42c1897f/core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java#L266
https://github.com/scalar-labs/scalardb/blob/c6dc9766f2cc55cc92d8cb039a4c50fd42c1897f/core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java#L269

If yes, please also add tests for it.

brfrn169 (commented):
Left a comment. Please take a look!

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
thongdk8: We need it as well, the scan test with the range also includes the order in the query but seems the `two` keyword I used does not affect the query in this case. I change the keyword to `select` then the test failed. I added the quote for ordering and changed the column name as well [commit](https://github.com/scalar-labs/scalardb/pull/441/commits/2d136ee2f33c57a51b6b960ad43363529968faa7)

brfrn169: So you mean the `two` keyword is not an actual reserved keyword, right? If so, please check if all the reserved keywords you use in the integration tests are really reserved keywords?

thongdk8: In the DataStax document, the `two` is mentioned in the reserved keyword list but seems it depends on where the keyword is placed in the query, some still working. For some obvious keywords like `select`, `update` should be really a keyword

brfrn169: Okay I see. It would be great if you re-check the reserve keywords for all the adapters just in case.

brfrn169 (approved):
LGTM! Thank you!

As I said in another comment, It would be great if you re-check the reserve keywords for all the adapters just in case.

feeblefakie (approved):
LGTM! Thank you!
It's hard to verify them all with review so it would be great if all the cases (or most cases if not doable) are covered by unit tests.

PR #349, opened 2021-10-20
Add end-to-end test for schema loader to CI

thongdk8:
This PR is made for adding e2e test for schema loader to CI. PTAL. Thank you.
[PR](https://github.com/scalar-labs/scalardb/pull/348) should be merged before this.

PR #444, opened 2021-12-25
Refactor schema loader integration test

brfrn169:
I refactored the code in #349. This PR will be merged to the branch for #349. Please take a look!


schema-loader/build.gradle
```diff
@@ -118,59 +118,55 @@ shadowJar {
...
-    dependsOn(shadowJarUnversioning)
+    dependsOn shadowJar
     description = 'Runs the schema loader integration tests for Cassandra.'
     group = 'verification'
     testClassesDirs = sourceSets.integrationTestCassandra.output.classesDirs
     classpath = sourceSets.integrationTestCassandra.runtimeClasspath
     outputs.upToDateWhen { false }  // ensures integration tests are run every time when called
     options {
         systemProperties(System.getProperties())
+        systemProperty("scalardb.schemaloader.jar_path", "${project.buildDir}/libs/${archivesBaseName}-${project.version}.jar")
```
brfrn169, line 130: Rather than renaming the jar file, we can use the jar file under the build dir.

schema-loader/src/integration-test/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
```diff
@@ -1,152 +1,211 @@
...
+                .build());
+  }
+
+  protected void writeSchemaFile(Map<String, Object> schemaJsonMap) throws IOException {
+    Gson gson = new Gson();
+    try (Writer writer =
+        new OutputStreamWriter(new FileOutputStream(SCHEMA_FILE), StandardCharsets.UTF_8)) {
+      gson.toJson(schemaJsonMap, writer);
+    }
   }
```
brfrn169, line 133: We can write the schema json file for each test run.

schema-loader/src/main/java/com/scalar/db/schemaloader/command/CosmosCommand.java
```diff
@@ -47,11 +47,19 @@
...
       hidden = true)
-  private String prefix;
+  private String tableMetadataDatabasePrefix;
+
+  // For test
+  @Option(
+      names = {"--coordinator-namespace-prefix"},
+      description = "Coordinator namespace prefix",
+      hidden = true)
+  private String coordinatorNamespacePrefix;
```
brfrn169, line 62: We can add two parameters for table metadata database prefix and coordinator namespace prefix.

schema-loader/src/integration-test/java/com/scalar/db/schemaloader/server/SchemaLoaderWithScalarDbServer.java
```diff
@@ -0,0 +1,38 @@
+package com.scalar.db.schemaloader.server;
```
brfrn169, line 1: I also added an integration test with Scalar DB Server.

schema-loader/src/integration-test/java/com/scalar/db/schemaloader/multistorage/MultiStorageSchemaLoaderIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
+package com.scalar.db.schemaloader.multistorage;
```
brfrn169, line 1: I also added an integration test for multi-storage.

schema-loader/build.gradle
thongdk8: Is there a way to set this property visible for all the tasks? Then we don't have to rewrite it in each one

thongdk8 (approved):
Thank you! LGTM!
(just left a question but it does block the PR being to be merged)

schema-loader/build.gradle
brfrn169: I'm not sure the way. Instead, I added a variable for the fat jar path and made use of it for the tasks.

PR #452, opened 2022-01-06
Should rollback a transaction when expiring it in Two-phase commit transactions

brfrn169:
I found we don't rollback a transaction when expiring it in Two-phase commit transactions. This PR fixes it. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #453, opened 2022-01-06
Refactor Scalar DB Server code

brfrn169:
This is very trivial things. Please take a look!

### DIFF
--- a/server/src/main/java/com/scalar/db/server/ServerConfig.java
+++ b/server/src/main/java/com/scalar/db/server/ServerConfig.java
@@ -1,11 +1,8 @@
-package com.scalar.db.server.config;
+package com.scalar.db.server;
 
 import static com.scalar.db.config.ConfigUtils.getInt;
 import static com.scalar.db.config.ConfigUtils.getString;
 
-import com.scalar.db.server.GateKeeper;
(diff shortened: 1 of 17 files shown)


server/src/main/java/com/scalar/db/server/ServerConfig.java
```diff
@@ -1,11 +1,8 @@
-package com.scalar.db.server.config;
+package com.scalar.db.server;
```
brfrn169, line 1: I moved `ServerConfig` from the `com.scalar.db.server.config` package to the `com.scalar.db.server` package.

server/src/main/java/com/scalar/db/server/ServerModule.java
```diff
@@ -1,4 +1,4 @@
-package com.scalar.db.server.service;
+package com.scalar.db.server;
```
brfrn169, line 1: I also moved `ServerModule` from the `com.scalar.db.server.service` package to the `com.scalar.db.server` package.

thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #454, opened 2022-01-07
Refactor TableMetadataManager

brfrn169:
Please take a look!


core/src/main/java/com/scalar/db/util/TableMetadataManager.java
```diff
@@ -1,12 +1,11 @@
-package com.scalar.db.storage.common;
+package com.scalar.db.util;
```
brfrn169, line 1: I moved the `TableMetadataManager` class from the `com.scalar.db.storage.common` package to the `com.scalar.db.util` package because it's used not only the classes in the `storage` package but also the classes in the `transaction` package.

core/src/main/java/com/scalar/db/util/TableMetadataManager.java
```diff
@@ -18,11 +17,10 @@
 
   private final LoadingCache<TableKey, Optional<TableMetadata>> tableMetadataCache;
 
-  public TableMetadataManager(DistributedStorageAdmin admin, DatabaseConfig config) {
+  public TableMetadataManager(DistributedStorageAdmin admin, long cacheExpirationTimeSecs) {
```
brfrn169, line 20: I also changed this constructor signature.

feeblefakie (dismissed):
LGTM! Thank you!

thongdk8 (dismissed):
LGTM! Thank you!

PR #455, opened 2022-01-07
Deprecate XXXService classes

brfrn169:
Since I think we no longer need the `XXXService` classes, this PR deprecates them. Please take a look!


feeblefakie (approved):
LGTM! Thank you!
BTW, I don't really remember why we had XXXService classes.

thongdk8 (approved):
LGTM! Thank you!

PR #461, opened 2022-01-13
Handling namespaces/tables/columns with reserved keywords in Cosmos DB adapter

thongdk8:
Please take a look! Thank you!


core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -306,7 +306,7 @@ public void truncateTable(String namespace, String table) throws ExecutionExcept
 
       CosmosPagedIterable<Record> records =
           container.queryItems(
-              "SELECT t." + ID + ", t." + CONCATENATED_PARTITION_KEY + " FROM " + table + " t",
+              "SELECT t." + ID + ", t." + CONCATENATED_PARTITION_KEY + " FROM " + " t",
```
brfrn169, line 309:
```suggestion
              "SELECT t." + ID + ", t." + CONCATENATED_PARTITION_KEY + " FROM " + "t",
```

core/src/main/java/com/scalar/db/storage/cosmos/CosmosUtils.java
```diff
@@ -0,0 +1,7 @@
+package com.scalar.db.storage.cosmos;
+
+public class CosmosUtils {
```
brfrn169, line 3:
```suggestion
public final class CosmosUtils {

  private CosmosUtils() {}

```

brfrn169 (commented):
Left very minor suggestions. Other than that, LGTM. Thank you.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosUtils.java
```diff
@@ -0,0 +1,7 @@
+package com.scalar.db.storage.cosmos;
+
+public final class CosmosUtils {
```
brfrn169, line 3: Please don't just ignore my suggestion. If you have reasons why you don't want to apply my suggestion, please tell me and we can discuss it:
```suggestion
public final class CosmosUtils {

  private CosmosUtils() {}

```

thongdk8: Ah sorry, I missed the method.

brfrn169: > Why do we need this method btw?

I found a good article for it. You can check it:
https://www.informit.com/articles/article.aspx?p=1216151&seqNum=4

thongdk8: Great! Thank you!

core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosWithReservedKeywordIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
+package com.scalar.db.storage.cosmos;
+
+import com.scalar.db.config.DatabaseConfig;
+import com.scalar.db.storage.StorageWithReservedKeywordIntegrationTestBase;
+import java.util.Map;
+import java.util.Optional;
+
+public class CosmosWithReservedKeywordIntegrationTest
```
feeblefakie, line 8: I don't fully get this but this class seems not referenced at all. Is it OK?

brfrn169 (approved):
LGTM! Thank you!

core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosWithReservedKeywordIntegrationTest.java
thongdk8: Seems the IDE still loading? There is no warning about that in my IDE. The CI running included this class already I think

feeblefakie (approved):
LGTM! Thank you!

PR #463, opened 2022-01-13
Update JDBC adapter to use executeBatch when multiple mutations are passed

brfrn169:
I think we can use `executeBatch` for mutations without any conditions. This PR makes the change for it. Please take a look!


core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
```diff
@@ -182,21 +193,80 @@ public boolean mutate(
...
       }
+      for (PreparedStatement preparedStatement : preparedStatementMap.values()) {
+        preparedStatement.executeBatch();
+      }
+      return true;
+    } finally {
+      for (PreparedStatement preparedStatement : preparedStatementMap.values()) {
+        preparedStatement.close();
+      }
+    }
```
brfrn169, line 222: This is the main change. But this optimization is only for mutations without any conditions, and we can pass mutations only for a single partition to the mutate API. So I wonder how many use cases can take advantage of this optimization.

brfrn169:
As discussed offline, we decided not to merge this PR because this optimization is not very useful for many use cases. Closing it.

PR #462, opened 2022-01-13
Handling namespaces/tables/columns with reserved keywords in DynamoDB adapter

thongdk8:
Please take a look! Thank you!

PR #464, opened 2022-01-13
Refactor #462

brfrn169:
This PR refactors the PR #462. Please take a look and check if this refactoring is correct!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/dynamo/PutStatementHandler.java
+++ b/core/src/main/java/com/scalar/db/storage/dynamo/PutStatementHandler.java
@@ -10,7 +10,6 @@ import com.scalar.db.exception.storage.NoMutationException;
 import com.scalar.db.exception.storage.RetriableExecutionException;
 import com.scalar.db.util.TableMetadataManager;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/scalar/db/storage/dynamo/PutStatementHandler.java
```diff
@@ -58,46 +57,40 @@ private void execute(Put put, TableMetadata tableMetadata) {
...
-      expressionAttributeNameMap.putAll(dynamoMutation.getColumnMap());
+      expressionAttributeNameMap = dynamoMutation.getColumnMap();
       bindMap = dynamoMutation.getValueBindMap();
     } else {
       expression = dynamoMutation.getUpdateExpression();
-      condition = dynamoMutation.getCondition();
-      expressionAttributeNameMap.putAll(dynamoMutation.getColumnMap());
+      condition = dynamoMutation.getIfExistsCondition() + " AND " + dynamoMutation.getCondition();
+      expressionAttributeNameMap = dynamoMutation.getColumnMap();
+      expressionAttributeNameMap.putAll(dynamoMutation.getConditionColumnMap());
```
thongdk8, line 81: Should we do combining maps into a new map instead of mutating a map from an existing one getting from another place?

thongdk8 (approved):
Just left a discussion. Other than that, LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/dynamo/PutStatementHandler.java
brfrn169: Thank you for asking that. Yes, that's a good point. Actually, it's basically not a good practice to mutate a map returned from another class's method because we don't know the implementation of the map (ex., it might be an immutable one). I mutate the map here because we already do the same thing in some other parts (ex., https://github.com/scalar-labs/scalardb/blob/dbe49adc20ab88f4d53fe68ab1c057557238abbe/core/src/main/java/com/scalar/db/storage/dynamo/PutStatementHandler.java#L82-L83), and we know `DynamoMutation.getColumnMap()` returns a mutable `HashMap` since `DynamoMutation` is our class. 

What do you think? If no objection, I'll merge it. Thanks.

thongdk8: I see. Let's merge it. Thanks!

PR #465, opened 2022-01-14
Add TransactionalTableMetadataManager

brfrn169:
This PR adds `TransactionalTableMetadataManager` that retrieves and caches `TransactionalTableMetadata` instances with table metadata and transactional meta info for transactional tables. We don't use these classes yet but will use them in another PR. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadata.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadata.java
@@ -1,0 +1,84 @@
+package com.scalar.db.transaction.consensuscommit;
+
+import com.scalar.db.api.Scan;
+import com.scalar.db.api.Scan.Ordering.Order;
+import com.scalar.db.api.TableMetadata;
+import com.scalar.db.io.DataType;
+import com.scalar.db.util.ImmutableLinkedHashSet;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class TransactionalTableMetadata {
+
+  private final TableMetadata tableMetadata;
+  private final LinkedHashSet<String> transactionalMetaColumnNames;
+  private final LinkedHashSet<String> beforeImageColumnNames;
+  private final LinkedHashSet<String> afterImageColumnNames;
+
+  public TransactionalTableMetadata(TableMetadata tableMetadata) {
+    this.tableMetadata = tableMetadata;
+    transactionalMetaColumnNames =
+        new ImmutableLinkedHashSet<>(
+            tableMetadata.getColumnNames().stream()
+                .filter(c -> ConsensusCommitUtils.isTransactionalMetaColumn(c, tableMetadata))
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadata.java
```diff
@@ -0,0 +1,80 @@
...
+    return transactionalMetaColumnNames;
+  }
+
+  public LinkedHashSet<String> getBeforeImageColumnNames() {
+    return beforeImageColumnNames;
+  }
+
+  public LinkedHashSet<String> getAfterImageColumnNames() {
+    return afterImageColumnNames;
+  }
```
brfrn169, line 79: We can get transactional meta column names, before image column names, and after image column names from `TransactionalTableMetadata`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadataManager.java
```diff
@@ -0,0 +1,87 @@
...
+import com.scalar.db.api.DistributedStorageAdmin;
+import com.scalar.db.api.Operation;
+import com.scalar.db.api.TableMetadata;
+import com.scalar.db.exception.storage.ExecutionException;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+
+public class TransactionalTableMetadataManager {
```
brfrn169, line 15: This class retrieves a `TableMetadata` instance from `DistributedStorageAdmin` and convert it to a `TransactionalTableMetadata` instance and cache it.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadataManager.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Returns a transactional table metadata corresponding to the specified operation.
+   *
+   * @param operation an operation
+   * @return a table metadata. null if the table is not found.
+   * @throws ExecutionException if the operation failed
+   */
+  public TransactionalTableMetadata getTransactionalTableMetadata(Operation operation)
```
feeblefakie, line 48: Isn't it more general to make it `getTransactionalTableMetadata(String namespace, String table)`?

feeblefakie (approved):
LGTM! Thank you!
I left one comment but it doesn't block it from being merged.

thongdk8 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadataManager.java
brfrn169: Actually, we already have `TableMetadataManager` similar to this class, and it also has `getTransactionalTableMetadata(Operation operation)`:
https://github.com/scalar-labs/scalardb/blob/554e408984380131ca8d270bb95352515fb5b077/core/src/main/java/com/scalar/db/util/TableMetadataManager.java#L42

For consistency, I will keep the method. And we can change it in the future if needed. Thank you!

PR #469, opened 2022-01-18
[WIP] Allow reading already written data in Consensus Commit

brfrn169:
(This PR is working in progress. Please don't merge it for now)

This PR allows reading already written data in Consensus Commit.

The main changes are in the following classes:
- CrudHandler
- Snapshot
- TableSnapshot (NEW)

This PR is working in progress, and I need to add more tests and do some refactoring. But I think we can have the first discussion based on it. 

Please take a look!


brfrn169:
As discussed, as a first step, we support **Get-After-Put** only. And, we support **Scan-After-Put** as a next step because we don't think many use cases do **Scan-After-Put**.

I will create a new PR for it. Closing this PR for now.

PR #470, opened 2022-01-19
Add projections to Get and Scan operations in GraphQL

ymorimo:
For get and scan operations, use the `withProjections()` according to the specified selections of the GraphQL inputs.

For example, the GraphQL query

```graphql
query {
  table1_scan(scan: { partitionKey: { c1: 1 }}) {
    table1 {
      c1
      c2
  }
}
```

will be converted to a `Scan` command like the following:

```java
new Scan(new Key(new IntValue("c1", 1)))
  .withProjections(Arrays.asList("c1", "c2"))
```

### DIFF
--- a/graphql/src/main/java/com/scalar/db/graphql/datafetcher/QueryGetDataFetcher.java
+++ b/graphql/src/main/java/com/scalar/db/graphql/datafetcher/QueryGetDataFetcher.java
@@ -65,12 +65,12 @@ public class QueryGetDataFetcher
                 helper.createPartitionKeyFromKeyArgument(key),
                 helper.createClusteringKeyFromKeyArgument(key))
             .forNamespace(helper.getNamespaceName())
-            .forTable(helper.getTableName());
+            .forTable(helper.getTableName())
+            .withProjections(DataFetcherHelper.getProjections(environment));
     String consistency = (String) getInput.get("consistency");
     if (consistency != null) {
       get.withConsistency(Consistency.valueOf(consistency));
     }
-    helper.addProjections(get, environment);
 
     return get;
   }
--- a/graphql/src/main/java/com/scalar/db/graphql/datafetcher/QueryGetDataFetcher.java
+++ b/graphql/src/main/java/com/scalar/db/graphql/datafetcher/QueryGetDataFetcher.java
@@ -43,7 +43,7 @@ public class QueryGetDataFetcher
       performGet(environment, get)
           .ifPresent(
               dbResult -> {
-                for (String fieldName : helper.getFieldNames()) {
+                for (String fieldName : get.getProjections()) {
                   dbResult.getValue(fieldName).ifPresent(value -> data.put(fieldName, value.get()));
                 }
               });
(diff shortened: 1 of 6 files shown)


graphql/src/main/java/com/scalar/db/graphql/datafetcher/QueryScanDataFetcher.java
```diff
@@ -42,7 +44,15 @@ public QueryScanDataFetcher(DistributedStorage storage, DataFetcherHelper helper
     LOGGER.debug("got scan argument: {}", scanInput);
     Scan scan = createScan(scanInput);
 
-    // TODO: scan.withProjections()
+    // Add the specified field names, filtered by "*/*", to the scan object as the projections
+    // since the fields of the scan input are represented in the following format:
+    // <object_type_name>_ScanPayLoad.<object_type_name>/<object_type_name>.<field_name>
+    List<String> selectedFieldNames =
```
brfrn169, line 50: Do we need to do the same thing for Get operations?

brfrn169 (commented):
Thank you! I left a question. Please take a look when you have time!

graphql/src/main/java/com/scalar/db/graphql/datafetcher/QueryScanDataFetcher.java
ymorimo: I did not notice that `Get` operations also had projections. Updated to do the same for them as `Scan`s, and did some refactoring. Please take a look again. Thank you!

brfrn169 (approved):
LGTM! Thank you!

graphql/src/main/java/com/scalar/db/graphql/datafetcher/QueryGetDataFetcher.java
```diff
@@ -69,6 +70,7 @@ Get createGet(Map<String, Object> getInput) {
     if (consistency != null) {
       get.withConsistency(Consistency.valueOf(consistency));
     }
+    helper.addProjections(get, environment);
```
feeblefakie, line 73: I'm not sure how this class and `DataFetcherHelper` are used but I have some thoughts about the class design.

First of all, is it OK that this class is not thread-safe? Multiple threads calling `createGet` would cause some issues due to the `helper.addProjections`?
Second, from a user's perspective, I wouldn't expect `createGet` to have a side effect on the  `DataFetcherHelper` object, so I think the interface might need to be revised.

Can this class create `projections` and return it by a (new) method?

feeblefakie (commented):
Thank you!
I left a comment.

graphql/src/main/java/com/scalar/db/graphql/datafetcher/QueryGetDataFetcher.java
ymorimo: Actually, the `helper.addProjections()` method doesn't have any side effects on the `helper` instance, and the method could have been static. I've just made a fix for that. Then I don't think there is a problem with thread safety.
But the design of `DataFetcherHelper` is getting not good. I made it to delegate common procedures for all data fetchers, but now the responsibilities for that class are not clear. I'm going to refactor the structures of data fetcher classes.

feeblefakie (approved):
LGTM! Thank you!

PR #471, opened 2022-01-21
GraphQL custom validations

ymorimo:
This PR adds custom validations for user inputs in addition to the GraphQL default validations against the schema.

1. Validations for `abort` mutation:

    To use `abort`,
    - it must be alone in a mutation operation
    - a `@transaction` directive with a non-null `txId` argument is required
    - a `commit` argument must be omitted or specified as `false`

2. Validations for scalar values:

    Where one of `intValue`, `bigIntValue`, `floatValue`, `doubleValue`, `textValue`, and `booleanValue` is needed, it is validated that only one value is specified.

    - in scan boundaries (`start`, `end`)
    - in conditional `expressions` for `PutIf`and `DeleteIf`
    - if the condition type is either `PutIfExists`, `PutIfNotExists`, or `DeleteIfExists`, the whole `expressions` key must be omitted

Note: The validation classes are placed in the package `com.scalar.db.graphql.instrumentation.validation`. I'm going to move the `TransactionInstrumentation` class into `com.scalar.db.graphql.instrumentation` later on. And the `com.scalar.db.graphql.schema.Constants` class is already not limited to the `schema` package, it might also be better to move it to the parent package.


graphql/src/main/java/com/scalar/db/graphql/instrumentation/validation/ScanStartAndEndValidation.java
```diff
@@ -0,0 +1,62 @@
...
+              clusteringKey -> {
+                validateClusteringKey(errors, "start", clusteringKey, fieldAndArguments);
+              });
+        }
+        List<Map<String, Object>> end = (List<Map<String, Object>>) scan.get("end");
+        if (end != null) {
+          end.forEach(
+              clusteringKey -> {
+                validateClusteringKey(errors, "end", clusteringKey, fieldAndArguments);
+              });
```
brfrn169, line 37: Ditto.
```suggestion
              clusteringKey ->
                  validateClusteringKey(errors, "end", clusteringKey, fieldAndArguments));

```

brfrn169 (approved):
I just left very minor comments. Other than that, LGTM!  Thank you!

graphql/src/main/java/com/scalar/db/graphql/instrumentation/validation/ScanStartAndEndValidation.java
```diff
@@ -0,0 +1,62 @@
...
+    for (FieldAndArguments fieldAndArguments : validationEnvironment.getFields()) {
+      String fieldName = fieldAndArguments.getField().getName();
+      if (fieldName.endsWith("_scan")) {
+        Map<String, Object> scan = fieldAndArguments.getArgumentValue("scan");
+        List<Map<String, Object>> start = (List<Map<String, Object>>) scan.get("start");
+        if (start != null) {
+          start.forEach(
+              clusteringKey -> {
+                validateClusteringKey(errors, "start", clusteringKey, fieldAndArguments);
+              });
```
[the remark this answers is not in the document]
ymorimo: Sorry, but other than the braces issue, I also adapted how the validation method reports an error to the way `ConditionalExpressionValidation` in this PR does. Now the method returns an `Optional` of an error instead of modifying the `errors` list that is passed by the argument. Can you please take another quick look?

graphql/src/main/java/com/scalar/db/graphql/instrumentation/validation/AbortFieldValidation.java
```diff
@@ -0,0 +1,77 @@
...
+    ImmutableList.Builder<GraphQLError> errors = ImmutableList.builder();
+    List<Directive> transactionDirectives =
+        operationDefinition.getDirectives(Constants.TRANSACTION_DIRECTIVE_NAME);
+    SourceLocation location = abortField.get().getSourceLocation();
+    if (transactionDirectives.isEmpty()) {
+      errors.add(
+          createError(
+              "transaction directive with txId is required to abort transaction", location));
+    } else {
+      Directive directive = transactionDirectives.get(0); // @transaction is not repeatable
```
feeblefakie, line 46: What do you mean by this? `@transaction is not repeatable`

feeblefakie (approved):
LGTM! Thank you!

graphql/src/main/java/com/scalar/db/graphql/instrumentation/validation/AbortFieldValidation.java
ymorimo: Directives can be repeatable if defined so, but we don't define `@transaction` as such. So, if the list is not empty, we can only pay attention to the first element (`get(0)`).

feeblefakie: I didn't get the meaning of `repeatable` here, but after some investigation, I got it. Thank you!

brfrn169 (approved):
LGTM! Thank you!

PR #472, opened 2022-01-21
GraphQL schema refactoring

ymorimo:
This PR contains some refactorings for the GraphQL schema package.

* Renames the `CommonSchema` class to `ScalarDbTypes` to avoid using a vague ("common") name
* Splits a large method that was building all GraphQL types into small parts using named static constants
* Removes repeating code creating a series of `InputObjectField`s for scalar types (`intValue`, `bigIntValue`, ...)
* Removes the unused object type (`primaryKeyOutputObjectType`), which was used to represent the results of old mutations


brfrn169 (approved):
LGTM! Thank you!

Can we also make the `Constants` class a `final` class?

feeblefakie (approved):
LGTM! Thank you!

PR #474, opened 2022-01-22
Allow getting already written data in Consensus Commit

brfrn169:
This PR allows getting already written data in Consensus Commit.

Note that we only support `get-after-put` in this PR, not `scan-after-put`. `scan-after-put` still throw `IllegalArgumentException` even after this change.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #476, opened 2022-01-22
Fix OperationChecker for mutation operations

brfrn169:
Currently, the check for mutation operations in `OperationChecker` doesn't take `Namespace` into account. So when we specify multiple mutations with the same table and partition key but a different namespace, `OperationChecker` doesn't throw `IllegalArgumentException`. This PR fixes this bug. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #479, opened 2022-01-25
Fix Scan logic in CrudHandler in Consensus Commit

brfrn169:
It looks like the current scan logic in `CrudHandler` is not correct when a record of the scan result is already deleted (the record is in the delete set). This PR fixes the issue. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #481, opened 2022-01-27
GraphQL should reject storage access to consensus commit transactional tables

ymorimo:
1. In the GraphQL schema generation, the consensus commit meta columns are hidden from the GraphQL schema only if a table has all meta columns. (`ConsensusCommitUtils.isTransactionalTableMetadata(tableMetadata)` is `true`).

2. Accessing a table with storage method will fail if it has consensus commit meta columns.

    |                  | No meta columns | CC meta columns |
    |------------------|-----------------|-----------------|
    | Storage          | OK              | **Error**       |
    | Consensus Commit | Error           | OK              |
    | JDBC             | OK              | OK (*)          |

    (*) Since there is no way to know that the transaction is backed by JDBC native transaction at runtime, the GraphQL interface doesn't reject access to CC tables with JDBC. Alternatively, it would be possible to produce warnings at the startup of the GraphQL server, if the transaction mode is JDBC while there are CC tables in the specified namespaces by the `scalar.db.graphql.namespaces` property.

As pointed out in https://github.com/scalar-labs/scalardb/pull/470#discussion_r788728658, the class design of `DataFetcherHelper` is not good. It should be fixed in the upcoming PR.

### DIFF
--- a/graphql/src/main/java/com/scalar/db/graphql/datafetcher/DataFetcherHelper.java
+++ b/graphql/src/main/java/com/scalar/db/graphql/datafetcher/DataFetcherHelper.java
@@ -31,6 +31,7 @@ import graphql.ErrorType;
 import graphql.GraphQLError;
 import graphql.GraphqlErrorBuilder;
 import graphql.Scalars;
+import graphql.execution.AbortExecutionException;
 import graphql.schema.DataFetchingEnvironment;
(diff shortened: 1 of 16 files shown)


brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #482, opened 2022-01-27
Use only id and version to check if a read record is not changed in the validation in EXTRA_READ

brfrn169:
I don't think we need to compare all the values of a read record and a re-read record in the validation in EXTRA_READ. I think it's enough to compare `id` and `version` values. This PR makes it. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -26,7 +26,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 import javax.annotation.concurrent.NotThreadSafe;
 import org.slf4j.Logger;
@@ -259,6 +258,7 @@ public class Snapshot {
     }
   }
 
+  @VisibleForTesting
   void toSerializableWithExtraRead(DistributedStorage storage)
       throws ExecutionException, CommitConflictException {
     if (!isExtraReadEnabled()) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -347,6 +348,18 @@ void toSerializableWithExtraRead(DistributedStorage storage)
...
+    if (latestResult.isPresent() != result.isPresent()) {
+      return true;
+    }
+    if (!latestResult.isPresent()) {
+      return false;
+    }
+    return !latestResult.get().getId().equals(result.get().getId())
+        || latestResult.get().getVersion() != result.get().getVersion();
+  }
+
```
brfrn169, line 362: This is the new comparison logic. It compares only `id` and `version` values of the read and re-read records.

feeblefakie: It's a super rare case, but a read record could be deleted, created, and updated several times to get to the read version by other transactions, so we need to compare the transaction ID as well?

feeblefakie (commented):
Thank you!
I left one question. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: Thank you for taking a look at it!

Yes, I think the logic already compare the transaction ID (`!latestResult.get().getId().equals(result.get().getId())`). I think the logic compares both the transaction ID and version. That's not enough?
```java
    return !latestResult.get().getId().equals(result.get().getId())
        || latestResult.get().getVersion() != result.get().getVersion();
```

feeblefakie: Oh, sorry! I was a bit confused with Scalar DL. Thank you!

feeblefakie (dismissed):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #484, opened 2022-01-27
GraphQL constants

ymorimo:
Refactoring of the "`Constants`" class.

1. Moved the class from the package `com.scalar.graphql.schema` to `com.scalar.graphql` (and renamed it to `GraphQlConstants `) since it is not limited to schema-related things.
2. Inlined constants of the schema element names (`"transaction"`, `"txId"`, and `"commit"`), since other keywords of the same kind, such as `"get"`, `"name"`, and `"PutCondition"`, are simply hard-coded.


feeblefakie (approved):
LGTM! Thank you!

PR #485, opened 2022-01-28
[WIP] Improve the transactional read logic in Consensus Commit

brfrn169:
(This is working in progress. Please don't merge it)

The basics of the new transactional read logic is as follows:

1. Read (Get, Scan) data without before image columns
2. If the read record is a COMMITTED state, use it
3. If the read record is a PREPARED/DELETED state
    1. Re-read the latest record with before image columns
        1. If the latest record doesn't exist, regard the record as deleted
        2. If the latest record is a COMMITTED state, use it
    2. Check the coordinator state of the record
        1. If the coordinator state is COMMITTED, roll forward the record and use the after image of the record
        2. If the coordinator state is ABORTED, roll back the record and use the before image of the record
        3. If the coordinator state doesn't exist, abort the transaction it expired and use the before image of the record

The main changes are in `CrudHandler` and `RecoveryHandler`.

Please take a look!


core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
```diff
@@ -4,53 +4,167 @@
...
+        } else {
+          values.put(afterImageColumnName, result.getValue(afterImageColumnName).get());
+        }
+      }
+      return Optional.of(
+          new TransactionResult(new ResultImpl(values, metadata.getTableMetadata())));
+    } else { // TransactionState.DELETED
+      return Optional.empty();
+    }
+  }
```
brfrn169, line 140: Note that this after image result returned by `getAfterImageResult()` doesn't have the `committed_at` column.

brfrn169:
As discussed, I'm making it pending to merge. We will go back in the future. Closing.

PR #489, opened 2022-02-04
Fix bug where projections don't work properly with transactions

ymorimo:
This PR fixes a bug where a get or scan query returns nothing if it is run with `ConsensusCommit` or `TwoPhaseConsensusCommit`.

This was because `projections` of a get or scan operation are cleared by `clearProjections()` when it is passed to `get()` or `scan()` of the above classes.


brfrn169:
Thank you! But I think it's a Scalar DB side issue. Scalar DB should not change the get or scan operations passed as arguments. I'll fix it in Scalar DB. Thanks.

brfrn169:
This issue should be fixed in #498.

ymorimo:
I have confirmed that the problem has been fixed with the latest master. I'm closing this PR. Thank you!

PR #490, opened 2022-02-04
Should not publish a fat jar for Schema Loader to the maven repository

brfrn169:
Currently, the `shadowJar` task for Schema Loader overwrites the original jar because we set `archiveClassifier` to empty in the `shadowJar` task as follows:
https://github.com/scalar-labs/scalardb/blob/cb65eaacd27795e98aa51c5208ac602b4e130a83/schema-loader/build.gradle#L141

As a result, we publish the fat jar to the maven repository, which can cause a dependency conflict issue. To address this issue, it seems like we can exclude the `shadowJar` task when we run the `publish` task (and the `publishToMavenLocal` task). This PR fixes it. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #492, opened 2022-02-04
Move ResultImpl to the util package

brfrn169:
Now that the `ResultImpl` class is not used only for the classes in the `storage` package, we should move it to the `util` package. This PR does that. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cassandra/ResultInterpreter.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/ResultInterpreter.java
@@ -12,7 +12,7 @@ import com.scalar.db.io.FloatValue;
 import com.scalar.db.io.IntValue;
 import com.scalar.db.io.TextValue;
 import com.scalar.db.io.Value;
-import com.scalar.db.storage.common.ResultImpl;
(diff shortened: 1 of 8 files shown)


feeblefakie (approved):
LGTM! Thank you!
I feel `util` doesn't seem to be the best place, but I'm not sure where to put it... Hmm

thongdk8 (approved):
LGTM! Thank you!

brfrn169:

> I feel `util` doesn't seem to be the best place, but I'm not sure where to put it... Hmm

Correct. `util` might not be the best place, and I'm also not sure where to put it for now. At this point, we put all the commonly used classes to `util`. Since the classes in `util` are internally used, we can refactor them easily in the future. So if we get a better idea, let's refactor them at that time. Thanks.

feeblefakie:
@brfrn169 
> Correct. util might not be the best place, and I'm also not sure where to put it for now. At this point, we put all the commonly used classes to util. Since the classes in util are internally used, we can refactor them easily in the future. So if we get a better idea, let's refactor them at that time. Thanks.

I agree! Thank you!

PR #495, opened 2022-02-07
Should not change states of operation instances passed as arguments

brfrn169:
Currently, we change the states of operation instances (`Get`/`Scan`/`Put`/`Delete`) passed as arguments, which could be confusing and might be causing a bug. To address issue, we can copy operation instances at the begging of the methods (`get()`, `scan()`, `put()`, `delete()`, and `mutate()`) of `DistributedStorage` and `DistributedTransaction` and use them.

One problem here is that Multi Storage and Consensus Commit use `DistributedStorage` instances internally, so we copy operation instances twice in that case. To avoid that situation, this PR introduces a new configuration `scalar.db. need_operation_copy` where the operation copy is disabled in `DistributedStorage` when it's set to `false`. And we can set it to `false` for the internally used storage instances when we use Multi Storage and Consensus Commit.

Actually, it's a bigger PR than I thought... So I will add inline comments to make it more readable. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/Delete.java
+++ b/core/src/main/java/com/scalar/db/api/Delete.java
@@ -32,6 +32,10 @@ public class Delete extends Mutation {
     super(partitionKey, clusteringKey);
   }
 
+  public Delete(Delete delete) {
+    super(delete);
+  }
+
   @Override
   public Delete forNamespace(String namespace) {
     return (Delete) super.forNamespace(namespace);
--- a/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
+++ b/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
@@ -1,6 +1,7 @@
 package com.scalar.db.config;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.scalar.db.config.ConfigUtils.getBoolean;
 import static com.scalar.db.config.ConfigUtils.getInt;
 import static com.scalar.db.config.ConfigUtils.getLong;
 import static com.scalar.db.config.ConfigUtils.getString;
@@ -52,6 +53,7 @@ public class DatabaseConfig {
   private Class<? extends DistributedTransactionManager> transactionManagerClass;
   private Class<? extends TwoPhaseCommitTransactionManager> twoPhaseCommitTransactionManagerClass;
   private long tableMetadataCacheExpirationTimeSecs;
+  private boolean needOperationCopy;
 
   public static final String PREFIX = "scalar.db.";
   public static final String CONTACT_POINTS = PREFIX + "contact_points";
@@ -62,6 +64,7 @@ public class DatabaseConfig {
   public static final String TRANSACTION_MANAGER = PREFIX + "transaction_manager";
   public static final String TABLE_METADATA_CACHE_EXPIRATION_TIME_SECS =
       PREFIX + "table_metadata.cache_expiration_time_secs";
+  public static final String NEED_OPERATION_COPY = PREFIX + "need_operation_copy";
 
   public DatabaseConfig(File propertiesFile) throws IOException {
     try (FileInputStream stream = new FileInputStream(propertiesFile)) {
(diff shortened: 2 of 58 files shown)


core/src/main/java/com/scalar/db/api/Delete.java
```diff
@@ -32,6 +32,10 @@ public Delete(Key partitionKey, Key clusteringKey) {
     super(partitionKey, clusteringKey);
   }
 
+  public Delete(Delete delete) {
```
brfrn169, line 35: I added copy constructors for all the operations.

core/src/main/java/com/scalar/db/config/DatabaseConfig.java
```diff
@@ -62,6 +64,7 @@
   public static final String TRANSACTION_MANAGER = PREFIX + "transaction_manager";
   public static final String TABLE_METADATA_CACHE_EXPIRATION_TIME_SECS =
       PREFIX + "table_metadata.cache_expiration_time_secs";
+  public static final String NEED_OPERATION_COPY = PREFIX + "need_operation_copy";
```
brfrn169, line 67: I added a new configuration `scalar.db.need_operation_copy` where the operation copy in `DistributedStorage` is disabled when it's set to `false`.
Honestly, I'm wondering if it's a good configuration name.

core/src/main/java/com/scalar/db/service/TransactionModule.java
```diff
@@ -14,12 +14,26 @@
...
-    this.config = config;
+    if (config.getTransactionManagerClass() == ConsensusCommitManager.class
+        || config.getTwoPhaseCommitTransactionManagerClass()
+            == TwoPhaseConsensusCommitManager.class) {
+      // since operations are copied in the Consensus Commit layer, we don't need to copy operations
+      // in the storage
+      Properties props = new Properties();
+      props.putAll(config.getProperties());
+      props.put(DatabaseConfig.NEED_OPERATION_COPY, "false");
+      this.config = new DatabaseConfig(props);
```
brfrn169, line 33: If we use Consensus Commit, we set `scalar.db.need_operation_copy` to `false` to avoid to copy operation instances twice.

core/src/main/java/com/scalar/db/storage/common/AbstractDistributedStorage.java
```diff
@@ -0,0 +1,71 @@
+package com.scalar.db.storage.common;
```
brfrn169, line 1: This PR introduces `AbstractDistributedStorage` that extracts the common logic of the `DistributedStorage` implementations.

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageConfig.java
```diff
@@ -81,14 +69,18 @@ private void loadDatabaseConfigs() {
...
           dbProps.put(
               propertyName.replace("multi_storage.storages." + storage + ".", ""),
-              props.getProperty(propertyName));
+              getProperties().getProperty(propertyName));
         }
       }
 
+      // since operations are copied in MultiStorage, we don't need to copy operations in the
+      // storage
+      dbProps.put(DatabaseConfig.NEED_OPERATION_COPY, "false");
```
brfrn169, line 82: If we use Multi Storage, we set `scalar.db.need_operation_copy` to `false` to avoid to copy operation instances twice.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransaction.java
```diff
@@ -0,0 +1,68 @@
+package com.scalar.db.transaction.common;
```
brfrn169, line 1: I also added `AbstractDistributedTransaction`, `AbstractDistributedTransactionManager`, `AbstractTwoPhaseCommitTransaction`, and `AbstractTwoPhaseCommitTransactionManager`.

brfrn169:
Since I would like to rethink it, I'm closing it. I will create another PR for the issue. Thanks.

PR #498, opened 2022-02-08
Should not change states of operation instances passed as arguments

brfrn169:
Currently, we change the states of operation instances (`Get`/`Scan`/`Put`/`Delete`) passed as arguments, which could be confusing and might be causing a bug. To address issue, we can copy the operation instances at the begging of the methods (`get()`, `scan()`, `put()`, `delete()`, and `mutate()`) of `DistributedStorage` and `DistributedTransaction` and `TwoPhaseCommitTransaction`, and use them.

The difference from #495 is the new configuration `scalar.db.need_operation_copy` is removed. This is because I think the operation copy cost is very small but introducing the new configuration makes the code more complicated, so it's not worth it.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/Delete.java
+++ b/core/src/main/java/com/scalar/db/api/Delete.java
@@ -32,6 +32,10 @@ public class Delete extends Mutation {
     super(partitionKey, clusteringKey);
   }
 
+  public Delete(Delete delete) {
+    super(delete);
+  }
+
   @Override
   public Delete forNamespace(String namespace) {
     return (Delete) super.forNamespace(namespace);
--- a/core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
@@ -17,6 +17,7 @@ import com.scalar.db.api.Scanner;
 import com.scalar.db.api.TableMetadata;
 import com.scalar.db.config.DatabaseConfig;
 import com.scalar.db.exception.storage.ExecutionException;
+import com.scalar.db.storage.common.AbstractDistributedStorage;
 import com.scalar.db.storage.common.checker.OperationChecker;
 import com.scalar.db.util.ScalarDbUtils;
 import com.scalar.db.util.TableMetadataManager;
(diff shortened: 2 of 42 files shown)


core/src/main/java/com/scalar/db/api/Delete.java
```diff
@@ -32,6 +32,10 @@ public Delete(Key partitionKey, Key clusteringKey) {
     super(partitionKey, clusteringKey);
   }
 
+  public Delete(Delete delete) {
```
brfrn169, line 35: I added copy constructors for all the operations.

core/src/main/java/com/scalar/db/storage/common/AbstractDistributedStorage.java
```diff
@@ -0,0 +1,68 @@
...
+import com.scalar.db.api.DistributedStorage;
+import com.scalar.db.api.Get;
+import com.scalar.db.api.Mutation;
+import com.scalar.db.api.Put;
+import com.scalar.db.api.Scan;
+import com.scalar.db.util.ScalarDbUtils;
+import java.util.List;
+import java.util.Optional;
+
+public abstract class AbstractDistributedStorage implements DistributedStorage {
```
brfrn169, line 13: This PR introduces AbstractDistributedStorage that extracts the common logic of the DistributedStorage implementations.

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
```diff
@@ -64,42 +63,13 @@ public Cassandra(DatabaseConfig config) {
...
-  public Optional<String> getTable() {
-    return tableName;
   }
 
   @Override
   @Nonnull
   public Optional<Result> get(Get get) throws ExecutionException {
     LOGGER.debug("executing get operation with " + get);
-    ScalarDbUtils.setTargetToIfNot(get, namespace, tableName);
+    get = copyAndSetTargetToIfNot(get);
```
brfrn169, line 72: We copy the operation at the begging of the methods (`get()`, `scan()`, `put()`, `delete()`, `mutation()`).

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransaction.java
```diff
@@ -0,0 +1,68 @@
+package com.scalar.db.transaction.common;
```
brfrn169, line 1: I also added `AbstractDistributedTransaction`, `AbstractDistributedTransactionManager`, `AbstractTwoPhaseCommitTransaction`, and `AbstractTwoPhaseCommitTransactionManager` similar to `AbstractDistributedStorage`.

thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #500, opened 2022-02-09
Move the rollbackRecords method from RecoveryHandler to CommitHandler

brfrn169:
This is a PR for refactoring. I think it's better to move the `RecoveryHandler.rollbackRecords()` method to `CommitHandler` because the method is only used in `CommitHandler`. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #501, opened 2022-02-09
Add ConfigUtils.getStringArray()

ymorimo:
This PR adds `ConfigUtils.getStringArray()` utility method that is for getting a string array property from a comma-separated string in a properties file.

It is applied to the existing `DatabaseConfig` and `MultiStorageConfig`. Note that there is a slight difference in how a trailing _regex_ in a property is treated between `split(regex, -1)` and `split(regex)` (equivalent to `split(regex, 0)`). In the latter case, trailing empty strings will be discarded (e.g., `scalar.db.contact_points = server1, , ,` is parsed into a one-element array). This should not be a problem.

### DIFF
--- a/core/src/main/java/com/scalar/db/config/ConfigUtils.java
+++ b/core/src/main/java/com/scalar/db/config/ConfigUtils.java
@@ -51,4 +51,12 @@ public final class ConfigUtils {
           "the specified value of '" + name + "' is not a boolean value");
     }
   }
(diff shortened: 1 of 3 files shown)


brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #447, opened 2021-12-28
Revisit Key construction

brfrn169:
This PR revisits the `Key` construction. After this change, we can follow the following guideline to construct `Key` objects:

1. If the key consists of a single value (column), use the constructors for a single value (column) as follows:
```
Key key1 = new Key("col1", 1);
Key key2 = new Key("col1", 100L);
Key key3 = new Key("col1", 1.3d);
Key key4 = new Key("col1", "value");
```
2. If the key consists of 2 - 5 values (columns), use the constructors for multiple values (columns) as follows:
```
Key key1 = new Key("col1", 1, "col2", 100L);
Key key2 = new Key("col1", 1, "col2", 100L, "col3", 1.3d);
Key key3 = new Key("col1", 1, "col2", 100L, "col3", 1.3d, "col4", "value");
Key key4 = new Key("col1", 1, "col2", 100L, "col3", 1.3d, "col4", "value", "col5", false);
```
3. If the key consists of more than 5 values (columns), use the builder as follows:
```
Key key = Key.newBuilder()
            .add("col1", 1)
            .add("col2", 100L)
            .add("col3", 1.3d)
            .add("col4", "value")
            .add("col5", false)
            .add("col6", 100)
            .build();
```

Please take a look!

PR #504, opened 2022-02-09
Change Key builder

brfrn169:
This PR actually reverts some of the changes in #447.

It seems like the current standard of Java API is more explicit method names rather than using overload.

After this reverting, We can use the Key builder as follows:
```java
Key key = Key.newBuilder()
            .addInt("col1", 1)
            .addBigInt("col2", 100L)
            .addDouble("col3", 1.3d)
            .addText("col4", "value")
            .addBoolean("col5", false)
            .addInt("col6", 100)
            .build();
```

Note that we don't need to consider its compatibility since we have not yet released the changes in #447.


thongdk8 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #505, opened 2022-02-09
GraphQL: Handling variable references in @transaction directive

ymorimo:
Previously, GraphQL queries containing variables in the `@transaction` directive, such as the following, were not supported even though they were syntactically correct. This PR addresses this issue.

```graphql
query balances($tx_id: String, $commit: Boolean) @transaction(txId: $tx_id, commit: $commit) {
   tx_transfer_get(get: { ... }) { ... }
}
```

This is important for clients to be able to reuse the "`query`" part of the request document (JSON) without any string formatting.

### DIFF
--- a/graphql/src/main/java/com/scalar/db/graphql/instrumentation/TransactionInstrumentation.java
+++ b/graphql/src/main/java/com/scalar/db/graphql/instrumentation/TransactionInstrumentation.java
@@ -26,6 +26,8 @@ import graphql.language.Argument;
 import graphql.language.BooleanValue;
 import graphql.language.Directive;
 import graphql.language.StringValue;
+import graphql.language.Value;
+import graphql.language.VariableReference;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


graphql/src/main/java/com/scalar/db/graphql/instrumentation/TransactionInstrumentation.java
```diff
@@ -78,7 +80,21 @@ public TransactionInstrumentation(DistributedTransactionManager transactionManag
...
     if (txIdArg == null) {
+      txId = null;
+    } else {
+      Value<?> value = txIdArg.getValue();
+      if (value instanceof StringValue) {
+        txId = ((StringValue) value).getValue();
+      } else if (value instanceof VariableReference) {
+        txId = (String) executionContext.getVariables().get(((VariableReference) value).getName());
+      } else {
+        throw new IllegalStateException("Unexpected value of txId: " + value);
```
ymorimo, line 94: Normally, execution will never reach this point since the document is validated for the schema in advance and the directive argument value can be safely cast to `StringValue` or `VariableReference` of a string.

feeblefakie (approved):
LGTM! Thank you!

PR #497, opened 2022-02-08
Add ByteBuffer support for Blob type

brfrn169:
This PR adds ByteBuffer support for Blob type. Please take a look!

PR #508, opened 2022-02-10
Add ByteBuffer support to Value

brfrn169:
In addition to #497, this PR adds `ByteBuffer` support to `Value`. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/io/Value.java
+++ b/core/src/main/java/com/scalar/db/io/Value.java
@@ -1,5 +1,6 @@
 package com.scalar.db.io;
 
+import java.nio.ByteBuffer;
 import java.util.Optional;
 import javax.annotation.Nonnull;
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/scalar/db/io/Value.java
```diff
@@ -102,4 +103,13 @@ default double getAsDouble() {
   default Optional<byte[]> getAsBytes() {
     throw new UnsupportedOperationException();
   }
+
+  /**
+   * Returns the content of this {@code Value} as a byte buffer
+   *
+   * @return the content of this {@code Value}
+   */
+  default Optional<ByteBuffer> getAsByteBuffer() {
```
brfrn169, line 112: Added the `getAsByteBuffer()` method to the `Value` class.

feeblefakie (approved):
LGTM! Thank you!

thongdk8 (approved):
LGTM! Thank you!

PR #512, opened 2022-02-17
Allow null values

brfrn169:
This PR allows null values. Since it's a very big change, I will add inline comments to make it more readable. Please take a look!


core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -26,7 +27,7 @@
  */
 @NotThreadSafe
 public class Put extends Mutation {
-  private final Map<String, Value<?>> values;
+  private final Map<String, Optional<Value<?>>> values;
```
brfrn169, line 30: To express NULL values, I changed the type of this map value to `Optional<Value<?>>`. `Optional.empty()` means a NULL value.

core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -168,7 +181,18 @@ public Put withValue(String name, @Nullable ByteBuffer value) {
...
+    return this;
+  }
+
+  /**
+   * Adds NULL value for the specified name of the value.
+   *
+   * @param name a name of the value
+   * @return this object
+   */
+  public Put withNullValue(String name) {
```
brfrn169, line 194: We use this `withNullValue()` method to set a NULL value.

core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -177,7 +201,7 @@ public Put withValues(Collection<Value<?>> values) {
    *
    * @return a map of {@code Value}s
    */
-  public Map<String, Value<?>> getValues() {
+  public Map<String, Optional<Value<?>>> getValues() {
```
brfrn169, line 204: This change is actually an incompatible change but we can assume this method is only used internally, so I think it's okay.

core/src/main/java/com/scalar/db/api/Result.java
```diff
@@ -29,6 +32,8 @@
   /**
    * Returns the {@link Value} which the specified name is mapped to
    *
+   * <p>Note that if the value is NULL, it will return a {@code Value} object with a default value.
```
brfrn169, line 35: This is for backward compatibility.

core/src/main/java/com/scalar/db/api/Result.java
```diff
@@ -37,7 +42,150 @@
   /**
    * Returns a map of {@link Value}s
    *
+   * <p>Note that if the value in the map is NULL, the value will be a {@code Value} object with a
+   * default value.
```
brfrn169, line 46: Ditto.

core/src/main/java/com/scalar/db/api/Result.java
```diff
@@ -37,7 +42,150 @@
...
    */
   Map<String, Value<?>> getValues();
+
+  /**
+   * Indicates whether the value for the specified name is NULL.
+   *
+   * @param name a name
+   * @return whether the value for the specified name is NULL
+   */
+  boolean isNull(String name);
```
brfrn169, line 58: We can check if the column is NULL or not with this method `isNull()`.

core/src/main/java/com/scalar/db/api/Result.java
```diff
@@ -37,7 +42,150 @@
...
+   * {@code LONG} object. If the columns is a FLOAT type, it returns a {@code FLOAT} object. If the
+   * columns is a DOUBLE type, it returns a {@code DOUBLE} object. If the columns is a TEXT type, it
+   * returns a {@code String} object. If the columns is a BLOB type, it returns a {@code ByteBuffer}
+   * object.
+   *
+   * @param name a name
+   * @return the value for the specified name as a Java Object type. If the value is NULL, null
+   */
+  @Nullable
+  Object getObject(String name);
```
brfrn169, line 175: Also, I added several method to get a value from the result. After this change, users no longer need to treat the `Value` class directly. So we can completely hide this class from users. I'm going to do that in another PR.

core/src/main/java/com/scalar/db/util/ProtoUtils.java
```diff
@@ -95,41 +99,52 @@ private static Key toKey(com.scalar.db.rpc.Key key) {
...
+          return null;
         }
       case BLOB_VALUE:
         if (value.getBlobValue().hasValue()) {
           return new BlobValue(value.getName(), value.getBlobValue().getValue().toByteArray());
         } else {
-          return new BlobValue(value.getName(), (byte[]) null);
+          return null;
         }
+      case VALUE_NOT_SET:
```
brfrn169, line 110: For the protocol buffer, if the value is not set (`VALUE_NOT_SET`), we tread it as a NULL value.

core/src/main/resources/cosmosdb_stored_procedure/mutate.js
```diff
@@ -15,7 +15,9 @@ function mutate(num, ...args) {
```
brfrn169, line 15: One pain point of this change is that we need to change the stored procedure to treat NULL values in Cosmos DB. But if we don't use NULL values, we don't need to update the stored procedure. So, we preserve the backward compatibility, and we can upgrade to a version including this change without any changes. But if we want to use NULL values after upgrading, we need to update the stored procedure.

core/src/main/java/com/scalar/db/api/Put.java
feeblefakie: In Scalar DL, `getValues` is not used in `src` code but is used a lot in `test` code, so we might need to think about introducing `getNullableValues()` or something for backward incompatibility?
Let's discuss it later.

core/src/main/java/com/scalar/db/storage/cosmos/Record.java
```diff
@@ -58,7 +58,7 @@ public String getConcatenatedPartitionKey() {
   }
 
   public Map<String, Object> getValues() {
-    return ImmutableMap.copyOf(values);
+    return Collections.unmodifiableMap(values);
```
feeblefakie, line 61: Just curious. Any reason for this change?

core/src/main/java/com/scalar/db/transaction/consensuscommit/MergedResult.java
```diff
@@ -0,0 +1,219 @@
+package com.scalar.db.transaction.consensuscommit;
```
feeblefakie, line 1: Is this expected to be here? I thought this file is added before.

feeblefakie (commented):
Thank you for the great refactoring! Overall looking good.
Left a comment and questions.

core/src/main/java/com/scalar/db/storage/cosmos/Record.java
brfrn169: Oh sorry, I forgot mentioned it. We need this because the guava `ImmutableMap` doesn't support null values. Putting null values to this `values` map indicates overwriting the existing values with null values.

core/src/main/java/com/scalar/db/transaction/consensuscommit/MergedResult.java
brfrn169: This class was originally an inter class in the `Snapshot` class. Because the change in this PR makes this class bigger, I separated it from `Snapshot`.

core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -168,19 +184,303 @@ public Put withValue(String name, @Nullable ByteBuffer value) {
...
    *
    * @return a map of {@code Value}s
+   * @deprecated As of release 3.6.0. Will be removed in release 5.0.0
    */
+  @Deprecated
   public Map<String, Value<?>> getValues() {
+    Map<String, Value<?>> ret = new HashMap<>();
+    values.forEach((k, v) -> ret.put(k, v.orElse(null)));
+    return ret;
+  }
```
brfrn169, line 213: I kept this method `getValues()` for backward compatibility.

core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -168,19 +184,303 @@ public Put withValue(String name, @Nullable ByteBuffer value) {
...
+
+  /**
+   * Returns a set of the contained column names for the values added to the list of put values.
+   *
+   * @return a set of the contained column names
+   */
+  public Set<String> getContainedColumnNames() {
+    return ImmutableSet.copyOf(values.keySet());
+  }
```
brfrn169, line 476: I also added these methods to get a value from `Put`. This methods assume to be used for tests.

feeblefakie (approved):
LGTM! Thank you!
 
I was able to run Scalar DL E2E test with this updated version of Scalar DB, but as I shared, I faced a minor backward incompatibility in the emulator code that implements `Result` interface, which I think is acceptable.

PR #513, opened 2022-02-18
Refactor ImmutableLinkedHashSet

brfrn169:
This PR changes `ImmutableLinkedHashSet` to use `ImmutableSet` internally and delegate to it. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/util/ImmutableLinkedHashSet.java
+++ b/core/src/main/java/com/scalar/db/util/ImmutableLinkedHashSet.java
@@ -1,72 +1,140 @@
 package com.scalar.db.util;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import com.google.common.collect.ImmutableSet;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.Spliterator;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 import javax.annotation.concurrent.Immutable;
 
-@SuppressFBWarnings("UR_UNINIT_READ_CALLED_FROM_SUPER_CONSTRUCTOR")
 @Immutable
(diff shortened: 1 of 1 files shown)


thongdk8 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/util/ImmutableLinkedHashSet.java
```diff
@@ -1,72 +1,140 @@
...
 public class ImmutableLinkedHashSet<E> extends LinkedHashSet<E> {
 
-  private final boolean immutable;
+  private final ImmutableSet<E> data;
 
-  public ImmutableLinkedHashSet() {
-    immutable = true;
+  public ImmutableLinkedHashSet(Collection<? extends E> data) {
+    super(0);
+    this.data = ImmutableSet.copyOf(data);
```
feeblefakie, line 22: Just wondering. Is it guaranteed to keep the insertion order even if it's copied (not actually copied behind the scene, though) to ImmutableSet?

feeblefakie (approved):
LGTM! Thank you!
But, I left one question.

core/src/main/java/com/scalar/db/util/ImmutableLinkedHashSet.java
brfrn169: Yes, `ImmutableSet` guarantees to keep the insertion order. Actually, I didn't know that before working on this change, but it does according to the following document:
https://guava.dev/releases/30.1-jre/api/docs/com/google/common/collect/ImmutableCollection.html

> Deterministic iteration. The iteration order is always well-defined, depending on how the collection was created. Typically this is insertion order unless an explicit ordering is otherwise specified (e.g. [ImmutableSortedSet.naturalOrder()](https://guava.dev/releases/30.1-jre/api/docs/com/google/common/collect/ImmutableSortedSet.html#naturalOrder())).

feeblefakie: Oh, cool. Thank you!

PR #516, opened 2022-03-02
Add more explicit methods to Put

brfrn169:
This PR adds more explicit methods to add values like `withBooleanValue()` and `withIntValue()` to the `Put` class and deprecates the existing methods `withValue()`. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #517, opened 2022-03-02
Add factory methods to Key

brfrn169:
This PR adds factory methods to the `Key` class.

We can use the factory methods as follows:
```java
// for a key that consists of a single column of Int
Key key1 = Key.ofInt("col1", 1);

// for a key that consists of a single column of BigInt
Key key2 = Key.ofBigInt("col1", 100L);

// for a key that consists of a single column of Double
Key key3 = Key.ofDouble("col1", 1.3d);

// for a key that consists of a single column of Text
Key key4 = Key.ofText("col1", "value");

// for a key that consists of 2 - 5 columns
Key key5 = Key.of("col1", 1, "col2", 100L);
Key key6 = Key.of("col1", 1, "col2", 100L, "col3", 1.3d);
Key key7 = Key.of("col1", 1, "col2", 100L, "col3", 1.3d, "col4", "value");
Key key8 = Key.of("col1", 1, "col2", 100L, "col3", 1.3d, "col4", "value", "col5", false);
```

We can still use the Key Builder class when we create a key that consists of more than 5 columns.

This PR also deprecates some existing constructors. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/io/Key.java
+++ b/core/src/main/java/com/scalar/db/io/Key.java
@@ -3,7 +3,6 @@ package com.scalar.db.io;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.MoreObjects;
-import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.Ordering;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


feeblefakie (approved):
It's nice! Thank you!

PR #520, opened 2022-03-02
Add factory methods to Scan.Ordering

brfrn169:
This PR adds factory methods to the `Scan.Ordering` class.

We can use them as follows:
```
// for asc ordering
Scan.Ordering ascOrdering = Scan.Ordering.asc("col1");

// for desc ordering
Scan.Ordering descOrdering = Scan.Ordering.desc("col2");
```

This PR also deprecates the constructor of `Scan.Ordering`. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #523, opened 2022-03-09
Add a method to get DataType of the column to Value

brfrn169:
This is a trivial improvement. This PR adds a method (`getDataType()`) to get DataType of the column to the `Value` class. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #526, opened 2022-03-11
Add creating and dropping index support to DistributedStorageAdmin

brfrn169:
This PR adds `createIndex()` and `dropIndex()` methods to `DistributedStorageAdmin`, and adds its implementations for all the adaptors. Please take a look!


core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -320,6 +325,53 @@ public void truncateTable(String namespace, String table) throws ExecutionExcept
...
+          database.createContainerIfNotExists(containerName, PARTITION_KEY_PATH);
+      CosmosContainerProperties properties = response.getProperties();
+
+      // set the new index policy to the container properties
+      properties.setIndexingPolicy(computeIndexingPolicy(newTableMetadata));
+
+      // update the container properties
+      database.getContainer(containerName).replace(properties);
+    } catch (RuntimeException e) {
+      throw new ExecutionException("dropping the secondary index failed", e);
```
feeblefakie, line 371: Maybe, not necessarily for dropping?

feeblefakie (dismissed):
LGTM! Thank you!
(but left one minor comment)

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
brfrn169: Ah, it looks like the error message is wrong. I will fix it. Thank you!

PR #515, opened 2022-02-28
Hide `Value` from users

brfrn169:
This PR hides the `Value` class from users. After this change, users basically no longer need to use `Value` directly.

What exactly this PR does is as follows:

- Change the terms, from `value` to `column`, and from `value content` to `column value`
- Make some methods to access the `Value` class deprecated or internal use
- Rename the argument name `name` to `columnName` to make it more explicit
- Rename the argument name `value` to `xxxValue` (like `booleanValue` and `intValue`) to make it more explicit
- Add several methods to get values in `ConditionalExpression` and `Key` so that users can access the values without the `Value` class

I will also add inline comments later. Please take a look!

PR #527, opened 2022-03-15
Introduce Column

brfrn169:
This PR introduce the `Column` class that replaces the `Value` class and supports `NULL` values natively. Since it's a huge change, I will add inline comments to make it more readable. Note that this PR doesn't remove all the `Value` class usages. I will do that in another PR. Please take a look!


core/src/integration-test/java/com/scalar/db/storage/StorageColumnValueIntegrationTestBase.java
```diff
@@ -453,13 +453,13 @@ public void put_WithNullValues_ShouldPutCorrectly() throws ExecutionException {
...
-            .withNullValue(COL_NAME5)
-            .withNullValue(COL_NAME6)
-            .withNullValue(COL_NAME7)
+            .withBooleanValue(COL_NAME1, null)
+            .withIntValue(COL_NAME2, null)
+            .withBigIntValue(COL_NAME3, null)
+            .withFloatValue(COL_NAME4, null)
+            .withDoubleValue(COL_NAME5, null)
+            .withTextValue(COL_NAME6, null)
+            .withBlobValue(COL_NAME7, (ByteBuffer) null)
```
brfrn169, line 462: I changed how to specify NULL values in this PR. After this change, we can specify NULL values as above.

core/src/main/java/com/scalar/db/api/ConditionalExpression.java
```diff
@@ -19,8 +21,7 @@
  */
 @Immutable
 public class ConditionalExpression {
-  private final String columnName;
-  private final Value<?> value;
+  private final Column<?> column;
```
brfrn169, line 24: I replaced these `String  columnName` and `Value<?> value` to one field `Column<?> column` here. Maybe we need to discuss it.

core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -30,7 +31,7 @@
  */
 @NotThreadSafe
 public class Put extends Mutation {
-  private final Map<String, Optional<Value<?>>> values;
+  private final Map<String, Column<?>> columns;
```
brfrn169, line 34: I changed `Map<String, Optional<Value<?>>> values` to `Map<String, Column<?>> columns` here.

core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -302,18 +371,21 @@ public Put withBlobValue(String columnName, ByteBuffer value) {
...
+   * Adds a column to the list of put values.
    *
-   * @param columnName a column name of the value
+   * <p>This method is primarily for internal use. Breaking changes can and will be introduced to
+   * this method. Users should not depend on it.
+   *
+   * @param column a column to put
    * @return this object
    */
-  public Put withNullValue(String columnName) {
```
brfrn169, line 315: As mentioned, how to specify NULL values is changed in this PR, so I removed this `withNullValue()`.

core/src/main/java/com/scalar/db/io/Column.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * An abstraction for a column.
+ *
+ * <p>This class and its implementation classes of it are primarily for internal use. Breaking
+ * changes can and will be introduced to them. Users should not depend on them.
+ */
+public interface Column<T> extends Comparable<Column<T>> {
```
brfrn169, line 13: I added `Column<T>` and its implementation classes `BooleanColumn`, `IntColumn`, `BigIntColumn`, `FloatColumn`, `DoubleColumn`, `TextColumn`, and `BlobColumn`.

core/src/main/java/com/scalar/db/io/Key.java
```diff
@@ -22,7 +25,7 @@
  */
 @Immutable
 public final class Key implements Comparable<Key>, Iterable<Value<?>> {
-  private final List<Value<?>> values;
+  private final List<Column<?>> columns;
```
brfrn169, line 28: Changed `List<Value<?>>` to `List<Column<?>>` here.

core/src/main/java/com/scalar/db/io/Key.java
```diff
@@ -300,8 +331,8 @@ public String getColumnName(int i) {
    * @param i the position of the column which this key is composed of
    * @return the BOOLEAN value of the i-th column which this key is composed of
    */
-  public boolean getBoolean(int i) {
-    return values.get(i).getAsBoolean();
+  public boolean getBooleanValue(int i) {
```
feeblefakie, line 334: Is this a backward-incompatible change?

core/src/main/java/com/scalar/db/storage/cassandra/ValueBinder.java
```diff
@@ -20,7 +20,7 @@
  * @author Hiroyuki Yamada
  */
 @NotThreadSafe
-public final class ValueBinder implements ValueVisitor {
+public final class ValueBinder implements ColumnVisitor {
```
feeblefakie, line 23: Just a question or topic for discussion. Should the class name be changed to `ColumnBinder`? Or, does it introduce too many changes?

feeblefakie (commented):
Left some questions. PTAL!

core/src/main/java/com/scalar/db/io/Key.java
brfrn169: Ah sorry, I should have mentioned it. It's not a backward-incompatible change because these methods are added in #515 and they are not released yet.

core/src/main/java/com/scalar/db/storage/cassandra/ValueBinder.java
brfrn169: I kept this class name `ValueVisitor` because I thought it was still a proper name since this class binds *value* to the statement. What do you think?

> Or, does it introduce too many changes?

I don't think it introduces too many changes.

brfrn169 (commented):
@feeblefakie Thank you for taking a look at this PR! I just left replies to your questions. Please take a look when you have time!

core/src/main/java/com/scalar/db/io/Key.java
feeblefakie: Ah, I got it. Thank you!

core/src/main/java/com/scalar/db/storage/cassandra/ValueBinder.java
feeblefakie: Right. Now, I agree!

feeblefakie (approved):
LGTM! Thank you!

PR #530, opened 2022-03-22
Add a integration test putting a null value for secondary index column

brfrn169:
This PR adds a integration test putting a null value for secondary index column. It looks like DynamoDB doesn't support putting a null value for a secondary index column, so I ignored the test for DynamoDB. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #533, opened 2022-03-25
Extract an interface for transaction CRUD operations

brfrn169:
This PR extracts an interface for transaction CRUD operations from `DistributedTransaction` and `TwoPhaseCommitTransaction`. I would like to discuss if the interface name `TransactionCrudOperable` is good. Please take a look!


feeblefakie (approved):
LGTM! Thank you!
(I'll also think about the name and let's discuss it later)

PR #534, opened 2022-03-26
Handle lastEvaluatedKey for query in DynamoDB

brfrn169:
According to the following document, we need to handle `LastEvaluatedKey` when the query result is larger than 1MB:
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html

But the current implementation of our DynamoDB adapter doesn't handle it and it returns wrong results when the query result is larger than 1MB. This PR fixes this bug. Please take a look!


core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -52,31 +53,30 @@
...
+    this.metadataManager = checkNotNull(metadataManager);
   }
 
   @Nonnull
-  @Override
-  public List<Map<String, AttributeValue>> handle(Operation operation) throws ExecutionException {
-    checkArgument(operation, Get.class, Scan.class);
-
-    TableMetadata tableMetadata = metadataManager.getTableMetadata(operation);
+  public Scanner handle(Selection selection) throws ExecutionException {
```
brfrn169, line 61: I changed this return type to `Scanner`.

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -92,16 +92,11 @@ public SelectStatementHandler(DynamoDbClient client, TableMetadataManager metada
...
     }
 
-    GetItemResponse getItemResponse = client.getItem(builder.build());
-    if (getItemResponse.hasItem()) {
-      return Collections.singletonList(getItemResponse.item());
-    } else {
-      return Collections.emptyList();
-    }
+    return new GetItemScanner(
+        client, builder.build(), new ResultInterpreter(get.getProjections(), tableMetadata));
```
brfrn169, line 96: When executing a `Get` operation, it returns `GetItemScanner`.

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -130,17 +125,17 @@ public SelectStatementHandler(DynamoDbClient client, TableMetadataManager metada
       }
     }
 
-    QueryResponse queryResponse = client.query(builder.build());
-    return new ArrayList<>(queryResponse.items());
+    return new QueryScanner(
+        client, builder.build(), new ResultInterpreter(selection.getProjections(), tableMetadata));
```
brfrn169, line 129: When executing an index scan, it returns `QueryScanner`.

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -163,8 +158,8 @@ public SelectStatementHandler(DynamoDbClient client, TableMetadataManager metada
       builder.consistentRead(true);
     }
 
-    QueryResponse queryResponse = client.query(builder.build());
-    return new ArrayList<>(queryResponse.items());
+    return new QueryScanner(
+        client, builder.build(), new ResultInterpreter(scan.getProjections(), tableMetadata));
```
brfrn169, line 162: When executing a `Scan` operation, it returns `QueryScanner`.

core/src/main/java/com/scalar/db/storage/dynamo/StatementHandler.java
```diff
@@ -1,60 +0,0 @@
-package com.scalar.db.storage.dynamo;
```
brfrn169, line 1: I deleted `StatementHandler` in this PR because I needed to change the return type of `handle()`.

core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Map;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import software.amazon.awssdk.services.dynamodb.model.QueryRequest;
+import software.amazon.awssdk.services.dynamodb.model.QueryResponse;
+
+public class QueryScanner implements Scanner {
```
brfrn169, line 18: I think this `QueryScanner` is the main change of this PR. This class handles the `lastEvaluatedKey`.

core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
```diff
@@ -0,0 +1,100 @@
...
+    return Optional.of(resultInterpreter.interpret(itemsIterator.next()));
+  }
+
+  private boolean hasNext() {
+    if (itemsIterator.hasNext()) {
+      return true;
+    }
+    if (lastEvaluatedKey != null) {
+      QueryRequest.Builder builder = request.toBuilder();
+      builder.exclusiveStartKey(lastEvaluatedKey);
```
brfrn169, line 55: If the query response has `lastEvaluatedKey`, we need to re-query with `exclusiveStartKey()`.

core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+    return false;
+  }
+
+  private void query(QueryRequest request) {
+    QueryResponse response = client.query(request);
+    List<Map<String, AttributeValue>> items = response.items();
+    totalResultCount += items.size();
+    itemsIterator = items.iterator();
+    if ((request.limit() == null || totalResultCount < request.limit())
```
brfrn169, line 67: We need special handling for `limit`. The dynamo query API also returns `lastEvaluatedKey` even when `limit` is specified and the total query results (without `limit`) have more than the `limit` count. So we need to record the total result count that we have retrieved at that time (`totalResultCount`), and if we have already retrieved the `limit` count of the results, we need to stop querying.

feeblefakie (approved):
LGTM! Thank you!

PR #535, opened 2022-03-27
Should return primary key when key is contained in WriteSet and ReadSet but the result in ReadSet is empty in Consensus Commit

brfrn169:
Currently, when a key is contained in WriteSet and ReadSet but the result in ReadSet is empty, it doesn't return primary key, which is not expected behavior. This PR fixes this issue. 

Note that this bug is only in the branch `3.5`, but the `master` branch doesn't have it. So this bug fix is only for the branch `3.5`.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -2,6 +2,7 @@ package com.scalar.db.transaction.consensuscommit;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Ordering;
 import com.scalar.db.api.Consistency;
 import com.scalar.db.api.Delete;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -468,12 +469,21 @@ public int compareTo(Key o) {
...
     public MergedResult(Optional<TransactionResult> result, Put put) {
-      this.result = result;
       this.put = put;
+
+      Map<String, Value<?>> tmp = new HashMap<>();
+      if (result.isPresent()) {
+        tmp.putAll(result.get().getValues());
+      } else {
+        put.getPartitionKey().get().forEach(v -> tmp.put(v.getName(), v));
+        put.getClusteringKey().ifPresent(c -> c.get().forEach(v -> tmp.put(v.getName(), v)));
```
brfrn169, line 483: This logic includes the primary key into `values` when the result in the read set is empty.

feeblefakie (approved):
LGTM! Thank you!

PR #539, opened 2022-03-31
Rename metadata cache expiration time configuration name

brfrn169:
This PR renames the metadata cache expiration time configuration name `scalar.db.table_metadata.cache_expiration_time_secs` to `scalar.db.metadata.cache_expiration_time_secs` to make it more general name. Since this configuration is only in the source code and it's not exposed in any documentation, I think it's no problem to change it. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #541, opened 2022-04-04
Migrate CI to GitHub actions

Torch3333:
This migrates the project CI from Circle-ci to GitHub actions. 
Please have a look!


Torch3333:
@feeblefakie @brfrn169 
Running multiple runners concurrently to execute the Cosmos DB integration test is not working as expected so I will need to investigate further. 
I will put back the PR status to Draft.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,472 @@
...
+          distribution: 'temurin'
+
+      - name: Generate unique prefix using the epoch
+        run: |
+          echo "db_prefix=$(date +%s%3N)" >> $GITHUB_ENV
+
+      - name: Setup and execute Gradle 'integrationTestCosmos' task
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: integrationTestCosmos -Dscalardb.cosmos.uri=${{ secrets.COSMOS_URI }} -Dscalardb.cosmos.password=${{ secrets.COSMOS_PASSWORD }} -Dscalardb.cosmos.database_prefix=${{ env.db_prefix }}
```
brfrn169, line 119: Please add underscore (`_`) to the end of `scalardb.cosmos.database_prefix`.

brfrn169 (commented):
Thank you! I left a comment. Other than that, LGTM!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,472 @@
...
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: gradle_cassandra_integration_test_reports
+          path: /tmp/gradle_integration_test_reports
+
+  integration-test-for-cosmos:
+    runs-on: self-hosted
```
Torch3333, line 101: We decided to limit the number of self-hosted runners to two. With 3 runners running the test in parallel, there are a lot of cosmos DB request metadata failure making the whole test fail.

.github/workflows/ci.yaml
Torch3333: Thank you for reviewing. Sure, let me add it.

brfrn169 (dismissed):
LGTM! Thank you!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,472 @@
+name: CI
+
+on: [ push ]
```
feeblefakie, line 3: Isn't it too much to run on `push`? (should it be `pull_request`?)

feeblefakie (commented):
Looking good! Thank you!
Left one comment on the trigger timing.
Also, one of the CIs is failing, but is it OK?

.github/workflows/ci.yaml
Torch3333: Yes, I could limit it to `pull_request` event. Let me confirm with Toshi.

@brfrn169 The current circle-ci config triggers on each `push` event. Do you see any problem limiting it to the `pull_request` event?

Torch3333:
> Looking good! Thank you! Left one comment on the trigger timing. Also, one of the CIs is failing, but is it OK?

The Ci is failing because it still requires a passing Circle-ci CI, but the related config file has been deleted. 
Once the PR is ready to merge, we will need to update the repository configuration to remove Circle-ci-related check.

.github/workflows/ci.yaml
feeblefakie: My only concern is cost, so if the cost doesn't increase linearly for the number of pushes, then it should be fine. (Github Actions cost is supposed to be free for public repositories, but I'm not fully sure about the other costs for the CIs)

So, I'm fine with `push` if it doesn't make much difference in cost and provides more development productivity.
If it costs too much, the alternative would be `pull_request` and `workflow_dispatch` so that we can still run the CIs when needed.

feeblefakie (dismissed):

.github/workflows/ci.yaml
brfrn169: In most of my cases, the timing of `push` is the timing where I want to run the CI. However, I'm not sure the same is true for others. So cost-wise, it's safe that the triggers of the CI are `pull_request` and `workflow_dispatch`.

@Torch3333 Can you please change the the triggers of the CI to `pull_request` and `workflow_dispatch`?

Torch3333: Sure

feeblefakie (approved):
LGTM!

brfrn169 (approved):
LGTM! Thank you!

PR #542, opened 2022-04-04
Support alias in SQL API

brfrn169:
This PR supports alias in SQL API.

After this change, we can specify projections with aliases (with the `as()` method) as follows:
```java
SelectStatement statement =
  StatementBuilder.select(Projection.column("col2").as("a"), Projection.column("col3").as("b"))
      .from("ns", "tbl")
      .where(Predicate.column("col1").isEqualTo(Value.ofInt(10)))
      .and(Predicate.column("col2").isGreaterThan(Value.ofText("aaa")))
      .and(Predicate.column("col2").isLessThan(Value.ofText("ccc")))
      .orderBy(
          ClusteringOrdering.column("col2").desc(), ClusteringOrdering.column("col3").desc())
      .limit(10)
      .build();
```

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/sql/DmlStatementExecutor.java
+++ b/core/src/main/java/com/scalar/db/sql/DmlStatementExecutor.java
@@ -57,10 +57,12 @@ public class DmlStatementExecutor
 
     Selection selection = convertSelectStatementToSelection(statement, metadata);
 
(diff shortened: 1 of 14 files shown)


feeblefakie (approved):
Getting close to SQ! LGTM! Thank you.

PR #544, opened 2022-04-05
Fix `Query condition missed key schema element` error in DynamoDB

brfrn169:
`Query condition missed key schema element` error happens when executing Scan/Get operations using secondary index and specifying projections in DynamoDB. This PR fixes this bug. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
+++ b/core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
@@ -2,7 +2,6 @@ package com.scalar.db.storage.dynamo;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.collect.ImmutableMap;
 import com.google.common.primitives.UnsignedBytes;
 import com.scalar.db.api.Consistency;
 import com.scalar.db.api.Get;
@@ -85,7 +84,9 @@ public class SelectStatementHandler {
             .key(dynamoOperation.getKeyMap());
 
     if (!get.getProjections().isEmpty()) {
-      projectionExpression(builder, get);
+      Map<String, String> expressionAttributeNames = new HashMap<>();
+      projectionExpression(builder, get, expressionAttributeNames);
+      builder.expressionAttributeNames(expressionAttributeNames);
     }
 
     if (get.getConsistency() != Consistency.EVENTUAL) {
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/scalar/db/storage/dynamo/SelectStatementHandlerTest.java
```diff
@@ -225,6 +225,42 @@ public void handle_ScanOperationWithIndexGiven_ShouldCallQuery() {
     assertThat(actualRequest.expressionAttributeValues()).isEqualTo(expectedBindMap);
   }
 
+  @Test
+  public void handle_ScanOperationWithIndexGiven_WithProjections_ShouldCallQuery() {
```
brfrn169, line 229: This is a unit test to reproduce the bug.

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -109,15 +110,17 @@ private Scanner executeQueryWithIndex(Selection selection, TableMetadata tableMe
     ValueBinder binder = new ValueBinder(DynamoOperation.VALUE_ALIAS);
     keyColumn.accept(binder);
     Map<String, AttributeValue> bindMap = binder.build();
-    builder
-        .keyConditionExpression(condition)
-        .expressionAttributeValues(bindMap)
-        .expressionAttributeNames(ImmutableMap.of(expressionColumnName, column));
```
brfrn169, line 115: This bug is introduced in #462.

We set `expressionAttributeNames` here, but we also set it in `projectionExpression()` if the operation has projections, so the `expressionAttributeNames` set here is overwritten, which is a problem.
https://github.com/scalar-labs/scalardb/pull/544/files#diff-59d53393b90fb6d234264b0143965e0670ab3a4e80341de6a4e79b3bf1d3c29aL177-R188

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -109,15 +110,17 @@ private Scanner executeQueryWithIndex(Selection selection, TableMetadata tableMe
...
+
+    Map<String, String> expressionAttributeNames = new HashMap<>();
+    expressionAttributeNames.put(expressionColumnName, column);
 
     if (!selection.getProjections().isEmpty()) {
-      projectionExpression(builder, selection);
+      projectionExpression(builder, selection, expressionAttributeNames);
     }
 
+    builder.expressionAttributeNames(expressionAttributeNames);
```
brfrn169, line 122: In this fix, we first merge the `expressionAttributeNames` and set it to the builder.

feeblefakie (dismissed):
LGTM! Thank you!

ypeckstadt (approved):
LGTM. Thank you.

thongdk8 (approved):
LGTM! Thank you!

PR #548, opened 2022-04-12
Introduce ScanAll operation

brfrn169:
This PR introduces a `ScanAll` operation that scans all the data of a database. This PR doesn't contain the actual implementation of `ScanAll` but it contains only  the `ScanAll` operation class.

I added the `ScanAll` as a subclass of `Scan` because I think that minimizes the total change cost.

I assume the `ScanAll` operation is used with the existing `scan()` API as follows:
```java
Scanner scanner =
    storage.scan(
        new ScanAll().withProjection("col1").withLimit(10).forNamespace("ns").forTable("tbl"));
```
Currently `ScanAll` only supports specifying projections and limit.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/ScanAll.java
+++ b/core/src/main/java/com/scalar/db/api/ScanAll.java
@@ -1,0 +1,101 @@
+package com.scalar.db.api;
+
+import com.google.common.base.MoreObjects;
+import com.scalar.db.io.Key;
+import java.util.Collection;
+import java.util.Objects;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/**
+ * A command to retrieve all the entries of the database. The scan range of clustering key and
+ * {@link Ordering} cannot be specified for this command. The number of {@link Result} can be
+ * limited.
+ */
+@NotThreadSafe
+public class ScanAll extends Scan {
+
+  private static final Key DUMMY_PARTITION_KEY = Key.of();
+
+  public ScanAll() {
+    super(DUMMY_PARTITION_KEY);
+  }
+
+  public ScanAll(ScanAll scanAll) {
+    super(scanAll);
+  }
+
+  @Override
+  public ScanAll withStart(Key clusteringKey) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public ScanAll withStart(Key clusteringKey, boolean inclusive) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public Scan withEnd(Key clusteringKey) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public Scan withEnd(Key clusteringKey, boolean inclusive) {
+    throw new UnsupportedOperationException();
+  }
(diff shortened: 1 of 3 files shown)


feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/io/Key.java
```diff
@@ -586,6 +586,15 @@ public static Key ofBlob(String columnName, ByteBuffer value) {
     return new Key(columnName, value);
   }
 
+  /**
+   * Create an empty {@code Key} object
+   *
+   * @return an empty {@code Key} object
+   */
+  public static Key of() {
```
Torch3333, line 594: Would `Key.empty()` be appropriate ?

core/src/main/java/com/scalar/db/api/ScanAll.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public Scan withEnd(Key clusteringKey) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public Scan withEnd(Key clusteringKey, boolean inclusive) {
+    throw new UnsupportedOperationException();
+  }
```
Torch3333, line 45: If I understood correctly, you would like to implement these methods for ScanAll later on. 
Though intuitively, as a user, If I want only to retrieve a subset of all the database entries, wouldn't I want to filter on the entire primary key and not only on the clustering key only?

Torch3333 (commented):
Thank you. I left some comments.

core/src/main/java/com/scalar/db/api/ScanAll.java
brfrn169: In this `ScanAll` operation, we won't support a clustering key range scan. Instead, we can add other methods filtering on arbitrary columns in the future. At this point, we support only specifying projections and limit.

core/src/main/java/com/scalar/db/io/Key.java
brfrn169: In `ImmutableXXX` in Guava, we can create empty collections with `of()` without arguments. I followed this naming. What do you think?

brfrn169 (commented):
@Torch3333 Thank you for taking a look at this PR. I reply for your comments. Please take a look when you have time!

core/src/main/java/com/scalar/db/api/ScanAll.java
Torch3333: All right. 
I wouldn't say I like that we have to expose the non-implemented `withEnd` and `withStart` methods to the end-user, but if it's a necessary evil to reduce the development cost, then I understand.

core/src/main/java/com/scalar/db/io/Key.java
Torch3333: All right. If it's a well known usage, that's good.

Torch3333 (approved):
Thank you!
LGTM

PR #549, opened 2022-04-13
Invalidate metadata cache when schema change happens in SQL API

brfrn169:
This PR adds a logic to invalidate metadata cache when schema change happens in SQL API. I will inline comments to make it more readable. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/sql/DdlStatementExecutor.java
+++ b/core/src/main/java/com/scalar/db/sql/DdlStatementExecutor.java
@@ -4,6 +4,9 @@ import com.scalar.db.api.DistributedTransactionAdmin;
 import com.scalar.db.api.Scan;
 import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.sql.exception.SqlException;
+import com.scalar.db.sql.metadata.CachedMetadata;
+import com.scalar.db.sql.metadata.CachedNamespaceMetadata;
+import com.scalar.db.sql.metadata.Metadata;
 import com.scalar.db.sql.statement.CreateCoordinatorTableStatement;
 import com.scalar.db.sql.statement.CreateIndexStatement;
 import com.scalar.db.sql.statement.CreateNamespaceStatement;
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/scalar/db/sql/DdlStatementExecutor.java
```diff
@@ -23,9 +26,11 @@
 public class DdlStatementExecutor implements DdlStatementVisitor<Void, Void> {
```
brfrn169, line 26: The cache invalidations are kicked when a schema change happens in `DdlStatementExecutor`.

core/src/main/java/com/scalar/db/sql/metadata/CachedMetadata.java
```diff
@@ -0,0 +1,58 @@
...
+import com.scalar.db.api.DistributedTransactionAdmin;
+import com.scalar.db.exception.storage.ExecutionException;
+import com.scalar.db.sql.exception.SqlException;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.ThreadSafe;
+
+@ThreadSafe
+public class CachedMetadata implements Metadata {
```
brfrn169, line 15: I also extract metadata related interfaces in this PR.

The extracted interfaces are `Metadata`, `NamespaceMetadata`, `TableMetadata`, `ColumnMetadata`, and `IndexMetadata`.

core/src/main/java/com/scalar/db/sql/DdlStatementExecutor.java
```diff
@@ -125,6 +142,17 @@ public Void visit(DropNamespaceStatement statement, Void context) {
   public Void visit(DropTableStatement statement, Void context) {
     try {
       admin.dropTable(statement.namespaceName, statement.tableName, statement.ifExists);
+
+      // Invalidate the metadata cache
+      metadata
```
feeblefakie, line 147: Should it be done in the finally clause to care for the case where dropTable fails for some reason?
(it's a minor case, but if `dropTable` fails after dropping the specified table, the corresponding metadata will be left un-invalidated?)

feeblefakie (commented):
Looking good!
I left one question.

core/src/main/java/com/scalar/db/sql/DdlStatementExecutor.java
brfrn169: Thank you for taking a look at this! Yes, that's better. I will modify it in that way. Thanks.

Torch3333 (approved):
LGTM, thank you.

feeblefakie (approved):
LGTM! Thank you!

PR #550, opened 2022-04-13
Fix SQL syntax error that happens when scanning a table without clustering key in JDBC adapter

brfrn169:
This PR is a fix for a bug where an SQL syntax error happens when scanning a table without clustering key in JDBC adapter. We basically don't need to scan a table without clustering key, but an error should not happen if we do that. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
@@ -86,7 +86,9 @@ public class SimpleSelectQuery implements SelectQuery {
   }
 
   private String orderBySqlString() {
-    if (!isRangeQuery || indexedColumn.isPresent()) {
+    if (!isRangeQuery
+        || indexedColumn.isPresent()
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -86,7 +86,9 @@ private String conditionSqlString() {
   }
 
   private String orderBySqlString() {
-    if (!isRangeQuery || indexedColumn.isPresent()) {
+    if (!isRangeQuery
+        || indexedColumn.isPresent()
+        || tableMetadata.getClusteringKeyNames().isEmpty()) {
```
brfrn169, line 91: We always added ORDER BY clause when we scan a table (except index scan) but if the table doesn't have clustering key, we should not have done that.

ypeckstadt (approved):
LGTM. Thank you.

feeblefakie (approved):
LGTM! Thank you!
(Should the PR title be changed to `Fix ...` ?)

Torch3333 (approved):
LGTM, thank you!

PR #553, opened 2022-04-13
Fix equals method of Value for bytes type

brfrn169:
Currently, the `equals` of the `Value` class in SQL API doesn't work expected for bytes type. This PR fixes this issue. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/sql/Value.java
+++ b/core/src/main/java/com/scalar/db/sql/Value.java
@@ -2,6 +2,7 @@ package com.scalar.db.sql;
 
 import com.google.common.base.MoreObjects;
 import java.nio.ByteBuffer;
+import java.util.Arrays;
 import java.util.Objects;
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/sql/Value.java
```diff
@@ -30,8 +31,14 @@ public boolean equals(Object o) {
...
     }
-    Value value1 = (Value) o;
-    return type == value1.type && Objects.equals(value, value1.value);
+    Value that = (Value) o;
+    if (type != that.type) {
+      return false;
+    }
+    if (type == Type.BLOB_BYTES) {
+      return Arrays.equals((byte[]) value, (byte[]) that.value);
+    }
```
brfrn169, line 40: When the `type` is `BLOB_BYTES`, the `value` is a `byte[]` type, so we need to use `Arrays.equals()`.

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #561, opened 2022-04-21
Release SNAPSHOT versions

brfrn169:
This PR adds a GitHub actions workflow to release SNAPSHOT versions. Please take a look!

### DIFF
--- a/.github/workflows/release-snapshot.yaml
+++ b/.github/workflows/release-snapshot.yaml
@@ -1,0 +1,47 @@
+name: Release SHAPSHOT
+
+on:
+  push:
+    branches:
+      - master
+      - "[0-9]+"
+      - "[0-9]+.[0-9]+"
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Set version
+        id: version
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release-snapshot.yaml
```diff
@@ -0,0 +1,54 @@
+name: Release SHAPSHOT
+
+on:
+  push:
+    branches:
+      - master
+      - "[0-9]+"
+      - "[0-9]+.[0-9]+"
```
brfrn169, line 8: This workflow targets the `master` branch, the support branches (like `3`), and the release branches (like `3.4` and `3.5`). The triggers are `push` to those branches.

.github/workflows/release-snapshot.yaml
```diff
@@ -0,0 +1,54 @@
...
+            ${{ runner.os }}-gradle-
+
+      - name: Set version
+        id: version
+        run: |
+          VERSION=$(gradle :core:properties -q | grep "version:" | awk '{print $2}')
+          echo ::set-output name=version::${VERSION}
+
+      - name: Upload SNAPSHOT versions for scalardb, scalardb-server, scalardb-rpc, and scalardb-schema-loader to Maven Central Repository
+        if: contains(steps.version.outputs.version, '-SNAPSHOT')
```
brfrn169, line 39: It uploads the SNAPSHOT versions only when the project version contains `-SNAPSHOT`.

Torch3333 (approved):
LGTM, thank you!

.github/workflows/release.yaml
```diff
@@ -94,10 +94,3 @@ jobs:
           asset_path: schema-loader/build/libs/scalardb-schema-loader-${{ steps.version.outputs.version }}.jar
           asset_name: scalardb-schema-loader-${{ steps.version.outputs.version }}.jar
           asset_content_type: application/java-archive
-
-      - name: Cleanup Gradle cache
```
feeblefakie, line 98: OK to delete them? (I think it's auto-generated but not really sure why they are auto-generated.)

feeblefakie (approved):
LGTM! Thank you!

.github/workflows/release.yaml
brfrn169: Actually, I'm also not sure they are needed, so I deleted it. I will restore it if we have any problems with it. Thank you.

PR #568, opened 2022-04-22
Add docker check to CI

brfrn169:
This PR adds a docker check job to the CI workflow. Please take a look!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,6 +60,74 @@ jobs:
           name: gradle_spotbugs_reports
           path: /tmp/gradle_spotbugs_reports
 
+  docker:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up JDK 8
+        uses: actions/setup-java@v3
+        with:
+          java-version: '8'
+          distribution: 'temurin'
+
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Dockerfile Lint for Scalar DB Server
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: :server:dockerfileLint
+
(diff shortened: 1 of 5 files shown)


schema-loader/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ghcr.io/scalar-labs/jre8:1.1.1
 
-ADD scalardb-schema-loader-*.jar /app.jar
+COPY scalardb-schema-loader-*.jar /app.jar
```
brfrn169, line 3: The Dockerfile Lint suggested using `COPY` instead of `ADD`.

.github/workflows/ci.yaml
```diff
@@ -60,6 +60,74 @@ jobs:
...
+
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
```
Torch3333, line 83: My understanding is that `gradle/gradle-build-action@v2` action will [already cache gradle dependency](https://github.com/gradle/gradle-build-action#caching) making the usage of `actions/cache@v2` redundant. What do you think ?

.github/workflows/ci.yaml
```diff
@@ -60,6 +60,74 @@ jobs:
...
+
+      - name: Dockerfile Lint for Scalar DB Server
```
Torch3333, line 85: If this action fails because a vulnerability is found, I think the following actions will be skipped. 
Should the following actions be configured to always run using the `if: always()` syntax?

Torch3333 (commented):
Thanks, it looks nice overall. I left comment regarding some details.

.github/workflows/ci.yaml
brfrn169: Removed the Cache part in the latest commit. Thanks.

.github/workflows/ci.yaml
brfrn169: Added the `if: always()` syntax. Thanks.

brfrn169:
@Torch3333 I fixed it based on your review. Please take a look at it again when you have time!

Torch3333 (approved):
Thank you!!
LGTM

feeblefakie (approved):
LGTM! Thank you!

PR #570, opened 2022-04-26
Add scanAll for cassandra storage

Torch3333:
This adds the implementation of the `ScanAll` command for the Cassandra storage. 
I will add the implementation for other storage in subsequent PRs.


core/src/integration-test/java/com/scalar/db/api/DistributedStorageIntegrationTestBase.java
```diff
@@ -1418,6 +1426,142 @@ public void scan_ScanLargeDataWithOrdering_ShouldRetrieveExpectedValues()
...
+                                  .partitionKey(Key.ofInt(COL_NAME1, i))
+                                  .clusteringKey(Key.ofInt(COL_NAME4, j))
+                                  .nonKeyColumns(
+                                      ImmutableList.of(
+                                          TextColumn.of(COL_NAME2, Integer.toString(i + j)),
+                                          IntColumn.of(COL_NAME3, i + j),
+                                          BooleanColumn.of(COL_NAME5, j % 2 == 0)));
+                          expectedResults.add(erBuilder.build());
+                        }));
+    assertResultsContainsExactlyInAnyOrder(results, expectedResults);
```
Torch3333, line 1458: Since the `scanAll` results are unordered, asserting the values are correct is not as straightforward as for `Scan` results, so I added two utility methods in `TestUtils` to do so : 
- `assertResultsContainsExactlyInAnyOrder(results, expectedResults)`
- `assertResultsAreASubsetOf(results, expectedResults)`

core/src/main/java/com/scalar/db/api/DistributedStorage.java
```diff
@@ -121,6 +121,16 @@ public interface DistributedStorage {
...
 
+  /**
+   * Retrieves all the records from the storage with the specified {@link ScanAll} command and
+   * returns {@link Scanner} to iterate the results.
+   *
+   * @param scanAll a {@code ScanAll} command
+   * @return {@link Scanner} to iterate results
+   * @throws ExecutionException if the operation failed
+   */
+  Scanner scanAll(ScanAll scanAll) throws ExecutionException;
```
brfrn169, line 132: I didn't think we need to add a new method for the `ScanAll` operations, and I thought we can use the existing `scan` method. What do you think?

brfrn169 (commented):
Thank you! I left a comment. Please take a look!

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
```diff
@@ -112,6 +117,16 @@ private Select prepare(Scan scan) {
     return select;
   }
 
+  private Select prepare(ScanAll scanAll) {
+    Select select = getSelect(scanAll);
```
feeblefakie, line 121: It doesn't seem to be doing "ALLOW FILTERING"; thus, simple "scan everything" works but "scan with some predicates" doesn't work?
Please add tests for scan with some predicate that doesn't include a partition key.

https://www.datastax.com/blog/allow-filtering-explained

core/src/main/java/com/scalar/db/api/DistributedStorage.java
Torch3333: Ok, I misunderstood your design idea. I will remove this method.

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
feeblefakie: Sorry, as discussed, I misunderstood the spec for this PR. So, forget it!

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
```diff
@@ -55,11 +56,13 @@ public SelectStatementHandler(Session session) {
   @Override
   @Nonnull
   protected PreparedStatement prepare(Operation operation) {
-    checkArgument(operation, Get.class, Scan.class);
+    checkArgument(operation, Get.class, Scan.class, ScanAll.class);
```
brfrn169, line 59: If we add `ScanAll.class` here, I think we should add the same to the following place:
https://github.com/scalar-labs/scalardb/pull/570/files#diff-5feb096f3b977bcff460e21a574fbb6b8a0f8f84a0540f84f63eda0be2a14e5cR76

Or we don't need it for both places?

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
```diff
@@ -73,6 +76,8 @@ protected BoundStatement bind(PreparedStatement prepared, Operation operation) {
     checkArgument(operation, Get.class, Scan.class);
     if (operation instanceof Get) {
       return bind(prepared.bind(), (Get) operation);
+    } else if (operation instanceof ScanAll) {
+      return bind(prepared.bind(), (ScanAll) operation);
```
brfrn169, line 80: It looks like we don't need this `else if` block for now. But it might be good to have it for consistency with the `prepare()`. What do you think?

brfrn169 (commented):
I left a couple of very minor comments. Other that, LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
Torch3333: Indeed, it's not necessary in both cases. I will remove this. 
Thank you.

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
Torch3333: Same here, I think we can remove the `else if` block for ScanAll.

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
Torch3333: Sorry, I spoke too fast. 😅 
After checking more in detail, Since I added specific `prepare()` method for ScanAll 
 ` private Select prepare(ScanAll scanAll)`. I think it's more cautious to keep the `checkArgument(operation, Get.class, Scan.class, ScanAll.class);` as is.

core/src/main/java/com/scalar/db/storage/cassandra/SelectStatementHandler.java
Torch3333: Sorry I did not think this through as well. 
Calling the `bind()` method for `ScanAll` is not required as there are no columns to bind. 

Eventually, we could still keep the `else if` block `ScanAll` but simplify it by the following. 
```
protected BoundStatement bind(PreparedStatement prepared, Operation operation) {
    checkArgument(operation, Get.class, Scan.class, ScanAll.class);
    if (operation instanceof Get) {
      return bind(prepared.bind(), (Get) operation);
    } else if (operation instanceof ScanAll) {
      return prepared.bind();
    } else {
      return bind(prepared.bind(), (Scan) operation);
    }
  }
  ```
  What do you think ?

brfrn169: Sounds good!

Torch3333:
This is ready for review again, please have a look when you can. Thank you.

feeblefakie (approved):
LGTM! Thank you!

PR #577, opened 2022-05-10
Add ScanAll for DynamoDB storage

Torch3333:
This implements ScanAll for the DynamoDB storage. 
Please have a look!


core/src/main/java/com/scalar/db/storage/dynamo/request/PaginatedRequest.java
```diff
@@ -0,0 +1,29 @@
+package com.scalar.db.storage.dynamo.request;
+
+import java.util.Map;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+
+/** This interface abstracts DynamoDB requests that are paginated */
+public interface PaginatedRequest {
```
brfrn169, line 7: 👍

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -60,12 +62,15 @@ public SelectStatementHandler(DynamoDbClient client, TableMetadataManager metada
...
-      if (ScalarDbUtils.isSecondaryIndexSpecified(selection, tableMetadata)) {
+      if (!(selection instanceof ScanAll)
+          && ScalarDbUtils.isSecondaryIndexSpecified(selection, tableMetadata)) {
         return executeQueryWithIndex(selection, tableMetadata);
       }
 
       if (selection instanceof Get) {
         return executeGet((Get) selection, tableMetadata);
+      } else if (selection instanceof ScanAll) {
+        return executeScan((ScanAll) selection, tableMetadata);
```
feeblefakie, line 73: I think the naming is getting confusing between `scan` and `query`.
Maybe, we can name it `executeFullScan`, rename `executeQuery` with `executeScan`.

feeblefakie (approved):
LGTM! Thank you!
I left one suggestion but it doesn’t block it from being merged.

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
Torch3333: These `executeXXX` methods are named after the corresponding Dynamo operation that will be executed. 
This explains why it is confusing from the Scalar DB perspective. 

Scalar DB get -> Dynamo get
Scalar DB scan -> Dynamo query
Scalar DB scanAll -> Dynamo scan

I guess renaming them as you suggest would be easier to understand.

Torch3333: I renamed the methods as suggested.

PR #580, opened 2022-05-11
Add new steps in the release.yaml to push container to Azure Container Registry

kota2and3kan:
This PR add new steps in the release.yaml to push scalardb-server container image to Azure Container Registry.
(We don't need to push scalardb-schema-loader, because it is not used in Azure Market Place.)

Depending on the specification (requirement) of Azure Market Place, we need to push container image not only `patch version (X.Y.Z)` but also `minor version (X.Y)` and `latest tag`.

So, this PR add new steps to push the following three containers to Azure Container Registry, when the tag pushed.

* `scalardb-server:X.Y.Z`
* `scalardb-server:X.Y`
* `scalardb-server:latest`

Also, I tested the behavior of logic to detect latest version and create `minor version` / `latest` tag using following yaml.

<details>
<summary>YAML for test</summary>

```yaml
name: Release Test for Azure Container Registry

on:
  push:
    tags:
      - "v[0-9]+.[0-9]+.[0-9]+"

jobs:
  release:
    runs-on: ubuntu-latest

    steps:
      - name: Set version
        id: version
        run: |
          VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/v##g")
          echo ::set-output name=version::${VERSION}
          MINOR_VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/v##g" | grep -o -E "^[0-9]+\.[0-9]+")
          echo ::set-output name=minor_version::${MINOR_VERSION}
          LATEST_VERSION=$(curl -s -u "${{ github.repository_owner }}:${{ secrets.GITHUB_TOKEN }}" https://api.github.com/repos/kota2and3kan/test-push-container-to-acr/tags | jq -r .[].name | sort --version-sort -r | head -n 1 | sed -e "s#v##g")
          echo ::set-output name=latest_version::${LATEST_VERSION}

      - uses: actions/checkout@v2

      - name: Login to Azure Container Registry
        uses: docker/login-action@v1
        with:
          registry: ${{ secrets.ACR_LOGIN_SERVER }}
          username: ${{ secrets.ACR_USER }}
          password: ${{ secrets.ACR_PASSWORD }}

      - name: Pull containers for test (use NGINX)
        run: docker pull nginx:latest

      - name: Add tag for Azure Container Registry
        run: |
          docker tag nginx:latest ${{ secrets.ACR_LOGIN_SERVER }}/test:${{ steps.version.outputs.version }}
          docker tag nginx:latest ${{ secrets.ACR_LOGIN_SERVER }}/test:${{ steps.version.outputs.minor_version }}
          # For testing
          docker image ls

      - name: Push containers to Azure Container Registry
        run: |
          docker push ${{ secrets.ACR_LOGIN_SERVER }}/test:${{ steps.version.outputs.version }}
          docker push ${{ secrets.ACR_LOGIN_SERVER }}/test:${{ steps.version.outputs.minor_version }}

      - name: Push latest tag
        if: ${{ steps.version.outputs.version == steps.version.outputs.latest_version }}
        run: |
          docker tag nginx:latest ${{ secrets.ACR_LOGIN_SERVER }}/test:latest
          # For testing
          docker image ls
          docker push ${{ secrets.ACR_LOGIN_SERVER }}/test:latest
```
</details>

<details>
<summary>Result (List in the Azure Container Registry)</summary>

Push_container_test_from_GitHubActions_to_ACR
</details>

I will add comments for details.
Please take a look!

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,6 +15,10 @@ jobs:
         run: |
           VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/v##g")
           echo ::set-output name=version::${VERSION}
+          MINOR_VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/v##g" | grep -o -E "^[0-9]+\.[0-9]+")
+          echo ::set-output name=minor_version::${MINOR_VERSION}
+          LATEST_VERSION=$(curl -s -u "${{ github.repository_owner }}:${{ secrets.GITHUB_TOKEN }}" https://api.github.com/repos/scalar-labs/scalardb/tags | jq -r .[].name | sort --version-sort -r | head -n 1 | sed -e "s#v##g")
+          echo ::set-output name=latest_version::${LATEST_VERSION}
 
       - uses: actions/checkout@v2
 
@@ -40,6 +44,13 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.CR_PAT }}
 
+      - name: Login to Azure Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ secrets.ACR_LOGIN_SERVER }}
+          username: ${{ secrets.ACR_USER }}
+          password: ${{ secrets.ACR_PASSWORD }}
+
       - name: Create containers
         run: gradle docker
 
@@ -47,6 +58,19 @@ jobs:
         run: |
           docker push ghcr.io/scalar-labs/scalardb-server:${{ steps.version.outputs.version }}
           docker push ghcr.io/scalar-labs/scalardb-schema-loader:${{ steps.version.outputs.version }}
+
+      - name: Push containers to Azure Container Registry
+        run: |
+          docker tag ghcr.io/scalar-labs/scalardb-server:${{ steps.version.outputs.version }} ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:${{ steps.version.outputs.version }}
+          docker tag ghcr.io/scalar-labs/scalardb-server:${{ steps.version.outputs.version }} ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:${{ steps.version.outputs.minor_version }}
+          docker push ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:${{ steps.version.outputs.version }}
+          docker push ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:${{ steps.version.outputs.minor_version }}
+
+      - name: Push latest tag to Azure Container Registry
+        if: ${{ steps.version.outputs.version == steps.version.outputs.latest_version }}
+        run: |
+          docker tag ghcr.io/scalar-labs/scalardb-server:${{ steps.version.outputs.version }} ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:latest
+          docker push ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:latest
 
       - name: Upload scalardb, scalardb-server, scalardb-rpc, and scalardb-schema-loader to Maven Central Repository
         run: |


.github/workflows/release.yaml
```diff
@@ -15,6 +15,10 @@ jobs:
         run: |
           VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/v##g")
           echo ::set-output name=version::${VERSION}
+          MINOR_VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/v##g" | grep -o -E "^[0-9]+\.[0-9]+")
+          echo ::set-output name=minor_version::${MINOR_VERSION}
```
kota2and3kan, line 19: These steps get `minor version` using `${{ github.ref }}`.

.github/workflows/release.yaml
```diff
@@ -15,6 +15,10 @@ jobs:
...
+          LATEST_VERSION=$(curl -s -u "${{ github.repository_owner }}:${{ secrets.GITHUB_TOKEN }}" https://api.github.com/repos/scalar-labs/scalardb/tags | jq -r .[].name | sort --version-sort -r | head -n 1 | sed -e "s#v##g")
+          echo ::set-output name=latest_version::${LATEST_VERSION}
```
kota2and3kan, line 21: In these steps, I use GitHub API to get the information of `latest version` of Scalar DB.
You can try it by following command in your local.
```console
curl -s https://api.github.com/repos/scalar-labs/scalardb/tags | jq -r .[].name | sort --version-sort -r | head -n 1 | sed -e "s#v##g"
```
For example, we can get the string `3.5.2` now.

.github/workflows/release.yaml
```diff
@@ -48,6 +59,19 @@ jobs:
           docker push ghcr.io/scalar-labs/scalardb-server:${{ steps.version.outputs.version }}
           docker push ghcr.io/scalar-labs/scalardb-schema-loader:${{ steps.version.outputs.version }}
 
+      - name: Push containers to Azure Container Registry
```
kota2and3kan, line 62: In this step, I create the following two images and push them to Azure Container Registry.

* `scalardb-server:X.Y.Z`
* `scalardb-server:X.Y`

.github/workflows/release.yaml
```diff
@@ -48,6 +59,19 @@ jobs:
...
+        run: |
+          docker tag ghcr.io/scalar-labs/scalardb-server:${{ steps.version.outputs.version }} ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:${{ steps.version.outputs.version }}
+          docker tag ghcr.io/scalar-labs/scalardb-server:${{ steps.version.outputs.version }} ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:${{ steps.version.outputs.minor_version }}
+          docker push ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:${{ steps.version.outputs.version }}
+          docker push ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:${{ steps.version.outputs.minor_version }}
+
+      - name: Push latest tag to Azure Container Registry
+        if: ${{ steps.version.outputs.version == steps.version.outputs.latest_version }}
```
kota2and3kan, line 70: In this step, I create the following image and push it to Azure Container Registry.

* `scalardb-server:latest`

This step only run, if the following two versions are same.

* The version of `tag` pushed as a trigger of workflow.
* The latest version in the entire repository.

brfrn169 (approved):
@kota2and3kan LGTM! Thank you! It's nice!

I think we can do the same docker image tagging for GitHub Container Registry as well. Can you please do that in another PR?

kota2and3kan:
Thank you for your review!

> I think we can do the same docker image tagging for GitHub Container Registry as well. Can you please do that in another PR?

Yes. We can do the same tagging in the GitHub Container Registry, but the `minor version` and `latest` tag are mutable.
So, I think that there is possibility that those tags lead some issues caused by the version difference, when users use them in the code (e.g. specify the latest tag in custom value file of Helm Charts).
However, Azure Market Place required the `latest` tag to provide containers.
So first, I decided to push the `minor version` and `latest` tags to Azure Container Registry only. 

I would like to discuss if the `minor version` and `latest` tags are necessary in the GitHub Container Registry.
(I think it is not better that there is difference about tags between several Container Registry, but I don't want users to use mutable tags because it causes some issues...)

@brfrn169 @feeblefakie  
What do you think?

feeblefakie (approved):
LGTM! Thank you!

feeblefakie:
@kota2and3kan @brfrn169 
Let’s discuss! 
IMHO, I also think using latest and minor latest is error-prone.

brfrn169:
@feeblefakie  @kota2and3kan 
I thought having `latest` and `minor latest` tags is the standard because many published docker images do that. But yes, I also think using latest and minor latest is error-prone and problematic, and I agree not to use those tags in the GitHub Container Registry.

kota2and3kan:
@feeblefakie @brfrn169  
Thank you for your comments!

At this time, I think to push `minor latest` and `latest` tags to the Azure Container Registry only.
We can push those tags to the GitHub Container Registry anytime if we need them in the future.

PR #582, opened 2022-05-13
Add ScanAll for JDBC storage

Torch3333:
This implements ScanAll for the JDBC storage. Please have a look.

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
@@ -8,6 +8,7 @@ import com.scalar.db.api.Mutation;
 import com.scalar.db.api.Put;
 import com.scalar.db.api.Result;
 import com.scalar.db.api.Scan;
+import com.scalar.db.api.ScanAll;
 import com.scalar.db.api.Scanner;
 import com.scalar.db.api.TableMetadata;
 import com.scalar.db.common.TableMetadataManager;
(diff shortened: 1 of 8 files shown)


brfrn169 (commented):
Thank you. Left a comment. Please take a look when you have time!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
```diff
@@ -84,11 +85,19 @@ public Optional<Result> get(Get get, Connection connection)
...
+    }
+
     TableMetadata tableMetadata = tableMetadataManager.getTableMetadata(scan);
     ScalarDbUtils.addProjectionsForKeys(scan, tableMetadata);
 
-    SelectQuery selectQuery = buildSelectQueryForScan(scan, tableMetadata);
+    SelectQuery selectQuery =
+        scan instanceof ScanAll
+            ? buildSelectQueryForScanAll((ScanAll) scan, tableMetadata)
+            : buildSelectQueryForScan(scan, tableMetadata);
```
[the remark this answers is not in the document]
Torch3333: I thought it was more appropriate to make the change for ScanAll in `scan()` later on inside a PR related to the transaction implementation.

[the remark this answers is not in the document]
brfrn169: Maybe I think we can do the change for `scan()` in this PR so that we can purely change the transaction side in the PR related to the transaction implementation. What do you think?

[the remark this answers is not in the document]
Torch3333: Understood 👍

brfrn169 (dismissed):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/query/SelectWithTop.java
```diff
@@ -3,18 +3,18 @@
...
+  public SelectWithTop(Builder builder, int limit) {
     super(builder);
     assert limit > 0;
     this.limit = limit;
   }
 
   @Override
   public String sql() {
-    return super.sql() + " OFFSET 0 ROWS FETCH NEXT " + limit + " ROWS ONLY";
+    return new StringBuilder(super.sql()).insert(7, "TOP " + limit + " ").toString();
```
feeblefakie, line 18: Haven't checked the other code, but the hard-coded `7` is always OK?

feeblefakie (dismissed):
LGTM! Thank you!
Left one question just in case.

core/src/main/java/com/scalar/db/storage/jdbc/query/SelectWithTop.java
Torch3333: Yes. 
The `super.sql()` always returns a select query like `SELECT ...`. This method inserts right after the `SELECT` the `TOP X` option. 
I used the `StringBuilder.insert` instead of something more readable like a `String.replace()` because it is less CPU intensive.

feeblefakie: Ah, OK. I didn't really know "SELECT TOP` clause. Thank you!
Maybe better to add a comment on the above to specify what `7` means for people who don't know the `SELECT TOP` clause like me. :)

Torch3333: Sure 🙂

feeblefakie (approved):

PR #583, opened 2022-05-18
Implement the checkPaused service according to scalar-admin 1.2.0

supl:
This PR implements the new Admin gRPC service `checkPaused` by which the client can get the pause status.

I tested it with the scalar-admin 1.2.0's executables and it works as expected.

@Torch3333 
Vincent, for your reference.
Scalar server has implemented a gRPC server for these services of this [Protocol Buffers](https://github.com/scalar-labs/scalar-admin/blob/main/admin.proto) by which we can pause/unpause the Scalar server.

This PR implements the new service `checkPaused` which is used for getting the pause status of the Scalar server.

### DIFF
--- a/server/src/main/java/com/scalar/db/server/AdminService.java
+++ b/server/src/main/java/com/scalar/db/server/AdminService.java
@@ -3,6 +3,7 @@ package com.scalar.db.server;
 import com.google.inject.Inject;
 import com.google.protobuf.Empty;
 import com.scalar.admin.rpc.AdminGrpc;
+import com.scalar.admin.rpc.CheckPausedResponse;
 import com.scalar.admin.rpc.PauseRequest;
 import com.scalar.admin.rpc.StatsResponse;
 import io.grpc.Status;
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/scalar/db/server/AdminService.java
```diff
@@ -66,4 +67,11 @@ public void stats(Empty request, StreamObserver<StatsResponse> responseObserver)
     responseObserver.onNext(StatsResponse.newBuilder().setStats("{}").build());
     responseObserver.onCompleted();
   }
+
+  @Override
+  public void checkPaused(Empty request, StreamObserver<CheckPausedResponse> responseObserver) {
+    responseObserver.onNext(
+        CheckPausedResponse.newBuilder().setPaused(!gateKeeper.letIn()).build());
```
supl, line 74: use `Gatekeeper.letIn()` as the current paused status

brfrn169: I don't think we can use `letIn()` to get the current paused status.

Maybe we need to add something like `isOpen()` method to `GateKeeper`, and make the new method return the current paused status (the `isOpen` field in `LockFreeGateKeeper` and `SynchronizedGateKeeper`).

brfrn169 (commented):
Thank you. I left a comment. Please take a look when you have time!

server/src/main/java/com/scalar/db/server/AdminService.java
supl: Understood.
Let me revise it.

supl: @brfrn169 
Sorry for my mistake.
I add `isOpen` method in https://github.com/scalar-labs/scalardb/pull/583/commits/acb118db9e08acc1cdd87c55946549363d91d75d

Please take a look!
Thank you 😄

brfrn169 (approved):
LGTM, Thank you!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #585, opened 2022-05-19
Use `max_pause_wait_time` in AdminService

supl:
This PR uses the gRPC optional argument max_pause_wait_time to drain the service.

The optional argument `max_pause_wait_time` was added in scalar-admin 1.1.0, but, for a release issue, it wasn't applied on the server side yet.

This PR does the makeup.

### DIFF
--- a/server/src/main/java/com/scalar/db/server/AdminService.java
+++ b/server/src/main/java/com/scalar/db/server/AdminService.java
@@ -17,7 +17,7 @@ import org.slf4j.LoggerFactory;
 public class AdminService extends AdminGrpc.AdminImplBase {
   private static final Logger LOGGER = LoggerFactory.getLogger(AdminService.class);
 
-  private static final long DEFAULT_MAX_PAUSE_WAIT_TIME_MILLIS = 10000; // 10 seconds
+  private static final long DEFAULT_MAX_PAUSE_WAIT_TIME_MILLIS = 30000; // 30 seconds
 
   private final GateKeeper gateKeeper;
 
--- a/server/src/main/java/com/scalar/db/server/AdminService.java
+++ b/server/src/main/java/com/scalar/db/server/AdminService.java
@@ -17,7 +17,7 @@ import org.slf4j.LoggerFactory;
 public class AdminService extends AdminGrpc.AdminImplBase {
   private static final Logger LOGGER = LoggerFactory.getLogger(AdminService.class);
 
-  private static final long MAX_PAUSE_WAIT_TIME_MILLIS = 10000; // 10 seconds
+  private static final long DEFAULT_MAX_PAUSE_WAIT_TIME_MILLIS = 10000; // 10 seconds
 
   private final GateKeeper gateKeeper;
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/scalar/db/server/AdminService.java
```diff
@@ -28,12 +29,17 @@ public AdminService(GateKeeper gateKeeper) {
   @Override
   public void pause(PauseRequest request, StreamObserver<Empty> responseObserver) {
     gateKeeper.close();
+    long maxPauseWaitTime =
+        request.getMaxPauseWaitTime() != 0
```
supl, line 33: if `max_pause_wait_time` is not set, getMaxPauseWaitTime will return `0`.

reference:
https://developers.google.com/protocol-buffers/docs/javatutorial

```
a system default is used: zero for numeric types, the empty string for strings, false for bools. 
```

server/src/main/java/com/scalar/db/server/AdminService.java
```diff
@@ -17,7 +17,7 @@
 public class AdminService extends AdminGrpc.AdminImplBase {
   private static final Logger LOGGER = LoggerFactory.getLogger(AdminService.class);
 
-  private static final long MAX_PAUSE_WAIT_TIME_MILLIS = 10000; // 10 seconds
+  private static final long DEFAULT_MAX_PAUSE_WAIT_TIME_MILLIS = 10000; // 10 seconds
```
brfrn169, line 20: Actually, I was thinking the original max wait time is too short. So can you please increase it to 30 seconds at this time?

brfrn169 (commented):
@supl Thank you so much! I thought I needed to do it, but you did it!

LGTM! Just left a request. Please take a look when you have time! Thanks.

server/src/main/java/com/scalar/db/server/AdminService.java
supl: Of course!
https://github.com/scalar-labs/scalardb/pull/585/commits/f3f1201fa6d284c716b29c7ce20a1192f41b2b1b

server/src/main/java/com/scalar/db/server/AdminService.java
```diff
@@ -33,8 +33,13 @@ public void pause(PauseRequest request, StreamObserver<Empty> responseObserver)
     if (request.getWaitOutstanding()) {
       LOGGER.warn("Pausing... waiting until outstanding requests are all finished");
       boolean drained = false;
+      long maxPauseWaitTime =
+          request.getMaxPauseWaitTime() != 0
```
supl, line 37: if `max_pause_wait_time` is not set, getMaxPauseWaitTime will return `0`.

reference:
https://developers.google.com/protocol-buffers/docs/javatutorial

```
a system default is used: zero for numeric types, the empty string for strings, false for bools. 
```

brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #586, opened 2022-05-19
Implement ScanAll for Consensus Commit

Torch3333:
This implements the ScanAll operation for Consensus Commit and Two Phase Consensus Commit
Please have a look.


core/src/main/java/com/scalar/db/transaction/consensuscommit/AbstractMutationComposer.java
```diff
@@ -42,4 +43,12 @@ protected Optional<Key> getClusteringKey(Operation base, TransactionResult resul
...
     }
   }
+
+  protected Key getPartitionKey(Operation base, TransactionResult result) {
+    if (base instanceof ScanAll) {
+      return result.getPartitionKey().get();
+    } else {
+      return base.getPartitionKey();
+    }
+  }
```
Torch3333, line 53: When a `ScanAll` operation triggers a record recovery, since it does not hold the partition key, we need to use the partition key from the result.

brfrn169 (commented):
It's looking good to me! Thank you!

Since it's a critical part, let me re-check it after some time. Thanks.

feeblefakie (approved):
LGTM! Thank you!

brfrn169 (approved):
LGTM! Thank you!

PR #588, opened 2022-05-20
Add suspend method to TwoPhaseCommitTransactionManager

brfrn169:
This is actually a backward incompatible change for Two-phase Commit Transactions. But I don't think the 2PC transaction feature is not widely used, so we can change it for now.

Currently, when we join a transaction (with `TwoPhaseCommitTransactionManager.join()`), the transaction object is automatically put in the object map (`activeTransactions`), and we can resume this transaction with `TwoPhaseCommitTransactionManager.resume()`. But after this change, we need to call `TwoPhaseCommitTransactionManager.suspend()` explicitly if you want to resume the transaction.

I also modified the document for this feature:
https://github.com/scalar-labs/scalardb/blob/add-suspend-to-2pc/docs/two-phase-commit-transactions.md#suspend-and-resume-the-transaction

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
+++ b/core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
@@ -81,8 +81,12 @@ public class ActiveExpiringMap<K, V> {
     return map.containsKey(key);
   }
 
-  public void remove(K key) {
-    map.remove(key);
+  public V remove(K key) {
+    ValueHolder<V> prev = map.remove(key);
+    if (prev == null) {
+      return null;
+    }
+    return prev.get();
   }
 
   public void updateExpirationTime(K key) {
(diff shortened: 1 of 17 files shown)


core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
```diff
@@ -81,8 +81,12 @@ public boolean contains(K key) {
...
-    map.remove(key);
+  public V remove(K key) {
+    ValueHolder<V> prev = map.remove(key);
+    if (prev == null) {
+      return null;
+    }
+    return prev.get();
   }
 
   public void updateExpirationTime(K key) {
```
Torch3333, line 92: This method does not appear to be used anymore, should we still keep it?

Torch3333 (approved):
Sorry for the late review. LGTM, thank you!
I just left a small comment. Please have a look.

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
brfrn169: Thank you for pointing it out!

I think we can keep it because it's a useful for some use cases. `ActiveExpiringMap` is a commonly used utility class and it might be used in another part in the future.

PR #591, opened 2022-05-25
Handle projections in query for Cosmos DB

Torch3333:
Currently, columns are not projected when doing a Selection operation for Cosmos DB, the projected columns are filtered on the client side after the query is executed.
With this PR, the projected columns are now parametrized in the query directly. 

Please have a look.

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/api/DistributedStorageIntegrationTestBase.java
+++ b/core/src/integration-test/java/com/scalar/db/api/DistributedStorageIntegrationTestBase.java
@@ -197,7 +197,10 @@ public abstract class DistributedStorageIntegrationTestBase {
 
     // Act
     Get get = prepareGet(pKey, cKey);
-    get.withProjection(COL_NAME1).withProjection(COL_NAME2).withProjection(COL_NAME3);
+    get.withProjection(COL_NAME1)
+        .withProjection(COL_NAME2)
+        .withProjection(COL_NAME3)
+        .withProjection(COL_NAME6);
     Optional<Result> actual = storage.get(get);
 
     // Assert
@@ -210,6 +213,8 @@ public abstract class DistributedStorageIntegrationTestBase {
         .isEqualTo(Optional.of(new IntValue(COL_NAME3, pKey + cKey)));
     assertThat(actual.get().getValue(COL_NAME4).isPresent()).isTrue(); // since it's clustering key
     assertThat(actual.get().getValue(COL_NAME5).isPresent()).isFalse();
+    assertThat(actual.get().contains(COL_NAME6)).isTrue();
+    assertThat(actual.get().isNull(COL_NAME6)).isTrue();
   }
 
   @Test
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java
```diff
@@ -135,6 +147,46 @@ private String makeQueryWithCondition(
...
+      // placeholder, each curly brace need to be doubled "{{" to have a single curly brace "{"
+      // present in the generated sql query
+      projectedFields.add("{{" + String.join(",", projectedValues) + "}} as values");
+    }
+
+    return DSL.using(SQLDialect.DEFAULT)
+        .select(projectedFields.stream().map(DSL::field).collect(Collectors.toList()))
+        .from("Record r");
+  }
+
```
Torch3333, line 189: The query projections are handled here. 
For example, to project the columns `c3` and `c4`, the request for a ScanAll will be like : 
`select r.id, 
             r.concatenatedPartitionKey, 
             r.partitionKey, 
             r.clusteringKey, 
             {"name3":r.values["name3"],"name4":r.values["name4"]} as values 
             from Record r`

core/src/integration-test/java/com/scalar/db/api/DistributedStorageIntegrationTestBase.java
```diff
@@ -210,6 +213,8 @@ public void get_GetWithProjectionsGiven_ShouldRetrieveSpecifiedValues()
         .isEqualTo(Optional.of(new IntValue(COL_NAME3, pKey + cKey)));
     assertThat(actual.get().getValue(COL_NAME4).isPresent()).isTrue(); // since it's clustering key
     assertThat(actual.get().getValue(COL_NAME5).isPresent()).isFalse();
+    assertThat(actual.get().contains(COL_NAME6)).isTrue();
+    assertThat(actual.get().isNull(COL_NAME6)).isTrue();
```
Torch3333, line 217: I updated some unit tests to be more robust by projecting as well the column `COL_NAME6` which value is null.

core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java
```diff
@@ -135,6 +147,46 @@ private String makeQueryWithCondition(
...
+    List<String> projectedFields = new ArrayList<>();
+
+    // To project the required columns, we build a JSON object with the same structure as the
+    // `Record.class`so that each field can be deserialized properly into a `Record.class` object.
+    // For example, the projected field "r.id" will be mapped to the `Record.id` attribute
+    projectedFields.add("r.id");
+    projectedFields.add("r.concatenatedPartitionKey");
+    projectedFields.add("r.partitionKey");
+    projectedFields.add("r.clusteringKey");
```
brfrn169, line 164: Just a question. Do we need project these fields in Cosmos DB even if we don't specify partition key and clustering key as projections?

brfrn169 (approved):
LGTM! Thank you!

I left a question. Please take a look when you have time!

core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java
Torch3333: For this PR, for the sake of being consistent with the other adapters (otherwise I would have needed to make an exception in the integration testing for Cosmos), I projected by default the partition key and clustering key even though I will need to update the logic again in another PR to not project by default the partition and clustering key for all adapters.

feeblefakie (approved):
LGTM! Thank you!

PR #593, opened 2022-05-31
For Dynamo storage, throw an exception when a Put set a secondary index column value to null

Torch3333:
Please have a look.

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/common/checker/DynamoOperationChecker.java
+++ b/core/src/main/java/com/scalar/db/storage/common/checker/DynamoOperationChecker.java
@@ -1,0 +1,26 @@
+package com.scalar.db.storage.common.checker;
+
+import com.scalar.db.api.Put;
+import com.scalar.db.api.TableMetadata;
+import com.scalar.db.common.TableMetadataManager;
+
+public class DynamoOperationChecker extends OperationChecker {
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/scalar/db/storage/common/checker/DynamoOperationChecker.java
```diff
@@ -0,0 +1,26 @@
+package com.scalar.db.storage.common.checker;
+
+import com.scalar.db.api.Put;
+import com.scalar.db.api.TableMetadata;
+import com.scalar.db.common.TableMetadataManager;
+
+public class DynamoOperationChecker extends OperationChecker {
```
brfrn169, line 7: I think we should put `DynamoOperationChecker` in the `com.scalar.db.storage.dynamo` package. What do you think?

brfrn169 (commented):
Thank you! LGTM overall. I left a comment. Please take a look when you have time!

core/src/main/java/com/scalar/db/storage/common/checker/DynamoOperationChecker.java
Torch3333: Yes, that makes more sense. Thank you

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #594, opened 2022-06-03
Add check for empty key

brfrn169:
Please take a look!


core/src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
```diff
@@ -285,15 +285,19 @@ private void checkClusteringKey(Operation operation, TableMetadata metadata) {
...
     } else {
-      if (key.getColumns().size() > keyNames.size()) {
+      if (key.size() > keyNames.size()) {
         return false;
       }
     }
 
+    if (key.size() == 0) {
+      return false;
+    }
```
brfrn169, line 299: Added a check for an empty key.

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #597, opened 2022-06-03
Upgrade SpotBugs

brfrn169:
Please take a look!


build.gradle
```diff
@@ -44,7 +44,7 @@ subprojects {
         commonsLangVersion = '3.12.0'
         assertjVersion = '3.22.0'
         mockitoVersion = '3.12.4'
-        spotbugsVersion = '4.3.0'
+        spotbugsVersion = '4.6.0'
```
Torch3333, line 47: If the goal of the PR is to update the spotbugs version used for the code analysis, I wonder if this change will have an impact as this variable is only used by this dependency. From the name, it might only be responsible for importing the spot bugs annotation but not running the code analysis
```
compileOnly "com.github.spotbugs:spotbugs-annotations:${spotbugsVersion}"
testCompileOnly "com.github.spotbugs:spotbugs-annotations:${spotbugsVersion}"
```
To upgrade the spotbug version, we might need to update the spotbug plugin version too.
[id 'com.github.spotbugs' version '4.7.1'](https://github.com/scalar-labs/scalardb/blob/72d373b13ecf59c5e446ca962de46c92e3ea53ae/core/build.gradle#L2)

Torch3333 (commented):
Thank you! I left a question. 
Please have a look when you can.

build.gradle
feeblefakie: Seems like it. (@Torch3333 👍 )
https://github.com/spotbugs/spotbugs-gradle-plugin#spotbugs-version-mapping

brfrn169: Ah, that's right. I will update the spotbug plugin version. Thank you!

brfrn169: @feeblefakie @Torch3333 I updated the spotbug plugin version to `5.0.7`. Please take a look again! Thanks.

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #598, opened 2022-06-03
Add a way to disable metadata cache

brfrn169:
Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/common/TableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/common/TableMetadataManager.java
@@ -21,7 +21,7 @@ public class TableMetadataManager {
 
   public TableMetadataManager(Admin admin, long cacheExpirationTimeSecs) {
     CacheBuilder<Object, Object> builder = CacheBuilder.newBuilder();
-    if (cacheExpirationTimeSecs > 0) {
+    if (cacheExpirationTimeSecs >= 0) {
       builder.expireAfterWrite(cacheExpirationTimeSecs, TimeUnit.SECONDS);
     }
     tableMetadataCache =
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadataManager.java
@@ -22,7 +22,7 @@ public class TransactionalTableMetadataManager {
       DistributedStorageAdmin admin, long cacheExpirationTimeSecs) {
 
     CacheBuilder<Object, Object> builder = CacheBuilder.newBuilder();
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/scalar/db/common/TableMetadataManager.java
```diff
@@ -21,7 +21,7 @@ public class TableMetadataManager {
 
   public TableMetadataManager(Admin admin, long cacheExpirationTimeSecs) {
     CacheBuilder<Object, Object> builder = CacheBuilder.newBuilder();
-    if (cacheExpirationTimeSecs > 0) {
+    if (cacheExpirationTimeSecs >= 0) {
```
brfrn169, line 24: After this change, we can completely disable the metadata cache when setting `cacheExpirationTimeSecs` to `0`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionalTableMetadataManager.java
```diff
@@ -22,7 +22,7 @@ public TransactionalTableMetadataManager(
       DistributedStorageAdmin admin, long cacheExpirationTimeSecs) {
 
     CacheBuilder<Object, Object> builder = CacheBuilder.newBuilder();
-    if (cacheExpirationTimeSecs > 0) {
+    if (cacheExpirationTimeSecs >= 0) {
```
brfrn169, line 25: Ditto.

core/src/main/java/com/scalar/db/common/TableMetadataManager.java
Torch3333: Without this change, when `cacheExpirationTimeSecs=0`, the `CacheLoader` would still have a default `expireAfterWrite` value ?

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/common/TableMetadataManager.java
brfrn169: If `cacheExpirationTimeSecs=0` without this change, as we don't set `expireAfterWrite` value, the `CacheLoader` doesn't expire any cache entries.

So after this change,
- if we set `cacheExpirationTimeSecs` to a negative value, enabling cache but no expiration
- if we set `cacheExpirationTimeSecs` to `0`, disabling cache
- if we set `cacheExpirationTimeSecs` to a positive value, enabling cache and expiration

Torch3333: Ok, thank your for the clarification. 👍

PR #603, opened 2022-06-10
Update the docs

feeblefakie:
The getting started guides build scalardb from the source code so we need Java 8 to do it instead of higher versions.


brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #605, opened 2022-06-18
Fix incompatible change

brfrn169:
Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/Put.java
+++ b/core/src/main/java/com/scalar/db/api/Put.java
@@ -11,6 +11,7 @@ import com.scalar.db.io.BigIntColumn;
 import com.scalar.db.io.BlobColumn;
 import com.scalar.db.io.BooleanColumn;
 import com.scalar.db.io.Column;
+import com.scalar.db.io.DataType;
 import com.scalar.db.io.DoubleColumn;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -497,7 +498,18 @@ public Put withValue(Column<?> column) {
   @Deprecated
   public Map<String, Value<?>> getValues() {
     Map<String, Value<?>> ret = new HashMap<>();
-    columns.forEach((k, v) -> ret.put(k, v.hasNullValue() ? null : ScalarDbUtils.toValue(v)));
+    columns.forEach(
+        (k, c) -> {
+          Value<?> value;
+          if (c.hasNullValue()
+              && c.getDataType() != DataType.TEXT
+              && c.getDataType() != DataType.BLOB) {
```
brfrn169, line 506: When a Text value and a Blob value are `null`, we should return `new TextValue("<column name>", null)` and `new BlobValue("<column name>", null)` respectively instead of `null`.

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #608, opened 2022-06-20
Refactor Guice modules

brfrn169:
Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/service/StorageModule.java
+++ b/core/src/main/java/com/scalar/db/service/StorageModule.java
@@ -7,6 +7,8 @@ import com.scalar.db.api.DistributedStorage;
 import com.scalar.db.api.DistributedStorageAdmin;
 import com.scalar.db.config.DatabaseConfig;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/service/StorageModule.java
```diff
@@ -17,8 +19,8 @@ public StorageModule(DatabaseConfig config) {
 
   @Override
   protected void configure() {
-    bind(DistributedStorage.class).to(config.getStorageClass()).in(Singleton.class);
-    bind(DistributedStorageAdmin.class).to(config.getStorageAdminClass()).in(Singleton.class);
+    bind(DistributedStorage.class).to(config.getStorageClass());
```
brfrn169, line 22: Also, removed the `Singleton` scope. This is because the storage and admin and transaction manager instances are closeable, so if they are closed, then Guice always returns already closed instances.

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #609, opened 2022-06-20
Refactor Scalar DB Server

brfrn169:
Please take a look!


rpc/src/main/proto/scalardb.proto
```diff
@@ -345,7 +345,7 @@ message TransactionResponse {
     enum ErrorCode {
       INVALID_ARGUMENT = 0;
       CONFLICT = 1;
-      UNKNOWN_TRANSACTION = 2;
+      UNKNOWN_TRANSACTION_STATUS = 2;
```
brfrn169, line 348: Renamed it. It's backward compatible because the index is not changed.

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #613, opened 2022-06-22
Add methods to repair table and coordinator table to the Admin interfaces

Torch3333:
This adds the `repairTable` and `repairCoordinatorTables` to the admin interface. Please refer to the comments below for the method signature.

Besides, the "repair table" option will be added to the schema loader in a following PR as this one is already big enough.


core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -332,4 +332,17 @@ default boolean indexExists(String namespace, String table, String columnName)
...
+   *
+   * @param namespace an existing namespace
+   * @param table an existing table
+   * @param metadata the metadata associated to the table to repair
+   * @param options options to repair
+   * @throws ExecutionException if the operation failed
+   */
+  void repairTable(
+      String namespace, String table, TableMetadata metadata, Map<String, String> options)
+      throws ExecutionException;
```
Torch3333, line 347: The definition of the new `repairTable` method available on all Admin interfaces.

core/src/main/java/com/scalar/db/api/DistributedTransactionAdmin.java
```diff
@@ -96,6 +96,14 @@ default void dropCoordinatorTables(boolean ifExist) throws ExecutionException {
...
   boolean coordinatorTablesExist() throws ExecutionException;
 
+  /**
+   * Repair coordinator tables which may be in an unknown state
+   *
+   * @param options options to repair
+   * @throws ExecutionException if the operation failed
+   */
+  void repairCoordinatorTables(Map<String, String> options) throws ExecutionException;
+
```
Torch3333, line 106: The definition of the new `repairCoordinatorTables` available on transactional admin interfaces.

rpc/src/main/proto/scalardb.proto
```diff
@@ -194,6 +194,8 @@ service DistributedStorageAdmin {
   }
   rpc NamespaceExists(NamespaceExistsRequest) returns (NamespaceExistsResponse) {
   }
+  rpc RepairTable(RepairTableRequest) returns (google.protobuf.Empty) {
+  }
```
Torch3333, line 198: New endpoints were added to the grpc interface.

rpc/src/main/proto/scalardb.proto
```diff
@@ -502,6 +512,10 @@ service DistributedTransactionAdmin {
   }
   rpc CoordinatorTablesExist(CoordinatorTablesExistRequest) returns (CoordinatorTablesExistResponse) {
   }
+  rpc RepairTable(RepairTableRequest) returns (google.protobuf.Empty) {
+  }
+  rpc RepairCoordinatorTables(RepairCoordinatorTablesRequest) returns (google.protobuf.Empty){
+  }
```
Torch3333, line 518: Here as well.

core/src/integration-test/java/com/scalar/db/util/AdminTestUtils.java
```diff
@@ -0,0 +1,41 @@
+package com.scalar.db.util;
+
+import com.scalar.db.config.DatabaseConfig;
+import java.util.Properties;
+
+/** Utility class used to delete or truncate the metadata table */
+public abstract class AdminTestUtils {
```
Torch3333, line 7: This adds utility method to delete and truncate the metadata table to be used in integration testing.

Torch3333:
> @Torch3333 Can you explain what `repair` means here?

If the metadata for the given table does not exist because either
- case 1: the metadata table does not exist
- case 2: the metadata table exists but does not contain the metadata for the given table
Then the `repairTable` method will create the metadata table if necessary and add the metadata for the given table. 

`repairTable(String namespace, String table, TableMetadata metadata, Map<String, String> options)` will repair a given table while `repairCoordinatorTables(Map<String, String> options)` will repair the coordinator table only.

Additionally, the implementation of the `repairTable` for the CosmosAdmin also restores if necessary the `mutate.js` stored procedure to the given table. The reason is that when a Cosmos DB table is restored after a backup, the stored procedure is not restored.

core/src/main/java/com/scalar/db/api/Admin.java
feeblefakie: So, it's repairing the metadata of tables, should it be `repairTableMetadata` or `repairMetadataTable`?
`repairTable` sounds like repairing a table itself like Cassandra's repair.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -460,6 +483,29 @@ public boolean namespaceExists(String namespace) throws ExecutionException {
...
+        throw new IllegalArgumentException(
+            "The table " + getFullTableName(namespace, table) + "  does not exist");
+      }
+    }
+    createMetadataDatabaseAndContainerIfNotExists();
+    if (getTableMetadata(namespace, table) == null) {
+      putTableMetadata(namespace, table, metadata);
+    }
+    if (!storedProcedureExists(namespace, table)) {
+      addStoredProcedure(namespace, table);
```
feeblefakie, line 505: Hmm, this is a little exceptional since stored procedures are not table metadata.

feeblefakie (commented):
Thank you!
Did a quick look and left some comments. 
(sorry, due to the time limit, I will take a look at it again later)

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
Torch3333: I will answer here for both of your comments. 
 
Our goal in creating this method was to address two needs : 
- Be able to repair table metadata which is common to all storage
- Be able to restore a table stored procedure which is Cosmos DB specific

Adding a separate method to the Admin interface to address the stored procedure repair does not fit the Admin interface philosophy so far. For example, specific storage options to configure table (like Dynamo DB RU or disabling Dynamo DB auto-backup)  are addressed in the `createTable` method using the `Map<String, String> options` parameter. We decided to have a similar approach for the `createTable` method.
That’s why the repairTable naming is quite vague, it indeed repairs metadata but might also repair other things.

core/src/integration-test/java/com/scalar/db/util/CosmosAdminTestUtils.java
```diff
@@ -0,0 +1,78 @@
...
+import com.azure.cosmos.models.CosmosQueryRequestOptions;
+import com.azure.cosmos.models.PartitionKey;
+import com.azure.cosmos.util.CosmosPagedIterable;
+import com.scalar.db.config.DatabaseConfig;
+import com.scalar.db.storage.cosmos.CosmosAdmin;
+import com.scalar.db.storage.cosmos.CosmosConfig;
+import com.scalar.db.storage.cosmos.Record;
+import java.util.Properties;
+
+public class CosmosAdminTestUtils extends AdminTestUtils {
```
brfrn169, line 19: Is it better that these `XXXAdminTestUtils` classes in the storage specific packages (`com.scalar.db.storage.xxx`)?

brfrn169 (commented):
Thank you! LGTM overall. Left several minor comments/questions. Please take a look when you have time!

server/src/integration-test/java/com/scalar/db/server/ServerEnv.java
```diff
@@ -9,8 +9,7 @@
 
 public final class ServerEnv {
 
-  private static final String PROP_SERVER_EXTERNAL_SERVER_USED =
-      "scalardb.server.external_server_used";
+  static final String PROP_SERVER_EXTERNAL_SERVER_USED = "scalardb.server.external_server_used";
```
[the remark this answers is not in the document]
Torch3333: My bad, this is a mistake. Thank you for catching it.

core/src/integration-test/java/com/scalar/db/api/DistributedStorageAdminRepairTableIntegrationTestBase.java
```diff
@@ -0,0 +1,139 @@
...
+    // Arrange
+    adminTestUtils.dropMetadataTable();
+
+    // Act
+    admin.repairTable(getNamespace(), getTable(), TABLE_METADATA, getCreateOptions());
+
+    // Assert
+    assertThat(admin.tableExists(getNamespace(), getTable())).isTrue();
+    assertThat(admin.getTableMetadata(getNamespace(), getTable())).isEqualTo(TABLE_METADATA);
+  }
```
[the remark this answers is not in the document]
Torch3333: As a matter of fact, I added the put data verification in `DistributedTransactionAdminRepairTableIntegrationTestBase` because I wanted to verify that the coordinator table metadata were present in the metadata table after the repair. In retrospect, I could have used `storageAdmin.getTableMetadata(coordinatorNamespace, coordinatorTable)` instead which is maybe more appropriate ?

In `DistributedStorageAdminRepairTableIntegrationTestBase`, I thought it was not necessary to add the put data verification as I could check the metadata for the table were present in the metadata table using the admin. If the metadata for the table is present, then the put data verification is superfluous maybe ?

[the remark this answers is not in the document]
brfrn169: Okay, I see. The reason why I thought we could put data in `DistributedStorageAdminRepairTableIntegrationTestBase` is that I thought putting data could check the stored procedure things for Cosmos DB Tables. But on second thought, we already have the tests for the stored procedure things in  `CosmosAdminRepairTableIntegrationTest` and `ConsensusCommitAdminRepairTableIntegrationTestWithCosmos`, so we don't need to put data.

> In retrospect, I could have used storageAdmin.getTableMetadata(coordinatorNamespace, coordinatorTable) instead which is maybe more appropriate?

Maybe yes, using `storageAdmin.getTableMetadata(coordinatorNamespace, coordinatorTable)` is straightforward to me.

core/src/integration-test/java/com/scalar/db/util/CosmosAdminTestUtils.java
Torch3333: Yes it would be but I don't think it is possible with the current integration-testing package structure since the factory like `AdminTestUtils.create()` method reference all storage implementation classes.

Let's say each `XXXAdminTestUtils` classes are moved to their respective `com.scalar.db.storage.xxx` storage package. Because of the `AdminTestUtils.create` method, all storage implementations are imported in `AdminTestUtils`. Because of this, when running the integration test in `com.scalar.db.storage.jdbc` that are using `com.scalar.db.util.AdminTestUtils`, it requires `com.scalar.db.storage.dynamo.DynamoAdminTestUtils` and others storage implementation to be present in the classpath, which is not the case because storage test package are compartmentalized from each other. 
If possible, I would like to keep the `AdminTestUtils.create` method as it simplify the instantiation. 
What do you think ?

core/src/integration-test/java/com/scalar/db/api/DistributedStorageAdminRepairTableIntegrationTestBase.java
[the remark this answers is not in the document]
Torch3333: All right, I will update `ConsensusCommitAdminRepairTableIntegrationTestWithCosmos` to rely on `storageAdmin.getTableMetadata(coordinatorNamespace, coordinatorTable)` and remove the put data.

core/src/integration-test/java/com/scalar/db/util/CosmosAdminTestUtils.java
brfrn169: I understand. Let's keep it as it is. Thanks.

core/src/integration-test/java/com/scalar/db/api/DistributedStorageAdminRepairTableIntegrationTestBase.java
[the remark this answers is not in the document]
Torch3333: I revised it in https://github.com/scalar-labs/scalardb/pull/613/commits/aa6395429ffd10b52eef7715bc36e0c8dc491a39 and https://github.com/scalar-labs/scalardb/pull/613/commits/c5beaab6ab8aed61f43c9474f107c17120a005aa

server/src/integration-test/java/com/scalar/db/server/ServerEnv.java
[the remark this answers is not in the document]
Torch3333: I revised it in https://github.com/scalar-labs/scalardb/pull/613/commits/aa6395429ffd10b52eef7715bc36e0c8dc491a39

server/src/integration-test/java/com/scalar/db/server/DistributedTransactionRepairTableAdminServiceIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.scalar.db.api.DistributedTransactionAdminRepairTableIntegrationTestBase;
+import com.scalar.db.config.DatabaseConfig;
+import java.io.IOException;
+import java.util.Properties;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIf;
+
+public class DistributedTransactionRepairTableAdminServiceIntegrationTest
```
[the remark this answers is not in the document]
Torch3333: Thank you, I revised it in https://github.com/scalar-labs/scalardb/pull/613/commits/aa6395429ffd10b52eef7715bc36e0c8dc491a39

server/src/integration-test/java/com/scalar/db/server/DistributedStorageAdminRepairTableServiceIntegrationTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.scalar.db.api.DistributedStorageAdminRepairTableIntegrationTestBase;
+import com.scalar.db.config.DatabaseConfig;
+import java.io.IOException;
+import java.util.Properties;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIf;
+
+public class DistributedStorageAdminRepairTableServiceIntegrationTest
```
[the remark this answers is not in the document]
Torch3333: Thank you, I revised it in https://github.com/scalar-labs/scalardb/pull/613/commits/aa6395429ffd10b52eef7715bc36e0c8dc491a39

core/src/integration-test/java/com/scalar/db/util/JdbcAdminTestUtils.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public void truncateMetadataTable() throws Exception {
+    String truncateTableStatement =
+        "TRUNCATE TABLE "
+            + QueryUtils.enclosedFullTableName(metadataNamespace, metadataTable, rdbEngine);
+    execute(truncateTableStatement);
+  }
+
+  @SuppressFBWarnings("OBL_UNSATISFIED_OBLIGATION")
```
brfrn169, line 52: I don't think we will need this after merging the latest changes from master into this branch.

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminRepairTableIntegrationTestBase.java
```diff
@@ -0,0 +1,175 @@
...
+    Properties properties = TestUtils.addSuffix(getStorageProperties(), TEST_NAME);
+    String coordinatorNamespace =
+        new ConsensusCommitConfig(new DatabaseConfig(properties))
+            .getCoordinatorNamespace()
+            .orElse(Coordinator.NAMESPACE);
+    String coordinatorTable = Coordinator.TABLE;
+    // Use the DistributedStorageAdmin instead of the DistributedTransactionAdmin because the latter
+    // expects the
+    // table to hold transactional table metadata columns which is not the case for the coordinator
+    // table
```
brfrn169, line 167:
```suggestion
    // Use the DistributedStorageAdmin instead of the DistributedTransactionAdmin because the latter
    // expects the table to hold transactional table metadata columns which is not the case for the
    // coordinator table
```

core/src/integration-test/java/com/scalar/db/api/DistributedStorageAdminRepairTableIntegrationTestBase.java
```diff
@@ -0,0 +1,139 @@
...
+    adminTestUtils = AdminTestUtils.create(TestUtils.addSuffix(getStorageProperties(), TEST_NAME));
+  }
+
+  @AfterEach
+  protected void afterEach() throws Exception {
+    dropTable();
+  }
+
+  @AfterAll
+  protected void afterAll() throws Exception {}
```
brfrn169, line 112: We should close `admin` just in case.
```suggestion
  protected void afterAll() throws Exception {
    admin.close();
  }
```

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminRepairTableIntegrationTestBase.java
```diff
@@ -0,0 +1,175 @@
...
+    adminTestUtils = AdminTestUtils.create(TestUtils.addSuffix(getStorageProperties(), TEST_NAME));
+  }
+
+  @AfterEach
+  protected void afterEach() throws Exception {
+    dropTable();
+  }
+
+  @AfterAll
+  protected void afterAll() throws Exception {}
```
brfrn169, line 118: Diito.
```suggestion
  protected void afterAll() throws Exception {
    admin.close();
  }
```

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminRepairTableIntegrationTestBase.java
```diff
@@ -0,0 +1,175 @@
...
+    // expects the
+    // table to hold transactional table metadata columns which is not the case for the coordinator
+    // table
+    DistributedStorageAdmin storageAdmin =
+        StorageFactory.create(TestUtils.addSuffix(getStorageProperties(), TEST_NAME))
+            .getStorageAdmin();
+
+    assertThat(storageAdmin.getTableMetadata(coordinatorNamespace, coordinatorTable))
+        .isEqualTo(Coordinator.TABLE_METADATA);
+  }
```
brfrn169, line 174: Ditto.
```suggestion

    storageAdmin.close();
  }
```

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminRepairTableIntegrationTestBase.java
```diff
@@ -0,0 +1,175 @@
...
+    assertThat(admin.tableExists(getNamespace(), TABLE)).isTrue();
+    assertThat(admin.getTableMetadata(getNamespace(), TABLE)).isEqualTo(TABLE_METADATA);
+    assertTableMetadataForCoordinatorTableArePresent();
+  }
+
+  private void assertTableMetadataForCoordinatorTableArePresent() throws Exception {
+    // There is no coordinator table when using JdbcTransaction
+    if (getStorageProperties().getProperty(DatabaseConfig.TRANSACTION_MANAGER, "").equals("jdbc")) {
+      return;
+    }
```
brfrn169, line 157: This is an implementation specific thing, so I don't think we should put it here.

One possible idea is we can add something like the following method:
```java
  protected boolean hasCoordinatorTables() {
    return true;
  }
```

And then, we can assert the coordinator table metadata when `hasCoordinatorTables()` returns `true`:
```java
  if (hasCoordinatorTables()) {
    assertTableMetadataForCoordinatorTableArePresent();
  }
```

And we can override the method in `JdbcTransactionAdminRepairTableIntegrationTest` and make it return `false`.

What do you think?

brfrn169 (commented):
LGTM overall. Left several minor comments. Please take a look when you have time!

core/src/integration-test/java/com/scalar/db/util/JdbcAdminTestUtils.java
Torch3333: Ok, I updated the `spotbugs-exclude.xml` file to add a matching file regex for `XXXTestUtils.java` files.

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminRepairTableIntegrationTestBase.java
Torch3333: Thank you, I revised it.

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminRepairTableIntegrationTestBase.java
Torch3333: Thank you, I revised it.

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminRepairTableIntegrationTestBase.java
Torch3333: Thank you, I revised it.

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminRepairTableIntegrationTestBase.java
Torch3333: Thank you. I implemented your suggestion.

core/src/integration-test/java/com/scalar/db/util/JdbcAdminTestUtils.java
brfrn169: Oh sorry, I thought we didn't need to update `spotbugs-exclude.xml` to suppress the warning.

brfrn169 (approved):
LGTM! Thank you!

PR #614, opened 2022-06-23
Add some useful methods

brfrn169:
While creating API guide, I found several methods that it would be better if we have. This PR adds those. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/GetBuilder.java
+++ b/core/src/main/java/com/scalar/db/api/GetBuilder.java
@@ -60,16 +60,16 @@ public class GetBuilder {
     }
 
     @Override
-    public Buildable projection(String projection) {
-      checkNotNull(projection);
-      projections.add(projection);
+    public Buildable clusteringKey(Key clusteringKey) {
+      checkNotNull(clusteringKey);
+      this.clusteringKey = clusteringKey;
       return this;
     }
 
     @Override
-    public Buildable clusteringKey(Key clusteringKey) {
-      checkNotNull(clusteringKey);
-      this.clusteringKey = clusteringKey;
+    public Buildable projection(String projection) {
+      checkNotNull(projection);
+      projections.add(projection);
       return this;
     }
 
@@ -162,6 +162,12 @@ public class GetBuilder {
 
     @Override
     public BuildableFromExisting projections(Collection<String> projections) {
+      super.projections(projections);
+      return this;
+    }
+
+    @Override
+    public BuildableFromExisting projections(String... projections) {
       super.projections(projections);
       return this;
     }
--- a/core/src/main/java/com/scalar/db/api/GetBuilder.java
+++ b/core/src/main/java/com/scalar/db/api/GetBuilder.java
@@ -9,6 +9,7 @@ import com.scalar.db.api.OperationBuilder.Projection;
 import com.scalar.db.api.OperationBuilder.TableBuilder;
 import com.scalar.db.io.Key;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import javax.annotation.Nullable;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/scalar/db/api/GetBuilder.java
```diff
@@ -79,6 +80,11 @@ public Buildable projections(Collection<String> projections) {
       return this;
     }
 
+    @Override
+    public Buildable projections(String... projections) {
+      return projections(Arrays.asList(projections));
+    }
```
Torch3333, line 86: In the `Get.BuildableFromExisting` class, I think it is necessary to override this method to make it return the `this` object of type `Get.BuildableFromExisting`. Otherwise, while calling this method from the `Get.BuildableFromExisting` object, it will return a `Get.Buildable` object.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -90,6 +91,11 @@ public BuildableScan projections(Collection<String> projections) {
       return this;
     }
 
+    @Override
+    public BuildableScan projections(String... projections) {
+      return projections(Arrays.asList(projections));
+    }
+
```
Torch3333, line 98: I think we need to override this method in `Scan.BuildableScanOrScanAllFromExisting` to make it return a `Scan.BuildableScanOrScanAllFromExisting` object.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -103,6 +109,18 @@ public BuildableScan ordering(Scan.Ordering ordering) {
...
+    public BuildableScan orderings(Collection<Scan.Ordering> orderings) {
+      checkNotNull(orderings);
+      this.orderings.addAll(orderings);
+      return this;
+    }
+
+    @Override
+    public BuildableScan orderings(Scan.Ordering... orderings) {
+      return orderings(Arrays.asList(orderings));
+    }
```
Torch3333, line 122: Same here.

core/src/main/java/com/scalar/db/api/GetBuilder.java
```diff
@@ -79,6 +80,11 @@ public Buildable projections(Collection<String> projections) {
       return this;
     }
 
+    @Override
+    public Buildable projections(String... projections) {
```
Torch3333, line 84: Is it ok not to add unit tests for the changes to the builders?

Torch3333 (commented):
Thank you for you adding these new useful methods. 
I left a few comments regarding the builders changes, please have a look.

core/src/main/java/com/scalar/db/api/GetBuilder.java
brfrn169: No, I should add unit tests. I will do that. Thank you!

brfrn169 (commented):
Thank you for taking a look at this!

I will revise it based on your feedback. Thanks.

core/src/main/java/com/scalar/db/api/GetBuilder.java
brfrn169: Added unit tests in the latest commit.

core/src/main/java/com/scalar/db/api/GetBuilder.java
brfrn169: Fixed in the latest commit.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
brfrn169: Fixed in the latest commit.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
brfrn169: Fixed in the latest commit.

brfrn169 (commented):
@Torch3333 I fixed it based on your feedback. Please take a look again when you have time!

Torch3333 (approved):
Thank for the revision. LGTM !

feeblefakie (approved):
LGTM! Thank you!

PR #617, opened 2022-06-25
Update Getting Started

brfrn169:
Please take a look!


docs/getting-started-with-scalardb-on-jdbc.md
```diff
@@ -69,3 +69,13 @@ scalar.db.storage=jdbc
 ```
 
 Please follow [Getting Started with Scalar DB](getting-started-with-scalardb.md) to run the application.
+
+## Use JDBC transaction
```
brfrn169, line 73: Moved JDBC transaction thing to this JDBC page.

docs/getting-started-with-scalardb.md
```diff
@@ -6,36 +6,34 @@ If you haven't done it, please configure them first by following [this](getting-
 ## Build
```
brfrn169, line 6: Deleted the Storage API thing in Getting Started because users usually don't use Storage API. Users usually use Transaction API only.

However, we have some use cases for Storage API mainly for performance reasons. So, I will add another document for Storage API (maybe for experts) in another PR.

docs/getting-started-with-scalardb.md
```diff
@@ -235,35 +170,67 @@ public class ElectronicMoneyWithTransaction extends ElectronicMoney {
...
+```shell
+$ ./gradlew run --args="-action getBalance -id user1"
 ```
 
-You don't need to set a key `transaction` to `true` in Scalar DB schema for the native transaction manager.
-So you can use the same schema file as **emoney-storage.json**.
+- Get the balance of `merchant1`:
+```shell
+$ ./gradlew run --args="-action getBalance -id merchant1"
+```
```
brfrn169, line 233: Also added the `getBalance` action that's a read only transaction to check the balance of a specified user.

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

kota2and3kan (approved):
LGTM! Thank you!

I was able to retrieve the balances!
```shell
# ./gradlew run --args="-action getBalance -id user1"

> Task :run
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
The balance for user1 is 900

BUILD SUCCESSFUL in 1s
```
I will update the [Getting started guide of Helm Charts](https://github.com/scalar-labs/helm-charts/blob/main/docs/getting-started-scalardb.md) since the guide explains the way to retrieve records from a backend database directly.

brfrn169:
Merging this.

@scalar-boney Please check if this change affects any QA things when you have time.

PR #618, opened 2022-06-26
Introduce GetWithIndex and ScanWithIndex operations

brfrn169:
This PR introduces `GetWithIndex` and `ScanWithIndex` operations that are read operations using secondary indexes.

We can build `GetWithIndex` and `ScanWithIndex` operations with the builder, and after this change, we can execute read operations using secondary indexes as follows:
```java
Get get =
    Get.newBuilder()
        .namespace("ns")
        .table("tbl")
        .indexKey(Key.ofInt("<index key column>", 0)) // specify the index key
        .build();
Optional<Result> result = transaction.get(get);

Scan scan =
    Scan.newBuilder()
        .namespace("ns")
        .table("tbl")
        .indexKey(Key.ofInt("<index key column>", 0)) // specify the index key
        .build();
List<Result> results = transaction.scan(scan);
```

We can already execute read operations using secondary indexes with the existing `Get` and `Scan` actually, but I think using the `GetWithIndex` and `ScanWithIndex` operations is easier to use and more readable.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/IndexGet.java
+++ b/core/src/main/java/com/scalar/db/api/IndexGet.java
@@ -1,0 +1,105 @@
+package com.scalar.db.api;
+
+import com.google.common.base.MoreObjects;
+import com.scalar.db.io.Key;
+import java.util.Collection;
+import java.util.Objects;
+
+public class IndexGet extends Get {
+
+  /**
+   * Constructs an {@code IndexGet} with the specified index {@code Key}.
+   *
+   * @param indexKey an index key
+   * @deprecated As of release 3.6.0. Will be removed in release 5.0.0. Use {@link Get#newBuilder()}
+   *     instead
+   */
+  @Deprecated
+  @SuppressWarnings("InlineMeSuggester")
+  public IndexGet(Key indexKey) {
+    super(indexKey);
+  }
+
+  /**
+   * Copy a IndexGet.
+   *
+   * @param indexGet an IndexGet
+   * @deprecated Use {@link Get#newBuilder(Get)} instead
+   */
+  @Deprecated
+  @SuppressWarnings("InlineMeSuggester")
+  public IndexGet(IndexGet indexGet) {
+    super(indexGet);
+  }
+
+  /**
+   * @deprecated As of release 3.6.0. Will be removed in release 5.0.0. Use the setter method of the
+   *     Get builder instead; to create a Get builder, use {@link Get#newBuilder()}
+   */
+  @Deprecated
+  @Override
+  public IndexGet forNamespace(String namespace) {
+    return (IndexGet) super.forNamespace(namespace);
+  }
+
+  /**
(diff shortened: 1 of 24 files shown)


core/src/main/java/com/scalar/db/api/IndexGet.java
```diff
@@ -0,0 +1,105 @@
+package com.scalar.db.api;
+
+import com.google.common.base.MoreObjects;
+import com.scalar.db.io.Key;
+import java.util.Collection;
+import java.util.Objects;
+
+public class IndexGet extends Get {
```
brfrn169, line 8: Added `IndexGet` that extends `Get`.

core/src/main/java/com/scalar/db/api/IndexScan.java
```diff
@@ -0,0 +1,174 @@
+package com.scalar.db.api;
+
+import com.google.common.base.MoreObjects;
+import com.scalar.db.io.Key;
+import java.util.Collection;
+import java.util.Objects;
+
+public class IndexScan extends Scan {
```
brfrn169, line 8: Also added `IndexScan` that extends `Scan`.

Torch3333 (approved):
Basically, LGTM, thank you!
I just left comments regarding naming and typos.

core/src/main/java/com/scalar/db/api/IndexScan.java
brfrn169: Thank you for the suggestion. I agree. I will rename `IndexScan` to `ScanWithIndex` and `IndexGet` to `GetWithIndex`

brfrn169: @Torch3333 I renamed them in the latest commit. Please take a look again when you have time!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!
Left a minor suggestion but it doesn't block it from being merged.

core/src/main/java/com/scalar/db/storage/common/checker/OperationChecker.java
```diff
@@ -53,21 +66,40 @@ public void check(Get get) throws ExecutionException {
...
+      check((ScanAll) scan);
+      return;
+    }
+
     TableMetadata metadata = getMetadata(scan);
 
     checkProjections(scan, metadata);
 
     if (ScalarDbUtils.isSecondaryIndexSpecified(scan, metadata)) {
+      if (scan.getPartitionKey().size() != 1) {
```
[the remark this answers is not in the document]
brfrn169: @feeblefakie I tried making the change you mentioned as I thought it was a nice idea. But currently, even if we use `Get`, `isSecondaryIndexSpecified()` can return `true` when specifying the index key as a partition key with `Get`. So I will leave it as it is for now. We can do the refactoring in the future after we delete the existing way of using a secondary index.

PR #619, opened 2022-06-28
Add Scanner paging support for scan operations to the Cosmos adapter

Torch3333:
This updates the Cosmos Adapter for Scan operation. Up until now, the `Scanner` object, returned by performing a Scan,  downloaded all at once the records eagerly. Now, the Scanner retrieves the records on demand page by page.

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cosmos/ScannerImpl.java
+++ b/core/src/main/java/com/scalar/db/storage/cosmos/ScannerImpl.java
@@ -2,10 +2,12 @@ package com.scalar.db.storage.cosmos;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.azure.cosmos.models.FeedResponse;
 import com.scalar.db.api.Result;
 import com.scalar.db.api.Scanner;
 import com.scalar.db.storage.common.ScannerIterator;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java
```diff
@@ -89,22 +98,16 @@ private List<Record> executeRead(Get get, TableMetadata tableMetadata) throws Co
                 DSL.field("r.id").eq(cosmosOperation.getId()))
             .getSQL(ParamType.INLINED);
 
-    return getContainer(get).queryItems(query, new CosmosQueryRequestOptions(), Record.class)
-        .stream()
-        .collect(Collectors.toList());
+    return executeQuery(get, tableMetadata, query);
```
brfrn169, line 101: I'm not sure but should we specify a partition key here?
```suggestion
    return executeQuery(get, tableMetadata, query, new CosmosQueryRequestOptions().setPartitionKey(cosmosOperation.getCosmosPartitionKey()));
```

brfrn169 (commented):
Thank you. LGTM overall. Left a question. Please take a look when you have time!

core/src/main/java/com/scalar/db/storage/cosmos/SelectStatementHandler.java
Torch3333: In the sdk [documentation](https://docs.microsoft.com/en-us/azure/cosmos-db/sql/sql-api-query-metrics#about-sql-query-execution) it says 

> When you issue a query to Azure Cosmos DB, the SDK performs these logical steps:
> ...
>     If the query includes a filter against the partition key, like SELECT * FROM c WHERE c.city = "Seattle", **it is routed to a single partition**. If the query does not have a the filter on partition key, then it is executed in all partitions, and results are merged client side.
> ...

So I assume that setting the partition key via the query options is redundant for this query as the where clause already filters on the partition key. This means the query will be routed to a single partition with or without the `CosmosQueryRequestOptions().setPartitionKey` anyway. 

Even though it does not constitute a benchmark, I executed from my laptop to the cosmos integration test machine the query with the added `CosmosQueryRequestOptions().setPartitionKey(cosmosOperation.getCosmosPartitionKey()` and did not notice a significant performance gain or loss compared to the original. A real benchmark would be necessary to see if it improves anything.

brfrn169: I see. So we don't need to specify a partition key. Thank you for the investigation!

brfrn169 (approved):
LGTM! Thank you! Left very minor comments. Please take a look when you have time!

feeblefakie (approved):
LGTM! nice work! Thank you!

core/src/main/java/com/scalar/db/storage/cosmos/ScannerImpl.java
```diff
@@ -14,34 +16,61 @@
...
   @Nonnull
   public Optional<Result> one() {
-    if (records.isEmpty()) {
-      return Optional.empty();
+    while (currentPageRecords.hasNext() || recordsPages.hasNext()) {
+      // Return the next record of the current page if there is one
+      if (currentPageRecords.hasNext()) {
+        Record currentRecord = currentPageRecords.next();
+        Optional<Result> ret = Optional.of(resultInterpreter.interpret(currentRecord));
+        currentPageRecords.remove();
```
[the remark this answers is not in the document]
Torch3333: No, I did not think this through enough. I will remove it.

core/src/main/java/com/scalar/db/storage/cosmos/ScannerImpl.java
```diff
@@ -14,34 +16,61 @@
...
+        return ret;
+        // Otherwise, advance to the next page
+      } else {
```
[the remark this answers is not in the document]
Torch3333: Yes!

Torch3333:
It appears the PR was merged but I still have a few changes to apply so I will do it in another PR.

PR #621, opened 2022-06-29
Add documentation for Consensus Commit configurations

brfrn169:
Please take a look!

### DIFF
--- a/docs/configurations-for-consensus-commit.md
+++ b/docs/configurations-for-consensus-commit.md
@@ -11,44 +11,24 @@ The following sections describes the configurations for Consensus Commit.
 
 ## Basic configurations
 
-The basic configurations for Consensus Commit is as follows:
+The basic configurations for Consensus Commit are as follows:
 
-```properties
-# Default isolation level for ConsensusCommit. Either SNAPSHOT or SERIALIZABLE can be specified. SNAPSHOT is used by default.
-scalar.db.consensus_commit.isolation_level=SNAPSHOT
+| name | description | default |
+| ---- | ----------- | ------- |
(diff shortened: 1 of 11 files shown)


docs/design.md
```diff
@@ -1,4 +1,4 @@
-## Scalar DB design document
+# Scalar DB design document
```
brfrn169, line 1: This PR includes some refactoring things for the existing documents.

Torch3333 (approved):
LGTM, thank you!

docs/configurations-for-consensus-commit.md
```diff
@@ -0,0 +1,54 @@
...
+The basic configurations for Consensus Commit is as follows:
+
+```properties
+# Default isolation level for ConsensusCommit. Either SNAPSHOT or SERIALIZABLE can be specified. SNAPSHOT is used by default.
+scalar.db.consensus_commit.isolation_level=SNAPSHOT
+
+# Default serializable strategy for ConsensusCommit transaction manager.
+# Either EXTRA_READ or EXTRA_WRITE can be specified. EXTRA_READ is used by default.
+# If SNAPSHOT is specified, this is ignored.
+scalar.db.consensus_commit.serializable_strategy=EXTRA_READ
```
feeblefakie, line 23: The comment explains but it’s a bit confusing to see both SNAPSHOT and EXTRA_READ are used in the configuration.
Should it explain each configuration separately?
How about making it a table-format like this? (can be simpler than this, though)
https://kafka.apache.org/documentation/#brokerconfigs

feeblefakie (commented):
The contents are looking good! Thank you!
I left a suggestion about the format. PTAL!

docs/configurations-for-consensus-commit.md
brfrn169: I fixed it in the latest commit based on your feedback. Please take a look again when you have time!

feeblefakie (approved):
LGTM! Thank you!

PR #632, opened 2022-07-07
Rename `transactional` to `transaction`

brfrn169:
Currently, we have used the two terms `transactional` and `transaction` mixed together. Though I don't think we need to fix it in the documentation, we should fix it in the code. This PR does that. Please take a look!


brfrn169:
@Torch3333 Merging this for now. Please take a look when you gat a chance!

PR #638, opened 2022-07-12
Add capability to add a new column to an existing table in the Administration interface

Torch3333:
This adds a new method to the Admin interface to be able to add a new non-primary-key column to an existing table schema.


core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -345,4 +346,16 @@ default boolean tableExists(String namespace, String table) throws ExecutionExce
...
+   * Add a new column to an existing table. The new column cannot be a partition or clustering key
+   *
+   * @param namespace the table namespace
+   * @param table the table name
+   * @param columnName the name of the new column
+   * @param columnType the type of the new column
+   * @throws ExecutionException if the operation failed
+   */
+  void addNewColumnToTable(String namespace, String table, String columnName, DataType columnType)
+      throws ExecutionException;
```
Torch3333, line 360: This adds this new method to the Admin interface.

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

BTW, I think most NoSQL databases can apply alter table for adding a column pretty quickly since that's one of the benefits of (schema-less) NoSQL.
However, RDBs might take a very long time to alter table or they might (or might not) take a table-level lock, so it shouldn't be done without planning, especially DBs are running in production and big.
Should we note such points somewhere?

Torch3333:
> LGTM! Thank you!
> 
> BTW, I think most NoSQL databases can apply alter table for adding a column pretty quickly since that's one of the benefits of (schema-less) NoSQL. However, RDBs might take a very long time to alter table or they might (or might not) take a table-level lock, so it shouldn't be done without planning, especially DBs are running in production and big. Should we note such points somewhere?



> LGTM! Thank you!
> 
> BTW, I think most NoSQL databases can apply alter table for adding a column pretty quickly since that's one of the benefits of (schema-less) NoSQL. However, RDBs might take a very long time to alter table or they might (or might not) take a table-level lock, so it shouldn't be done without planning, especially DBs are running in production and big. Should we note such points somewhere?

Indeed, that is a good point. 
In other PR, I will add documentation mentioning your point for this new method in the [API guide](https://github.com/scalar-labs/scalardb/blob/master/docs/api-guide.md) and I will also update the method Javadoc as well.

PR #641, opened 2022-07-16
Deprecate table mapping in Multi-storage

brfrn169:
This PR deprecates table mapping in Multi-storage. I do this because there are some cases where the table mapping is ambiguous.

For example, let's say we have the following two table mappings:

`ns1.tbl1` -> `mysql`
`ns1.tbl2` -> `cassandra`

And if we create a namespace `ns1`, we are not sure if we should create the namespace for `mysql` or `cassandra`. In that case, we can create the namespace for both `mysql` and `cassandra`, of course. But rather than that complexity, I think we can deprecate the table mapping. So after that, Multi-storage only supports the namespace mapping, which I think would be enough when I imagine the use cases.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #642, opened 2022-07-17
Change default value of is-transaction-table in Schema Loader

brfrn169:
Currently, if we don't specify the `transaction` field in a table definition in a schema file, the table is considered a non-transaction table. But I think the default should be a transaction table in Scalar DB because Scalar DB users usually use Transaction API and Storage API is an advanced feature for experts. This PR changes the default. Please take a look!

### DIFF
--- a/schema-loader/README.md
+++ b/schema-loader/README.md
@@ -1,7 +1,7 @@
 # Scalar DB Schema Loader
 
 Scalar DB Schema Loader creates, deletes and repairs Scalar DB schemas (namespaces and tables) on the basis of a provided schema file.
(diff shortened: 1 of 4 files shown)


Torch3333 (commented):
Thank you, it looks good overall. 
Should the schema loader readme be updated as well?

brfrn169:
@Torch3333 Thank you for the feedback. I just modified the README in the latest commit. Please take a look again when you have time!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #643, opened 2022-07-19
Add documentation for admin.addNewColumnToTable()

Torch3333:
This adds documentation for the `Admin.addNewColumnToTable` in the API guide and updates the javadoc.

### DIFF
--- a/core/src/main/java/com/scalar/db/api/Admin.java
+++ b/core/src/main/java/com/scalar/db/api/Admin.java
@@ -348,7 +348,21 @@ public interface Admin {
       throws ExecutionException;
 
   /**
-   * Add a new column to an existing table. The new column cannot be a partition or clustering key
+   * Add a new column to an existing table. The new column cannot be a partition or clustering key.
+   * <br>
+   * <br>
+   * <strong>Warning :</strong> this should be executed with significant consideration as the
+   * execution time may vary greatly depending on the underlying storage. Please plan accordingly
+   * especially if the database runs in production:
+   *
+   * <ul>
+   *   <li>for Cosmos and Dynamo DB: this operation is almost instantaneous as the table schema is
+   *       not modified. Only the table metadata stored in a separated table are updated.
+   *   <li>for Cassandra : adding a column is a constant-time operation based on the amount of data
+   *       in the table.
+   *   <li>for relational databases (Mysql, Oracle, etc.): it may take a very long time to execute
+   *       and a table-lock may be performed.
+   * </ul>
    *
    * @param namespace the table namespace
    * @param table the table name
--- a/core/src/main/java/com/scalar/db/api/Admin.java
+++ b/core/src/main/java/com/scalar/db/api/Admin.java
@@ -358,9 +358,11 @@ public interface Admin {
    * <ul>
    *   <li>for Cosmos and Dynamo DB: this operation is almost instantaneous as the table schema is
    *       not modified. Only the table metadata stored in a separated table are updated.
-   *   <li>for Cassandra : adding a column is a constant-time operation based on the amount of data
-   *       in the table.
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -348,7 +348,21 @@ void repairTable(
...
+   * <strong>Warning :</strong> this should be executed with significant consideration as the
+   * execution time may vary greatly depending on the underlying storage. Please plan accordingly
+   * especially if the database runs in production:
+   *
+   * <ul>
+   *   <li>for Cosmos and Dynamo DB: this operation is almost instantaneous as the table schema is
+   *       not modified. Only the table metadata stored in a separated table are updated.
+   *   <li>for Cassandra : adding a column is a constant-time operation based on the amount of data
+   *       in the table.
+   *   <li>for relational databases (Mysql, Oracle, etc.): it may take a very long time to execute
```
brfrn169, line 363:
```suggestion
   *   <li>for relational databases (MySQL, Oracle, etc.): it may take a very long time to execute
```

brfrn169 (commented):
Thank you. I left several minor suggestions. Please take a look when you have time!

Also, can you please add the same documentation to `storage-abstraction.md` as well?

core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -348,7 +348,21 @@ void repairTable(
...
+   * <br>
+   * <br>
+   * <strong>Warning :</strong> this should be executed with significant consideration as the
+   * execution time may vary greatly depending on the underlying storage. Please plan accordingly
+   * especially if the database runs in production:
+   *
+   * <ul>
+   *   <li>for Cosmos and Dynamo DB: this operation is almost instantaneous as the table schema is
+   *       not modified. Only the table metadata stored in a separated table are updated.
+   *   <li>for Cassandra : adding a column is a constant-time operation based on the amount of data
```
feeblefakie, line 361: Is it a constant time or linear time?
It says constant time but it also says the time is based on the amount of data, which sounds like a linear time.

feeblefakie (commented):
Thank you!
Left a clarifying question.

Also, can you also add the warning message to the schema loader doc after this PR is merged?
(maybe we can add an interactive prompt to ask users if they know what they are doing)

core/src/main/java/com/scalar/db/api/Admin.java
Torch3333: Indeed, it is ambiguous. I copied this from the official Cassandra [documentation](https://cassandra.apache.org/doc/latest/cassandra/cql/ddl.html#alter-table-statement). 
I haven't found online any other information besides this about the add column performance. 
I added a [question](https://stackoverflow.com/questions/73060026/is-altering-a-table-to-add-a-column-a-constant-time-or-linear-time-operation) on StackOverflow to ask about it.

Torch3333: My StackOverflow question received a detailed [answer](https://stackoverflow.com/a/73060202/9538402) which clarified things. 

What do you think about updating the text as follows?
```suggestion
   *   <li>for Cassandra: adding a column will only update the schema metadata and do not modify existing schema records. The cluster topology is the main factor for the execution time. Since the schema metadata change propagates to each cluster node via a gossip protocol, the larger the cluster, the longer it will take for all nodes to be updated.

feeblefakie: Looking good. Thank you!

feeblefakie (approved):
LGTM! Thank you!

Torch3333:
I revised the documentation according to your comments and added it as well to the `storage-abstraction.md`. 
Please have a look, thank you.

brfrn169 (approved):
LGTM! Thank you!

PR #645, opened 2022-07-26
In the Admin create and drop index methods, alter column type if necessary

Torch3333:
**Problem**
Creating a secondary index using the `Admin.createIndex()` will fail for the storage and Scalar DB column type combination below because the current native RDB column type is not compatible with becoming an index.
- MySQL:  TEXT and BLOB
- PostgreSQL : TEXT
- Oracle: TEXT and BLOB

**Implemented fixes**
- Update the `JdbcAdmin.createIndex()` implementation to alter the column type if necessary to be compatible with a secondary index and then create the index. If the column holds data whose size is too big for the new altered type, an exception will be thrown.
- When an index is deleted using `JdbcAdmin.dropIndex()` the column type will be altered from the index compatible type back to the regular column type if necessary.
- Since Oracle does not natively support the column type alteration from BLOB (Scalar DB BLOB regular column type) to RAW(64) (Scalar DB BLOB primary key/index column type) which is something we wanted to be able to do if a column were to converted to a secondary index. We decided to change the regular default column type from `BLOB` to `RAW(2000)`.

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
+++ b/core/src/integration-test/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
@@ -66,7 +66,7 @@ public abstract class DistributedStorageAdminIntegrationTestBase {
           .addSecondaryIndex(COL_NAME5)
           .addSecondaryIndex(COL_NAME6)
           .build();
-
+  private StorageFactory storageFactory;
   private DistributedStorageAdmin admin;
   private DistributedStorage storage;
   private String namespace1;
@@ -76,13 +76,13 @@ public abstract class DistributedStorageAdminIntegrationTestBase {
   @BeforeAll
   public void beforeAll() throws Exception {
     initialize();
-    StorageFactory factory = StorageFactory.create(TestUtils.addSuffix(getProperties(), TEST_NAME));
-    admin = factory.getAdmin();
+    storageFactory = StorageFactory.create(TestUtils.addSuffix(getProperties(), TEST_NAME));
+    admin = storageFactory.getAdmin();
     namespace1 = getNamespace1();
     namespace2 = getNamespace2();
     namespace3 = getNamespace3();
     createTables();
-    storage = factory.getStorage();
+    storage = storageFactory.getStorage();
   }
 
   protected void initialize() throws Exception {}
@@ -119,6 +119,7 @@ public abstract class DistributedStorageAdminIntegrationTestBase {
   public void afterAll() throws ExecutionException {
     dropTables();
     admin.close();
+    storage.close();
   }
 
   private void dropTables() throws ExecutionException {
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -777,19 +777,101 @@ private boolean hasDescClusteringOrder(TableMetadata metadata) {
...
+
     try (Connection connection = dataSource.getConnection()) {
+      alterToIndexColumnTypeIfNecessary(connection, namespace, table, columnName);
       createIndex(connection, namespace, table, columnName);
       updateTableMetadata(connection, namespace, table, columnName, true);
-    } catch (SQLException e) {
+    } catch (ExecutionException | SQLException e) {
       throw new ExecutionException("creating the secondary index failed", e);
     }
   }
```
Torch3333, line 788: The updated `JdbcAdmin.createIndex()`

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -777,19 +777,101 @@ private boolean hasDescClusteringOrder(TableMetadata metadata) {
...
+  }
+
   @Override
   public void dropIndex(String namespace, String table, String columnName)
       throws ExecutionException {
     try (Connection connection = dataSource.getConnection()) {
       dropIndex(connection, namespace, table, columnName);
+      alterToRegularColumnTypeIfNecessary(connection, namespace, table, columnName);
       updateTableMetadata(connection, namespace, table, columnName, false);
     } catch (SQLException e) {
```
Torch3333, line 876: The updated `JdbcAdmin.dropIndex()`

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -767,6 +767,11 @@ public void truncateTable(String namespace, String table) throws ExecutionExcept
   public void createIndex(
       String namespace, String table, String columnName, Map<String, String> options)
       throws ExecutionException {
+    if (getTableMetadata(namespace, table).getColumnDataType(columnName) == DataType.BOOLEAN) {
+      throw new IllegalArgumentException(
+          "Currently, BOOLEAN type is not supported for a secondary index in DynamoDB: "
+              + columnName);
+    }
```
Torch3333, line 775: I added this check because the message of the exception thrown by the Dynamo SDK is not clear enough boolean index are not supported.

core/src/integration-test/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
```diff
@@ -400,40 +401,155 @@ public void tableExists_ShouldReturnCorrectResults() throws ExecutionException {
...
       // Act
-      admin.createIndex(namespace1, TABLE4, COL_NAME7, options);
+      admin.createIndex(namespace1, TABLE4, COL_NAME2, options);
+      admin.createIndex(namespace1, TABLE4, COL_NAME3, options);
+      admin.createIndex(namespace1, TABLE4, COL_NAME4, options);
+      admin.createIndex(namespace1, TABLE4, COL_NAME5, options);
+      admin.createIndex(namespace1, TABLE4, COL_NAME6, options);
+      if (isIndexOnBooleanColumnSupported()) {
+        admin.createIndex(namespace1, TABLE4, COL_NAME7, options);
+      }
```
Torch3333, line 448: Since Dynamo DB does not support boolean index, I added this `if`.

core/src/integration-test/java/com/scalar/db/api/DistributedStorageIntegrationTestBase.java
```diff
@@ -95,6 +95,10 @@ private void createTable() throws ExecutionException {
         options);
   }
 
+  protected int getLargeDataSizeInBytes() {
+    return 5000;
+  }
```
Torch3333, line 100: Since the maximum blob size is limited to 2000 bytes for Oracle, I had to create an exception here.

schema-loader/README.md
```diff
@@ -356,15 +356,15 @@ By default, the schema loader enables auto-scaling of RU for all tables: RU is s
...
+| BIGINT | BIGINT | BIGINT | NUMBER(19)     | BIGINT |
 | TEXT | LONGTEXT | TEXT | VARCHAR2(4000) | VARCHAR(8000) |
-| FLOAT | DOUBLE | FLOAT | BINARY_FLOAT | FLOAT(24) |
-| DOUBLE | DOUBLE | DOUBLE PRECISION | BINARY_DOUBLE | FLOAT |
-| BOOLEAN | BOOLEAN | BOOLEAN | NUMBER(1) | BIT |
-| BLOB | LONGBLOB | BYTEA | BLOB | VARBINARY(8000) |
+| FLOAT | DOUBLE | FLOAT | BINARY_FLOAT   | FLOAT(24) |
+| DOUBLE | DOUBLE | DOUBLE PRECISION | BINARY_DOUBLE  | FLOAT |
+| BOOLEAN | BOOLEAN | BOOLEAN | NUMBER(1)      | BIT |
+| BLOB | LONGBLOB | BYTEA | RAW(2000)      | VARBINARY(8000) |
```
Torch3333, line 367: Since Oracle does not natively support the column type alteration from `BLOB` (Scalar DB BLOB regular column type) to `RAW(64)` (Scalar DB BLOB primary key/index column type) which is something we wanted to be able to do if a column were to converted to a secondary index. 
We decided to change the regular default column type from `BLOB` to `RAW(2000)`.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -777,19 +777,101 @@ private boolean hasDescClusteringOrder(TableMetadata metadata) {
...
+        getAlterColumnTypeStatement(namespace, table, columnName, indexedColumnType);
+
+    execute(connection, alterColumnStatement);
+  }
+
+  private void alterToRegularColumnTypeIfNecessary(
+      Connection connection, String namespace, String table, String columnName)
+      throws ExecutionException, SQLException {
+    ImmutableMap<DataType, String> typeMappingForIndexes = DATA_TYPE_MAPPING_FOR_KEY.get(rdbEngine);
+    DataType indexType = getTableMetadata(namespace, table).getColumnDataType(columnName);
```
brfrn169, line 812:
```suggestion
    DataType indexType = getTableMetadata(namespace, table).getColumnDataType(columnName);
    ImmutableMap<DataType, String> typeMappingForIndexes = DATA_TYPE_MAPPING_FOR_KEY.get(rdbEngine);
```

core/src/integration-test/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
```diff
@@ -66,7 +66,7 @@ public abstract class DistributedStorageAdminIntegrationTestBase {
           .addSecondaryIndex(COL_NAME5)
           .addSecondaryIndex(COL_NAME6)
           .build();
-
+  private StorageFactory storageFactory;
   private DistributedStorageAdmin admin;
   private DistributedStorage storage;
```
brfrn169, line 71: Can you please also get a storage instance in `truncateTable_ShouldTruncateProperly()` similar to the test methods you added/modified in this PR? So we can remove this `storage` field.

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminIntegrationTestBase.java
```diff
@@ -67,7 +67,7 @@ public abstract class DistributedTransactionAdminIntegrationTestBase {
           .addSecondaryIndex(COL_NAME5)
           .addSecondaryIndex(COL_NAME6)
           .build();
-
+  private TransactionFactory transactionFactory;
   private DistributedTransactionAdmin admin;
   private DistributedTransactionManager manager;
```
brfrn169, line 72: Ditto. Can you please also get a manager instance in `truncateTable_ShouldTruncateProperly()` similar to the test methods you added/modified in this PR? So we can remove this `manager ` field.

brfrn169 (dismissed):
Thank you. I left several minor suggestions/comments. Please take a look when you have time!

Torch3333:
@brfrn169 Thank you for your comments. I revised the PR accordingly, please have a look.

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -777,19 +777,101 @@ private boolean hasDescClusteringOrder(TableMetadata metadata) {
...
+    ImmutableMap<DataType, String> typeMappingForIndexes = DATA_TYPE_MAPPING_FOR_KEY.get(rdbEngine);
+    assert typeMappingForIndexes != null;
+
+    String indexedColumnType = typeMappingForIndexes.get(indexType);
+    if (indexedColumnType == null) {
+      // The column type does not need to be altered to be compatible with being secondary index
+      return;
+    }
+    String alterColumnStatement =
+        getAlterColumnTypeStatement(namespace, table, columnName, indexedColumnType);
```
feeblefakie, line 803: I feel that triggering alter table by creating an index might be a bit too tricky?
I think it is safer to make users decide instead of changing schema automatically.

feeblefakie (commented):
Thank you!
Left one comment on the spec.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
feeblefakie: @Torch3333 Sorry for the late feedback.
I discussed it with Toshi and now think it's OK to add.

feeblefakie (approved):
LGTM! Thank you!
(Discussed it with Toshi and understand the Pros and Cons, and now think it's OK to go. Sorry for making you wait)

PR #649, opened 2022-08-03
Add new configuration to include transaction metadata on Get and Scan operations results when using Consensus Commit

Torch3333:
By adding the following to the Scalar DB properties, when using Consensus Commit, `Scan` and `Get` operations results will contain transaction metadata. 
```
scalar.db.consensus_commit.include_metadata.enabled=true
```


brfrn169 (commented):
Thank you! LGTM overall. I left several minor comments. Please take a look when you have time!

One thing is, I'm wondering if the debug parameter (`scalar.db.debug`) should be consensus commit specific (in that case, maybe `scalar.db.consensus_commit.debug`) or not since the parameter affects only the consensus commit. If we add another debug feature enabled by `scalar.db.debug` in the future, we can justify having the global configuration `scalar.db.debug`, but if not, it might be better to make the parameter scope more narrow. What do you think?

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminIntegrationTestBase.java
```diff
@@ -69,24 +71,29 @@ public abstract class DistributedTransactionAdminIntegrationTestBase {
...
     namespace2 = getNamespace2();
     namespace3 = getNamespace3();
     createTables();
   }
 
-  protected void initialize() throws Exception {}
+  protected void initialize() throws Exception {
+    Properties debugProperties = TestUtils.addSuffix(getProperties(), TEST_NAME);
+    debugProperties.setProperty(DatabaseConfig.DEBUG, "true");
+    adminWithDebug = TransactionFactory.create(debugProperties).getTransactionAdmin();
```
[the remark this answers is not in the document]
Torch3333: First, thank you for all your suggestions.

Sorry, it's a bit confusing why the `adminWithDebug` initialization is in the `initialize` method only for this class. 
I put it here because in `DistributedTransactionAdminServiceIntegrationTest.initialize` the `adminWithDebug` instance needs to be configured specifically to target an instance of Scalar DB server with debug mode turned on. Since I am also creating a second server instance, I thought it was more appropriate to put it in `initialize` instead of having to overwrite the `beforeAll` method.
https://github.com/scalar-labs/scalardb/blob/849a58e2e90490d0e5080f182dd40d3848ee3b13/server/src/integration-test/java/com/scalar/db/server/DistributedTransactionAdminServiceIntegrationTest.java#L26-L40

Torch3333:
@brfrn169 
> One thing is, I'm wondering if the debug parameter (scalar.db.debug) should be consensus commit specific (in that case, maybe scalar.db.consensus_commit.debug) or not since the parameter affects only the consensus commit. If we add another debug feature enabled by scalar.db.debug in the future, we can justify having the global configuration scalar.db.debug, but if not, it might be better to make the parameter scope more narrow. What do you think?

Good point, thank you. I aggree, I will update it to `scalar.db.consensus_commit.debug`

docs/scalardb-server.md
```diff
@@ -60,6 +60,11 @@ scalar.db.password=cassandra
 # Storage implementation. "cassandra" or "cosmos" or "dynamo" or "jdbc" or "grpc" can be set. Default storage is "cassandra".
 scalar.db.storage=cassandra
 
+# If set to "true", Get and Scan operations results will contain transaction metadata. To see the transaction metadata columns details for a given table, you can use the `DistributedTransactionAdmin.getTableMetadata` 
+# method which will return the table metadata augmented with the transaction metadata columns. Using the debug mode can be useful to investigate transaction related issues.
+# The default is false.
+scalar.db.debug=false
```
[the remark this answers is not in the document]
Torch3333: Right, thank you. I will do as you suggest.

docs/api-guide.md
```diff
@@ -747,6 +747,17 @@ Also, the `validate()` API could throw `ValidationException` and `ValidationConf
 If you catch `ValidationException`, like the `CrudException` case, you should cancel the transaction or retry the transaction after the failure/error is fixed.
 If you catch `ValidationConflictException`, like the `CrudConflictException` case, you can retry the transaction.
 
+## Debug transactions errors
+
+By adding the following configuration, `Get` and `Scan` operations results will contain [transaction metadata](../schema-loader/README.md/#internal-metadata-for-consensus-commit).
```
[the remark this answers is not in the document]
Torch3333: I did not know, thank you.

Torch3333:
@brfrn169 I revised according to your comments. Please have a look when you can.

docs/api-guide.md
```diff
@@ -747,6 +747,17 @@ Also, the `validate()` API could throw `ValidationException` and `ValidationConf
 If you catch `ValidationException`, like the `CrudException` case, you should cancel the transaction or retry the transaction after the failure/error is fixed.
 If you catch `ValidationConflictException`, like the `CrudConflictException` case, you can retry the transaction.
 
+## Debug transactions errors
+
+By adding the following configuration, `Get` and `Scan` operations results will contain [transaction metadata](https://github.com/scalar-labs/scalardb/blob/master/schema-loader/README.md#internal-metadata-for-consensus-commit).
```
brfrn169, line 752: I think we should also mention that the configuration is for Consensus Commit.

docs/scalardb-server.md
```diff
@@ -70,6 +70,11 @@ scalar.db.consensus_commit.isolation_level=SNAPSHOT
 # Either EXTRA_READ or EXTRA_WRITE can be specified. EXTRA_READ is used by default.
 # If SNAPSHOT is specified in the property "scalar.db.consensus_commit.isolation_level", this is ignored.
 scalar.db.consensus_commit.serializable_strategy=
+
+# If set to "true", Get and Scan operations results will contain transaction metadata. To see the transaction metadata columns details for a given table, you can use the `DistributedTransactionAdmin.getTableMetadata()`
```
brfrn169, line 74: Ditto.

brfrn169 (commented):
Thank you. Left several comments. Please take a look when you have time!

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
```diff
@@ -840,6 +850,87 @@ public void abort_forOngoingTransaction_ShouldAbortCorrectly() throws Transactio
     assertResultsContainsExactlyInAnyOrder(results, Collections.singletonList(expectedResult));
   }
 
+  @Test
+  public void scan_UsingDebugMode_ShouldReturnTransactionMetadataColumns()
```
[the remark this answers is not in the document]
Torch3333: Ok, I will move these tests to `ConsensusCommitIntegrationTestBase`

core/src/integration-test/java/com/scalar/db/api/TwoPhaseCommitTransactionIntegrationTestBase.java
```diff
@@ -1035,6 +1047,91 @@ public void scan_ScanAllGivenForNonExisting_ShouldReturnEmpty() throws Transacti
     assertResultsContainsExactlyInAnyOrder(results, Collections.singletonList(expectedResult));
   }
 
+  @Test
+  public void scan_UsingDebugMode_ShouldReturnTransactionMetadataColumns()
```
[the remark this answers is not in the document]
Torch3333: Ok, I will move these tests to `TwoPhaseConsensusCommitIntegrationTestBase`

core/src/integration-test/java/com/scalar/db/api/DistributedTransactionAdminIntegrationTestBase.java
```diff
@@ -131,6 +139,21 @@ private void dropTables() throws ExecutionException {
     admin.dropCoordinatorTables();
   }
 
+  @Test
+  public void
+      getTableMetadata_WhenDebugging_ShouldReturnCorrectMetadataWithTransactionMetadataColumns()
```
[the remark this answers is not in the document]
Torch3333: If I move the test to `ConsensusCommitAdminIntegrationTestBase`, the test won't be available from the server integration test since there is only a `DistributedTransactionAdminServiceIntegrationTest` which inherits from `DistributedTransactionAdminIntegrationTestBase`. 
Should we add a new class `ConsensusCommitAdminIntegrationTestWithDistributedStorageService` which inherits from `ConsensusCommitAdminIntegrationTestBase` mainly for testing `getTableMetadata_WhenDebugging_ShouldReturnCorrectMetadataWithTransactionMetadataColumns` or is it too much ?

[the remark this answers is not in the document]
brfrn169: Ah okay, that's good point. If we add the new class `ConsensusCommitAdminIntegrationTestWithDistributedStorageService`, we have a lot of duplicated tests between it and `DistributedTransactionAdminServiceIntegrationTest`.

IMO, for now, we can add the new class `ConsensusCommitAdminIntegrationTestWithDistributedStorageService` in this PR, and after that we should refactor and reorganize the integration tests in the `server` module. What do you think?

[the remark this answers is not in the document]
Torch3333: Ok, I will add `ConsensusCommitAdminIntegrationTestWithDistributedStorageService`. Ok, let's talk about the test reorganization on the server side afterward.

brfrn169 (approved):
LGTM! Thank you!

conf/database.properties
```diff
@@ -21,3 +21,9 @@ scalar.db.password=cassandra
 # Either EXTRA_READ or EXTRA_WRITE can be specified. EXTRA_READ is used by default.
 # If SNAPSHOT is specified in the property "scalar.db.consensus_commit.isolation_level", this is ignored.
 #scalar.db.consensus_commit.serializable_strategy=
+
+# Debug mode usable for Consensus Commit.
+# If set to "true", Get and Scan operations results will contain transaction metadata. To see the transaction metadata columns details for a given table, you can use the `DistributedTransactionAdmin.getTableMetadata()`
+# method which will return the table metadata augmented with the transaction metadata columns. Using the debug mode can be useful to investigate transaction related issues.
+# The default is false.
+#scalar.db.consensus_commit.debug=false
```
feeblefakie, line 29: Is it more consistent to name it like this for boolean configurations?
```suggestion
#scalar.db.consensus_commit.debug.enabled=false
```

This is just a comment, but `debug` might sound too broad unless we are planning to add more debug features.
I don't come up with a good naming, but we could make it more specific like `consensus_commit.filter_metadata.enabled` or something.
(in such a case, true by default, though)

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
```diff
@@ -134,4 +139,8 @@ public boolean isAsyncCommitEnabled() {
   public boolean isAsyncRollbackEnabled() {
     return asyncRollbackEnabled;
   }
+
+  public boolean isDebugging() {
```
feeblefakie, line 143:
```suggestion
  public boolean isDebugEnabled() {
```
This might be better to make it consistent with other boolean configurations.

feeblefakie (commented):
Overall, looking good! I think it's very useful, especially in SQL mode.
Left some comments/suggestions in the wording.

conf/database.properties
Torch3333: Ok, I agree, updating the name to something more specific is a good idea I think. . I will add the `enabled` as well too, thank you for pointing it out. 
`consensus_commit.filter_metadata.enabled` -> Personally I would understand that metadata are kept in this case, not removed. Using `filter_out` might get the point across more clearly.
What do you think about one of the following?
- `consensus_commit.filter_out_metadata_columns.enabled`
- `consensus_commit.omit_metadata_columns.enabled`

feeblefakie: I think `filter` itself also means removing something as below, so `out` might not be necessary.
https://www.oxfordlearnersdictionaries.com/definition/english/filter_2

`filter_metadata_columns.enabled` or making it more nouny like `metadata_columns_filtering.enabled` ?

Torch3333: Ok, both seem good but for consistency `metadata_columns_filtering.enabled` is probably better.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
Torch3333: I will rename this `isMetadataColumnsFilteringEnabled` and update the rest of the code in consequence.

conf/database.properties
brfrn169: IMHO, as we filter the metadata columns by default, if we use the `enabled` wording, I feel `include metadata` or `show metadata` or something like that is more intuitive than `filter metadata`. What do you think? @feeblefakie @Torch3333

feeblefakie: That's a good point to discuss. I don't have a strong preference for that.
Is it common to use `false` by default for this kind of boolean configuration?

As far as I can see in the Spark doc, both `true` and `false` are set by default.
(e.g., `spark.network.crypto.enabled` (false by default), `spark.ui.xContentTypeOptions.enabled` (true by default))
https://spark.apache.org/docs/latest/security.html

Torch3333: `include metadata` is a good option I think. It feels more clear-cut than `filter`

brfrn169: > As far as I can see in the Spark doc, both true and false are set by default.

@feeblefakie Ah okay, so maybe it doesn't depend on the default value. `Enable metadata filtering` sounds to me that we should originally return the metadata, but we don't when metadata filtering is enabled. However, as we hide the metadata from users so that users can access transaction tables transparently, I think the correct is that we should not return the metadata basically, but we do when enabling including (or showing) metadata is enabled.

feeblefakie: @brfrn169 @Torch3333 I think it makes sense, so something like `include_metadata` also sounds good to me. (show sounds a bit weird to me because it's up to users if they show the metadata or not. but it makes sense if it is SQL command-line tool option)

Ideally, it can be a nouny form, but `metadata_inclusion` sounds a bit unusual?

brfrn169: I'm okay with `metadata_inclusion`. But I'm not sure it sounds unusual. What do you think? @Torch3333

Torch3333: `metadata_inclusion` sounds indeed a bit unusual. Even though `include_metadata` may not be consistent with the rest of the configuration naming, I think using terms developers are more familiar with is more important.

brfrn169: @Torch3333 Okay, sure. Let's use `include_metadata` then.

Torch3333: To sum up, the full configuration name will be `scalar.db.consensus_commit.include_metadata.enabled`

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
Torch3333: After our discussion above, I will rename it to `isIncludeMetadataEnabled()`

Torch3333:
The PR has been revised with the configuration name change to `scalar.db.consensus_commit.include_metadata.enabled`. 
Please have a look again. 
Thank you.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -128,12 +133,15 @@ public void delete(Delete delete) {
...
     try {
-      // get only after image columns
       get.clearProjections();
-      LinkedHashSet<String> afterImageColumnNames =
-          tableMetadataManager.getTransactionTableMetadata(get).getAfterImageColumnNames();
-      get.withProjections(afterImageColumnNames);
-
+      // Retrieve only the after images columns when including the metadata is disabled, otherwise
+      // retrieve all the
+      // columns
```
brfrn169, line 139:
```suggestion
      // retrieve all the columns
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -143,12 +151,15 @@ private Optional<TransactionResult> getFromStorage(Get get) throws CrudException
...
     try {
-      // get only after image columns
       scan.clearProjections();
-      LinkedHashSet<String> afterImageColumnNames =
-          tableMetadataManager.getTransactionTableMetadata(scan).getAfterImageColumnNames();
-      scan.withProjections(afterImageColumnNames);
-
+      // Retrieve only the after images columns when including the metadata is disabled, otherwise
+      // retrieve all the
+      // columns
```
brfrn169, line 157:
```suggestion
      // retrieve all the columns
```

brfrn169 (approved):
I left several minor suggestions, but other than that, LGTM! Thank you!

brfrn169 (approved):

feeblefakie (approved):
LGTM! Thank you!

PR #651, opened 2022-08-08
Refactor TableMetadataManager

brfrn169:
Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/common/TableMetadataManager.java
+++ b/core/src/main/java/com/scalar/db/common/TableMetadataManager.java
@@ -22,7 +22,7 @@ public class TableMetadataManager {
 
   public TableMetadataManager(Admin admin, long cacheExpirationTimeSecs) {
     this(
-        (key) -> Optional.ofNullable(admin.getTableMetadata(key.namespace, key.table)),
+        key -> Optional.ofNullable(admin.getTableMetadata(key.namespace, key.table)),
         cacheExpirationTimeSecs);
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/scalar/db/common/TableMetadataManager.java
```diff
@@ -20,6 +21,14 @@ public class TableMetadataManager {
   private final LoadingCache<TableKey, Optional<TableMetadata>> tableMetadataCache;
 
   public TableMetadataManager(Admin admin, long cacheExpirationTimeSecs) {
+    this(
+        (key) -> Optional.ofNullable(admin.getTableMetadata(key.namespace, key.table)),
+        cacheExpirationTimeSecs);
+  }
+
+  public TableMetadataManager(
+      ThrowableFunction<Optional<TableMetadata>, TableKey, Exception> getTableMetadataFunc,
```
brfrn169, line 30: After this change, we can also specify a function to retrieve a table metadata in `TableMetadataManager`.

Torch3333 (dismissed):
LGTM, thank you. 
I am curious, what is the end goal of this refactoring?

brfrn169:
> I am curious, what is the end goal of this refactoring?

Sorry for the lack explanation. Actually, I wanted to use the new constructor in another part as follows:

```java
TableMetadataManager tableMetadataManager = new TableMetadataManager(
    key -> {
      // Retrieve table metadata
      ...
    },
    cacheExpirationTimeSecs);
```

We can pass only the `Admin` interface in the existing constructor, but it didn't suit my case.

Torch3333:
@brfrn169 I see, thank you for the details.

feeblefakie (approved):
LGTM! Thank you!

PR #652, opened 2022-08-08
Admin "repairTable()" handle metadata corruption

Torch3333:
This updates the `Admin.repairTable()` method implementation to handle the case if the metadata for the table is corrupted. 
So far, the method implementation was only creating the metadata for the table if it did not exist. This PR updates it so that if the metadata are present, they are : 
- for the `JdbcAdmin`, they are deleted and then added
- for the `CosmosAdmin` and `DynamoAdmin`, they are upserted.


brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #655, opened 2022-08-11
Refactor Scalar DB proto

brfrn169:
This PR basically renames the fields of protocol buffers messages. Since it just does renaming, it should not violate backward compatibility. I added several inline comments for the details. Please take a look!

### DIFF
--- a/rpc/src/main/proto/scalardb.proto
+++ b/rpc/src/main/proto/scalardb.proto
@@ -4,7 +4,7 @@ option java_multiple_files = true;
 option java_package = "com.scalar.db.rpc";
 option java_outer_classname = "ScalarDbProto";
 
-package rpc;
+package scalardb.rpc;
 
 import "google/protobuf/empty.proto";
 
@@ -46,7 +46,7 @@ message Column {
 
 message Key {
   repeated Value value = 1 [deprecated = true];
-  repeated Column column = 2;
+  repeated Column columns = 2;
 }
 
 enum Consistency {
@@ -61,7 +61,7 @@ message Get {
   Key partition_key = 3;
   Key clustering_key = 4;
   Consistency consistency = 5;
-  repeated string projection = 6;
+  repeated string projections = 6;
 }
 
 enum Order {
@@ -79,12 +79,12 @@ message Scan {
   string table = 2;
   Key partition_key = 3;
   Consistency consistency = 4;
-  repeated string projection = 5;
+  repeated string projections = 5;
   Key start_clustering_key = 6;
   bool start_inclusive = 7;
   Key end_clustering_key = 8;
   bool end_inclusive = 9;
-  repeated Ordering ordering = 10;
+  repeated Ordering orderings = 10;
   int32 limit = 11;
 }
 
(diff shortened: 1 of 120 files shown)


rpc/src/main/proto/scalardb.proto
```diff
@@ -4,7 +4,7 @@ option java_multiple_files = true;
 option java_package = "com.scalar.db.rpc";
 option java_outer_classname = "ScalarDbProto";
 
-package rpc;
+package scalardb.rpc;
```
brfrn169, line 7: Changed the package name.

rpc/src/main/proto/scalardb.proto
```diff
@@ -46,7 +46,7 @@ message Column {
 
 message Key {
   repeated Value value = 1 [deprecated = true];
-  repeated Column column = 2;
+  repeated Column columns = 2;
```
brfrn169, line 49: Made the names `repeated` of fields plural.

rpc/src/main/proto/scalardb.proto
```diff
@@ -211,11 +211,11 @@ enum DataType {
 }
 
 message TableMetadata {
-  map<string, DataType> column = 1;
-  repeated string partition_key_name = 2;
-  repeated string clustering_key_name = 3;
-  map<string, Order> clustering_order = 4;
-  repeated string secondary_index_name = 5;
+  map<string, DataType> columns = 1;
```
brfrn169, line 214: Also made the names of `map` fields plural.

rpc/src/main/proto/scalardb.proto
```diff
@@ -357,13 +356,13 @@ message TransactionResponse {
...
   message ScanResponse {
-    repeated Result result = 1;
+    repeated Result results = 1;
   }
 
   message Error {
     enum ErrorCode {
       INVALID_ARGUMENT = 0;
-      CONFLICT = 1;
+      TRANSACTION_CONFLICT = 1;
```
brfrn169, line 365: Renamed `CONFLICT` to `TRANSACTION_CONFLICT`.

rpc/src/main/proto/scalardb.proto
```diff
@@ -468,14 +467,14 @@ message TwoPhaseCommitTransactionResponse {
...
+    repeated Result results = 1;
   }
 
   message Error {
     enum ErrorCode {
       INVALID_ARGUMENT = 0;
-      CONFLICT = 1;
-      UNKNOWN_TRANSACTION = 2;
+      TRANSACTION_CONFLICT = 1;
+      UNKNOWN_TRANSACTION_STATUS = 2;
```
brfrn169, line 477: Also renamed `UNKNOWN_TRANSACTION` to `UNKNOWN_TRANSACTION_STATUS`.

feeblefakie (approved):
LGTM! Thank you!

I feel adding only one element with `addXXXs` method is pretty unusual, but seems like it's the way of the protobuf ...

Torch3333 (approved):
LGTM, thank you!

brfrn169:
> I feel adding only one element with `addXXXs` method is pretty unusual, but seems like it's the way of the protobuf ...

Yeah, I thought so too. [The protocol-buffers style guide](https://developers.google.com/protocol-buffers/docs/style#repeated_fields) says we should do that...

PR #659, opened 2022-08-18
Add @Nullable to Admin.getTableMetadata()

brfrn169:
Since `Admin.getTableMetadata()` may return null, we should add `@Nullable` to it. Please take a look!


Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #666, opened 2022-08-24
Use toUpperCase() for valueOf() of enum type

brfrn169:
Please take a look!


feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -309,7 +310,9 @@ void createTableInternal(
     // Add compaction strategy
     CompactionStrategy compactionStrategy =
         CompactionStrategy.valueOf(
-            options.getOrDefault(COMPACTION_STRATEGY, CompactionStrategy.STCS.toString()));
+            options
+                .getOrDefault(COMPACTION_STRATEGY, CompactionStrategy.STCS.toString())
+                .toUpperCase(Locale.ROOT));
```
brfrn169, line 315: Use `toUpperCase()` so that uses can use both lower case and upper case for the configuration.

PR #669, opened 2022-08-25
Add logging for UnknownTransactionStatusException in Scalar DB Server

brfrn169:
We should log when UnknownTransactionStatusException is thrown for debugging in Scalar DB Server. This PR adds logging for it. Please take a look!

### DIFF
--- a/server/src/main/java/com/scalar/db/server/DistributedTransactionService.java
+++ b/server/src/main/java/com/scalar/db/server/DistributedTransactionService.java
@@ -473,6 +473,7 @@ public class DistributedTransactionService
                 .setMessage(e.getMessage())
                 .build());
       } catch (UnknownTransactionStatusException e) {
+        logger.error("the transaction status is unknown", e);
         responseBuilder.setError(
             TransactionResponse.Error.newBuilder()
                 .setErrorCode(ErrorCode.UNKNOWN_TRANSACTION_STATUS)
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/scalar/db/server/DistributedTransactionService.java
```diff
@@ -473,6 +473,7 @@ private void execute(
                 .setMessage(e.getMessage())
                 .build());
       } catch (UnknownTransactionStatusException e) {
+        logger.error("the transaction status is unknown", e);
```
Torch3333, line 476: If `UnknowTransactionStatusException.getUnknownTransactionId()` is present, should we append the transaction id to the log message, or it is redundant?

Torch3333 (approved):
Thank you, I left a question. Please take a look when you can.

server/src/main/java/com/scalar/db/server/DistributedTransactionService.java
brfrn169: Thank you for pointing it out. Yes, we should show the transaction ID as well. I'll fix it. Thanks.

brfrn169: @Torch3333 I just fixed it. Please take a look again when you have time!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #672, opened 2022-08-25
Rename endpoint override config in Dynamo

brfrn169:
This PR renames the endpoint override config in the Dynamo adapter `scalar.db.dynamo.endpoint-override` to `scalar.db.dynamo.endpoint_override` for consistency. Please take a look!


core/src/main/java/com/scalar/db/storage/dynamo/DynamoConfig.java
```diff
@@ -6,12 +6,15 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Immutable
 public class DynamoConfig {
+  private static final Logger logger = LoggerFactory.getLogger(DynamoConfig.class);
 
   public static final String PREFIX = DatabaseConfig.PREFIX + "dynamo.";
-  public static final String ENDPOINT_OVERRIDE = PREFIX + "endpoint-override";
+  public static final String ENDPOINT_OVERRIDE = PREFIX + "endpoint_override";
```
brfrn169, line 17: Renamed `scalar.db.dynamo.endpoint-override` to `scalar.db.dynamo.endpoint_override` for consistency.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
```diff
@@ -48,7 +48,7 @@ public class ConsensusCommitConfig {
   private final boolean isIncludeMetadataEnabled;
 
   public ConsensusCommitConfig(DatabaseConfig databaseConfig) {
-    if (databaseConfig.getProperties().containsValue("scalar.db.isolation_level")) {
+    if (databaseConfig.getProperties().containsKey("scalar.db.isolation_level")) {
```
brfrn169, line 51: Found a bug here. Fixed it as well. Also fixed the same bug in other places.

feeblefakie: Should it be a bugfix PR and cherry-picked to bugfix releases? (to 3.4.x and 3.5.x)

feeblefakie (approved):
LGTM! Thank you!
Left one question/comment about the bugfix handling.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
brfrn169: I didn't think so actually because it's a trivial bug where the warning message isn't shown even when the old config is used. Do you think we should backport it?

feeblefakie: Ah, if it's just a warning message thing, then it can be fine as it is.

Torch3333 (approved):
LGTM, thank you!

PR #678, opened 2022-08-29
Add method to retrieve the namespace names for Cosmos Admin

Torch3333:
This implement the `admin.getNamespaceNames()` for the CosmosAdmin .

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
@@ -493,7 +493,17 @@ public class CosmosAdmin implements DistributedStorageAdmin {
 
   @Override
   public boolean namespaceExists(String namespace) throws ExecutionException {
-    return databaseExists(namespace);
+    try {
+      getNamespacesContainer()
+          .readItem(namespace, new PartitionKey(namespace), CosmosNamespace.class);
+      return true;
+    } catch (RuntimeException e) {
+      if (e instanceof CosmosException
+          && ((CosmosException) e).getStatusCode() == CosmosErrorCode.NOT_FOUND.get()) {
+        return false;
+      }
+      throw new ExecutionException("checking if the namespace exists failed", e);
+    }
   }
 
   @Override
--- a/core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
@@ -53,7 +53,8 @@ public class CosmosAdmin implements DistributedStorageAdmin {
   public static final String DEFAULT_NO_SCALING = "false";
 
   public static final String METADATA_DATABASE = "scalardb";
-  public static final String METADATA_CONTAINER = "metadata";
+  public static final String TABLE_METADATA_CONTAINER = "metadata";
+  public static final String NAMESPACES_CONTAINER = "namespaces";
   private static final String ID = "id";
   private static final String CONCATENATED_PARTITION_KEY = "concatenatedPartitionKey";
   private static final String PARTITION_KEY_PATH = "/" + CONCATENATED_PARTITION_KEY;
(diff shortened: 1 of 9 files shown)


feeblefakie (approved):
LGTM! Thank you!

brfrn169 (commented):
LGTM overall. Left a question. Please take a look when you have time!

I also have the following questions:

- Should we have a method to repair the namespaces metadata table?
- Just a confirmation, for backward compatibility, you are going to add something like upgrade command in the Schema Loader to load the existing namespaces to the namespaces metadata table, right?

Please take a look!

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -319,12 +320,24 @@ private void deleteTableMetadata(String namespace, String table) throws Executio
...
+      // namespaces
+      if (!namespacesContainer
+          .queryItems(
+              "SELECT 1 FROM container OFFSET 0 LIMIT 1",
+              new CosmosQueryRequestOptions(),
+              Object.class)
+          .stream()
+          .findFirst()
+          .isPresent()) {
+        client.getDatabase(metadataDatabase).delete();
```
brfrn169, line 339: Don't we need to delete the namespaces container before deleting the metadata database?

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -53,7 +53,8 @@ public class CosmosAdmin implements DistributedStorageAdmin {
   public static final String DEFAULT_NO_SCALING = "false";
```
brfrn169, line 53: Don't we need to change the implementation of `namespaceExists()` as well?

brfrn169: Ah, we may need to change `namespaceExists()` in `JdbcAdmin` as well.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -319,12 +320,24 @@ private void deleteTableMetadata(String namespace, String table) throws Executio
 
   @Override
   public void dropNamespace(String namespace) throws ExecutionException {
-    if (!databaseExists(namespace)) {
-      throw new ExecutionException("the database does not exist");
-    }
```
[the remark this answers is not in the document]
Torch3333: I feel like this check was not necessary as the deletion call `client.getDatabase(namespace).delete();` will throw an exception if the database does not exists.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
Torch3333: Deleting the database will delete all its container so deleting the container first is not required.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
Torch3333: Yes, good point thank you. I will update `namespaceExists()` in all the admins.

Torch3333:
@brfrn169 
>Should we have a method to repair the namespaces metadata table?

I think so. Concretely, this method would be an alias for `admin.createNamespace("my_ns", ifNotExists=true, options)`. 
We could then update the Schema Loader `--repair-all` command to also execute `admin.repairNamespace()`   

> * Just a confirmation, for backward compatibility, you are going to add something like upgrade command in the Schema Loader to load the existing namespaces to the namespaces metadata table, right?

Yes. I started working on implementing the `admin.upgrade()` method for now.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
Torch3333: I revised the `namespaceExists()` implementation in the latest commit. Please have a look when you can.

Torch3333:
I changed the PR target branch to the [feat/namespaces_management](https://github.com/scalar-labs/scalardb/tree/feat/namespaces_management) branch that will contain all the namespaces management-related changes.

Sorry for the trouble. Please have a look again

brfrn169 (approved):
LGTM! Thank you!

PR #680, opened 2022-08-31
Commit/rollback records should always be done for all records even on error

brfrn169:
We faced the following situation:

1. Execute a transaction with two record writes
2. Preparing the two records successfully
3. But commit fails due to something conflict, and need to rollback the two records
4. Try to roll forward the first record, but it fails due to something conflict
5. The second record isn't rolled back

In this case, we should also roll back the second record even if rolling back the first record fails.

BTW, this situation happens only when the parallel commit/rollback is disabled. The workaround for this bug is enabling the parallel commit/rollback.

This PR fixes this bug. Please take a look!


core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
```diff
@@ -49,79 +54,144 @@ public ParallelExecutor(ConsensusCommitConfig config) {
...
     }
+  }
+
+  private void executeTasksInParallel(
+      String name, List<ParallelExecutorTask> tasks, boolean noWait, boolean stopOnError)
+      throws ExecutionException, CommitConflictException {
+    assert parallelExecutorService != null;
+
+    CompletionService<Void> completionService =
+        new ExecutorCompletionService<>(parallelExecutorService);
```
brfrn169, line 114: Use `CompletionService` to handle `Future`s once each task is finished.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
```diff
@@ -5,17 +5,22 @@
...
 import java.util.concurrent.Future;
-import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @ThreadSafe
 public class ParallelExecutor {
+  private static final Logger logger = LoggerFactory.getLogger(BatchHandler.class);
```
jnmt, line 23:
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(ParallelExecutor.class);
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
```diff
@@ -49,79 +54,144 @@ public ParallelExecutor(ConsensusCommitConfig config) {
...
     } else {
-      futures = Collections.emptyList();
-      for (ParallelExecutorTask task : tasks) {
-        task.run();
-      }
+      executeTasksSerially(name, tasks, stopOnError);
     }
+  }
+
+  private void executeTasksInParallel(
```
jnmt, line 108: With a heavy TPC-C workload, I'm facing an issue that some records remain in prepare-state even if using this branch with parallel rollback (and using the current master as well). I will check further.

jnmt (commented):
Thank you! Overall looks good to me, but as I commented, I'm facing an issue that might be related to this. So, let me check further.

feeblefakie (dismissed):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
brfrn169: Oh okay. Thank you for checking this. 🙇

brfrn169:
Converting it to draft since we should handle the issue reported by @jnmt.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
```diff
@@ -47,81 +51,168 @@ public ParallelExecutor(ConsensusCommitConfig config) {
...
   }
 
-  public void prepare(List<ParallelExecutorTask> tasks) throws ExecutionException {
+  public void prepare(List<ParallelExecutorTask> tasks, String transactionId)
+      throws ExecutionException {
     try {
-      executeTasks(tasks, config.isParallelPreparationEnabled(), false);
+      // when the parallel preparation disabled, we stop running the tasks when one of them fails.
+      // when not, we need to wait for all the tasks to finish even if some of them fail
+      boolean stopOnError = !config.isParallelPreparationEnabled();
```
brfrn169, line 59: In the preparation phase, we should switch `stopOnError` depending on whether the parallel preparation disabled or not.

brfrn169:
@jnmt Thank you for the investigation! I fixed it based on your feedback.

@feeblefakie @Torch3333 @jnmt Can you please take a look at it again when you have time?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
```diff
@@ -47,81 +51,168 @@ public ParallelExecutor(ConsensusCommitConfig config) {
     this.parallelExecutorService = parallelExecutorService;
   }
 
-  public void prepare(List<ParallelExecutorTask> tasks) throws ExecutionException {
+  public void prepare(List<ParallelExecutorTask> tasks, String transactionId)
+      throws ExecutionException {
     try {
-      executeTasks(tasks, config.isParallelPreparationEnabled(), false);
+      // when the parallel preparation disabled, we stop running the tasks when one of them fails.
```
feeblefakie, line 57: The description basically explains what it does, but I think it's obvious and not much meaningful. So, it would be great if we can say something like why we do this and what kind of bad consequences would happen if `stopOnError=true` in prepare in the parallel mode.

Something like this;

Enabling stopOnError in parallel preparation would cause the corresponding rollback to be executed earlier than the preparation of records, which could result in left-unrecovered records even in a normal case. Thus, we disable stopOnError when parallel preparation is enabled.

feeblefakie (commented):
LGTM! Thank you!
Left a minor comment on the description.

@jnmt Great work!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
brfrn169: Thank you for the feedback! Updated the comment in the latest commit. Please take a look again!

feeblefakie (approved):
LGTM! Thank you!

jnmt (approved):
LGTM! Thank you! I confirmed unrecovered records are no longer left with this fix.

brfrn169:
> LGTM! Thank you! I confirmed unrecovered records are no longer left with this fix.

Thank you for your confirmation!

Torch3333 (approved):
LGTM, thank you!

PR #682, opened 2022-09-01
Delete setDriver() for BasicDataSource in JDBC

brfrn169:
As we have found how to avoid the "No suitable driver" error in https://github.com/scalar-labs/kelpie/pull/28, we no longer need to call `setDriver()` for `BasicDataSource`.

Please take a look!


Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #671, opened 2022-08-25
Refactor CosmosAdminTest

brfrn169:
This PR does the same refactoring for `CosmosAdminTest` as the one for `DynamoAdminTest` in #658. Please take a look!

PR #670, opened 2022-08-25
Refactor JdbcAdminTest

brfrn169:
This PR does the same refactoring for `JdbcAdminTest` as the one for `DynamoAdminTest` in #658. Please take a look!

PR #687, opened 2022-09-02
Add method to retrieve the namespace names for Cassandra Admin

Torch3333:
This implements the `admin.getNamespaceNames()` for the Cassandra admin.

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
@@ -117,7 +117,7 @@ public class CassandraAdmin implements DistributedStorageAdmin {
     String insertQuery =
         QueryBuilder.insertInto(
                 quoteIfNecessary(metadataKeyspace), quoteIfNecessary(KEYSPACES_TABLE))
-            .value(quoteIfNecessary(KEYSPACES_NAME_COL), keyspace)
+            .value(KEYSPACES_NAME_COL, quoteIfNecessary(keyspace))
             .toString();
     clusterManager.getSession().execute(insertQuery);
   }
@@ -156,7 +156,7 @@ public class CassandraAdmin implements DistributedStorageAdmin {
     String deleteQuery =
         QueryBuilder.delete()
             .from(quoteIfNecessary(metadataKeyspace), quoteIfNecessary(KEYSPACES_TABLE))
-            .where(QueryBuilder.eq(quoteIfNecessary(KEYSPACES_NAME_COL), keyspace))
+            .where(QueryBuilder.eq(KEYSPACES_NAME_COL, quoteIfNecessary(keyspace)))
             .toString();
     clusterManager.getSession().execute(deleteQuery);
   }
@@ -337,7 +337,7 @@ public class CassandraAdmin implements DistributedStorageAdmin {
 
       Set<String> keyspaceNames = new HashSet<>();
       String selectQuery =
-          QueryBuilder.select(quoteIfNecessary(KEYSPACES_NAME_COL))
+          QueryBuilder.select(KEYSPACES_NAME_COL)
               .from(quoteIfNecessary(metadataKeyspace), quoteIfNecessary(KEYSPACES_TABLE))
               .getQueryString();
       for (Row row : clusterManager.getSession().execute(selectQuery).all()) {
@@ -362,7 +362,7 @@ public class CassandraAdmin implements DistributedStorageAdmin {
 
   private void dropKeyspacesTableIfEmpty() {
     String selectQuery =
-        QueryBuilder.select(quoteIfNecessary(KEYSPACES_NAME_COL))
+        QueryBuilder.select(KEYSPACES_NAME_COL)
             .from(quoteIfNecessary(metadataKeyspace), quoteIfNecessary(KEYSPACES_TABLE))
             .limit(1)
             .getQueryString();
--- a/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
@@ -6,6 +6,7 @@ import static com.scalar.db.util.ScalarDbUtils.getFullTableName;
 import com.datastax.driver.core.ClusteringOrder;
 import com.datastax.driver.core.ColumnMetadata;
 import com.datastax.driver.core.KeyspaceMetadata;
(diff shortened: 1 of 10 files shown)


core/src/test/java/com/scalar/db/storage/cassandra/CassandraAdminTestBase.java
```diff
@@ -43,20 +52,42 @@
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
-public class CassandraAdminTest {
+/**
+ * Abstraction that defines unit tests for the {@link CassandraAdmin}. The class purpose is to be
+ * able to run the {@link CassandraAdmin} unit tests with different values for the {@link
+ * CassandraAdmin}, notably {@link CassandraAdmin#METADATA_KEYSPACE}.
+ */
+public abstract class CassandraAdminTestBase {
```
Torch3333, line 60: The `CassandraAdminTest` was refactored the same way as the other storage Admin classes. cf https://github.com/scalar-labs/scalardb/pull/670 https://github.com/scalar-labs/scalardb/pull/671

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -60,7 +70,20 @@ public void createTable(
   @Override
   public void createNamespace(String namespace, Map<String, String> options)
       throws ExecutionException {
-    CreateKeyspace query = SchemaBuilder.createKeyspace(quoteIfNecessary(namespace));
+    try {
+      createKeyspace(namespace, options);
+      createMetadataKeyspaceIfNotExists();
```
feeblefakie, line 75: Just a question. 
I think we haven't managed a metadata keyspace for Cassandra since Cassandra manages keyspace metadata by itself, but now are we going to do it for some reason?
If so, what is the reason?

feeblefakie (commented):
Thank you! Left a question. PTAL when you get a chance!

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
Torch3333: The goal is to be able to manage keyspaces that are only created through Scalar DB. We don't think we should be returning keyspaces that are not Scalar DB compatible with `admin.getNamespaceNames()`. 
Since with the built-in Cassandra keyspace metadata manager, we can't know if a keyspace/table was created through Scalar DB, so adding a table to manage Scalar DB keyspace was necessary.  

Similarly, we are also thinking of making `admin.getNamespaceTableNames()` only return tables created through Scalar DB, which implies we will need to create a table to manage table metadata for Cassandra as it is done on other admin implementations.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -6,6 +6,7 @@
 import com.datastax.driver.core.ClusteringOrder;
```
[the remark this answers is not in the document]
Torch3333: Yes, I will update it.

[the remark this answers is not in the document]
Torch3333: I updated the `namespaceExists()` implementation in https://github.com/scalar-labs/scalardb/pull/687/commits/5aac1e24ebd0fc2d5163427d832e7e85d39bc279

Besides, I realized using `quoteIfNecessary()` was not necessary for `CosmosAdmin.KEYSPACES_NAME_COL` as it is not a reserved keyword and added it in places were it was missing for user inputted variable. These modifications are in https://github.com/scalar-labs/scalardb/pull/687/commits/5d61f93cc4f28dbfd1a7bf9d03fd87c8b9d62c91

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -221,13 +277,18 @@ public Set<String> getNamespaceTableNames(String namespace) throws ExecutionExce
   @Override
   public boolean namespaceExists(String namespace) throws ExecutionException {
```
brfrn169, line 278: Ah, it looks like it's backward incompatible change. `namespaceExists()` won't work after upgrading Scalar DB until running the upgrade command. 

Ideally, `namespaceExists()` should work after upgrading Scalar DB even before running the upgrade command. But I'm wondering if we can do that. What do you think?

brfrn169 (commented):

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
feeblefakie: OK, that sounds good to me.
So, it's a bit off-topic, but I'm wondering how Scalar DB would work if it uses JDBC native transactions.
Do we also manage metadata even if it uses JDBC native transactions?
(cc: @brfrn169)

Torch3333:
I changed the PR target branch to the [feat/namespaces_management](https://github.com/scalar-labs/scalardb/tree/feat/namespaces_management) branch that will contain all the namespaces management related changes. 

Sorry for the trouble. Please have a look again.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
Torch3333: Since we decided not to handle backward compatibility, I think the `namespaceExists()` logic is ok as is. Please have a look gain.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
Torch3333: For JDBC native transactions, table and namespace metadata are also managed, via the [JdbcTransactionAdmin](https://github.com/scalar-labs/scalardb/blob/master/core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransactionAdmin.java) 

As with the other usage mode (consensus-commit or storage), the table metadata are used to validate if the given operation is valid (check wrong keys, check wrong projected columns names, etc.).

feeblefakie: Oh, OK. Thank you!

feeblefakie (approved):
LGTM! Thank you!

brfrn169 (approved):
LGTM! Thank you!

PR #693, opened 2022-09-07
Update JdbcAdmin.namespaceExists() 

Torch3333:
This PR updates the `JdbcAdmin.namespaceExists()` method to only return true if the namespace was created through Scalar DB.

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
@@ -706,36 +706,28 @@ public class JdbcAdmin implements DistributedStorageAdmin {
 
   @Override
   public boolean namespaceExists(String namespace) throws ExecutionException {
-    String namespaceExistsStatement = "";
-    switch (rdbEngine) {
-      case POSTGRESQL:
-      case MYSQL:
-        namespaceExistsStatement =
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -706,36 +706,28 @@ public Set<String> getNamespaceTableNames(String namespace) throws ExecutionExce
...
-      return preparedStatement.executeQuery().next();
+        PreparedStatement statement = connection.prepareStatement(selectQuery); ) {
+      statement.setString(1, namespace);
+      try (ResultSet resultSet = statement.executeQuery()) {
+        return resultSet.next();
+      }
     } catch (SQLException e) {
+      // An exception will be thrown if the namespaces table does not exist when executing the
+      // select
+      // query
```
brfrn169, line 724:
```suggestion
      // select query
```

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
Torch3333: Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #696, opened 2022-09-09
Add upgrade method to jdbc admin

Torch3333:
This adds the `admin.upgrade()` in relation to the namespace management feature. FYI, the target branch is [feat/namespaces_management](https://github.com/scalar-labs/scalardb/tree/feat/namespaces_management)

Please refer to the inline comments for additional explanations.


rpc/src/main/proto/scalardb.proto
```diff
@@ -200,6 +200,7 @@ service DistributedStorageAdmin {
   }
   rpc GetNamespaceNames(GetNamespaceNamesRequest) returns (GetNamespaceNamesResponse) {
   }
+  rpc Upgrade(UpgradeRequest) returns (google.protobuf.Empty);
```
brfrn169, line 203: Oh, I didn't know we can omit curly brackets `{}` here. But please don't omit for consistency for now:
```suggestion
  rpc Upgrade(UpgradeRequest) returns (google.protobuf.Empty) {
  }
```

brfrn169 (commented):
LGTM other than what I commented. But they are very minor style things. Please take a look when you have time!

core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -388,4 +388,13 @@ void addNewColumnToTable(String namespace, String table, String columnName, Data
    * @throws ExecutionException if the operation failed
    */
   Set<String> getNamespaceNames() throws ExecutionException;
+
+  /**
+   * Upgrades the Scalar DB environment to support the latest version of the Scalar DB API.
```
feeblefakie, line 393: Can you elaborate on what is upgrade and why we need this?
I just don't know about it, but it helps me to understand and review it properly.

Torch3333: Thank you for your question. Sorry, I had prepared inlined comments but failed to publish them.

The motivation for adding this is that we wanted that users who update their existing Scalar DB application to the 4.0.0 version (planned release version for the namespace management feature) will be able to have `admin.getNamespaceNames()` returns namespaces created before and after the 4.0.0.
This means we needed to find a way for users to import their existing namespaces into the metadata namespaces table so that they can be managed via the admin interface.
Additionally, we may develop another feature in the future that requires as well to update the Scalar DB environment as the namespace management does. Therefore, we decided to keep a generic name for the `upgrade()` so that this method can be used for future features.

Torch3333:
@brfrn169 Thank you for your comments. I fixed the format issues in `scalardb.proto`. 

@brfrn169 @feeblefakie I also merged the `feat/namespaces_branch` in this PR branch which includes the refactoring of integration tests (https://github.com/scalar-labs/scalardb/pull/697) that had a couple of conflicts in integration tests files.
Please have a look again when you get the chance.

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/api/Admin.java
feeblefakie: Got it.Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #661, opened 2022-08-22
Use ServiceLoader to load storage and transaction modules

brfrn169:
Currently, we have the mappings between the storage adapter names and the implementation classes ([here](https://github.com/scalar-labs/scalardb/blob/4157d0991d89fe84f7ac15a7cb90bb0077b3a039/core/src/main/java/com/scalar/db/config/DatabaseConfig.java#L103-L131)) and the mappings between the transaction manager names and the implementation classes ([here](https://github.com/scalar-labs/scalardb/blob/4157d0991d89fe84f7ac15a7cb90bb0077b3a039/core/src/main/java/com/scalar/db/config/DatabaseConfig.java#L140-L176)) in `DatabaseConfig`. Due to this, when we add a new storage adapter or transaction manager, we need to add its implementation to the same project and add its mapping to `DatabaseConfig`. In other words, currently, we cannot add a new storage adapter or transaction manager to another project, which is inconvenient in some situations.

To address this issue, we can use `ServiceLoader` to load storage and transaction modules. Please see [here](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) for the details of `ServiceLoader`. In the change in this  PR, we introduce `DistributedStorageFactory` and `DistributedTransactionFactory`, which each adapter needs to implement,  and we load implementations of `DistributedStorageFactory` and `DistributedTransactionFactory` with `ServiceLoader`.

I will add inline comments for the details as well. Please take a look!

PR #697, opened 2022-09-09
Move integration test base code to subproject

brfrn169:
After #661, we are now able to dynamically load storage adapters and add a new storage adapter in another project. And, the next task is to make the integration test be able to run in another project.

To do that, we need to move the integration test base code to a subproject and publish the integration test jar to the Maven repository.

What I did in this change is as follows:

- Created `integration-test` subproject
- Moved the integration test base code (from `core` and `schema-loader`) to the `integration-test` subproject
- Moved the integration test code in `schema-loader` to `core`
- Did necessary refactoring for the integration test code to move it to the `integration-test` subproject

I will also add inline comments for the details. Please take a look!


server/src/integration-test/java/com/scalar/db/server/ServerAdminTestUtils.java
```diff
@@ -9,21 +9,24 @@
...
+import com.scalar.db.util.AdminTestUtils;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Properties;
 import org.apache.commons.dbcp2.BasicDataSource;
 
-public class JdbcAdminTestUtils extends AdminTestUtils {
+public class ServerAdminTestUtils extends AdminTestUtils {
```
brfrn169, line 20: A `XXXAdminTestUtils` class for server.

server/src/integration-test/java/com/scalar/db/server/ServerEnv.java
```diff
@@ -1,6 +1,8 @@
 package com.scalar.db.server;
```
brfrn169, line 1: Changed to use JDBC in the server integration tests.

Torch3333 (dismissed):
LGTM, thank you!

core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraAdminTestUtils.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class CassandraAdminTestUtils extends AdminTestUtils {
+
+  public CassandraAdminTestUtils(Properties properties) {
+    super(properties);
+  }
+
+  @Override
+  public void dropMetadataTable() {
+    // Do nothing
```
feeblefakie, line 14: Just a question.
I thought the metadata table is now managed for Cassandra as well for consistency, is it OK with empty?

feeblefakie (dismissed):
LGTM! Thank you!
Left one question but doesn't block it from being merged.

core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraAdminTestUtils.java
brfrn169: Thank you for taking a look! 

The change is not yet merged to the `master` branch. In the `feat/namespaces_management` branch, the metadata table is managed for Cassandra.

Torch3333: Maybe there is some confusion but the `feat/namespaces_management` only implements the namespace management for Cassandra. 
Managing the table metadata using a metadata table for Cassandra is not yet implemented.

feeblefakie: Thank you for clarification!
So, I understand `Do nothing` parts will eventually be implemented after the table metadata for Cassandra is implemented.

Torch3333: Yes, exactly.

PR #699, opened 2022-09-12
Update release workflow to push container to ECR of Marketplace

kota2and3kan:
This PR updates release workflow to push container image to ECR of AWS Marketplace.
I use Identity Provider (OpenID Connect) instead of access key and secret access key to authenticate ECR in the CI.
I will add some comment in the code.

Please take a look!

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,10 @@ on:
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+"
 
+permissions:
+  id-token: write
+  contents: read
+
 jobs:
   release:
     runs-on: ubuntu-latest
@@ -51,6 +55,17 @@ jobs:
           username: ${{ secrets.ACR_USER }}
           password: ${{ secrets.ACR_PASSWORD }}
 
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.AWS_MARKETPLACE_ROLE_OIDC }}
+          aws-region: us-east-1
+
+      - name: Login to Amazon Elastic Container Registry
+        uses: aws-actions/amazon-ecr-login@v1
+        with:
+          registries: ${{ secrets.AWS_MARKETPLACE_ECR_ID }}
+
       - name: Create containers
         run: gradle docker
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -5,6 +5,10 @@ on:
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+"
 
+permissions:
+  id-token: write
+  contents: read
```
kota2and3kan, line 10: To get `ID token` from GitHub OIDC Provider, we need to add permission configuration (allow) `id-token: write`.
https://docs.github.com/ja/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers

Also, if we add `permissions` in the YAML file, the default permission `contents: read` is NOT allowed.
So, in this case, we need to set (allow) `contents: read` explicitly.
The `contents: read` is required to checkout this repository in the CI (it's required for `actions/checkout@v2`).

.github/workflows/release.yaml
```diff
@@ -51,6 +55,17 @@ jobs:
           username: ${{ secrets.ACR_USER }}
           password: ${{ secrets.ACR_PASSWORD }}
 
+      - name: Configure AWS credentials
```
kota2and3kan, line 58: This step gets credentials from AWS to login ECR using OpenID Connect.

.github/workflows/release.yaml
```diff
@@ -51,6 +55,17 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.AWS_MARKETPLACE_ROLE_OIDC }}
+          aws-region: us-east-1
+
+      - name: Login to Amazon Elastic Container Registry
```
kota2and3kan, line 64: This step login to ECR of AWS Marketplace using credentials that we got in the above step.

.github/workflows/release.yaml
```diff
@@ -66,12 +81,19 @@ jobs:
...
 
+      # Push the latest tag image since it is required from the perspective of the specification of Azure Marketplace.
       - name: Push latest tag to Azure Container Registry
         if: ${{ steps.version.outputs.version == steps.version.outputs.latest_version }}
         run: |
           docker tag ghcr.io/scalar-labs/scalardb-server:${{ steps.version.outputs.version }} ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:latest
           docker push ${{ secrets.ACR_LOGIN_SERVER }}/scalardb-server:latest
 
+      # Push the patch version image only since AWS Marketplace restricts using the latest tag.
+      - name: Push containers to Amazon Elastic Container Registry
```
kota2and3kan, line 92: We push the patch version image only since AWS Marketplace restricts using the latest tag.

kota2and3kan:
CI is failed since the Trivy vulnerability test is failed.

Torch3333 (approved):
LGTM, thank you!

brfrn169 (approved):
LGTM! Thank you!

I realized that we should not have add such company things originally as this is published as OSS. Let's discuss it later. (CC: @feeblefakie)

kota2and3kan:
@brfrn169  
Thank you for your review!
Indeed, we should take care of the license. 

I think we can separate the container pushing part from this repository's CI to another private repository's CI. Let's discuss it later!

feeblefakie (approved):
LGTM! Thank you!

PR #704, opened 2022-09-16
Refactor ConsensusCommitAdmin unit test

Torch3333:
This PR refactor the `ConsensusCommitAdmin` unit test class to be able to write tests easily for different values of `ConsensusCommitConfig.COORDINATOR_NAMESPACE`. All the storage admin classes were refactored the same way, for example, the [JdbcAdmin](https://github.com/scalar-labs/scalardb/pull/670/files)

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
@@ -5,7 +5,6 @@ import static com.scalar.db.transaction.consensuscommit.ConsensusCommitUtils.get
 import static com.scalar.db.transaction.consensuscommit.ConsensusCommitUtils.removeTransactionMetaColumns;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.inject.Inject;
 import com.scalar.db.api.DistributedStorageAdmin;
 import com.scalar.db.api.DistributedTransactionAdmin;
 import com.scalar.db.api.TableMetadata;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
```diff
@@ -25,14 +24,6 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
   private final String coordinatorNamespace;
   private final boolean isIncludeMetadataEnabled;
 
-  @Inject
-  public ConsensusCommitAdmin(DistributedStorageAdmin admin, DatabaseConfig databaseConfig) {
-    this.admin = admin;
-    ConsensusCommitConfig config = new ConsensusCommitConfig(databaseConfig);
-    coordinatorNamespace = config.getCoordinatorNamespace().orElse(Coordinator.NAMESPACE);
-    isIncludeMetadataEnabled = config.isIncludeMetadataEnabled();
-  }
```
brfrn169, line 34: I don't think think we should remove this constructor because it's used by the Guice module:
https://github.com/scalar-labs/scalardb/blob/733e167368f18c50c5790691c56bd63c9b57d221/core/src/main/java/com/scalar/db/service/TransactionModule.java

brfrn169 (commented):
Left a comment. Other than that, LGTM. Thank you!

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
Torch3333: Right, thank you for catching this. 
I reverted the deletion, please have a look again.

brfrn169 (approved):
LGTM! Thank you!

PR #705, opened 2022-09-20
In the Cosmos DB job of the CI, add a step to clean up gradle daemon logs file

Torch3333:
The Azure VM running the CI for the Cosmos integration test ended up running out of disk space.
One of the reasons was that Gradle daemon creates a lot of log files. In this case, 20GB of logs file were created in the span of 5 months. 
So we decided to add a step to the CI to delete these log files. Below is the output produced by the new step.

image

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -184,7 +184,7 @@ jobs:
 
       - name : Delete gradle daemon log files
         if: always()
-#       Delete all files created more than 3 days ago with the ".out.log" file extension located in the "/home/azureuser/.gradle/daemon"
+#       Delete all files modified more than 3 days ago with the ".out.log" file extension located in the "/home/azureuser/.gradle/daemon"
 #       folder hierarchy. These files accumulate over time and can end up using a lot of disk space
         run : find /home/azureuser/.gradle/daemon -name "*.out.log" -type f -mtime +3 -exec rm -vf {} +
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -184,9 +184,9 @@ jobs:
 
       - name : Delete gradle daemon log files
         if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -182,6 +182,12 @@ jobs:
         with:
           arguments: integrationTestCosmos -Dscalardb.cosmos.uri=${{ secrets.COSMOS_URI }} -Dscalardb.cosmos.password=${{ secrets.COSMOS_PASSWORD }} -Dscalardb.cosmos.database_prefix=${{ env.db_prefix }}_
 
+      - name : Delete gradle daemon log files
+        if: always()
+#       Delete all files with the ".out.log" file extension in the "/home/azureuser/.gradle/daemon"
+#       folder hierarchy. These files accumulate over time and can end up using a lot of disk space
+        run : find /home/azureuser/.gradle/daemon -name "*.out.log" -type f -exec rm -vf {} +
```
brfrn169, line 189: I'm wondering if it's really okay to delete all the logs if we are running multiple integration tests at the same time. Maybe adding something like `-mtime +3` is safer?

```suggestion
        run : find /home/azureuser/.gradle/daemon -name "*.out.log" -type f -mtime +3 -exec rm -vf {} +
```

brfrn169 (commented):

.github/workflows/ci.yaml
Torch3333: Indeed, thank you. I updated it.

brfrn169 (approved):
LGTM! Thank you!

Torch3333:
A precision, after checking the log file content more in detail. The logs contain mostly exception stack traces created by the integration test scenario testing failure case.

feeblefakie (approved):
LGTM! Thank you!

PR #709, opened 2022-09-22
Add upgrade() method to the Cassandra Admin

Torch3333:
This implement the `Cassandra.upgrade()` method in relation to the namespace management feature as it was done for the `JdbcAdmin` (cf. https://github.com/scalar-labs/scalardb/pull/696). 
FYI, the target branch is [feat/namespaces_management](https://github.com/scalar-labs/scalardb/tree/feat/namespaces_management)

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraAdminTestUtils.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraAdminTestUtils.java
@@ -1,12 +1,24 @@
 package com.scalar.db.storage.cassandra;
 
+import static com.datastax.driver.core.Metadata.quoteIfNecessary;
+
+import com.datastax.driver.core.schemabuilder.SchemaBuilder;
+import com.scalar.db.config.DatabaseConfig;
 import com.scalar.db.util.AdminTestUtils;
 import java.util.Properties;
 
 public class CassandraAdminTestUtils extends AdminTestUtils {
 
+  private final ClusterManager clusterManager;
+  private final String metadataKeyspace;
+
   public CassandraAdminTestUtils(Properties properties) {
     super(properties);
+    DatabaseConfig databaseConfig = new DatabaseConfig(properties);
+    CassandraConfig cassandraConfig = new CassandraConfig(databaseConfig);
+    metadataKeyspace =
+        cassandraConfig.getMetadataKeyspace().orElse(CassandraAdmin.METADATA_KEYSPACE);
+    clusterManager = new ClusterManager(databaseConfig);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraAdminTestUtils.java
```diff
@@ -25,7 +37,9 @@ public void corruptMetadata(String namespace, String table) {
   }
 
   @Override
-  public void dropNamespacesTable() throws Exception {
-    throw new UnsupportedOperationException("Not yet implemented");
+  public void dropNamespacesTable() {
+    String dropKeyspaceQuery =
+        SchemaBuilder.dropKeyspace(quoteIfNecessary(metadataKeyspace)).getQueryString();
+    clusterManager.getSession().execute(dropKeyspaceQuery);
```
brfrn169, line 43: Don't we need to add the same to `MultiStorageAdminTestUtils`?

brfrn169 (commented):

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -542,6 +542,23 @@ public String toString() {
 
   @Override
   public void upgrade(Map<String, String> options) throws ExecutionException {
-    throw new UnsupportedOperationException("Not yet implemented");
+    createMetadataKeyspaceIfNotExists();
+    createKeyspacesTableIfNotExists();
+    // Retrieve user keyspace and filter out system ones. A downside is that this may include
```
feeblefakie, line 547: Just a question.
Are we going to check the tables of a namespace to see if the namespace is created for Scalar DB?

feeblefakie (approved):
LGTM! Thank you!
Left one question for clarification.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
Torch3333: If the table has transaction metadata columns, we can infer it is from Scalar DB. Otherwise, it is either a Scalar DB storage mode table or not a Scalar DB table.
So even if we check each table of the namespace, we can't be sure the namespace is not for Scalar DB.
Also, if a namespace not for Scalar DB is being added to the list of managed Scalar DB namespaces, it causes no harm besides being a false positive.

core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraAdminTestUtils.java
Torch3333: Good catch, thank you. I revised it. Please have a look again.

brfrn169 (approved):
LGTM! Thank you!

PR #711, opened 2022-09-23
Disable include metadata tests if external server used in server integration tests

brfrn169:
Please take a look!

### DIFF
--- a/server/src/integration-test/java/com/scalar/db/server/ConsensusCommitAdminIntegrationTestWithDistributedTransactionAdminService.java
+++ b/server/src/integration-test/java/com/scalar/db/server/ConsensusCommitAdminIntegrationTestWithDistributedTransactionAdminService.java
@@ -8,6 +8,8 @@ import com.scalar.db.transaction.consensuscommit.Coordinator;
 import java.io.IOException;
 import java.util.Properties;
 import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIf;
 
 public class ConsensusCommitAdminIntegrationTestWithDistributedTransactionAdminService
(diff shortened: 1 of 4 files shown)


Torch3333 (approved):
LGTM, thank you!

server/src/integration-test/java/com/scalar/db/server/ConsensusCommitAdminIntegrationTestWithDistributedTransactionAdminService.java
```diff
@@ -41,6 +44,8 @@ protected void initialize(String testName) throws IOException {
       properties.setProperty(ServerConfig.PORT, PORT_FOR_SERVER_WITH_INCLUDE_METADATA_ENABLED);
       serverWithIncludeMetadataEnabled = new ScalarDbServer(properties);
       serverWithIncludeMetadataEnabled.start();
+    } else {
+      isExternalServerUsed = true;
```
feeblefakie, line 48: I'm not really sure what the external server means here. Is it true when the Scalar DB server is used in the tests?

feeblefakie (approved):
LGTM! Thank you!
The code looks good, but I left one question for clarification.

server/src/integration-test/java/com/scalar/db/server/ConsensusCommitAdminIntegrationTestWithDistributedTransactionAdminService.java
brfrn169: By default, we start the Scalar DB Server in the integration test JVM internally and run the tests on it. But when we specify the JVM option `scalardb.server.external_server_used=true`, we don't start the Scalar DB Server in the integration test JVM and run the tests on the external server.

PR #713, opened 2022-09-26
Refactor consensus commit specific integration tests

brfrn169:
Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
@@ -42,7 +42,6 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
   private final CommitHandler commit;
   private final RecoveryHandler recovery;
   private Runnable beforeRecoveryHook;
-  private Runnable beforeCommitHook;
 
   @SuppressFBWarnings("EI_EXPOSE_REP2")
   public ConsensusCommit(CrudHandler crud, CommitHandler commit, RecoveryHandler recovery) {
@@ -50,7 +49,6 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
     this.commit = checkNotNull(commit);
     this.recovery = checkNotNull(recovery);
     this.beforeRecoveryHook = () -> {};
-    this.beforeCommitHook = () -> {};
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -42,15 +42,13 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
   private final CommitHandler commit;
   private final RecoveryHandler recovery;
   private Runnable beforeRecoveryHook;
-  private Runnable beforeCommitHook;
```
brfrn169, line 45: Deleted `beforeCommitHook`. 

We used it in the integration test and executed another transaction in this hook. But, actually, instead of using it, we can simply execute another transaction before commit the main transaction. That's why I deleted it.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommit.java
```diff
@@ -54,7 +54,6 @@ enum Status {
 
   // For test
   private Runnable beforeRecoveryHook = () -> {};
-  private Runnable beforePrepareHook = () -> {};
```
brfrn169, line 57: Ditto. Deleted `beforePrepareHook`.

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -59,18 +60,43 @@ public abstract class TwoPhaseConsensusCommitSpecificIntegrationTestBase {
...
   private static final String ANY_ID_2 = "id2";
 
-  private TwoPhaseConsensusCommitManager manager;
-  private DistributedStorage storage;
-  private DistributedStorageAdmin admin;
-  private ConsensusCommitAdmin consensusCommitAdmin;
-  private Coordinator coordinator;
-  private String namespace;
+  private TwoPhaseConsensusCommitManager manager1;
+  private TwoPhaseConsensusCommitManager manager2;
```
brfrn169, line 64: Also separated a `TwoPhaseCommitTransactionManager` instance per table because we assume users execute 2PC transactions with different instances. It's more like an actual situation.

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #719, opened 2022-10-11
Add transaction state management

brfrn169:
This PR adds transaction state management to improve usability. The purpose of this state management is to unify the call sequence of the transaction object (e.g. CRUD -> prepare -> validate -> commit or CRUD -> prepare (failed) -> rollback, etc.). I'll add inline comments for the details. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
@@ -115,7 +115,7 @@ public abstract class AbstractDistributedTransactionManager
 
   protected DistributedTransaction wrap(DistributedTransaction transaction)
       throws TransactionException {
-    return new ActiveTransaction(new TransactionStateManagement(transaction));
+    return new ActiveTransaction(new StateManagedTransaction(transaction));
   }
 
   private class ActiveTransaction extends AbstractDistributedTransaction
@@ -194,7 +194,7 @@ public abstract class AbstractDistributedTransactionManager
   }
 
   @VisibleForTesting
-  static class TransactionStateManagement extends AbstractDistributedTransaction
+  static class StateManagedTransaction extends AbstractDistributedTransaction
       implements WrappedDistributedTransaction {
 
     private enum Status {
@@ -208,7 +208,7 @@ public abstract class AbstractDistributedTransactionManager
     private Status status;
 
     @VisibleForTesting
-    TransactionStateManagement(DistributedTransaction transaction) {
+    StateManagedTransaction(DistributedTransaction transaction) {
       this.transaction = transaction;
       status = Status.ACTIVE;
     }
--- a/core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
@@ -17,8 +17,10 @@ import com.scalar.db.exception.transaction.TransactionException;
 import com.scalar.db.exception.transaction.UnknownTransactionStatusException;
 import com.scalar.db.util.ActiveExpiringMap;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -111,13 +113,18 @@ public abstract class AbstractDistributedTransactionManager
                         + "It might have been expired"));
   }
 
-  @VisibleForTesting
-  public class ActiveTransaction extends AbstractDistributedTransaction {
+  protected DistributedTransaction wrap(DistributedTransaction transaction)
+      throws TransactionException {
+    return new ActiveTransaction(new TransactionStateManagement(transaction));
+  }
+
+  private class ActiveTransaction extends AbstractDistributedTransaction
+      implements WrappedDistributedTransaction {
 
     private final DistributedTransaction transaction;
 
     @SuppressFBWarnings("EI_EXPOSE_REP2")
-    public ActiveTransaction(DistributedTransaction transaction) throws TransactionException {
+    private ActiveTransaction(DistributedTransaction transaction) throws TransactionException {
       this.transaction = transaction;
       addActiveTransaction(this);
     }
@@ -177,8 +184,119 @@ public abstract class AbstractDistributedTransactionManager
       }
     }
 
+    @Override
+    public DistributedTransaction getOriginalTransaction() {
+      if (transaction instanceof WrappedDistributedTransaction) {
+        return ((WrappedDistributedTransaction) transaction).getOriginalTransaction();
+      }
+      return transaction;
+    }
+  }
+
+  @VisibleForTesting
+  static class TransactionStateManagement extends AbstractDistributedTransaction
+      implements WrappedDistributedTransaction {
+
+    private enum Status {
+      ACTIVE,
+      COMMITTED,
+      COMMIT_FAILED,
+      ROLLED_BACK
+    }
+
+    private final DistributedTransaction transaction;
+    private Status status;
+
     @VisibleForTesting
-    public DistributedTransaction getActualTransaction() {
+    TransactionStateManagement(DistributedTransaction transaction) {
+      this.transaction = transaction;
+      status = Status.ACTIVE;
+    }
+
+    @Override
+    public String getId() {
+      return transaction.getId();
+    }
+
+    @Override
+    public Optional<Result> get(Get get) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      return transaction.get(get);
+    }
+
+    @Override
+    public List<Result> scan(Scan scan) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      return transaction.scan(scan);
+    }
+
+    @Override
+    public void put(Put put) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      transaction.put(put);
+    }
+
+    @Override
+    public void put(List<Put> puts) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      transaction.put(puts);
+    }
+
+    @Override
+    public void delete(Delete delete) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      transaction.delete(delete);
+    }
+
+    @Override
+    public void delete(List<Delete> deletes) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      transaction.delete(deletes);
+    }
+
+    @Override
+    public void mutate(List<? extends Mutation> mutations) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      transaction.mutate(mutations);
+    }
+
+    @Override
+    public void commit() throws CommitException, UnknownTransactionStatusException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      try {
+        transaction.commit();
+        status = Status.COMMITTED;
+      } catch (Exception e) {
+        status = Status.COMMIT_FAILED;
+        throw e;
+      }
+    }
+
+    @Override
+    public void rollback() throws RollbackException {
+      if (status == Status.COMMITTED || status == Status.ROLLED_BACK) {
+        throw new IllegalStateException(
+            "The transaction has already been committed or rolled back");
+      }
+      try {
+        transaction.rollback();
+      } finally {
+        status = Status.ROLLED_BACK;
+      }
+    }
+
+    private void checkStatus(@Nullable String message, Status... expectedStatus) {
+      boolean expected = Arrays.stream(expectedStatus).anyMatch(s -> status == s);
+      if (!expected) {
+        throw new IllegalStateException(message);
+      }
+    }
+
+    @Override
+    public DistributedTransaction getOriginalTransaction() {
+      if (transaction instanceof WrappedDistributedTransaction) {
+        return ((WrappedDistributedTransaction) transaction).getOriginalTransaction();
+      }
       return transaction;
     }
   }
--- a/core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
@@ -193,6 +193,11 @@ public abstract class AbstractDistributedTransactionManager
     }
   }
 
+  /**
+   * This class is to unify the call sequence of the transaction object. It doesn't care about the
+   * potential inconsistency between the status field on JVM memory and the underlying persistent
+   * layer.
+   */
   @VisibleForTesting
   static class StateManagedTransaction extends AbstractDistributedTransaction
       implements WrappedDistributedTransaction {
--- a/core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
@@ -113,7 +113,7 @@ public abstract class AbstractDistributedTransactionManager
                         + "It might have been expired"));
   }
 
-  protected DistributedTransaction wrap(DistributedTransaction transaction)
+  protected DistributedTransaction activate(DistributedTransaction transaction)
       throws TransactionException {
     return new ActiveTransaction(new StateManagedTransaction(transaction));
   }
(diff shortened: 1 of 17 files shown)


core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
```diff
@@ -177,8 +184,119 @@ public void rollback() throws RollbackException {
...
+    public DistributedTransaction getOriginalTransaction() {
+      if (transaction instanceof WrappedDistributedTransaction) {
+        return ((WrappedDistributedTransaction) transaction).getOriginalTransaction();
+      }
+      return transaction;
+    }
+  }
+
+  @VisibleForTesting
+  static class TransactionStateManagement extends AbstractDistributedTransaction
```
brfrn169, line 197: Added `TransactionStateManagement` to manage the transaction state.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
```diff
@@ -177,8 +184,119 @@ public void rollback() throws RollbackException {
...
+      implements WrappedDistributedTransaction {
+
+    private enum Status {
+      ACTIVE,
+      COMMITTED,
+      COMMIT_FAILED,
+      ROLLED_BACK
+    }
```
brfrn169, line 205: These are the transaction states.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
```diff
@@ -177,8 +184,119 @@ public void rollback() throws RollbackException {
...
+
+    @Override
+    public String getId() {
+      return transaction.getId();
+    }
+
+    @Override
+    public Optional<Result> get(Get get) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
```
brfrn169, line 223: Allow to call these CRUD methods only when the status is `ACTIVE`.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
```diff
@@ -177,8 +184,119 @@ public void rollback() throws RollbackException {
...
+
+    @Override
+    public void mutate(List<? extends Mutation> mutations) throws CrudException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
+      transaction.mutate(mutations);
+    }
+
+    @Override
+    public void commit() throws CommitException, UnknownTransactionStatusException {
+      checkStatus("The transaction is not active", Status.ACTIVE);
```
brfrn169, line 265: Allow to call this `commit` method only when the status is `ACTIVE`.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
```diff
@@ -177,8 +184,119 @@ public void rollback() throws RollbackException {
...
+        status = Status.COMMITTED;
+      } catch (Exception e) {
+        status = Status.COMMIT_FAILED;
+        throw e;
+      }
+    }
+
+    @Override
+    public void rollback() throws RollbackException {
+      if (status == Status.COMMITTED || status == Status.ROLLED_BACK) {
```
brfrn169, line 277: Allow to call this `rollback()` method only when the status is `ACTIVE` or `COMMIT_FAILED`.

core/src/main/java/com/scalar/db/transaction/common/AbstractTwoPhaseCommitTransactionManager.java
```diff
@@ -188,8 +195,148 @@ public void rollback() throws RollbackException {
...
+    public TwoPhaseCommitTransaction getOriginalTransaction() {
+      if (transaction instanceof WrappedTwoPhaseCommitTransaction) {
+        return ((WrappedTwoPhaseCommitTransaction) transaction).getOriginalTransaction();
+      }
+      return transaction;
+    }
+  }
+
+  @VisibleForTesting
+  static class TransactionStateManagement extends AbstractTwoPhaseCommitTransaction
```
brfrn169, line 208: Added `TransactionStateManagement` to manage the transaction state for 2PC.

core/src/main/java/com/scalar/db/transaction/common/AbstractTwoPhaseCommitTransactionManager.java
```diff
@@ -188,8 +195,148 @@ public void rollback() throws RollbackException {
...
+    private enum Status {
+      ACTIVE,
+      PREPARED,
+      PREPARE_FAILED,
+      VALIDATED,
+      VALIDATION_FAILED,
+      COMMITTED,
+      COMMIT_FAILED,
+      ROLLED_BACK
+    }
```
brfrn169, line 220: These are the transaction states for 2PC.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -183,7 +183,7 @@ DistributedTransaction begin(String txId, Isolation isolation, SerializableStrat
     ConsensusCommit consensus = new ConsensusCommit(crud, commit, recovery);
     getNamespace().ifPresent(consensus::withNamespace);
     getTable().ifPresent(consensus::withTable);
-    return new ActiveTransaction(consensus);
+    return wrap(consensus);
```
brfrn169, line 186: Wrap the transaction object.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
```diff
@@ -142,7 +142,7 @@ private TwoPhaseCommitTransaction createNewTransaction(
         new TwoPhaseConsensusCommit(crud, commit, recovery, isCoordinator);
     getNamespace().ifPresent(transaction::withNamespace);
     getTable().ifPresent(transaction::withTable);
-    return new ActiveTransaction(transaction);
+    return wrap(transaction);
```
brfrn169, line 145: Wrap the transaction object.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransactionManager.java
```diff
@@ -79,7 +79,7 @@ public DistributedTransaction begin(String txId) throws TransactionException {
           new JdbcTransaction(txId, jdbcService, dataSource.getConnection(), rdbEngine);
       getNamespace().ifPresent(transaction::withNamespace);
       getTable().ifPresent(transaction::withTable);
-      return new ActiveTransaction(transaction);
+      return wrap(transaction);
```
brfrn169, line 82: Ditto.

core/src/main/java/com/scalar/db/transaction/rpc/GrpcTransactionManager.java
```diff
@@ -103,7 +103,7 @@ private DistributedTransaction beginInternal(@Nullable String txId) throws Trans
           GrpcTransaction transaction = new GrpcTransaction(transactionId, stream);
           getNamespace().ifPresent(transaction::withNamespace);
           getTable().ifPresent(transaction::withTable);
-          return new ActiveTransaction(transaction);
+          return wrap(transaction);
```
brfrn169, line 106: Wrap the transaction object.

core/src/main/java/com/scalar/db/transaction/rpc/GrpcTransactionManager.java
```diff
@@ -126,7 +126,7 @@ private DistributedTransaction startInternal(@Nullable String txId) throws Trans
...
```
brfrn169, line 129: Ditto.

core/src/main/java/com/scalar/db/transaction/rpc/GrpcTwoPhaseCommitTransactionManager.java
```diff
@@ -89,7 +89,7 @@ private TwoPhaseCommitTransaction beginInternal(@Nullable String txId)
               new GrpcTwoPhaseCommitTransaction(transactionId, stream);
           getNamespace().ifPresent(transaction::withNamespace);
           getTable().ifPresent(transaction::withTable);
-          return new ActiveTransaction(transaction);
+          return wrap(transaction);
```
brfrn169, line 92: Wrap the transaction object.

core/src/main/java/com/scalar/db/transaction/rpc/GrpcTwoPhaseCommitTransactionManager.java
```diff
@@ -114,7 +114,7 @@ private TwoPhaseCommitTransaction startInternal(@Nullable String txId)
...
```
brfrn169, line 117: Ditto.

core/src/main/java/com/scalar/db/transaction/rpc/GrpcTwoPhaseCommitTransactionManager.java
```diff
@@ -129,7 +129,7 @@ public TwoPhaseCommitTransaction join(String txId) throws TransactionException {
               new GrpcTwoPhaseCommitTransaction(txId, stream);
           getNamespace().ifPresent(transaction::withNamespace);
           getTable().ifPresent(transaction::withTable);
-          return new ActiveTransaction(transaction);
+          return wrap(transaction);
```
brfrn169, line 132: Ditto.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
```diff
@@ -111,13 +113,18 @@ public DistributedTransaction resume(String txId) throws TransactionException {
                         + "It might have been expired"));
   }
 
-  @VisibleForTesting
-  public class ActiveTransaction extends AbstractDistributedTransaction {
+  protected DistributedTransaction wrap(DistributedTransaction transaction)
```
feeblefakie, line 116: Personally, `wrap` doesn't sound very meaningful, so how about renaming it to `activate` or something?
```suggestion
  protected DistributedTransaction activate(DistributedTransaction transaction)
```

feeblefakie (commented):
The code looks good to me!
I left one suggestion on the naming or `wrap` method. PTAL!

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
```diff
@@ -177,8 +184,119 @@ public void rollback() throws RollbackException {
...
+    @Override
+    public void rollback() throws RollbackException {
+      if (status == Status.COMMITTED || status == Status.ROLLED_BACK) {
+        throw new IllegalStateException(
+            "The transaction has already been committed or rolled back");
+      }
+      try {
+        transaction.rollback();
+      } finally {
+        status = Status.ROLLED_BACK;
```
komamitsu, line 284: Just a confirmation. Is it always safe to change the status to ROLLED_BACK even when `transaction.rollback()` fails?

komamitsu: My concern is any inconsistency between `status` field on JVM memory and underlying persistent layer.

core/src/main/java/com/scalar/db/transaction/common/AbstractTwoPhaseCommitTransactionManager.java
komamitsu, line 208: This class extends (and implements) `XxxxxxCommitTransaction`, but the class name sounds a bit different layer.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
komamitsu: `withActiveStatus` also sounds okay? It might be a bit verbose.

core/src/main/java/com/scalar/db/transaction/common/AbstractTwoPhaseCommitTransactionManager.java
feeblefakie: Something like StateManagedTransaction or StateManageableTransaction?
(both don't sound very cool, though...)

BTW, it might not make sense, but managing active or not is part of state management, so can we unify `ActiveTransaction` and `TransactionStateManagement`?

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
```diff
@@ -177,8 +184,119 @@ public void rollback() throws RollbackException {
...
+
+    private enum Status {
+      ACTIVE,
+      COMMITTED,
+      COMMIT_FAILED,
+      ROLLED_BACK
+    }
+
+    private final DistributedTransaction transaction;
+    private Status status;
```
komamitsu, line 208: It's better to make this `volatile` or use AtomicReference so that an update status is visible to other threads? For instance, a thread processing `commit()` method could fail to read the latest `ROLLED_BACK` status just updated by another thread that called `rollback()` method.

komamitsu: Ah, ConsensusCommit and etc. which inherit AbstractDistributedTransaction have `@NotThreadSafe` annotation, so it's okay?

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
brfrn169: This method wraps the transaction object with `ActiveTransaction` and `TransactionStateManagement` for now (we might want to add another wrapper in the future). I feel the names `activate`, `withActiveStatus` and `asActive` focus on only `ActiveTransaction`. Is that correct?

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
brfrn169: The purpose of this state management is just to unify the call sequence of the transaction object (e.g. CRUD -> prepare -> validate -> commit or CRUD -> prepare (failed) -> rollback, etc.). So we don't need to synchronize the `status` field on JVM memory and underlying persistent layer strictly.

core/src/main/java/com/scalar/db/transaction/common/AbstractTwoPhaseCommitTransactionManager.java
brfrn169: Honestly, I didn't think it was a good name for this class, but I didn't have any other idea. 

> BTW, it might not make sense, but managing active or not is part of state management, so can we unify ActiveTransaction and TransactionStateManagement?

Maybe, I feel it's better to have separate classes for those. I will rename it to `StateManagedTransaction`. Thank you. BTW, it's a private inner class so we can rename it whenever we want in the future.

brfrn169 (commented):
@feeblefakie @komamitsu Thank you for your feedback! I replied your comments. I will modify this PR based on your feed back. Thanks.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
komamitsu: Sounds good! Thanks for the explanation.

I might forget the background in the future. Could you add a comment to describe that an instance of this class won't be reused and we don't need to consider the potential inconsistency?

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
komamitsu: I thought it's related to TransactionStateManagement as well because the initial value set in the constructor is ACTIVE.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
brfrn169: I added the comment in 3d7ab472db58e7e0c1d91896b1427c7f01effd44. Thanks.

core/src/main/java/com/scalar/db/transaction/common/AbstractDistributedTransactionManager.java
brfrn169: Okay, so I will rename it to `activate`. Thanks.

brfrn169: Renamed it in f97d5eeb724d40d80f7ad21812bf10757ae00a0e. Thanks.

core/src/main/java/com/scalar/db/transaction/common/AbstractTwoPhaseCommitTransactionManager.java
brfrn169: Renamed it in 76c9ae03b1695099a82c97cc7ac84b978378eda2. Thanks.

komamitsu (approved):
👍

feeblefakie (approved):
LGTM! Thank you!

brfrn169:
Merging this for now.

@Torch3333 Can you please take a look when you come back?

PR #721, opened 2022-10-12
Add description of limitations of data types

kota2and3kan:
This PR adds the description of limitations of data types in the docs of the schema. For example, the range of `BIGINT` is described in the javadoc only. So, I added the link of javadoc to see the details of data types.

Please take a look!

### DIFF
--- a/docs/schema-loader.md
+++ b/docs/schema-loader.md
@@ -446,9 +446,9 @@ However, the following types in JDBC databases are converted differently when th
 | TEXT     | VARCHAR(64)   | VARCHAR(10485760) | VARCHAR2(64) |
 | BLOB     | VARBINARY(64) |                   | RAW(64)      |
 
-If this data type mapping doesn't match your application, please alter the tables to change the data types after creating them with this tool.
+Also, `BIGINT` of Scalar DB supports a value from -2^53 to 2^53 even if you use any databases.
 
-Also, the data types of Scalar DB have some limitations. For example, BIGINT has a limitation of range. Please refer to the [javadoc](https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html) (`com.scalar.db.io` package) for more details of the data types of Scalar DB.
+If this data type mapping doesn't match your application, please alter the tables to change the data types after creating them with this tool.
 
 ## Internal metadata for Consensus Commit
 
--- a/docs/schema-loader.md
+++ b/docs/schema-loader.md
@@ -446,7 +446,7 @@ However, the following types in JDBC databases are converted differently when th
 | TEXT     | VARCHAR(64)   | VARCHAR(10485760) | VARCHAR2(64) |
 | BLOB     | VARBINARY(64) |                   | RAW(64)      |
 
-Also, `BIGINT` of Scalar DB supports a value from -2^53 to 2^53 even if you use any databases.
+The value range of `BIGINT` in Scalar DB is from -2^53 to 2^53 regardless of the underlying database.
 
 If this data type mapping doesn't match your application, please alter the tables to change the data types after creating them with this tool.
 
--- a/docs/schema-loader.md
+++ b/docs/schema-loader.md
@@ -448,6 +448,8 @@ However, the following types in JDBC databases are converted differently when th
 
 If this data type mapping doesn't match your application, please alter the tables to change the data types after creating them with this tool.
 
+Also, the data types of Scalar DB have some limitations. For example, BIGINT has a limitation of range. Please refer to the [javadoc](https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html) (`com.scalar.db.io` package) for more details of the data types of Scalar DB.
+
 ## Internal metadata for Consensus Commit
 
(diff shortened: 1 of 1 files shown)


docs/schema-loader.md
```diff
@@ -448,6 +448,8 @@ However, the following types in JDBC databases are converted differently when th
 
 If this data type mapping doesn't match your application, please alter the tables to change the data types after creating them with this tool.
 
+Also, the data types of Scalar DB have some limitations. For example, BIGINT has a limitation of range. Please refer to the [javadoc](https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html) (`com.scalar.db.io` package) for more details of the data types of Scalar DB.
```
komamitsu, line 451: I tried to find the limitation on the Javadoc, but honestly speaking it was not easy to find a proper content.

This is just an idea, but how about adding a new column `note` or something to the above large table on this page and just writing down the actual value range in the column for BIGINT?

komamitsu: Or how about using https://javadoc.io/static/com.scalar-labs/scalardb/3.7.0/com/scalar/db/io/BigIntColumn.html instead?

kota2and3kan: @komamitsu  
Thank you for your review!

I added only a link to javadoc in this PR. This is because if we describe limitations in both javadoc and this document, it is duplicated and we need to maintain both docs when the limitation will be changed in the future. In other words, I would like to avoid redundancy part in the several docs.

Also, I considered adding a direct link to the `BIGINT` as follows (specify the URL `https://javadoc.io/static/com.scalar-labs/` instead of `https://javadoc.io/doc/com.scalar-labs/`).
https://javadoc.io/static/com.scalar-labs/scalardb/3.7.0/com/scalar/db/io/BigIntColumn.html

However, it seems that we need to specify the concrete version (e.g., `3.7.0`) in the URL if we access a specific class's javadoc (https://javadoc.io/static/). In other words, I was not able to specify `latest` instead of `3.7.0` in the above URL. In this case, we need to update docs when the new version of Scalar DB will be released in the future. So, I didn't add this URL in this document...
(I'm not familiar with the javadoc, but it would be grateful help for me if you know the way to specify `latest` in the above URL.)

If the changing limitation of BIGINTis a rare case, I think we can add the concrete limitation in this document directly even if it is redundant content.

@brfrn169  
Regarding the range of `BIGINT` (from -2^53 to 2^53), is there a possibility that it will be changed in the future frequently?
If no, I will consider adding a concrete description of the limitation of `BIGINT` to this document directly.

komamitsu: > I was not able to specify latest instead of 3.7.0 in the above URL. In this case, we need to update docs when the new version of Scalar DB will be released in the future. So, I didn't add this URL in this document...

Ah, good point..

brfrn169: > Regarding the range of BIGINT (from -2^53 to 2^53), is there a possibility that it will be changed in the future frequently?

@kota2and3kan No, it won't. So I think we can add a concrete description of the limitation of BIGINT to this document directly.

kota2and3kan: @brfrn169  
OK. I will do that. Thank you!

@komamitsu  
Thank you for your suggestion! At this time, I will add the limitation on this document directly.

kota2and3kan:
I updated docs based on the review comment.

First I tried to add the column in the table of data mapping. 
However, the data type that has Scalar DB specific limitation is `BIGINT` only at this time. 
So, the above new column has a lot of empty values and it looked like a bit strange.

Therefore, I added a simple one sentence to explain the limitation of `BIGINT`.

Please take a look again!

docs/schema-loader.md
```diff
@@ -446,6 +446,8 @@ However, the following types in JDBC databases are converted differently when th
 | TEXT     | VARCHAR(64)   | VARCHAR(10485760) | VARCHAR2(64) |
 | BLOB     | VARBINARY(64) |                   | RAW(64)      |
 
+Also, `BIGINT` of Scalar DB supports a value from -2^53 to 2^53 even if you use any databases.
```
komamitsu, line 449: Looks good.

Just another option: I think you can simply say "the value range of BIGINT is ..." without the term `supports` like
```
The value range of BIGINT in Scalar DB is from -2^53 to 2^53 regardless the underlying database.
```

komamitsu (approved):

docs/schema-loader.md
kota2and3kan: I see. Thank you for another option. I will keep this in mind for future reference.

It is difficult to simply and accurately describe information in a document...

feeblefakie: @kota2and3kan 
The one from @komamitsu sounds a bit more right in English (except for the missing `of`), so it would be great if you can update it as follows. (`even if` part sounds a bit weird to me, but English is difficult...)

```
The value range of BIGINT in Scalar DB is from -2^53 to 2^53 regardless of the underlying database.
```

feeblefakie (commented):
Thank you!
Left one suggestion.

docs/schema-loader.md
kota2and3kan: @feeblefakie  
Thank you for your review!
I updated it. Please take a look again!

feeblefakie (approved):
LGTM! Thank you!

PR #722, opened 2022-10-13
Add several methods to ActiveExpiringMap

brfrn169:
This PR adds several methods (`containsValue()`, `keySet()`, `values()`, and `entrySet()`) to `ActiveExpiringMap`. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
+++ b/core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
@@ -1,13 +1,23 @@
 package com.scalar.db.util;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.Uninterruptibles;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.AbstractCollection;
+import java.util.AbstractMap;
+import java.util.AbstractSet;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
 @ThreadSafe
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
```diff
@@ -83,6 +93,10 @@ public boolean contains(K key) {
     return map.containsKey(key);
   }
 
+  public boolean containsValue(V value) {
+    return map.values().stream().map(ValueHolder::get).anyMatch(value::equals);
```
komamitsu, line 97: It looks to create a whole intermediate collection by `map`.
```
    return map.values().stream().anyMatch(holder -> value.equals(holder.get()));
```
Maybe this can avoid it?

brfrn169: > t looks to create a whole intermediate collection by `map`.

Oh really? the `map()` method is an intermediate operation and also returns `Stream<R>`. Does it create a while intermediate collection?

komamitsu: Oops, I may have confused with other language's one. Sorry for the confusion.

core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
```diff
@@ -91,14 +105,98 @@ public V remove(K key) {
...
+          public boolean hasNext() {
+            return entryIterator.hasNext();
+          }
+
+          @SuppressFBWarnings("SE_BAD_FIELD")
+          @Override
+          public Entry<K, V> next() {
+            Entry<K, ValueHolder<V>> next = entryIterator.next();
+            return new AbstractMap.SimpleEntry<K, V>(next.getKey(), next.getValue().get()) {
+              @Override
```
komamitsu, line 164: Is this override needed?

brfrn169: You mean whether or not we really need to override the `setValue()` method and make it throw `UnsupportedOperationException`, right?

We expect that `setValue()` replaces the actual value of the map's entry, but I don't think we need it now, so I override it.

Do you think we need to implement it?

komamitsu: > but I don't think we need it now, so I override it.

I got your point. Sounds good!

The class doesn't seem to guarantee immutability, so I thought we also can allow users to modify values. But I don't have any strong opinion on it.

komamitsu (approved):
👍

core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
komamitsu: >  I thought we also can allow users to modify values
Ah, unlike java.util.HashMap which exposes internal HashMap.Node via iterator, ActiveExpiringMap creates an entry copying the original key and value. So, we need to do something if we want ActiveExpiringMap to behave like HashMap. I agree with your original idea now.

feeblefakie (approved):
LGTM! Thank you!

brfrn169:
Merging this for now.

@Torch3333 Can you please take a look when you come back?

Torch3333 (commented):
LGTM, thank you!

PR #724, opened 2022-10-19
Small integration test modifications for the resume feature

brfrn169:
Please take a look!

### DIFF
--- a/integration-test/src/main/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
+++ b/integration-test/src/main/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
@@ -910,7 +910,7 @@ public abstract class DistributedTransactionIntegrationTestBase {
     DistributedTransaction resumed = manager.resume(transaction.getId());
 
     // Assert
-    assertThat(resumed).isEqualTo(transaction);
+    assertThat(resumed.getId()).isEqualTo(transaction.getId());
 
     transaction.commit();
   }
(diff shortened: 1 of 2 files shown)


integration-test/src/main/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
```diff
@@ -910,7 +910,7 @@ public void resume_WithBeginningTransaction_ShouldReturnBegunTransaction()
     DistributedTransaction resumed = manager.resume(transaction.getId());
 
     // Assert
-    assertThat(resumed).isEqualTo(transaction);
+    assertThat(resumed.getId()).isEqualTo(transaction.getId());
```
brfrn169, line 913: Instead of comparing a transaction object itself, we should compare the transaction ID.

integration-test/src/main/java/com/scalar/db/api/TwoPhaseCommitTransactionIntegrationTestBase.java
```diff
@@ -1134,7 +1134,7 @@ public void resume_WithBeginningTransaction_ShouldReturnBegunTransaction()
     TwoPhaseCommitTransaction resumed = manager1.resume(transaction.getId());
 
     // Assert
-    assertThat(resumed).isEqualTo(transaction);
+    assertThat(resumed.getId()).isEqualTo(transaction.getId());
```
brfrn169, line 1137: Ditto.

integration-test/src/main/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
komamitsu: Looks good. Another option is overriding `DistributedTransaction#equals` to compare transaction ID?

komamitsu (approved):
LGTM! 👍

integration-test/src/main/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
brfrn169: I think, yes, that's an option. But I'm not sure it's okay to consider only transaction ID for the equivalence of transactions. So I will keep it for now. Thank you for the comment.

feeblefakie (approved):
LGTM! Thank you!

PR #728, opened 2022-10-25
Fix dependency and spotbugs issues for >JDK 8

komamitsu:
A similar change to https://github.com/scalar-labs/scalardb-sql/pull/74. PTAL~

> I noticed we can't build this project with JDK 9 or later, but JDK 8 isn't actively supported. So I tried supporting JDK 9 or later by fixing build dependency and spotbugs issues while I tested it only with JDK 11.
> The package of this project is built on CI using JDK 8, so I don't think this change affects the Maven package artifacts.

~"spotlessApply" gradle task changed the Javadoc comments this time, though.~ 
Hmm. ~This~ `spotlessCheck` error happened only on my laptop. Reverted the changes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,6 +34,7 @@ subprojects {
         sqlserverDriverVersion = '8.4.1.jre8'
         grpcVersion = '1.46.0'
         protobufVersion = '3.20.1'
+        annotationVersion = '1.3.2'
         picocliVersion = '4.1.4'
         scalarAdminVersion = '1.2.0'
         dropwizardMetricsVersion = '4.2.2'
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/scalar/db/api/Isolation.java
```diff
@@ -1,6 +1,8 @@
 package com.scalar.db.api;
 
-/** @deprecated As of release 3.5.0. Will be removed in release 5.0.0 */
+/**
```
komamitsu, line 3: "spotlessApply" gradle task changed these Javadoc comments

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #729, opened 2022-10-27
Throw an exception when putting an empty value to a secondary index column in Dynamo DB

brfrn169:
Since Dynamo DB doesn't allow putting an empty value to a secondary index column, we should throw an exception in that situation. Please take a look!


feeblefakie (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM. Thanks!

Torch3333 (approved):
LGTM, thank you!

brfrn169:
The vulnerability scanner error (CVE-2022-40153) is caused by `com.fasterxml.woodstox:woodstox-core`, and the cosmos DB client depends on it. We need to wait for the dependency of the cosmos client to be updated.

PR #730, opened 2022-10-31
Automatically terminate idle parallel execution threads not to block JVM shutdown

komamitsu:
In current implementation, `scalar.db.consensus_commit.parallel_commit.enabled` is true by default. As a result, JVM process won't finish without explicitly calling `System.exit()` or receiving signals. This might block some kinds of applications or components.

In this PR, I'm introducing a background monitor task to terminate `ParallelExecutor#parallelExecutorService` when it's idle. The terminated `parallelExecutorService` will be re-setup when it's needed. This won't cause any error.


komamitsu:
Closing this as we go with https://github.com/scalar-labs/scalardb/pull/733

PR #733, opened 2022-11-01
Make ParallelExecutor workers daemon thread

komamitsu:
In current implementation, `scalar.db.consensus_commit.parallel_commit.enabled` is true by default. As a result, JVM process won't finish without explicitly calling `ParallelExecutor#close`, `System.exit` or receiving signals. This might block some kinds of applications or components which don't have chance to call `ParallelExecutor#close` for some reasons.

In this PR, we've changed ParallelExecutor workers to daemon threads not to block JVM process termination. Even if a user don't have chance to call `ParallelExecutor#close` for some reasons, `java.lang.Runtime#addShutdownHook` can be used to call the method as it's guaranteed that daemon threads will continue to run during the shutdown sequence https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Runtime.html#addShutdownHook(java.lang.Thread)

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
@@ -38,7 +38,14 @@ public class ParallelExecutor {
       parallelExecutorService =
           Executors.newFixedThreadPool(
               config.getParallelExecutorCount(),
-              new ThreadFactoryBuilder().setNameFormat("parallel-executor-%d").build());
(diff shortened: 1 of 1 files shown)


brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

komamitsu:
Thanks for the reviews!

But I noticed the CVE check failed and I can't merge this PR.

@brfrn169 Could you merge this for me?

komamitsu:
Thanks!

PR #734, opened 2022-11-02
Make JDK11+ use the same google java format version as JDK8 uses

komamitsu:
I noticed there were sometimes differences between the spotlessCheck results on my laptop (JDK11) and CI (JDK8.) It seems different versions of https://github.com/google/google-java-format are internally used via the spotless plugin. This PR specifies the google-java-format version to make JDK11+ use the same version as JDK8 uses.


Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #736, opened 2022-11-07
Add upgrade command to schema loader

Torch3333:
**Context**
This PR is part of the namespace management feature and targets its feature branch [feat/namespaces_management](https://github.com/scalar-labs/scalardb/tree/feat/namespaces_management)

**Changes** 
This PR adds the `--upgrade` command to the schema loader which is identical to calling the [admin.upgrade(Map<String, String options)](https://github.com/scalar-labs/scalardb/blob/9fce0686aa0a2d2b07f6562a5ea17c5d6638ed6c/core/src/main/java/com/scalar/db/api/Admin.java#L392-L399) method. 
For a concrete explanation of what running this method does, please refer to this [PR](https://github.com/scalar-labs/scalardb/pull/696#issue-1367431598)

### DIFF
--- a/integration-test/src/main/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
+++ b/integration-test/src/main/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
@@ -414,7 +414,7 @@ public abstract class SchemaLoaderIntegrationTestBase {
 
     // Assert
     assertThat(exitCodeUpgrade).isZero();
-    waitForKeyspacesTableCreation();
+    waitForNamespacesTableCreation();
     assertThat(transactionAdmin.getNamespaceNames()).containsOnly(namespace1, namespace2);
   }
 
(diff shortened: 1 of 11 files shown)


komamitsu (approved):
LGTM! 👍

integration-test/src/main/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
```diff
@@ -423,4 +447,8 @@ private void deleteTables_ShouldDeleteTablesWithCoordinator() throws Exception {
   private int executeWithArgs(List<String> args) {
     return SchemaLoader.mainInternal(args.toArray(new String[0]));
   }
+
+  protected void waitForKeyspacesTableCreation() {
```
brfrn169, line 451: I know it's only for Cassandra for now, but I don't think we should use the Cassandra term `keyspaces` here? Maybe we should use `namespaces` instead?

```suggestion
  protected void waitForNamespacesTableCreation() {
```

brfrn169 (commented):
LGTM overall. Thank you. Left one minor comment. Please take a look when you have time!

integration-test/src/main/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
Torch3333: Good point, thank you. Let me update it.

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #737, opened 2022-11-14
Move vuln check to a separate daily workflow

komamitsu:
Current main CI workflow contains vulnerability check tasks and they sometimes results in CI failure due to third party security issue that we can't always handle on our side. This PR moves the check to a separate daily scheduled workflow not to block our PR flows.

We'll take care of vuln check for some released versions as a different PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,47 +87,6 @@ jobs:
         with:
           arguments: :schema-loader:dockerfileLint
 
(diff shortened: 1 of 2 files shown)


brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #740, opened 2022-11-16
Should drop namespace only when no tables are in the namespace in Schema Loader

brfrn169:
Currently, when Schema Loader runs with `-D`, it always tries to drop all the namespaces in the schema file. So, if a namespace has tables other than the ones in the schema file, it shows an error or drops the namespace with the tables in the namespace, depending on the storage implementations. I think Schema Loader should drop the namespace only when no tables are in the namespace. This PR fixes this issue. Please take a look!


feeblefakie (approved):
LGTM! Thank you!
Making the behavior consistent is definitely the first step to go, but I feel it would probably be useful (since I do it during my development) if we have the option to delete a namespace with tables for debugging purposes.

komamitsu (approved):
LGTM! 👍

kota2and3kan (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

brfrn169:
> Making the behavior consistent is definitely the first step to go,

@feeblefakie Correct, we should do that. I will put it into our task list. Thanks.

brfrn169:
It's actually a bug fix, so I changed the label to `bugfix`. Also, we will include this change in `3.6.8` and `3.7.7`.

PR #742, opened 2022-11-18
Remove unnecessary key definition from API guide

kota2and3kan:
It seems that there are some key definitions in the sample of the `all()` method of the `Scan` API.
However, I think we don't need to set the partition key and clustering key if we use the `Scan.all()`.

So, this PR removes the above unnecessary key definitions in the sample of `Scan.all()`.
Please take a look!


brfrn169 (approved):
LGTM! Thank you!

komamitsu (approved):
Good catch! Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #744, opened 2022-11-28
Move common classes

brfrn169:
This PR just moves the common classes under `com.scalar.db.storage.common` and `com.scalar.db.storage.common` to `com.scalar.db.common`. The classes are only for internal use, so it should not be an incompatible change. Please take a look!


komamitsu (approved):
LGTM! 👍

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #746, opened 2022-11-30
Throw TransactionNotFoundException when resuming a transaction but it's not found

brfrn169:
This PR adds `TransactionNotFoundException` and throws it when resuming a transaction but it's not found. We need this because we need special handling for the transaction not found situation, so it is helpful that we can catch a dedicated exception for that situation. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
@@ -1,6 +1,7 @@
 package com.scalar.db.api;
 
 import com.scalar.db.exception.transaction.TransactionException;
+import com.scalar.db.exception.transaction.TransactionNotFoundException;
 import java.util.Optional;
 
 public interface DistributedTransactionManager {
@@ -182,9 +183,10 @@ public interface DistributedTransactionManager {
    *
    * @param txId the transaction ID
    * @return {@link DistributedTransaction}
-   * @throws TransactionException if resuming the transaction failed
+   * @throws TransactionNotFoundException if the transaction associated with the specified
+   *     transaction ID is not found
    */
-  DistributedTransaction resume(String txId) throws TransactionException;
+  DistributedTransaction resume(String txId) throws TransactionNotFoundException;
 
   /**
    * Returns the state of a given transaction.
--- a/core/src/main/java/com/scalar/db/api/TwoPhaseCommitTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/api/TwoPhaseCommitTransactionManager.java
@@ -1,6 +1,7 @@
 package com.scalar.db.api;
 
 import com.scalar.db.exception.transaction.TransactionException;
+import com.scalar.db.exception.transaction.TransactionNotFoundException;
 import java.util.Optional;
 
 /**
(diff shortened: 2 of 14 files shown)


core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
```diff
@@ -182,9 +183,10 @@ DistributedTransaction start(String txId, Isolation isolation, SerializableStrat
    *
    * @param txId the transaction ID
    * @return {@link DistributedTransaction}
-   * @throws TransactionException if resuming the transaction failed
+   * @throws TransactionNotFoundException if the transaction associated with the specified
+   *     transaction ID is not found
    */
-  DistributedTransaction resume(String txId) throws TransactionException;
+  DistributedTransaction resume(String txId) throws TransactionNotFoundException;
```
brfrn169, line 189: Throw `TransactionNotFoundException` if the transaction associated with the specified transaction ID is not found.

core/src/main/java/com/scalar/db/api/TwoPhaseCommitTransactionManager.java
```diff
@@ -116,9 +117,10 @@ default TwoPhaseCommitTransaction start(String txId) throws TransactionException
    *
    * @param txId the transaction ID
    * @return {@link TwoPhaseCommitTransaction}
-   * @throws TransactionException if resuming the transaction failed
+   * @throws TransactionNotFoundException if the transaction associated with the specified
+   *     transaction ID is not found
    */
-  TwoPhaseCommitTransaction resume(String txId) throws TransactionException;
+  TwoPhaseCommitTransaction resume(String txId) throws TransactionNotFoundException;
```
brfrn169, line 123: Ditto.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTransactionManager.java
```diff
@@ -48,7 +49,7 @@ public ActiveTransactionManagedTransactionManager(DatabaseConfig config) {
   private void add(DistributedTransaction transaction) throws TransactionException {
     if (activeTransactions.putIfAbsent(transaction.getId(), transaction) != null) {
       transaction.rollback();
-      throw new IllegalStateException(
+      throw new TransactionException(
```
komamitsu, line 52: There is no change to expect or catch TransactionException in this PR. Does it mean we don't have any test case for this path? This might be a bit off-topic, though.

komamitsu (approved):
I added a minor comment. But this PR itself LGTM! 👍

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTransactionManager.java
brfrn169: Ah correct, we don't have tests for it. I will add them in this PR. Thanks!

brfrn169: Added unit tests in 75bbd48d22c581069545ad6d7acb1787b99598bb.

@komamitsu Please take a look again when you have time!

komamitsu: LGTM! Thank you!

komamitsu (approved):

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #750, opened 2022-12-05
Deprecate getPartitionKey() and getClusteringKey() of Result

brfrn169:
This PR deprecates the `getPartitionKey()` and `getClusteringKey()` methods of the `Result` class. This is because those methods are more like utility ones, so we should separate them into another utility class. I will add that utility class in another PR, and this PR just deprecates the methods. Please take a look!


Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

PR #752, opened 2022-12-07
Validate contact points size in XxxxConfig to avoid "java.lang.ArrayIndexOutOfBoundsException: 0"

komamitsu:
When I forgot to set `scalar.db.contact_points`, I ran into `java.lang.ArrayIndexOutOfBoundsException: 0` exception. It's not so user friendly for users. This PR adds a validation to confirm if `contactPoints` isn't empty in XxxxConfig classes.

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cosmos/CosmosConfig.java
+++ b/core/src/main/java/com/scalar/db/storage/cosmos/CosmosConfig.java
@@ -23,6 +23,9 @@ public class CosmosConfig {
       throw new IllegalArgumentException(DatabaseConfig.STORAGE + " should be 'cosmos'");
     }
 
+    if (databaseConfig.getContactPoints().isEmpty()) {
+      throw new IllegalArgumentException(DatabaseConfig.CONTACT_POINTS + " is empty");
+    }
     endpoint = databaseConfig.getContactPoints().get(0);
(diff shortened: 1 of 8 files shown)


brfrn169 (commented):
Thank you for handling this! Maybe we can add the validation to [DatabaseConfig](https://github.com/scalar-labs/scalardb/blob/1709a213defdbdc143b0bb1b2f44408eb9d65c84/core/src/main/java/com/scalar/db/config/DatabaseConfig.java) instead of the `XXXConfig`s?

komamitsu:
@brfrn169 I first thought the idea, but found the property seemed an optional (e.g. ConsensusCommitConfig doesn't use it.) Should we treat the property as a mandatory one by changing those usages which don't set the property?

brfrn169 (approved):
@komamitsu Ah, correct, the property is optional. Then I think the change is fine with me. Thanks.

komamitsu:
Another option is to create a class inheriting DatabaseConfig that treats contact points as a mandatory with a validation and make the 4 XxxxConfigs which use the property of DatabaseConfig use the new one?

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #759, opened 2022-12-17
docs: Clearer description on when to use online backups

laysakura:
## Problem

`backup-restore.md` could be read as the following:

- It is safe to make backups for MySQL and PostgreSQL (both are under a Scalar DB instance) using `mysqldump` and `pg_dump` while the Scalar DB instance is accepting new transactions

But it should (quite naturally) cause inconsistent tables among the MySQL and PostgreSQL.

## What I did

- Divide two ways to create backups: online or using pause duration
- Add a flowchart for when to use online backup
- Clearly write the responsibility of creating an online backup belongs to users
- Break "JDBC database" sections into ones for each DB product


laysakura:
Close this fork version in favor of: https://github.com/scalar-labs/scalardb/pull/761

PR #761, opened 2022-12-18
docs: Clearer description on when to use online backups

laysakura:
## Problem

`backup-restore.md` could be read as the following:

- It is safe to make backups for MySQL and PostgreSQL (both are under a Scalar DB instance) using `mysqldump` and `pg_dump` while the Scalar DB instance is accepting new transactions

But it should (quite naturally) cause inconsistent tables among the MySQL and PostgreSQL.

## What I did

- Divide two ways to create backups: online or using pause duration
- Add a flowchart for when to use online backup
- Clearly write the responsibility of creating an online backup belongs to users
- Break "JDBC database" sections into ones for each DB product

PR #765, opened 2022-12-18
docs: add SQLite data type mapping in schema-loader doc

laysakura:
Indeed these types are valid in SQLite 3:

```console
% sqlite3
SQLite version 3.37.0 2021-12-09 01:34:53
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> create table t2 (c1 boolean, c2 int, c3 bigint, c4 float, c5 double, c6 text, c7 blob);
```

Note that types in SQLite are quite loose and column types do not affect value types.
(see: https://www.sqlite.org/datatype3.html and https://www.sqlite.org/flextypegood.html if interested)

### DIFF
--- a/docs/schema-loader.md
+++ b/docs/schema-loader.md
@@ -429,15 +429,15 @@ By default, the schema loader enables auto-scaling of RU for all tables: RU is s
 
 Here are the supported data types in Scalar DB and their mapping to the data types of other databases.
 
-| Scalar DB | Cassandra | Cosmos DB      | DynamoDB | MySQL    | PostgreSQL       | Oracle         | SQL Server      |
-|-----------|-----------|----------------|----------|----------|------------------|----------------|-----------------|
-| BOOLEAN   | boolean   | boolean (JSON) | BOOL     | boolean  | boolean          | number(1)      | bit             |
-| INT       | int       | number (JSON)  | N        | int      | int              | int            | int             |
-| BIGINT    | bigint    | number (JSON)  | N        | bigint   | bigint           | number(19)     | bigint          |
-| FLOAT     | float     | number (JSON)  | N        | double   | float            | binary_float   | float(24)       |
-| DOUBLE    | double    | number (JSON)  | N        | double   | double precision | binary_double  | float           |
-| TEXT      | text      | string (JSON)  | S        | longtext | text             | varchar2(4000) | varchar(8000)   |
-| BLOB      | blob      | string (JSON)  | B        | longblob | bytea            | RAW(2000)      | varbinary(8000) |
+| Scalar DB | Cassandra | Cosmos DB      | DynamoDB | MySQL    | PostgreSQL       | Oracle         | SQL Server      | SQLite  |
+|-----------|-----------|----------------|----------|----------|------------------|----------------|-----------------|---------|
+| BOOLEAN   | boolean   | boolean (JSON) | BOOL     | boolean  | boolean          | number(1)      | bit             | boolean |
+| INT       | int       | number (JSON)  | N        | int      | int              | int            | int             | int     |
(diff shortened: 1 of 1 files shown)


brfrn169 (approved):
LGTM! Thank you!

docs/schema-loader.md
```diff
@@ -429,15 +429,15 @@ By default, the schema loader enables auto-scaling of RU for all tables: RU is s
...
-| FLOAT     | float     | number (JSON)  | N        | double   | float            | binary_float   | float(24)       |
-| DOUBLE    | double    | number (JSON)  | N        | double   | double precision | binary_double  | float           |
-| TEXT      | text      | string (JSON)  | S        | longtext | text             | varchar2(4000) | varchar(8000)   |
-| BLOB      | blob      | string (JSON)  | B        | longblob | bytea            | RAW(2000)      | varbinary(8000) |
+| Scalar DB | Cassandra | Cosmos DB      | DynamoDB | MySQL    | PostgreSQL       | Oracle         | SQL Server      | SQLite  |
+|-----------|-----------|----------------|----------|----------|------------------|----------------|-----------------|---------|
+| BOOLEAN   | boolean   | boolean (JSON) | BOOL     | boolean  | boolean          | number(1)      | bit             | boolean |
+| INT       | int       | number (JSON)  | N        | int      | int              | int            | int             | int     |
+| BIGINT    | bigint    | number (JSON)  | N        | bigint   | bigint           | number(19)     | bigint          | bigint  |
```
komamitsu, line 436: It seems SQLite3 has 5 Type Affinities https://www.sqlite.org/datatype3.html and `bigint` and `float` are interpreted as one of the type affinities. Using the type affinity name is a bit more correct?

komamitsu (approved):
LGTM! 👍 although I left a minor comment.

docs/schema-loader.md
komamitsu: Ah, you've already mentioned the page above. Thanks. Noticed Column Data type (INTEGER, REAL, TEXT, BLOB) might be better.

laysakura: I understand what you say.

On the other hand, we can even state all types are mapped to INTEGER because SQLite in reality has type for each value (not for column)

image

(https://www.sqlite.org/flextypegood.html).

Therefore, we may choose the one from the following (and I feel 1. is the best).

1. Use the same type names as the Scalar DB
2. Use the storage classes (INTEGER, REAL, TEXT, or BLOB)
3. Use only INTEGER (should be the worst but possible choice)

komamitsu: Thanks for the explanation! That makes sense.

docs/backup-restore.md
```diff
@@ -5,19 +5,41 @@ This document sets out some guidelines for backing up and restoring the database
```
feeblefakie, line 5: Can you exclude the (non-related) changes from other PRs?
I have several comments on the backup guide.

feeblefakie (approved):
LGTM! Thank you!
The change for this PR is good, but the PR mixes up other changed (from other PRs), so it would be great if you  could exclude them.

docs/backup-restore.md
laysakura: Sorry, I got it.
I will arrange the branch after https://github.com/scalar-labs/scalardb/pull/761 is merged.

Torch3333 (approved):
LGTM, thank you!

docs/backup-restore.md
laysakura: I fixed the commit logs. I'll merge this PR after the CI passes.

PR #770, opened 2022-12-22
Allow to use placeholders in configuration values

brfrn169:
This PR allows to use placeholders in configuration values.

The placeholders are replaced with environment variables (`${env:<environment variable name>}`) or system properties (`${sys:<system property name>}`). Also, we can also specify default values in placeholders like `${sys:<system property name>:-defaultValue}`.

Please take a look!


core/build.gradle
```diff
@@ -97,6 +97,7 @@ dependencies {
     implementation "org.postgresql:postgresql:${postgresqlDriverVersion}"
     implementation "com.oracle.database.jdbc:ojdbc8:${oracleDriverVersion}"
     implementation "com.microsoft.sqlserver:mssql-jdbc:${sqlserverDriverVersion}"
+    implementation "org.apache.commons:commons-text:${commonsTextVersion}"
```
brfrn169, line 100: Use Apache Commons Text to use placeholders in the configuration values.

core/src/main/java/com/scalar/db/config/ConfigUtils.java
```diff
@@ -1,22 +1,49 @@
...
+ * name&#62;}). You can also specify default values in placeholders like ${sys:&#60;system property
+ * name&#62;:-defaultValue}.
+ */
 public final class ConfigUtils {
 
+  /**
+   * A {@link StringSubstitutor} instance with the environment variable string lookup and system
+   * property string lookup.
+   */
+  private static final StringSubstitutor stringSubstitutor =
```
brfrn169, line 24: Use `StringSubstitutor` in Apache Commons Text with the environment variable string lookup and system property string lookup.

By using it, we replace placeholders with environment variables and system properties.

komamitsu (approved):
LGTM! 👍

kota2and3kan (approved):
LGTM! Thank you!
This is a great feature since we can make the docker image more secure!

---

Just to be sure, I tested that the `Apache Commons Text syntax` and existing `dockerize (Go text/template) syntax` are conflicted or not.

As a result, there is no error or unexpected modification (replacing) by dockerize.

So, we can use both `Apache Commons Text syntax` and `dockerize (Go text/template) syntax` in the same configuration.
(i.e., it can keep the backward compatibility.)


1. Create a test `.tmpl` file that includes `Apache Commons Text syntax` and `dockerize (Go text/template) syntax` 
   ```shell
   $ cat << EOF > database.properties.tmpl
   test1.properties.from.env=\${env:FOO}
   test2.properties.from.env=\${env:FOO:-"aaa"}
   test3.properties.from.sys=\${env:BAR}
   test4.properties.from.sys=\${env:BAR:-"bbb"}
   
   test5.properteis.from.env.dockerize={{ default .Env.SCALAR_DB_USERNAME "" }}
   test6.properteis.from.env.dockerize={{ default .Env.SCALAR_DB_PASSWORD "" }}
   EOF
   ```
1. Set environment variables for `dockerize` testing.
   ```shell
   $ export SCALAR_DB_USERNAME=user
   $ export SCALAR_DB_PASSWORD=pass
   ```
1. Run `dockerize` with the above `.tmpl` file.
   ```shell
   $ ./dockerize -template "database.properties.tmpl:database.properties"
   ```
1. Confirm the generated `database.properties` file.
   ```shell
   $ cat database.properties
   test1.properties.from.env=${env:FOO}
   test2.properties.from.env=${env:FOO:-"aaa"}
   test3.properties.from.sys=${env:BAR}
   test4.properties.from.sys=${env:BAR:-"bbb"}
   
   test5.properteis.from.env.dockerize=user
   test6.properteis.from.env.dockerize=pass
   ```
1. Confirm the diff between the original file and the generated file.
   ```diff
   $ diff -u ./database.properties.tmpl ./database.properties
   --- ./database.properties.tmpl  2022-12-22 15:34:04.456190828 +0900
   +++ ./database.properties       2022-12-22 15:34:15.486190746 +0900
   @@ -3,5 +3,5 @@
    test3.properties.from.sys=${env:BAR}
    test4.properties.from.sys=${env:BAR:-"bbb"}
   
   -test5.properteis.from.env.dockerize={{ default .Env.SCALAR_DB_USERNAME "" }}
   -test6.properteis.from.env.dockerize={{ default .Env.SCALAR_DB_PASSWORD "" }}
   +test5.properteis.from.env.dockerize=user
   +test6.properteis.from.env.dockerize=pass
   ```
   * Only the `dockerize (Go template) syntax` strings are replaced by the `dockerize` command. The `Apache Commons Text syntax` strings are not replaced by the `dockerize` command.

Torch3333 (approved):
LGTM ,thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #771, opened 2022-12-22
Refactor prepare logic

brfrn169:
This PR is just a refactoring thing, so it basically doesn't change the behavior. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -67,7 +67,6 @@ public class CommitHandler {
     } catch (RetriableExecutionException e) {
       throw new PreparationConflictException("conflict happened when preparing records", e);
     } catch (ExecutionException e) {
-      logger.warn("preparing records failed", e);
       throw new PreparationException("preparing records failed", e);
     }
   }
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -10,6 +10,8 @@ import com.scalar.db.exception.storage.NoMutationException;
 import com.scalar.db.exception.storage.RetriableExecutionException;
 import com.scalar.db.exception.transaction.CommitConflictException;
 import com.scalar.db.exception.transaction.CommitException;
+import com.scalar.db.exception.transaction.PreparationConflictException;
+import com.scalar.db.exception.transaction.PreparationException;
 import com.scalar.db.exception.transaction.UnknownTransactionStatusException;
 import com.scalar.db.transaction.consensuscommit.ParallelExecutor.ParallelExecutorTask;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -41,38 +43,37 @@ public CommitHandler(
...
+
     preCommitValidation(snapshot, true);
     commitState(snapshot);
     commitRecords(snapshot);
   }
 
-  public void prepare(Snapshot snapshot, boolean abortIfError)
-      throws CommitException, UnknownTransactionStatusException {
-    String id = snapshot.getId();
+  public void prepare(Snapshot snapshot) throws PreparationException {
```
brfrn169, line 62: Made this `prepare()` method throw `PreparationException` instead of `CommitException`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -41,38 +43,37 @@ public CommitHandler(
   }
 
   public void commit(Snapshot snapshot) throws CommitException, UnknownTransactionStatusException {
-    prepare(snapshot, true);
+    try {
+      prepare(snapshot);
+    } catch (PreparationException e) {
+      abort(snapshot.getId());
+      rollbackRecords(snapshot);
```
brfrn169, line 50: Moved this rollback from `prepare()` to here.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -415,8 +417,8 @@ private boolean isChanged(
         || latestResult.get().getVersion() != result.get().getVersion();
   }
 
-  private void throwExceptionDueToPotentialAntiDependency() throws CommitConflictException {
-    throw new CommitConflictException(
+  private void throwExceptionDueToPotentialAntiDependency() throws PreparationConflictException {
```
brfrn169, line 420: Made this method throw `PreparationConflictException` instead of `CommitConflictException`.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -41,38 +43,37 @@ public CommitHandler(
...
+  public void prepare(Snapshot snapshot) throws PreparationException {
     try {
       prepareRecords(snapshot);
-    } catch (Exception e) {
+    } catch (NoMutationException e) {
+      throw new PreparationConflictException("preparing record exists", e);
+    } catch (RetriableExecutionException e) {
+      throw new PreparationConflictException("conflict happened when preparing records", e);
+    } catch (ExecutionException e) {
       logger.warn("preparing records failed", e);
```
komamitsu, line 70: This warning won't be logged in the new source code if the exception isn't ExecutionException. Is it okay?

BTW, the cause exception is contained in `PreparationException`, so I'm not sure this warning is needed.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -265,7 +267,7 @@ private boolean isWriteSetOverlappedWith(Scan scan) {
 
   @VisibleForTesting
   void toSerializableWithExtraWrite(MutationComposer composer)
-      throws ExecutionException, CommitConflictException {
+      throws ExecutionException, PreparationConflictException {
```
komamitsu, line 270: Just a confirmation. I think this is a part of pre-commit validation phase not prepare phase. Is it okay to throw PreparationConflictException?

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: Agree. I will remove this log. Thanks.

brfrn169: Deleted the log in 77c422608ec17743dbd74fd3a3f858df71ca40c0. Thanks.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: Yes, Extra-write is a part of validation logically, but it actually executes in prepare phase. That's why this method throws `PreparationConflictException`.

komamitsu (approved):

feeblefakie (approved):
LGTM! Thank you!

PR #776, opened 2022-12-25
Add configurations for gRPC

brfrn169:
This PR adds the following configurations for gRPC:

For the client side:
```properties
# The maximum message size allowed for a single gRPC frame. If not specified, use the gRPC default value.
scalar.db.grpc.max_inbound_message_size=

# The maximum size of metadata allowed to be received. If not specified, use the gRPC default value.
scalar.db.grpc.max_inbound_metadata_size=
```

For the server side:
```properties
# The maximum message size allowed to be received. If not specified, use the gRPC default value.
scalar.db.server.grpc.max_inbound_message_size=

# The maximum size of metadata allowed to be received. If not specified, use the gRPC default value.
scalar.db.server.grpc.max_inbound_metadata_size=
```

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/config/ConfigUtils.java
+++ b/core/src/main/java/com/scalar/db/config/ConfigUtils.java
@@ -50,7 +50,8 @@ public final class ConfigUtils {
     try {
       return Integer.parseInt(value);
     } catch (NumberFormatException ignored) {
-      throw new IllegalArgumentException("the specified value of '" + name + "' is not a number");
+      throw new IllegalArgumentException(
+          "the specified value of '" + name + "' is not a number. value: " + value);
     }
   }
 
@@ -62,7 +63,8 @@ public final class ConfigUtils {
     try {
       return Integer.parseInt(value);
     } catch (NumberFormatException ignored) {
-      throw new IllegalArgumentException("the specified value of '" + name + "' is not a number");
+      throw new IllegalArgumentException(
+          "the specified value of '" + name + "' is not a number. value: " + value);
     }
   }
 
@@ -74,7 +76,8 @@ public final class ConfigUtils {
     try {
       return Long.parseLong(value);
     } catch (NumberFormatException ignored) {
-      throw new IllegalArgumentException("the specified value of '" + name + "' is not a number");
+      throw new IllegalArgumentException(
+          "the specified value of '" + name + "' is not a number. value: " + value);
     }
   }
 
@@ -86,7 +89,8 @@ public final class ConfigUtils {
     try {
       return Long.parseLong(value);
     } catch (NumberFormatException ignored) {
-      throw new IllegalArgumentException("the specified value of '" + name + "' is not a number");
+      throw new IllegalArgumentException(
+          "the specified value of '" + name + "' is not a number. value: " + value);
     }
   }
 
@@ -100,7 +104,7 @@ public final class ConfigUtils {
       return Boolean.parseBoolean(value);
     } else {
       throw new IllegalArgumentException(
-          "the specified value of '" + name + "' is not a boolean value");
+          "the specified value of '" + name + "' is not a boolean value. value: " + value);
     }
   }
 
@@ -114,7 +118,7 @@ public final class ConfigUtils {
       return Boolean.parseBoolean(value);
     } else {
       throw new IllegalArgumentException(
-          "the specified value of '" + name + "' is not a boolean value");
+          "the specified value of '" + name + "' is not a boolean value. value: " + value);
     }
   }
 
--- a/core/src/main/java/com/scalar/db/config/ConfigUtils.java
+++ b/core/src/main/java/com/scalar/db/config/ConfigUtils.java
@@ -54,6 +54,18 @@ public final class ConfigUtils {
     }
   }
 
+  public static Integer getInt(Properties properties, String name, Integer defaultValue) {
+    String value = trimAndReplace(properties.getProperty(name));
+    if (Strings.isNullOrEmpty(value)) {
+      return defaultValue;
+    }
+    try {
+      return Integer.parseInt(value);
+    } catch (NumberFormatException ignored) {
+      throw new IllegalArgumentException("the specified value of '" + name + "' is not a number");
+    }
+  }
+
   public static long getLong(Properties properties, String name, long defaultValue) {
     String value = trimAndReplace(properties.getProperty(name));
     if (Strings.isNullOrEmpty(value)) {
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/scalar/db/storage/rpc/GrpcConfig.java
```diff
@@ -1,15 +1,20 @@
...
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 @Immutable
 public class GrpcConfig {
 
   public static final String PREFIX = DatabaseConfig.PREFIX + "grpc.";
   public static final String DEADLINE_DURATION_MILLIS = PREFIX + "deadline_duration_millis";
+  public static final String MAX_INBOUND_MESSAGE_SIZE = PREFIX + "max_inbound_message_size";
+  public static final String MAX_INBOUND_METADATA_SIZE = PREFIX + "max_inbound_metadata_size";
```
brfrn169, line 17: Added these configurations for gRPC on the client side here.

server/src/main/java/com/scalar/db/server/ServerConfig.java
```diff
@@ -19,13 +21,21 @@ public class ServerConfig {
   public static final String PORT = PREFIX + "port";
   public static final String PROMETHEUS_EXPORTER_PORT = PREFIX + "prometheus_exporter_port";
 
+  public static final String GRPC_MAX_INBOUND_MESSAGE_SIZE =
+      PREFIX + "grpc.max_inbound_message_size";
+  public static final String GRPC_MAX_INBOUND_METADATA_SIZE =
+      PREFIX + "grpc.max_inbound_metadata_size";
```
brfrn169, line 27: Also added these configurations for gRPC on the server side here.

core/src/main/java/com/scalar/db/storage/rpc/GrpcUtils.java
```diff
@@ -0,0 +1,17 @@
...
+
+  private GrpcUtils() {}
+
+  public static ManagedChannel createChannel(GrpcConfig config) {
+    NettyChannelBuilder builder =
+        NettyChannelBuilder.forAddress(config.getHost(), config.getPort()).usePlaintext();
+    config.getMaxInboundMessageSize().ifPresent(builder::maxInboundMessageSize);
+    config.getMaxInboundMetadataSize().ifPresent(builder::maxInboundMetadataSize);
+    return builder.build();
+  }
```
brfrn169, line 16: Extracted a method to create a channel. Here we set the configurations to a channel.

server/src/main/java/com/scalar/db/server/ScalarDbServer.java
```diff
@@ -107,6 +107,9 @@ public void start() throws IOException {
               + "\" to 'jdbc'");
     }
 
+    config.getGrpcMaxInboundMessageSize().ifPresent(builder::maxInboundMessageSize);
+    config.getGrpcMaxInboundMetadataSize().ifPresent(builder::maxInboundMetadataSize);
```
brfrn169, line 111: We set the configurations to the server here.

core/src/main/java/com/scalar/db/config/ConfigUtils.java
```diff
@@ -54,6 +54,18 @@ public static int getInt(Properties properties, String name, int defaultValue) {
...
 
+  public static Integer getInt(Properties properties, String name, Integer defaultValue) {
+    String value = trimAndReplace(properties.getProperty(name));
+    if (Strings.isNullOrEmpty(value)) {
+      return defaultValue;
+    }
+    try {
+      return Integer.parseInt(value);
+    } catch (NumberFormatException ignored) {
+      throw new IllegalArgumentException("the specified value of '" + name + "' is not a number");
```
komamitsu, line 65: How about showing the actual `value` in the log message? Just containing the cause exception is enough?

core/src/main/java/com/scalar/db/config/ConfigUtils.java
```diff
@@ -66,6 +78,18 @@ public static long getLong(Properties properties, String name, long defaultValue
...
 
+  public static Long getLong(Properties properties, String name, Long defaultValue) {
+    String value = trimAndReplace(properties.getProperty(name));
+    if (Strings.isNullOrEmpty(value)) {
+      return defaultValue;
+    }
+    try {
+      return Long.parseLong(value);
+    } catch (NumberFormatException ignored) {
+      throw new IllegalArgumentException("the specified value of '" + name + "' is not a number");
```
komamitsu, line 89: Same as above

core/src/main/java/com/scalar/db/config/ConfigUtils.java
```diff
@@ -80,6 +104,20 @@ public static boolean getBoolean(Properties properties, String name, boolean def
...
+    String value = trimAndReplace(properties.getProperty(name));
+    if (Strings.isNullOrEmpty(value)) {
+      return defaultValue;
+    }
+    if (Boolean.TRUE.toString().equalsIgnoreCase(value)
+        || Boolean.FALSE.toString().equalsIgnoreCase(value)) {
+      return Boolean.parseBoolean(value);
+    } else {
+      throw new IllegalArgumentException(
+          "the specified value of '" + name + "' is not a boolean value");
```
komamitsu, line 117: Same as above

Torch3333 (approved):
LGTM, thank you!

server/src/main/java/com/scalar/db/server/ScalarDbServer.java
komamitsu: It means GrpcConfig and GrpcUtils aren't used in `server`, right?

komamitsu (approved):
LGTM! 👍 

I left some minor comments, though.

core/src/main/java/com/scalar/db/config/ConfigUtils.java
brfrn169: I think showing the actual `value` in the log message is better. I will do that in this RP. Thank you!

server/src/main/java/com/scalar/db/server/ScalarDbServer.java
brfrn169: Yes, `GrpcConfig` and `GrpcUtils` are for the client side, and they are not used on the `server` side. The class for the configurations for the `server` side is `ServerConfig` actually.

core/src/main/java/com/scalar/db/config/ConfigUtils.java
brfrn169: Fixed it in b090379c3e52e4f39f9b05fb5ea9d83eaa73d41d. 

@komamitsu Can you please take a look again when you have time? Thanks.

komamitsu (approved):

feeblefakie (approved):
LGTM! Thank you!

kota2and3kan (approved):
LGTM! Thank you!

I have a question. 
If we want to change the value of `GRPC_ARG_MAX_RECEIVE_MESSAGE_LENGTH` of gRPC via ScalarDB, I think we need to set these new properties both client (application) side and server (ScalarDB Server) side.
Is my understanding correct?

brfrn169:
> If we want to change the value of `GRPC_ARG_MAX_RECEIVE_MESSAGE_LENGTH` of gRPC via ScalarDB, I think we need to set these new properties both client (application) side and server (ScalarDB Server) side. Is my understanding correct?

Yes. I think we need the properties on the both sides (`scalar.db.grpc.max_inbound_message_size` on the client side and `scalar.db.server.grpc.max_inbound_message_size` on the server side).

kota2and3kan:
@brfrn169  
I understand. Thank you!

PR #779, opened 2022-12-27
Add description how to build docker image of ScalarDB Server

kota2and3kan:
I was not able to find the description of how to build a Docker image of ScalarDB Server from the source code.
From the perspective of customers/users, there is no problem since they can get the docker image from Github Packages or each marketplace.
However, from the perspective of developers (mainly new development members), I think this is useful information.
So, I added it to the document. Please take a look!

### DIFF
--- a/docs/scalardb-server.md
+++ b/docs/scalardb-server.md
@@ -27,6 +27,12 @@ $ ./gradlew installDist
 
 Of course, you can archive the jar and libraries by `./gradlew distZip` and so on.
 
+Also, you can build a Docker image from the source as follows.
+
+```shell
+$ ./gradlew :server:docker
+```
+
 ## Configure Scalar DB Server
 
 You need a property file holding the configuration for Scalar DB Server. 


komamitsu (approved):
Thanks! LGTM.

But how about using a similar guide to https://github.com/scalar-labs/scalardb/blob/master/docs/schema-loader.md#build ? Either content is fine with me, but unified something would be better.

brfrn169 (approved):
LGTM! Thank you!

kota2and3kan:
@komamitsu  
Thank you for your review!
Please let me confirm the details of your suggestion.

> But how about using a similar guide to https://github.com/scalar-labs/scalardb/blob/master/docs/schema-loader.md#build ? Either content is fine with me, but unified something would be better.

Do you mean it is better to update the table of contents like the following?

* Table of contents (example)
  ```
  # Scalar DB Server
  ## Usage
  ### Install
  ### Build
  ### Docker
  ### Run
  #### Configure Scalar DB Server
  #### Start Scalar DB Server
  #### Zip archives
  #### Usage of the Java client of Scalar DB Server
  ## Further reading
  ```

If yes, I agree with you since the unified structure document is readable for users.
And, I will update it in this PR.

komamitsu:
@kota2and3kan Oops, I pasted the wrong URL. I should've put https://github.com/scalar-labs/scalardb/blob/master/docs/schema-loader.md#docker. I also just noticed it would make no sense if we unify partial formats without taking care of the whole table of content when I read your comment. I think it's too much to ask you to restructure the table of content. Please ignore my above comment. Sorry for the confusion.

kota2and3kan:
@komamitsu  
> Oops, I pasted the wrong URL. I should've put https://github.com/scalar-labs/scalardb/blob/master/docs/schema-loader.md#docker.

Ah, I see. I understand.

> I also just noticed it would make no sense if we unify partial formats without taking care of the whole table of content when I read your comment. I think it's too much to ask you to restructure the table of content. Please ignore my above comment. Sorry for the confusion.

OK. No problem!

Torch3333 (approved):
LGTM, thank you.

feeblefakie (approved):
LGTM! Thank you!

PR #783, opened 2023-01-03
Revisit commit and abort state logic

brfrn169:
I'll add inline comments for the details. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -48,7 +48,7 @@ public class CommitHandler {
     try {
       prepare(snapshot);
     } catch (PreparationException e) {
-      abort(snapshot.getId());
+      abortState(snapshot.getId());
       rollbackRecords(snapshot);
       if (e instanceof PreparationConflictException) {
         throw new CommitConflictException(e.getMessage(), e);
@@ -59,7 +59,7 @@ public class CommitHandler {
     try {
       validate(snapshot);
     } catch (ValidationException e) {
-      abort(snapshot.getId());
+      abortState(snapshot.getId());
       rollbackRecords(snapshot);
       if (e instanceof ValidationConflictException) {
         throw new CommitConflictException(e.getMessage(), e);
@@ -110,21 +110,32 @@ public class CommitHandler {
       throws CommitException, UnknownTransactionStatusException {
     String id = snapshot.getId();
     try {
-      commitState(snapshot.getId());
-    } catch (CoordinatorException e) {
-      TransactionState state = abort(id);
-      if (state.equals(TransactionState.ABORTED)) {
-        rollbackRecords(snapshot);
-        throw new CommitException(
-            "committing state in coordinator failed. the transaction is aborted", e);
+      Coordinator.State state = new Coordinator.State(id, TransactionState.COMMITTED);
+      coordinator.putState(state);
+      logger.debug(
+          "transaction {} is committed successfully at {}", id, System.currentTimeMillis());
+    } catch (CoordinatorConflictException e) {
+      try {
+        Optional<Coordinator.State> s = coordinator.getState(id);
+        if (s.isPresent()) {
+          TransactionState state = s.get().getState();
+          if (state.equals(TransactionState.ABORTED)) {
+            rollbackRecords(snapshot);
+            throw new CommitException(
+                "committing state in coordinator failed. the transaction is aborted", e);
+          }
+        } else {
+          throw new UnknownTransactionStatusException(
+              "committing state failed with NoMutationException but the coordinator status doesn't exist",
+              e,
+              id);
+        }
+      } catch (CoordinatorException e1) {
+        throw new UnknownTransactionStatusException("can't get the state", e1, id);
       }
+    } catch (CoordinatorException e) {
+      throw new UnknownTransactionStatusException("coordinator status is unknown", e, id);
     }
-    logger.debug("transaction {} is committed successfully at {}", id, System.currentTimeMillis());
-  }
-
-  private void commitState(String id) throws CoordinatorException {
-    Coordinator.State state = new Coordinator.State(id, TransactionState.COMMITTED);
-    coordinator.putState(state);
   }
 
   public void commitRecords(Snapshot snapshot) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -145,30 +156,30 @@ public void commitRecords(Snapshot snapshot) {
     }
   }
 
-  public TransactionState abort(String id) throws UnknownTransactionStatusException {
+  public TransactionState abortState(String id) throws UnknownTransactionStatusException {
```
brfrn169, line 159: Renamed `abort` to `abortState`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -119,7 +119,7 @@ private void put(Put put) throws CoordinatorException {
         break;
       } catch (NoMutationException e) {
         logger.warn("mutation seems applied already", e);
-        throw new CoordinatorException("mutation seems applied already.", e);
+        throw new CoordinatorConflictException("mutation seems applied already.", e);
```
brfrn169, line 122: Changed to throw `CoordinatorConflictException` when catching `NoMutationException` because I think we should handle when caching `NoMutationException` or `ExecutionException` differently.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -110,21 +110,32 @@ public void commitState(Snapshot snapshot)
...
+          throw new UnknownTransactionStatusException(
+              "committing state failed with NoMutationException but the coordinator status doesn't exist",
+              e,
+              id);
+        }
+      } catch (CoordinatorException e1) {
+        throw new UnknownTransactionStatusException("can't get the state", e1, id);
       }
+    } catch (CoordinatorException e) {
+      throw new UnknownTransactionStatusException("coordinator status is unknown", e, id);
```
brfrn169, line 137: I think we can just throw `UnknownTransactionStatusException` when catching `CoordinatorException` because in that case, something unexpected database error is happening, so there is probably nothing we can do.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -110,21 +110,32 @@ public void commitState(Snapshot snapshot)
...
-        rollbackRecords(snapshot);
-        throw new CommitException(
-            "committing state in coordinator failed. the transaction is aborted", e);
+      Coordinator.State state = new Coordinator.State(id, TransactionState.COMMITTED);
+      coordinator.putState(state);
+      logger.debug(
+          "transaction {} is committed successfully at {}", id, System.currentTimeMillis());
+    } catch (CoordinatorConflictException e) {
+      try {
+        Optional<Coordinator.State> s = coordinator.getState(id);
```
brfrn169, line 119: I think we should check the state only when catching `CoordinatorConflictException` because when we catch `CoordinatorException`, something unexpected database error is happening, so we probably can't get the state from the coordinator as well.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -145,30 +156,30 @@ public void commitRecords(Snapshot snapshot) {
...
 
-  public TransactionState abort(String id) throws UnknownTransactionStatusException {
+  public TransactionState abortState(String id) throws UnknownTransactionStatusException {
     try {
-      abortState(id);
+      Coordinator.State state = new Coordinator.State(id, TransactionState.ABORTED);
+      coordinator.putState(state);
       return TransactionState.ABORTED;
-    } catch (CoordinatorException e) {
+    } catch (CoordinatorConflictException e) {
```
brfrn169, line 164: Similar to `commitState()`, I think we should check the state only when catching `CoordinatorConflictException`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -110,21 +110,32 @@ public void commitState(Snapshot snapshot)
...
+          "transaction {} is committed successfully at {}", id, System.currentTimeMillis());
+    } catch (CoordinatorConflictException e) {
+      try {
+        Optional<Coordinator.State> s = coordinator.getState(id);
+        if (s.isPresent()) {
+          TransactionState state = s.get().getState();
+          if (state.equals(TransactionState.ABORTED)) {
+            rollbackRecords(snapshot);
+            throw new CommitException(
+                "committing state in coordinator failed. the transaction is aborted", e);
```
komamitsu, line 125: TxID in the log message would be useful info for trouble shootings. I think adding transaction ID field to CommitException is also good.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -110,21 +110,32 @@ public void commitState(Snapshot snapshot)
...
+          }
+        } else {
+          throw new UnknownTransactionStatusException(
+              "committing state failed with NoMutationException but the coordinator status doesn't exist",
```
komamitsu, line 129: Just out of curiosity: is there any possible cause for this case?

komamitsu (approved):
LGTM! 👍 
I left a few minor comments, though.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: Thank you. Yes, you are right. I think we should do the same thing in other parts, so I will do that in another PR.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: Basically, it's impossible unless something external deletes the status record. Honestly, in that case, I'm not sure we should throw `UnknownTransactionStatusException`, but I think it's okay as it's a very rare case.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
feeblefakie: I'm fine with this, but I think we could have `CoordinatorException` due to a temporary glitch, which is not necessarily an unexpected database error.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
feeblefakie: If an underlying database is somewhat broken, it could go to the state.
I think it's reasonable to mask the unexpected behavior with `UnknownTransactionStatusException`.

feeblefakie (approved):
LGTM! Thank you!
Left some comments, but they don't block it from being merged.

PR #785, opened 2023-01-06
Should synchronize in active transaction management

brfrn169:
In the active transaction management, we hold active transactions in the `ActiveExpiringMap`, which expires and rolls back a transaction in a dedicated thread when the transaction is timed out. 

ActiveTransactionManagedDistributedTransactionManager:
https://github.com/scalar-labs/scalardb/blob/should-synchronize-in-active-transaction-management/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java#L42

ActiveTransactionManagedTwoPhaseCommitTransactionManager:
https://github.com/scalar-labs/scalardb/blob/should-synchronize-in-active-transaction-management/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitTransactionManager.java#L44

In that case, a transaction object can be accessed by multiple threads at the same time, but the transaction implementations are basically thread-unsafe, so something unexpected can happen. To address this issue, we should synchronize the transaction operations.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
@@ -29,7 +29,7 @@ public abstract class ActiveTransactionManagedDistributedTransactionManager
   private static final Logger logger =
       LoggerFactory.getLogger(ActiveTransactionManagedDistributedTransactionManager.class);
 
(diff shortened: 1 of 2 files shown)


Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #786, opened 2023-01-06
Allow commit() and rollback() to be called in parallel in all coordinator/participant processes in 2PC transactions

brfrn169:
Before this change, one of the limitations in 2PC transaction was that we need to call `commit()`/`rollback()` in a coordinator process before calling `commit()`/`rollback()` in participant processes. This PR gets rid of this limitation. After this change, we can call `commit()`/`rollback()` in parallel in all coordinator/participant processes.

I'll add inline comments for the details. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommit.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommit.java
@@ -31,7 +31,6 @@ public class TwoPhaseConsensusCommit extends AbstractTwoPhaseCommitTransaction {
   private final CrudHandler crud;
   private final CommitHandler commit;
   private final RecoveryHandler recovery;
-  private final boolean isCoordinator;
 
   private boolean validated;
   private boolean needRollback;
@@ -40,12 +39,10 @@ public class TwoPhaseConsensusCommit extends AbstractTwoPhaseCommitTransaction {
   private Runnable beforeRecoveryHook = () -> {};
 
   @SuppressFBWarnings("EI_EXPOSE_REP2")
-  public TwoPhaseConsensusCommit(
-      CrudHandler crud, CommitHandler commit, RecoveryHandler recovery, boolean isCoordinator) {
+  public TwoPhaseConsensusCommit(CrudHandler crud, CommitHandler commit, RecoveryHandler recovery) {
     this.crud = crud;
     this.commit = commit;
     this.recovery = recovery;
-    this.isCoordinator = isCoordinator;
   }
 
   @Override
@@ -144,15 +141,13 @@ public class TwoPhaseConsensusCommit extends AbstractTwoPhaseCommitTransaction {
               + " before calling commit()");
     }
 
-    if (isCoordinator) {
-      try {
-        commit.commitState(crud.getSnapshot());
-      } catch (Exception e) {
-        // no need to rollback because the transaction has already been rolled back
-        needRollback = false;
+    try {
+      commit.commitState(crud.getSnapshot());
+    } catch (CommitException | UnknownTransactionStatusException e) {
+      // no need to rollback because the transaction has already been rolled back
+      needRollback = false;
 
-        throw e;
-      }
+      throw e;
     }
 
     commit.commitRecords(crud.getSnapshot());
@@ -164,12 +159,10 @@ public class TwoPhaseConsensusCommit extends AbstractTwoPhaseCommitTransaction {
       return;
     }
 
-    if (isCoordinator) {
-      try {
-        commit.abortState(crud.getSnapshot().getId());
-      } catch (UnknownTransactionStatusException e) {
-        throw new RollbackException("rollback failed", e);
-      }
+    try {
+      commit.abortState(crud.getSnapshot().getId());
+    } catch (UnknownTransactionStatusException e) {
+      throw new RollbackException("rollback failed", e);
     }
 
     commit.rollbackRecords(crud.getSnapshot());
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommit.java
```diff
@@ -144,15 +141,13 @@ public void commit() throws CommitException, UnknownTransactionStatusException {
...
     }
 
-    if (isCoordinator) {
-      try {
-        commit.commitState(crud.getSnapshot());
-      } catch (Exception e) {
-        // no need to rollback because the transaction has already been rolled back
-        needRollback = false;
+    try {
+      commit.commitState(crud.getSnapshot());
```
brfrn169, line 145: Regardless that it's a coordinator process or a participant process, we always call this `commit.commitState()` so that we can call it in parallel in all coordinator/participant processes.

Only one process can succeed to putting a transaction state to a coordinator table as follows:
https://github.com/scalar-labs/scalardb/blob/allow-commit-and-rollback-to-be-called-in-parallel-in-2pc/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java#L114

And the rest of the processes catch `CoordinatorConflictException`, and they need to get a transaction state:
https://github.com/scalar-labs/scalardb/blob/allow-commit-and-rollback-to-be-called-in-parallel-in-2pc/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java#L119

So, this change can impact the performance a bit.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommit.java
```diff
@@ -164,12 +159,10 @@ public void rollback() throws RollbackException {
...
     }
 
-    if (isCoordinator) {
-      try {
-        commit.abortState(crud.getSnapshot().getId());
-      } catch (UnknownTransactionStatusException e) {
-        throw new RollbackException("rollback failed", e);
-      }
+    try {
+      commit.abortState(crud.getSnapshot().getId());
```
brfrn169, line 163: Ditto. Regardless that it's a coordinator or a participant, we always call this `commit. abortState()` here.

Torch3333 (approved):
LGTM, thank you!

komamitsu:
> we need to call commit()/rollback() in a coordinator process before calling commit()/rollback() in participant processes

Just a confirmation: In what case does this limitation matter? The following way can't avoid the limitation?
```
class Coordinator {
  def commit()
    local.commit()
    futures = participants.map { p -> p.commit() }
    joinAll(futures)
```

brfrn169:
> Just a confirmation: In what case does this limitation matter? The following way can't avoid the limitation?

@komamitsu Yes, the limitation is that we first need to call `commit()` on the coordinator side and wait for it to be done, and then we can call `commit()` on the participant sides in parallel:
```
// First, we need to call `commit()` on the coordinator side and wait for it to be done
local.commit()

// And then, we can call `commit()` on the participant sides in parallel
futures = participants.map { p -> p.commit() }
joinAll(futures)
```

After this change, we can call `commit()` on the coordinator side and the participant side in parallel:
```
// We can call `commit()` on the coordinator side and the participant side in parallel
future = local.commit()
futures = participants.map { p -> p.commit() }
joinAll(future, futures)
```

Thanks.

komamitsu:
@brfrn169 Thanks for the confirmation! It sounds cool performance improvment.

On the other hand, my slight concern is it might introduce some uncertainity like this case

```
class Coordinator {
  def commit()
    local.commit()
    futures = participants.map { p -> p.commit() }
    // One of participants fails to commit and updates the status to `abort`
    // local.commit() is executed after the status update and the final status will be `abort` becasue of the first win rule.
    // Maybe some other participants already commit?
    joinAll(futures)
```

In current implementation, the above case that the final result depends on the timing of commits won't happen (the final status should be always `commit` updated by the local while there is also inconsistency between the final status and failed participants)

But I don't want to block you. Just wanted to share the slight concern.

komamitsu (approved):

feeblefakie (approved):
LGTM! Thank you!

brfrn169:
@komamitsu If one of the participants fails to commit and updates the status to `abort`, the other participants also fail to commit. This is because the commit logic puts the status into the coordinator table with conditional update (`IfNotExists`), so if anyone of the participants puts the status with `abort`, the other participants fail to put the state and cancel to commit the transaction.

komamitsu:
@brfrn169 The point of my above comment is just the order of coordinator's commit state and participants ones. In current impl, coordinator's commit always wins. But with the change of this PR, IIUC any participant's commit state (ABORT) could happen before coordinator's commit. If so, new case that coordinator tries to commit (COMMIT) while commite (ABORT) is recorded already might happen. I'm not sure my understanding is correct or it matters, though.

komamitsu:
BTW, I just remembered our recent benchmarks of ScalarDB and some other transaction managers. IIRC we observed some performance bottlenecks that seemed to come from IO bound (e.g. WAL.) In case of PostgreSQL, any commit/abort even without update are recorded in `clog` (commit log) file (also in WAL files in case of commit.) I'm wondering if our aggressive parallel operations like this PR might increase IO (and IO wait latency.)

I think it would be great if we improve our daily benchmark system to run on a topic branch (and hopefully with all the underlying databases) someday so that we can move forward on with confident... 😅

brfrn169:
@komamitsu As discussed, `commit()` should be called after `prepare()` succeeds in all the processes (the coordinator and participants). So, basically, putting COMMIT state and ABORT state don't happen at the same time unless users wrongly use the 2PC API. One possibility is that the lazy recovery is kicked for the prepared record by another transaction. Even in that case, the decision of the transaction that's successfully putting the state wins, and no inconsistency occurs.

PR #762, opened 2022-12-18
feat: SQLite support via JDBC

laysakura:
This is a base (huge) pull request to fully support SQLite.
See separate pull requests for code review.

- #763 
- #764 
- #765 
- #769 
- #819

Issue #758, opened 2022-12-17
Refactor request: branching by SQLSTATE


laysakura:
## Refactoring target

https://github.com/scalar-labs/scalardb/blob/5768435efbbe0ea13280178f3123d450f33eb6ca/core/src/main/java/com/scalar/db/storage/jdbc/ConditionalMutator.java#L103-L105

## Problem

May easily forget to add a necessary condition when adding a new RDB engine support.

## Solution idea

Use a strategy pattern, where an interface has `detectDuplicateKeyError(e)` and concrete classes implement the logics to check SQLSTATE, for example.

PR #787, opened 2023-01-07
refactor: use strategy pattern to avoid branching by `RdbEngine` in `JdbcAdmin`

laysakura:
This PR is the first attempt to remove some of the conditional branches by `RdbEngine` in `JdbcAdmin`.
After this PR is merged, I'll try to:

- eliminate all the conditional branches by `RdbEngine` in `JdbcAdmin`.
- make DB error handling better (see #758, for example).

## Motivation

SQLite support would require a bit different logics compared to other JDBC database engines (c.f. [internal slack posts](https://scalar-labs.slack.com/archives/C04FHGU3AN6/p1672551832218709)).

I felt it would be better to make refactoring before implementing #762.

## Overview

Policy: Keep public APIs in JdbcAdmin.

- Add `abstract class RdbEngineStrategy` that has abstract methods where each specific database engine should have implementation details.
- Add `class RdbEngine{Mysql, Postgresql, Oracle, SqlServer} extends RdbEngineStrategy`

## Notes for reviewers

- https://github.com/scalar-labs/scalardb/pull/787/commits/b2ff9375f17810a204a85b8754dedd76f6e3e559 is a good example of what this PR would achieve.


laysakura:
I now feel:

- `RdbEngineStrategy` should only have abstract methods
- Diffs in `JdbcAdmin` should be less

So I made this PR back to draft.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+import com.scalar.db.io.DataType;
+import com.scalar.db.storage.jdbc.query.QueryUtils;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+abstract class RdbEngineStrategy {
+
+  static RdbEngineStrategy create(JdbcConfig config) {
+    switch (config.getRdbEngine()) {
+      case MYSQL:
```
komamitsu, line 13: [trivial] A future potential improvement might be like the following so that we don't need to update this file

```java
public enum RdbEngine {
  MYSQL(RdbEngineMysql::new),
  POSTGRESQL(RdbEngineMysql::new),
  ORACLE(RdbEngineMysql::new),
  SQL_SERVER(RdbEngineMysql::new);

  private final Supplier<RdbEngineStrategy> factory;

  RdbEngine(Supplier<RdbEngineStrategy> factory) {
    this.factory = factory;
  }
  
  RdbEngineStrategy createStrategy() {
    return factory.get();
  }
}
```

komamitsu (approved):
LGTM! Great refactoring!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -293,7 +183,7 @@ private void createMetadataSchemaAndTableIfNotExists(Connection connection) thro
   }
 
   private void createMetadataSchemaIfNotExists(Connection connection) throws SQLException {
-    switch (rdbEngine) {
+    switch (rdbEngineType) {
       case MYSQL:
```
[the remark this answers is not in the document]
laysakura: Yes.

From the PR description,

> After this PR is merged, I'll try to:
> eliminate all the conditional branches by RdbEngine in JdbcAdmin.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
```diff
@@ -221,4 +221,18 @@ public int getAdminConnectionPoolMaxIdle() {
   public int getAdminConnectionPoolMaxTotal() {
     return adminConnectionPoolMaxTotal;
   }
+
+  public RdbEngine getRdbEngine() {
```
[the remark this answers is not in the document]
laysakura: Pretty clean.

Actually, I'm trying to remove the `enum RdbEngine` by the further refactoring PRs (use subclasses of `RdbEngineStrategy` instead).
But currently I'm not sure whether I can remove the whole usage of `enum RdbEngine`.

If I need to remain `enum RdbEngine`, I'll use your definition.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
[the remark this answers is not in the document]
laysakura: FYI: will be done in https://github.com/scalar-labs/scalardb/pull/790/

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
feeblefakie: Similar to the above comment and it's a minor thing, but I feel it might be slightly better to separate the creation of `RdbEngineStrategy` instance and its interface from the "separation of concerns" perspective.

It depends on how `RdbEngineStrategy` is going to be extended, but we can make it just an interface instead of the abstract class by extracting the `create` method and making `enclose` default method.

feeblefakie (approved):
LGTM! Thank you!
It's a very nice first step. I hope all the `if` clauses using RdbEngineType are removed.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
laysakura: @feeblefakie I fully agree with the idea that we can (and should) make `RdbEngineStrategy` interface rather than `abstract class`.

I did it in another pull-request (will ask your review after the base branch is merged to master).
https://github.com/scalar-labs/scalardb/pull/791/files#diff-f2858564c4aa48bdf1bc09aa29ebb2f9def9f18245b75de17415d4f3ac36a261

feeblefakie: Sounds good to me! Thanks!

brfrn169 (approved):
LGTM! Thank you!

PR #792, opened 2023-01-12
Load admins lazily in Schema Loader

brfrn169:
Currently, `SchemaOperator` in Schema Loader has two admin instances (`storageAdmin` and `transactionAdmin`), and it creates the instances in the constructor even if they are used or not. But we can create them lazily only when they are really used. This PR fixes it. Please take a look!

### DIFF
--- a/schema-loader/src/main/java/com/scalar/db/schemaloader/SchemaOperator.java
+++ b/schema-loader/src/main/java/com/scalar/db/schemaloader/SchemaOperator.java
@@ -1,6 +1,7 @@
 package com.scalar.db.schemaloader;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Suppliers;
 import com.scalar.db.api.DistributedStorageAdmin;
 import com.scalar.db.api.DistributedTransactionAdmin;
 import com.scalar.db.api.TableMetadata;
@@ -17,6 +18,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
 import javax.annotation.concurrent.ThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


schema-loader/src/main/java/com/scalar/db/schemaloader/SchemaOperator.java
```diff
@@ -25,23 +28,33 @@
...
+            () -> {
+              storageAdminLoaded.set(true);
+              return storageFactory.getStorageAdmin();
+            });
+    transactionAdmin =
+        Suppliers.memoize(
+            () -> {
+              transactionAdminLoaded.set(true);
+              return transactionFactory.getTransactionAdmin();
+            });
```
brfrn169, line 57: Create the admin instances lazily by using `Suppliers.memoize()` in Guava.

Torch3333 (approved):
LGTM, thank you!

schema-loader/src/main/java/com/scalar/db/schemaloader/SchemaOperator.java
```diff
@@ -25,23 +28,33 @@
...
-    transactionAdmin = transactionFactory.getTransactionAdmin();
+    this(StorageFactory.create(properties), TransactionFactory.create(properties));
+  }
+
+  private SchemaOperator(StorageFactory storageFactory, TransactionFactory transactionFactory) {
+    storageAdmin =
+        Suppliers.memoize(
+            () -> {
+              storageAdminLoaded.set(true);
+              return storageFactory.getStorageAdmin();
```
komamitsu, line 50: [minor] Maybe we can write a test to confirm `storageFactory.getStorageAdmin()` or `transactionFactory.getTransactionAdmin()` isn't called.

komamitsu (approved):
LGTM! 👍

komamitsu:
BTW, looked into `com.google.common.base.Suppliers` and confirmed it doesn't cause memory leak (I had a concern that it might use static variable internally)

schema-loader/src/main/java/com/scalar/db/schemaloader/SchemaOperator.java
brfrn169: Correct. I'll do that. Thanks.

brfrn169: @komamitsu Added unit tests in 7fee183d7e48234446e27350059fffff57edccfb. Please take a look again when you have time!

komamitsu (approved):

PR #797, opened 2023-01-15
Revert async commit default value

brfrn169:
This PR reverts the async commit default value to `false` (disabled) because enabling the async commit feature can cause `UncommittedRecordException` when reading a record right after writing the record in another transaction, which can be an incompatible behavior change. After we add some optimization to avoid that behavior change to the consensus commit, we can re-enable the async commit feature by default. Please take a look.


feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM, thank you!

PR #798, opened 2023-01-16
refactor: make `enum RdbEngine` test-only

laysakura:
## Motivation and background

- `enum RdbEngine` has led to low cohesion codes with conditional branches
  - Then, I wanted to remove `enum RdbEngine`
- Luckily, I succeeded in removing all the dependencies to the `RdbEngine` in main codes but some test codes are making full use of it
  - So I decieded to leave `enum RdbEngine` in test package

## Major changes

- move `enum RdbEngine` from main to test
  - all main classes no longer have `RdbEngine` fields, parameters, and result types
- remove `JdbcConfig.getRdbEngine()`


komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -77,4 +79,8 @@ void dropIndexExecute(Connection connection, String schema, String table, String
   default String encloseFullTableName(String schema, String table) {
     return enclose(schema) + "." + enclose(table);
   }
+
+  SelectQuery buildSelectQuery(SelectQuery.Builder builder, int limit);
+
+  UpsertQuery buildUpsertQuery(UpsertQuery.Builder builder);
```
brfrn169, line 85: Not sure these methods are proper abstractions. But, I have no specific ideas for now, so let's keep them. And we can revisit it in the future if needed. Thanks.

brfrn169 (approved):
Left one comment, but LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #802, opened 2023-01-18
Increase Trivy timeout

komamitsu:
Trivy workflow often fails these days. It seems the root cause is that `search.maven.org` API is recently unstable. https://github.com/aquasecurity/trivy/issues/3421

`--offline-scan` might be an option, but it looks to have some security trade offs.

This PR increases Trivy timeout using `--timeout` option without any negative impact.

### DIFF
--- a/.github/workflows/vuln-check.yaml
+++ b/.github/workflows/vuln-check.yaml
@@ -61,7 +61,7 @@ jobs:
           exit-code: '1'
           ignore-unfixed: true
           severity: 'CRITICAL,HIGH'
-          timeout: '30m'
+          timeout: '60m'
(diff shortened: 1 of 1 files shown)


brfrn169 (approved):
LGTM! Thank you!

feeblefakie:
Ah, OK. That's why we got to the 3,000 mins limit faster this month.
Hmm...

feeblefakie (approved):
LGTM! Thank you!

komamitsu:
@Torch3333 I just merged this small improvement. Please take a look at it and let me know if you notice something when you get a chance later.

PR #804, opened 2023-01-18
Update transaction documents

brfrn169:
This PR updates (and improves) the transaction documents. Please take a look!


docs/api-guide.md
```diff
@@ -708,14 +704,14 @@ public class Sample {
...
+      // Begin a transaction
       DistributedTransaction tx;
       try {
-        tx = manager.start();
+        tx = transactionManager.begin();
       } catch (TransactionException e) {
-        // If starting a transaction fails, it indicates some failure happens during a transaction,
-        // so you should cancel the transaction or retry the transaction after the failure/error is
-        // fixed
+        // If starting a transaction fails, it indicates some failure happens during the
```
Torch3333, line 712:
```suggestion
        // If beginning a transaction fails, it indicates some failure happens during the
```
To stay consistent with the change from `start()` to `begin()` in the code right above.

Torch3333 (approved):
LGTM, thank you!
I just left a minor suggestion.

core/src/main/java/com/scalar/db/api/DistributedTransaction.java
```diff
@@ -69,8 +69,8 @@ public interface DistributedTransaction extends TransactionCrudOperable {
   /**
    * Commits a transaction.
    *
-   * @throws CommitConflictException if conflicts happened. You can retry the transaction in this
-   *     case
+   * @throws CommitConflictException if conflicts happened. You can retry the same transaction from
```
komamitsu, line 72: `same` sounds the previously failed and rollbacked transaction to me. "the same operations in another transaction" or something would be less confusing?

komamitsu: I googled with `""retry the transaction"` keyword and it seems very common. So, probably okay. Never mind.

brfrn169: I'm wondering if it could be better to remove the `same`?
```suggestion
   * @throws CommitConflictException if conflicts happened. You can retry the transaction from
```

komamitsu: Yeah, `retry the transaction` doesn't sound the previously failed transaction itself compared to `retry the same transaction` to me.

docs/api-guide.md
```diff
@@ -729,21 +725,21 @@ public class Sample {
...
       } catch (CrudException | CommitException e) {
         // If you catch CrudException or CommitException, it indicates some failure happens, so you
         // should cancel the transaction or retry the transaction after the failure/error is fixed
         try {
-          tx.abort();
-        } catch (AbortException ex) {
-          // Aborting the transaction fails. You can log it here
+          tx.rollback();
+        } catch (RollbackException ex) {
+          // Rolling back the transaction fails. You can log it here
```
komamitsu, line 743: > Rolling back the transaction fails
`failed` sounds a bit better?

core/src/main/java/com/scalar/db/api/DistributedTransaction.java
brfrn169: Thank you! I'll fix it that way.

docs/two-phase-commit-transactions.md
```diff
@@ -195,6 +195,111 @@ Similar to `prepare()`, you can call `validate()` in the coordinator/participant
...
+public class Sample {
+  public static void main(String[] args) throws IOException, InterruptedException {
+    TransactionFactory factory = TransactionFactory.create("<configuration file path>");
+    TwoPhaseCommitTransactionManager transactionManager =
+        factory.getTwoPhaseCommitTransactionManager();
+
+    int retryCount = 0;
+
+    while (true) {
+      if (retryCount > 0) {
```
komamitsu, line 212: [trivial] Incrementing the counter here might be more robust considering future modifications of the exception handling blocks below (e.g. developers notice CrudException should be retried and they change `return` to `break` without incrementing the counter)

core/src/main/java/com/scalar/db/api/DistributedTransaction.java
brfrn169: @komamitsu Fixed in b512a20f0a0c4148ae578d1bbc062b547200c1ff. Please take a look again when you have time!

docs/two-phase-commit-transactions.md
```diff
@@ -195,6 +195,111 @@ Similar to `prepare()`, you can call `validate()` in the coordinator/participant
...
+        return;
+      }
+    }
+  }
+}
+```
+
+The APIs for CRUD operations (`get()`/`scan()`/`put()`/`delete()`/`mutate()`) could throw `CrudException` and `CrudConflictException`.
+If you catch `CrudException`, it indicates some failure (e.g., database failure and network error) happens during a transaction, so you should cancel the transaction or retry the transaction after the failure/error is fixed.
+If you catch `CrudConflictException`, it indicates a conflict occurs during the transaction so that you can retry the transaction from the beginning, preferably with well-adjusted exponential backoff based on your application and environment.
```
komamitsu: 👍

docs/two-phase-commit-transactions.md
```diff
@@ -195,6 +195,111 @@ Similar to `prepare()`, you can call `validate()` in the coordinator/participant
...
+    TransactionFactory factory = TransactionFactory.create("<configuration file path>");
+    TwoPhaseCommitTransactionManager transactionManager =
+        factory.getTwoPhaseCommitTransactionManager();
+
+    int retryCount = 0;
+
+    while (true) {
+      if (retryCount > 0) {
+        // Retry the transaction three times maximum in this sample code
+        if (retryCount == 3) {
```
komamitsu, line 214: [trivial] `retryCount >= 3` would be a little bit robust

docs/two-phase-commit-transactions.md
```diff
@@ -195,6 +195,111 @@ Similar to `prepare()`, you can call `validate()` in the coordinator/participant
...
+If you catch `ValidationException`, like the `CrudException` case, you should cancel the transaction or retry the transaction after the failure/error is fixed.
+If you catch `ValidationConflictException`, like the `CrudConflictException` case, you can retry the transaction from the beginning.
+
+The `commit()` API could throw `CommitException`, `CommitConflictException`, and `UnknownTransactionStatusException`.
+If you catch `CommitException`, like the `CrudException` case, you should cancel the transaction or retry the transaction after the failure/error is fixed.
+If you catch `CommitConflictException`, like the `CrudConflictException` case, you can retry the transaction from the beginning.
+If you catch `UnknownTransactionStatusException`, you are not sure if the transaction succeeds or not.
+In such a case, you need to check if the transaction is committed successfully or not and retry it if it fails.
+How to identify a transaction status is delegated to users.
+You may want to create a transaction status table and update it transactionally with other application data so that you can get the status of a transaction from the status table.
```
komamitsu, line 302: I don't really get this. Could you tell me a simple example?

core/src/main/java/com/scalar/db/api/DistributedTransaction.java
komamitsu: 💯

docs/two-phase-commit-transactions.md
brfrn169: This is explained in the 3rd way in the following article (Japanese):
http://nippondanji.blogspot.com/2009/02/blog-post_23.html

You can have a status (or log) table and insert a record into this table in addition to the usual operations in a transaction. And when you get `UnknownTransactionStatusException` and want to check if the transaction is really committed, you can check if the record of the status table exists or not. If the record exists, the transaction has been successfully committed.

komamitsu: Great! Thank you for sharing it.

The blog post seems really helpful for our customers as I think it's hard for them to come up with the idea without the details. How about adding a link to similar articles here? I with if the blog post were written in English...

docs/api-guide.md
brfrn169: Fixed in 1041dda610d74ac00596e1ab1f69d6bbdeb32106.

docs/two-phase-commit-transactions.md
brfrn169: Fixed in 1041dda610d74ac00596e1ab1f69d6bbdeb32106.

docs/two-phase-commit-transactions.md
brfrn169: Fixed in 1041dda610d74ac00596e1ab1f69d6bbdeb32106.

docs/two-phase-commit-transactions.md
komamitsu: BTW (a bit off-topic), do you think it would be great if ScalarDB maintains this kind of transaction history in `coordinator.history` or something instead of asking application side? ScalarDB error handling can check the table when trying to throw UnknownTransactionStatusException.

komamitsu: > coordinator.history

Ah, checking `coordinator.state` might be enough. But the point in this case is that maybe something wrong with the underlying database is happening and it's likely ScalarDB can't access the table... 🤔

komamitsu: [off-topic]

> This is explained in the 3rd way in the following article (Japanese):
> http://nippondanji.blogspot.com/2009/02/blog-post_23.html

Just remembered I did the same way before.
https://github.com/treasure-data/digdag/blob/42ea7cb71270957ea6424c2df88dd7c65d248789/digdag-standards/src/main/java/io/digdag/standards/operator/redshift/RedshiftConnection.java#L297-L303 The workflow operator for Redshift does CRUD in the user's database and implicitly inserts a status log record to a secret table in the same transaction

brfrn169: > Ah, checking coordinator.state might be enough. But the point in this case is that maybe something wrong with the underlying database is happening and it's likely ScalarDB can't access the table... 🤔

Yes, we have `coordinator.state` which has a state record for each transaction, and we can check the record to see if the transaction is committed or not. And we actually expose it through [DistributedTransactionManager.getState()](https://github.com/scalar-labs/scalardb/blob/18c8b20984c97b67b70b0a4f532c7e0ec23f698a/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java#L198).

However, I would like to deprecate (and remove) this API in the future because of the following reasons:

1. The `coordinator.state` table is an implementation-specific thing (the Consensus Commit transaction has it, but the JDBC transaction doesn't have it).
2. We don't want to keep the state records in the `coordinator.state` table permanently. After successfully committing or rolling back a transaction, we can essentially remove the state record of the transaction (except for the `getState()` reason). So in the future, I want to remove a state record after the transaction is complete.
3. We plan to implement the one-phase commit optimization. In this optimization, we don't need to write the state record to the `coordinator.state` table.

komamitsu: I see. Regarding 1, I don't think we need to expose it to users as far as we can call it internally considering the mode. As for 2 and 3, it would matter. Also, as I mentioned above, automatically internal checking `state` (or `history`) table doesn't work when underlying database isn't working fine. So, it's difficult to suggest the best way to handle this kind of situation. I think this PR overall looks good. Probably we can get feedbacks from customers about this section in the future and let's discuss it deeply again.

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/DistributedTransaction.java
```diff
@@ -69,8 +69,8 @@ public interface DistributedTransaction extends TransactionCrudOperable {
   /**
    * Commits a transaction.
    *
-   * @throws CommitConflictException if conflicts happened. You can retry the transaction in this
-   *     case
+   * @throws CommitConflictException if a transaction conflict occurs, You can retry the transaction
```
feeblefakie, line 72:
```suggestion
   * @throws CommitConflictException if a transaction conflict occurs. You can retry the transaction
```

feeblefakie (approved):
Found one typo, but other than that, LGTM! Thank you!

core/src/main/java/com/scalar/db/api/DistributedTransaction.java
brfrn169: Ah it's not typo actually. The format `@throws XXXException if ...` is frequently used in this project.

PR #805, opened 2023-01-22
refactor: stop performing `execute()` in `RdbEngineStrategy` implementations

laysakura:
## Motivation

This comment: https://github.com/scalar-labs/scalardb/pull/790#pullrequestreview-1262971631

## Major changes

- `RdbEngineStrategy` and its implementations no longer need to import `java.sql.Connection` and `org.apache.commons.dbcp2.BasicDataSource`
  - They now basically just return SQL strings


core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -191,7 +205,15 @@ private void createMetadataTableIfNotExists(Connection connection) throws SQLExc
             + enclose(METADATA_COL_COLUMN_NAME)
             + "))";
 
-    rdbEngine.createMetadataTableIfNotExistsExecute(connection, createTableStatement);
+    String stmt = rdbEngine.tryAddIfNotExistsToCreateTableSql(createTableStatement);
+    try {
+      execute(connection, stmt);
+    } catch (SQLException e) {
+      // Suppress the exception thrown when the table already exists
+      if (!rdbEngine.isDuplicateTableError(e)) {
```
komamitsu, line 213: `RdbEngineMysql` and `RdbEnginePostgresql` still throw UnsupportedOperationException from `isDuplicateTableError()`. How about making them return `false`?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -33,40 +35,34 @@ public interface RdbEngineStrategy {
...
-  void createMetadataTableIfNotExistsExecute(Connection connection, String createTableStatement)
-      throws SQLException;
+  String tryAddIfNotExistsToCreateTableSql(String createTableSql);
 
-  void createMetadataSchemaIfNotExists(Connection connection, String metadataSchema)
-      throws SQLException;
+  String[] createMetadataSchemaIfNotExistsSql(String metadataSchema);
 
-  void deleteMetadataSchema(Connection connection, String metadataSchema) throws SQLException;
+  boolean isCreateMetadataSchemaDuplicateSchemaError(SQLException e);
```
brfrn169, line 50: It looks like we can unify this `isCreateMetadataSchemaDuplicateSchemaError()` and `isDuplicateUserError()` and `isDuplicateSchemaError()`?

brfrn169 (commented):
LGTM overall. Left a comment. PTAL when you have time!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
laysakura: The same reason as this: https://github.com/scalar-labs/scalardb/pull/788#discussion_r1071754726

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
laysakura: f36f6426

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -33,40 +31,34 @@ public interface RdbEngineStrategy {
 
   String computeBooleanValue(boolean value);
 
-  void createNamespaceExecute(Connection connection, String fullNamespace) throws SQLException;
+  String[] createNamespaceExecuteSqls(String fullNamespace);
```
brfrn169, line 34: Maybe we don't need `Execute`?
```suggestion
  String[] createNamespaceSqls(String fullNamespace);
```

brfrn169 (approved):
Much better abstraction! Left a very minor comment, but LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
laysakura: 🙆‍♀️ 
2629af5
b9e3c876

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
komamitsu: I think the situation has been changed. In the case of the previous PR, whether to call that method (`isDuplicateUserError()`) was controlled by JdbcAdmin https://github.com/scalar-labs/scalardb/blob/25a6465739937507b543b66db641ce497dfe3617/core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java#L187-L212. But `rdbEngine.isDuplicateTableError(e)` here in this PR seems to be called regardless of RdbEngineStrategy type?

I just tested the behavior with the following test code

```java
    Properties properties = new Properties();
    properties.put("scalar.db.storage", "mysql");
    properties.put("scalar.db.contact_points", "jdbc:mysql://localhost");
    properties.put("scalar.db.username", "root");
    properties.put("scalar.db.password", "mysql");
    properties.put("scalar.db.transaction_manager", "jdbc");
    JdbcAdmin ja = new JdbcAdmin(new DatabaseConfig(properties));
    String namespace = "ns";
    String table = "tbl";
    ja.dropNamespace(namespace);
    ja.createNamespace(namespace, true);
    TableMetadata metadata =
        TableMetadata.newBuilder().addColumn("c1", DataType.INT).addPartitionKey("c1").build();
    try {
      ja.createTable(namespace, table, metadata);
    } catch (ExecutionException e) {
      e.printStackTrace();
    }
```
I needed to suspend the test process at https://github.com/scalar-labs/scalardb/pull/805/files#diff-d79cc6031633d67d741355267fef8120873fc11f76b10f911b7f38b2920b806dR210 and shut down local MySQL, thought. As a result, UnsupportedOperationException was thrown.

```
java.lang.UnsupportedOperationException
	at com.scalar.db.storage.jdbc.RdbEngineMysql.isDuplicateTableError(RdbEngineMysql.java:129)
	at com.scalar.db.storage.jdbc.JdbcAdmin.createMetadataTableIfNotExists(JdbcAdmin.java:213)
	at com.scalar.db.storage.jdbc.JdbcAdmin.createMetadataSchemaAndTableIfNotExists(JdbcAdmin.java:156)
	at com.scalar.db.storage.jdbc.JdbcAdmin.addTableMetadata(JdbcAdmin.java:143)
	at com.scalar.db.storage.jdbc.JdbcAdmin.createTable(JdbcAdmin.java:89)
```

With `master` branch, it throws `Caused by: com.mysql.cj.jdbc.exceptions.CommunicationsException: Communications link failure`.

laysakura: @komamitsu I understand your point.

The following **was** my points:

- `RdbEngineStrategy.is*Error` are low-level functions to directly interpret SQLExceptions. They will be additionally used in the codebase.
- Current codebases (in master and this topic branch) do not call either `RdbEngineMysql.isDuplicateTableError` nor `RdbEnginePostgresql.isDuplicateTableError` because both issue `CREATE TABLE IF NOT EXISTS` statement.
- It's better to throw exceptions in the functions rather than to return false in case someone changes the codebase to call either of the functions (he/she takes responsibility to correctly implement the functions).

---

But your test code showed that `JdbcAdmin.createTable` is a public function so scalardb users can be affected by the exceptions in `isDuplicateTableError`.

I'm not willing to returning false in unimplemented `RdbEngineStrategy.is*Error` functions because I want to avoid potential (and hard-to-find) bugs.

Which do you think is better?

1. Implement `RdbEngineStrategy.is*Error` correctly now.
2. Stay unimplemented RdbEngineStrategy.is*Error` as-is (throwing exceptions).

komamitsu: IMO, current behavior in this topic branch is unexpected and it affects client-side error handling, so 2 shouldn't be chose.

Maybe another option is:
3. Add `canDuplicateTableErrorHappen` which returns false from `RdbEngineMysql` and `RdbEnginePostgresql` and let `JdbcAdmin` call it before calling `isDuplicateTableError`. We can leave the impl of `isDuplicateTableError` as is.

I think either of 1 or 3 is fine. 1 might be a bit better?

BTW, my concern is we can't detect the behavior change in this topic branch on CI. Maybe it's worth checking current test cases.

laysakura: @komamitsu Got it. I will soon proceed the choice 1.

> BTW, my concern is we can't detect the behavior change in this topic branch on CI. Maybe it's worth checking current test cases.

I agree with it and such a test case to throw `CommunicationsException` would be realized with mocks.
I'll try to add it.

komamitsu: Thanks a lot!

laysakura: https://github.com/scalar-labs/scalardb/compare/2e3c99f332b21efe19e463593487c2f0076347bf..20a40dc

komamitsu (approved):
LGTM, thank you!!

PR #822, opened 2023-02-22
Replace usage of gradle by gradle wrapper in Github workflows

Torch3333:
**Problem** 
All vulnerability checks actions are failing because the "set-version" step of the `vuln-check` workflow fails. 

**Root cause**
This step is using gradle that comes preinstalled with the workflow image. Lately, the pre-installed gradle version was updated to gradle 8 which does not seem compatible with  the `com.palantir.docker` plugin.
image

**Fix**
Use the project `./gradlew` instead of `gradle` which will use gradle 7 in the `vuln-check` WF as well in all other WF as a preemptive measure.

**Validation**
I confirmed the step is working properly with `./gradlew`  on this commit https://github.com/scalar-labs/scalardb/pull/822/commits/3a34f7b128d5ae44f5ebac8eda5d6599fb996555. 
You can see the full `vuln-check` workflow run for this commit here : https://github.com/scalar-labs/scalardb/actions/runs/4238143702/jobs/7364890062
image

### DIFF
--- a/.github/workflows/release-snapshot.yaml
+++ b/.github/workflows/release-snapshot.yaml
@@ -33,14 +33,14 @@ jobs:
       - name: Set version
         id: version
         run: |
-          VERSION=$(gradle :core:properties -q | grep "version:" | awk '{print $2}')
+          VERSION=$(./gradlew :core:properties -q | grep "version:" | awk '{print $2}')
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
 
       - name: Upload SNAPSHOT versions for scalardb, scalardb-server, scalardb-rpc, scalardb-schema-loader, and scalardb-integration-test to Maven Snapshot Repository
         if: contains(steps.version.outputs.version, '-SNAPSHOT')
         run: |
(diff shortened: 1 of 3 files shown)


komamitsu (approved):
Great investigation! Thanks!

komamitsu:
It seems some other workflows are using `gradle` not `gradlew`. Maybe we can fix them as well in this PR (I think fixing them in a separate PR is also okay)

```
$ git grep 'gradle ' | awk -F: '{print $1}' | sort -u
workflows/ci.yaml
workflows/release-snapshot.yaml
workflows/release.yaml
workflows/vuln-check.yaml
```

Torch3333:
@komamitsu Yes, we need to fix them too. Thank you

@komamitsu @brfrn169 @feeblefakie I will also replace the other usage of `gradle` by `gradlew` in the other workflow in this PR.

komamitsu:
Thanks. I'll take care of other ScalarDB family repositories.

Torch3333:
> Thanks. I'll take care of other ScalarDB family repositories.

Ok, thank you!

komamitsu (approved):
LGTM! 👍

brfrn169 (approved):
LGTM! Thank you!

Should we also change the RP subject?

PR #825, opened 2023-02-24
Add repair section for cosmosdb restore

kota2and3kan:
This PR adds how to repair stored procedures of Cosmos DB for NoSQL.
We must repair tables manually after we restore tables using PITR since it doesn't restore stored procedures automatically.

Please take a look!

### DIFF
--- a/docs/backup-restore.md
+++ b/docs/backup-restore.md
@@ -101,6 +101,8 @@ Please see [the doc](https://github.com/scalar-labs/cassy/blob/master/docs/getti
 You can follow the [azure official guide](https://docs.microsoft.com/en-us/azure/cosmos-db/restore-account-continuous-backup#restore-account-portal). After restoring backups. change the default consistencies of the restored databases to `STRONG`
 It is recommended to use the mid-time of paused duration as a restore point as we explained earlier.
 
+ScalarDB tables use stored procedures of Cosmos DB for NoSQL. However, the PITR feature of Cosmos DB for NoSQL doesn't restore stored procedures. So, you must create stored procedures in all tables. You can do it using ScalarDB Schema Loader the `--repair-all` flag. Please refer to the Schema Loader document [Repair tables](https://github.com/scalar-labs/scalardb/blob/master/docs/schema-loader.md#repair-tables).
+
 ### DynamoDB (restore)
 
 You can basically follow [the official doc](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/PointInTimeRecovery.Tutorial.html). However, a table can only be restored with an alias, so you need to restore a table with an alias and delete the original table and rename the alias to the original name to restore the same named tables.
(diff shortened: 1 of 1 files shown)


Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM, thank you!

brfrn169 (approved):
LGTM! Thank you!

docs/backup-restore.md
```diff
@@ -101,6 +101,8 @@ Please see [the doc](https://github.com/scalar-labs/cassy/blob/master/docs/getti
 You can follow the [azure official guide](https://docs.microsoft.com/en-us/azure/cosmos-db/restore-account-continuous-backup#restore-account-portal). After restoring backups. change the default consistencies of the restored databases to `STRONG`
 It is recommended to use the mid-time of paused duration as a restore point as we explained earlier.
 
+ScalarDB tables use stored procedures of Cosmos DB for NoSQL. However, the PITR feature of Cosmos DB for NoSQL doesn't restore stored procedures. So, you must create stored procedures in all tables. You can do it using ScalarDB Schema Loader with the `--repair-all` flag. Please refer to the Schema Loader document [Repair tables](https://github.com/scalar-labs/scalardb/blob/master/docs/schema-loader.md#repair-tables).
```
feeblefakie, line 104: It's not our issue, but `Cosmos DB for NoSQL` looks confusing when it's integrated into a sentence. (I was actually pretty confused when I first looked at the sentence.)
So, I think we can use only `Cosmos DB` when it makes sense.
(for example, we should say `Cosmos DB for NoSQL` as one of the supported databases, but we might be able to commit `for NoSQL` otherwise.)
```suggestion
ScalarDB implements the Cosmos DB adapter by using its stored procedures, which are installed when creating schemas with ScalarDB Schema Loader. However, the PITR feature of Cosmos DB doesn't restore stored procedures. So, you need to re-install the required stored procedures for all tables after restoration. You can do it using ScalarDB Schema Loader with the `--repair-all` option. Please refer to the Schema Loader document [Repair tables](https://github.com/scalar-labs/scalardb/blob/master/docs/schema-loader.md#repair-tables) for more details.
```

feeblefakie (commented):
Thank you for the update!
Left some suggestions, so PTAL!

docs/backup-restore.md
kota2and3kan: @feeblefakie  
Thank you for your suggestion!
I applied it.

> So, I think we can use only Cosmos DB when it makes sense.
> (for example, we should say Cosmos DB for NoSQL as one of the supported databases, but we might be able to commit for NoSQL otherwise.)

I see. I understood.
I will omit `for NoSQL` if I can when I create/update some documents in the future.

feeblefakie (approved):
LGTM! Thank you!

PR #826, opened 2023-03-04
Make overridden methods throw Exception in integration tests

brfrn169:
This PR makes overridden methods in the integration tests throw `Exception`. We do this because we can't assume which exception will be thrown in the overriding methods. Please take a look!


komamitsu (approved):
LGTM, thank you!

Torch3333 (approved):
LGTM, thank you!

PR #829, opened 2023-03-14
Make transaction ID required in UnknownTransactionStatusException

brfrn169:
This PR makes the transaction ID required in `UnknownTransactionStatusException`. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/exception/transaction/UnknownTransactionStatusException.java
+++ b/core/src/main/java/com/scalar/db/exception/transaction/UnknownTransactionStatusException.java
@@ -17,7 +17,7 @@ public class UnknownTransactionStatusException extends TransactionException {
   }
 
   /**
-   * @return a transaction ID that associated with the transaction whose status is unknown
+   * @return a transaction ID associated with the transaction whose status is unknown
    * @deprecated As of release 3.9.0. Will be removed in release 5.0.0
    */
   @Deprecated
@@ -26,9 +26,9 @@ public class UnknownTransactionStatusException extends TransactionException {
   }
 
   /**
-   * Returns a transaction ID that associated with the transaction whose status is unknown.
+   * Returns a transaction ID associated with the transaction whose status is unknown.
    *
-   * @return a transaction ID that associated with the transaction whose status is unknown
+   * @return a transaction ID associated with the transaction whose status is unknown
    */
   public String getTransactionId() {
     return unknownTxId;
--- a/core/src/main/java/com/scalar/db/exception/transaction/UnknownTransactionStatusException.java
+++ b/core/src/main/java/com/scalar/db/exception/transaction/UnknownTransactionStatusException.java
@@ -1,25 +1,20 @@
 package com.scalar.db.exception.transaction;
 
-import java.util.Optional;
-
 /** An exception thrown when the transaction status of the committing transaction is unknown. */
 public class UnknownTransactionStatusException extends TransactionException {
-  private Optional<String> unknownTxId = Optional.empty();
+  private final String unknownTxId;
 
-  public UnknownTransactionStatusException(String message) {
+  public UnknownTransactionStatusException(String message, String txId) {
     super(message);
-  }
-
-  public UnknownTransactionStatusException(String message, Throwable cause) {
-    super(message, cause);
+    this.unknownTxId = txId;
   }
 
   public UnknownTransactionStatusException(String message, Throwable cause, String txId) {
     super(message, cause);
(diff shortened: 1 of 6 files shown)


komamitsu (approved):
LGTM, thanks!

core/src/main/java/com/scalar/db/exception/transaction/UnknownTransactionStatusException.java
```diff
@@ -1,25 +1,20 @@
...
   }
 
   public UnknownTransactionStatusException(String message, Throwable cause, String txId) {
     super(message, cause);
-    this.unknownTxId = Optional.ofNullable(txId);
+    this.unknownTxId = txId;
   }
 
-  public Optional<String> getUnknownTransactionId() {
+  public String getUnknownTransactionId() {
```
feeblefakie, line 17: I'm wondering what the effect would be to users.
If an application catches `UnknownTransactionStatusException` and uses `getUnknownTransactionId`  method, then the application has to be modified?

Is it strongly required for future changes?
I'm wondering if the benefit outweighs the downside.

feeblefakie (commented):
Thank you!
The code looks good but I left a question for clarification.

core/src/main/java/com/scalar/db/exception/transaction/UnknownTransactionStatusException.java
brfrn169: > If an application catches UnknownTransactionStatusException and uses getUnknownTransactionId method, then the application has to be modified?

Yes, it's an incompatible change. So that situation will cause a compile error. But I didn't think that many users actually use `getUnknownTransactionId()` at the moment.

> Is it strongly required for future changes?
> I'm wondering if the benefit outweighs the downside.

I thought I needed this change when I was working on another private project that depends on ScalarDB, and I was wondering why the transaction ID is optional, and when it's null and when it's not. Then, I checked the source code again and found out that we can always return the transaction ID in case of `UnknownTransactionStatusException`. That's the background of this change.

To summarize, the benefit is that the current behavior of the optional transaction ID is a bit confusing for users, and this change fixes this. And the downside is that it's an incompatible change, so if users are using `getUnknownTransactionId`, they have to modify their code when upgrading ScalarDB. And we shouldn't make such an incompatible change between minor versions. But as I said, I didn't think many users were using it at the moment, so I didn't think the impact would be significant.

What do you think?

komamitsu: This might be an off-topic comment, but it would be great if any exception thrown from ScalarDB layer contains a transaction ID information not only in case of `UnknownTransactionStatusException`.

I'll take care of https://github.com/scalar-labs/scalardb-sql/pull/133#discussion_r1135043578, but I'm thinking we might need a consistent way (not only for Spring Data integration 2PC API) to get a transaction ID information from a thrown exception.

brfrn169: As discussed it offline, and we concluded that we don't want to break the backward compatibility so we will leave the original method as it is and add another method that returns non-optional transaction ID.

I will fix this PR. Thanks.

brfrn169: Fixed in 51311d0bb8642461a63506d2e6d41b166cede76b.

@feeblefakie Can you please take a look again when you have time?

core/src/main/java/com/scalar/db/exception/transaction/UnknownTransactionStatusException.java
```diff
@@ -4,22 +4,33 @@
...
-    this.unknownTxId = Optional.ofNullable(txId);
+    this.unknownTxId = txId;
   }
 
+  /**
+   * @return a transaction ID that associated with the transaction whose status is unknown
+   * @deprecated As of release 3.9.0. Will be removed in release 5.0.0
+   */
+  @Deprecated
   public Optional<String> getUnknownTransactionId() {
```
brfrn169, line 24: To maintain the backward compatibility, we leave this method as it is.

Torch3333 (approved):
LGTM, thank you!

brfrn169:
@feeblefakie @Torch3333 

I discussed [the comment](https://github.com/scalar-labs/scalardb/pull/829#discussion_r1136437860) with @komamitsu and we came to the conclusion that we should add a transaction ID not only for `UnknownTransactionStatusException` but for all exceptions. So I will close this PR and create another one for that. Thanks for the review.

PR #830, opened 2023-03-15
Add transaction ID to exceptions

brfrn169:
This PR adds a transaction ID to the exceptions. It also add a transaction ID to the log messages. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
@@ -42,7 +42,7 @@ public abstract class ActiveTransactionManagedDistributedTransactionManager
               try {
                 t.rollback();
               } catch (Exception e) {
-                logger.warn("rollback failed", e);
+                logger.warn("rollback failed. transactionId: {}", t.getId(), e);
               }
             });
   }
@@ -50,8 +50,7 @@ public abstract class ActiveTransactionManagedDistributedTransactionManager
   private void add(ActiveTransaction transaction) throws TransactionException {
     if (activeTransactions.putIfAbsent(transaction.getId(), transaction) != null) {
       transaction.rollback();
-      throw new TransactionException(
-          "The transaction already exists. transactionId: " + transaction.getId());
+      throw new TransactionException("The transaction already exists", transaction.getId());
     }
   }
 
@@ -68,7 +67,8 @@ public abstract class ActiveTransactionManagedDistributedTransactionManager
                 new TransactionNotFoundException(
                     "A transaction associated with the specified transaction ID is not found. "
                         + "It might have been expired. transactionId: "
-                        + txId));
+                        + txId,
+                    txId));
   }
 
   @Override
--- a/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
@@ -38,11 +38,11 @@ public abstract class ActiveTransactionManagedDistributedTransactionManager
             config.getActiveTransactionManagementExpirationTimeMillis(),
             TRANSACTION_EXPIRATION_INTERVAL_MILLIS,
             t -> {
-              logger.warn("the transaction is expired. transactionId: {}", t.getId());
+              logger.warn("the transaction is expired. transaction ID: {}", t.getId());
               try {
                 t.rollback();
               } catch (Exception e) {
-                logger.warn("rollback failed. transactionId: {}", t.getId(), e);
+                logger.warn("rollback failed. transaction ID: {}", t.getId(), e);
               }
             });
   }
@@ -66,8 +66,7 @@ public abstract class ActiveTransactionManagedDistributedTransactionManager
             () ->
                 new TransactionNotFoundException(
                     "A transaction associated with the specified transaction ID is not found. "
-                        + "It might have been expired. transactionId: "
-                        + txId,
+                        + "It might have been expired",
                     txId));
   }
 
--- a/core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java
+++ b/core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java
@@ -9,12 +9,12 @@ public class TransactionException extends Exception {
   @Nullable private final String transactionId;
 
   public TransactionException(String message, @Nullable String transactionId) {
-    super(message);
+    super(addTransactionIdToMessage(message, transactionId));
     this.transactionId = transactionId;
   }
 
   public TransactionException(String message, Throwable cause, @Nullable String transactionId) {
-    super(message, cause);
+    super(addTransactionIdToMessage(message, transactionId), cause);
     this.transactionId = transactionId;
   }
 
@@ -25,5 +25,20 @@ public class TransactionException extends Exception {
    */
   public Optional<String> getTransactionId() {
     return Optional.ofNullable(transactionId);
+  }
+
+  private static String addTransactionIdToMessage(String message, @Nullable String transactionId) {
+    if (transactionId == null) {
+      return message;
+    }
+
+    String suffix = ". transaction ID: " + transactionId;
+
+    // To avoid a duplicated transaction ID, check if the message already has a transaction ID
+    if (!message.endsWith(suffix)) {
+      return message + suffix;
+    } else {
+      return message;
+    }
   }
 }
--- a/core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java
+++ b/core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java
@@ -1,13 +1,29 @@
 package com.scalar.db.exception.transaction;
 
+import java.util.Optional;
+import javax.annotation.Nullable;
+
 /** Base class for all exceptions thrown by the Transaction API. */
 public class TransactionException extends Exception {
 
-  public TransactionException(String message) {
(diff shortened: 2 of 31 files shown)


core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java
```diff
@@ -1,13 +1,29 @@
 package com.scalar.db.exception.transaction;
 
+import java.util.Optional;
+import javax.annotation.Nullable;
+
 /** Base class for all exceptions thrown by the Transaction API. */
 public class TransactionException extends Exception {
 
-  public TransactionException(String message) {
+  @Nullable private final String transactionId;
```
brfrn169, line 9: Added the `transactionId` field to `TransactionException` that's a base class for all exceptions thrown by the Transaction API.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
```diff
@@ -42,7 +42,7 @@ public ActiveTransactionManagedDistributedTransactionManager(DatabaseConfig conf
               try {
                 t.rollback();
               } catch (Exception e) {
-                logger.warn("rollback failed", e);
+                logger.warn("rollback failed. transactionId: {}", t.getId(), e);
```
Torch3333, line 45: Since all exception extends `TransactionException`, would it be possible to append the `transactionId:<txId>`  to the message in the `TransactionException` constructor instead?

core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java
komamitsu, line 9: How about overriding `getMessage()` to include the transaction ID so that we can pass a message without transaction ID to constructer like this?

```java
  @Override
  public String getMessage() {
    return String.format("%s. transactionID:%s", super.getMessage(), transactionId);
  }
```

```java
    throw new TransactionException("The transaction already exists", transaction.getId());
```

brfrn169: @komamitsu @Torch3333 

I have found a problem with this. We sometimes catch a transaction exception and wrap it with its `getMessage()` and rethrow it like this:
https://github.com/scalar-labs/scalardb/blob/3e5751acd96a5937464a70362d9d8d9610193928/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java#L53-L56

In this case, if we override `TransactionException.getMessage()` that way, the exception message shows `transactionId` twice. And I haven't come up with a smart solution for this problem...

Do you have any idea?

Torch3333: It is not a smart solution but I think its the simplest to check if the message does not already contains the transaction ID. 

```
  @Override
  public String getMessage() {
    String message = super.getMessage();
    String suffix = "transaction ID: " + transactionId;
    if (!message.endsWith(suffix)) {
      return String.format("%s. %s", message, suffix);
    } else {
      return message;
    }
  }
  ```
Another option would be to add another method, `getOriginalMessage()` to `TransactionException`,  to retrieve the original message without the transaction ID, but I prefer the solution above for its simplicity.

Comment (komamitsu) [reply core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java:9]: Another option might be checking the existence of TransactionException-ish cause like this?

```java
    @Override
  public String getMessage() {
    if (hasTransactionExceptionInCauses()) {
      // The cause already contains transaction ID in its message
      return super.getMessage();
    }
    else {
      return String.format("%s. transactionID:%s", super.getMessage(), transactionId);
    }
  }
```

komamitsu: Ah, I think I had the wrong idea. Let me see again...

komamitsu: On second thought, it works as follows?

#### Case#1

```java
 if (e instanceof PreparationConflictException) { 
   throw new CommitConflictException(e.getMessage(), e, e.getTransactionId().orElse(null)); 
 } 
```

In this case, the new CommitConflictException instance should have
- message: "Preparation failed. transactionID=xxxx"
- cause: PreparationConflictException
  - message: "Preparation failed. transactionID=xxxx"
  - cause: Null or something
  - transactionId: xxxx
- transactionId: xxxx

With suggested the above `getMessage()`,
- CommitConflictException.getMessage() returns "Preparation failed. transactionID=xxxx"
  - as `hasTransactionExceptionInCauses()` returns true
- PreparationConflictException.getMessage() returns "Preparation failed. transactionID=xxxx"

#### Case#2

```java
 if (e instanceof IllegalArgumentException) { 
   throw new CommitConflictException(e.getMessage(), e, transactionId); 
 } 
```

In this case, the new CommitConflictException instance should have
- message: "An invalid parameter"
- cause: IllegalArgumentException
  - message: "An invalid parameter"
  - cause: Null or something
- transactionId: xxxx

With suggested the above `getMessage()`,
- CommitConflictException.getMessage() returns "An invalid parameter. transactionID=xxxx"
  - as `hasTransactionExceptionInCauses()` returns false
- IllegalArgumentException.getMessage() returns "An invalid parameter"

komamitsu: BTW, IMO I don't think it's a so severe issue to happen to show `transactionId` twice

brfrn169: It looks like we have one more case as follows:
https://github.com/scalar-labs/scalardb/blob/3e5751acd96a5937464a70362d9d8d9610193928/core/src/main/java/com/scalar/db/transaction/rpc/GrpcTransactionOnBidirectionalStream.java#L248

This is the case when we use ScalarDB Server. In this case, `error.getMessage()` should have the transaction ID, but the exception doesn't have any cause exceptions.

> BTW, IMO I don't think it's a so severe issue to happen to show transactionId twice

It looks like when we use ScalarDB Server (and ScalarDB Cluster), the transaction ID is always duplicated. I agree that it's not very severe, but I think we should handle it.

Example of the transaction ID duplicated messages:
```
The transaction already exists. transaction ID: 5d37796a-0e43-4ebb-8b4b-3da60d5a69a6. transaction ID: 5d37796a-0e43-4ebb-8b4b-3da60d5a69a6
```

So maybe, the Vincent's idea is more reasonable so far, though it's not perfect?

```java
  @Override
  public String getMessage() {
    String message = super.getMessage();
    String suffix = "transaction ID: " + transactionId;
    if (!message.endsWith(suffix)) {
      return String.format("%s. %s", message, suffix);
    } else {
      return message;
    }
  }
```

If there are no objections, I will go ahead with Vincent's idea.

komamitsu: LGTM!

brfrn169: @komamitsu @Torch3333 Thank you for the discussion! Fixed in 4e8dae7da1883defb20a0a4196be33a84c1bfbe9.

Please take a look again when you have time!

core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java
```diff
@@ -1,13 +1,44 @@
...
+  /**
+   * Returns a transaction ID that associated with the transaction whose status is unknown.
+   *
+   * @return a transaction ID that associated with the transaction whose status is unknown
+   */
+  public Optional<String> getTransactionId() {
+    return Optional.ofNullable(transactionId);
+  }
+
+  private static String addTransactionIdToMessage(String message, @Nullable String transactionId) {
```
brfrn169, line 30: I added this method to add a transaction ID to the message based on Vincent's idea. The constructors convert the message with the method and pass it to the constructors of the superclass.

komamitsu (approved):
LGTM! Thanks!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/exception/transaction/TransactionException.java
```diff
@@ -1,13 +1,44 @@
 package com.scalar.db.exception.transaction;
 
+import java.util.Optional;
+import javax.annotation.Nullable;
+
 /** Base class for all exceptions thrown by the Transaction API. */
 public class TransactionException extends Exception {
 
-  public TransactionException(String message) {
```
jnmt, line 6: @brfrn169 Sorry, it might be a stupid question, but is this change backward-compatible? Or is the usage like the following uncommon or unexpected?

https://github.com/scalar-labs/scalardb-benchmarks/blob/b236c1f75dfb5e95337e973a84fe26c38dfd15b8/src/main/java/com/scalar/db/benchmarks/tpcc/transaction/NewOrderTransaction.java#L169

brfrn169: @jnmt (CC: @feeblefakie)

The PR changed the signatures of the constructors of the transaction-related exceptions, so applications that instantiate the exceptions will encounter a compile error due to the change. In that sense, it's not backward-compatible.

Currently, ScalarDB doesn't have an explicit rule about which classes/methods/constructors are considered as public (exposed to users) or private (internal use). Mature OSS basically has such a rule, and the developers can make incompatible changes to the private ones, but not to the public ones.

And I thought we didn't assume that end users would use the constructors of the exceptions (so the constructors are considered as private (internal use)), so I made the (backward-incompatible) changes.

PR #840, opened 2023-04-11
Update supported Java version in getting started

kota2and3kan:
This PR updates the description of the supported Java version in the getting started guide.

In the previous getting started guide, we built ScalarDB itself in that steps. So, we must use Java 8 to build ScalarDB.
However, in the current getting started guide, we can use `Java 8 or higher` since it gets pre-compiled ScalarDB from the Maven repository.

And, I think there is a possibility that the current description may mislead users that `ScalarDB only supports Java 8`.
(Actually, I got a such question from a user.)
So, I think it is better to describe `Java 8 or higher` in the getting started guide.

Also, Gradle 7.1 doesn't support Java 17 or higher.
So, I updated the Gradle version of the getting started project to 7.6 (current latest version).
https://docs.gradle.org/current/userguide/compatibility.html

Please take a look!

### DIFF
--- a/docs/getting-started-with-scalardb-on-cassandra.md
+++ b/docs/getting-started-with-scalardb-on-cassandra.md
@@ -7,7 +7,7 @@ This document briefly explains how you can get started with ScalarDB on Cassandr
 
 ScalarDB is written in Java and uses Cassandra as an underlying storage implementation, so the following software is required to run it.
 
-* [Oracle JDK 8](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) (or OpenJDK 8)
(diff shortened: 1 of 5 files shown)


brfrn169 (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

feeblefakie:
@kota2and3kan Sorry, can you fix the conflict?

kota2and3kan:
@feeblefakie
Thank you for your review!
I will fix the conflict. And after that, I will merge this PR.

PR #841, opened 2023-04-19
Make transaction metadata nullable to support existing databases

jnmt:
This PR adds a feature to handle null transaction metadata. To support existing databases, we assume that we ask users to add transaction metadata columns (with null values in records) in existing tables. So, as this PR does, we would like to handle those records with null transaction metadata as a sort of 'deemed' committed state.


core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
```diff
@@ -137,28 +160,35 @@ private void add(Get base) {
...
-            v -> {
-              if (isBeforeRequired(v, base.getPartitionKey(), base.getClusteringKey())) {
-                values.add(v.copyWith(Attribute.BEFORE_PREFIX + v.getName()));
+            column -> {
+              if (isBeforeRequired(column, base.getPartitionKey(), base.getClusteringKey())) {
+                if (column.getName().equals(Attribute.VERSION) && column.hasNullValue()) {
+                  // For preparing an update of a deemed-commit state record, we need to use
+                  // version 0 rather than NULL since we want to distinguish the preparation from
+                  // an initial record insertion.
+                  columns.add(IntColumn.of(Attribute.BEFORE_VERSION, 0));
```
jnmt, line 175: As commented in the code, we don't use the original NULL value for `before_version` in prepare records
to distinguish them from prepare records for newly-inserted ones.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RollbackMutationComposer.java
```diff
@@ -77,26 +80,39 @@ private Put composePut(Operation base, TransactionResult result) throws Executio
...
     result
-        .getValues()
+        .getColumns()
         .forEach(
             (k, v) -> {
               if (beforeImageColumnNames.contains(k)) {
                 String key = k.substring(Attribute.BEFORE_PREFIX.length());
-                map.put(key, v.copyWith(key));
+                if (key.equals(Attribute.VERSION) && v.getIntValue() == 0) {
+                  columns.add(IntColumn.ofNull(Attribute.VERSION));
```
jnmt, line 91: For rollback, we need to change 0 -> null as the opposite side of prepare.

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
```diff
@@ -137,28 +160,35 @@ private void add(Get base) {
...
             });
-    return values;
+    return columns;
   }
 
-  private boolean isBeforeRequired(Value<?> value, Key primary, Optional<Key> clustering) {
-    return !value.getName().startsWith(Attribute.BEFORE_PREFIX)
-        && !isValueInKeys(value, primary, clustering);
+  private boolean isBeforeRequired(Column<?> column, Key primary, Optional<Key> clustering) {
+    return !column.getName().startsWith(Attribute.BEFORE_PREFIX)
```
komamitsu, line 185: This might be a silly question, but what if an original column name starts with `before_` prefix like `before_update_balance` (non-primary or clustering key)? This logic seems to return false, but I think `before_before_update_balance` column should be taken care of.

jnmt: Oh, good point. Yeah, it should be taken care of **if we allow users to use a column name starting with `before_`.** For new projects, at least, prohibiting such column names would not cause serious problems. For existing databases, some users might be using such column names, but I guess it would be rare, and we could ask users to rename them. So, IMHO, we can prohibit user columns with `before_` in both cases for now.

@brfrn169 By the way, I think we don't have any guard for those names in the schema-loader, and I was able to create a `before_xx` column and faced an issue that its **before** column was not updated in the prepare phase. Is my understanding correct?

komamitsu: Ah, I can reproduce it with the current code.

**ScalarDB SQL CLI**
```
0: scalardb> select * from example_app.account2;
+----+----------+-----+
| id | before_x |  y  |
+----+----------+-----+
| 1  | 10       | 100 |
+----+----------+-----+
1 row selected (0.005 seconds)
0: scalardb> update example_app.account2 set before_x = 20, y = 200 where id = 1;
1 row affected (0.04 seconds)
```

**psql**
```
postgres=# select id, before_x, y, tx_state, before_before_x, before_y from example_app.account2;
 id | before_x |  y  | tx_state | before_before_x | before_y 
----+----------+-----+----------+-----------------+----------
  1 |       20 | 200 |        3 |                 |      100
(1 row)
```

jnmt: @komamitsu Thanks! Yeah, I faced the same issue.

brfrn169: I thought we accept to create the `before_xx` column as a normal column, and in this case, its metadata column should be `before_before_xx`. But, we might miss something to handle that. 

The logic to detect the before column is here:
https://github.com/scalar-labs/scalardb/blob/master/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java#L179-L204

The basic idea is, if there are both `xxx` and `before_xx` (before prefixed) columns in the table, `xxx` is an after image column, and `before_xx` is a before image column.

brfrn169: > By the way, I think we don't have any guard for those names in the schema-loader,

Ah, yes, it looks like we don't have any guard for that. I think we should add some validation for that in `ConsensusCommitAdmin.createTable()`:
https://github.com/scalar-labs/scalardb/blob/44a7baf62937b9a44e19d36516b20dcdd7130264/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java#L83-L88

jnmt: > Ah, yes, it looks like we don't have any guard for that. I think we should add some validation for that in ConsensusCommitAdmin.createTable():

@brfrn169 @komamitsu Thank you for the comment. So, let's fix it in another PR.

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
```diff
@@ -51,68 +50,92 @@ public void add(Operation base, @Nullable TransactionResult result) {
...
+    PutBuilder.Buildable putBuilder =
+        Put.newBuilder()
+            .namespace(base.forNamespace().get())
+            .table(base.forTable().get())
+            .partitionKey(base.getPartitionKey())
+            .consistency(Consistency.LINEARIZABLE);
+
+    if (base.getClusteringKey().isPresent()) {
+      putBuilder.clusteringKey(base.getClusteringKey().get());
+    }
```
Torch3333, line 62:
```suggestion
    base.getClusteringKey().ifPresent(putBuilder::clusteringKey);
```
It is detail but this can be simplified.

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
```diff
@@ -51,68 +50,92 @@ public void add(Operation base, @Nullable TransactionResult result) {
...
```
Torch3333, line 107:
```suggestion
    base.getClusteringKey().ifPresent(putBuilder::clusteringKey);
```
Idem.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RollbackMutationComposer.java
```diff
@@ -77,26 +80,39 @@ private Put composePut(Operation base, TransactionResult result) throws Executio
...
+            .partitionKey(result.getPartitionKey().get())
+            .condition(
+                ConditionBuilder.putIf(ConditionBuilder.column(ID).isEqualToText(id))
+                    .and(ConditionBuilder.column(STATE).isEqualToInt(result.getState().get()))
+                    .build())
+            .consistency(Consistency.LINEARIZABLE);
+
+    if (result.getClusteringKey().isPresent()) {
+      putBuilder.clusteringKey(result.getClusteringKey().get());
+    }
```
Torch3333, line 111:
```suggestion
    result.getClusteringKey().ifPresent(putBuilder::clusteringKey);
```
Idem.

Torch3333 (approved):
LGTM, thank you. 
I only left minor refactoring suggestions.

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
feeblefakie: We can probably think of introducing configurable prefix?
(e.g., if `before_` is already used, we set `prev_` as the prefix for before columns)
It's gonna be one time configuration before starting though.

komamitsu: > configurable prefix

Sounds reasonable to me! Though it depends on the balance of the benefit and engineering side effort (Analytics also needs to know the configuration.)

BTW, it's a bit off-topic, but I noticed `metadata` table doesn't have any information on whether it's a `before` column. I think it would be useful for `before` column check in a robust way. Maybe adding a new type `BEFORE` to the existing `key_type`...?

komamitsu:
```java
       return tableMetadata 
           .getColumnNames() 
           .contains(columnName.substring(Attribute.BEFORE_PREFIX.length())); 
```
https://github.com/scalar-labs/scalardb/pull/841#discussion_r1171961632

I also think this logic works for most cases except that a user has used both `x` and `before_x` in their database and wants to migrate the table to ScalarDB.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RollbackMutationComposer.java
```diff
@@ -77,26 +80,39 @@ private Put composePut(Operation base, TransactionResult result) throws Executio
...
-                new ConditionalExpression(STATE, toStateValue(result.getState()), Operator.EQ)))
-        .withConsistency(Consistency.LINEARIZABLE)
-        .withValues(map.values());
+    PutBuilder.Buildable putBuilder =
+        Put.newBuilder()
+            .namespace(base.forNamespace().get())
+            .table(base.forTable().get())
+            .partitionKey(result.getPartitionKey().get())
+            .condition(
+                ConditionBuilder.putIf(ConditionBuilder.column(ID).isEqualToText(id))
```
komamitsu, line 104: I'm wondering if `version` should be contained in this conditions. It doesn't need to be?

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
brfrn169: @jnmt Sorry, just a confirmation. In the current logic, does creating the `before_xx` column as a normal column work or not? I think it works.

> I think we should add some validation for that in `ConsensusCommitAdmin.createTable()`:

I meant here that we should add the validation to prevent from being created both `xxx` and `before_xxx` as normal columns. I think just creating a `before_xxx` column works currently. Is there a bug in this?

jnmt: > > I think we should add some validation for that in `ConsensusCommitAdmin.createTable()`:
> 
> I meant here that we should add the validation to prevent from being created both `xxx` and `before_xxx` as normal columns. I think just creating a `before_xxx` column works currently. Is there a bug in this?

@brfrn169 Oh, that's my misunderstanding, but anyway, I don't think it works well even in the current code, according to reproductions by Mitsu(https://github.com/scalar-labs/scalardb/pull/841#discussion_r1171065262) and me. Probably a bug unfortunately.

brfrn169: Ah, okay. It should be a bug. Let me check. Thanks.

brfrn169: Ah, okay we should use `ConsensusCommitUtils.isAfterImageColumn()` here:

https://github.com/scalar-labs/scalardb/blob/44a7baf62937b9a44e19d36516b20dcdd7130264/core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java#L155

I thought I had fixed it 😞

brfrn169: Oh, we already have the validation to check if a table has transaction meta columns when creating the table: 
https://github.com/scalar-labs/scalardb/blob/44a7baf62937b9a44e19d36516b20dcdd7130264/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java#L46-L48

core/src/main/java/com/scalar/db/transaction/consensuscommit/RollbackMutationComposer.java
jnmt: It's enough to check only the transaction ID and the status to ensure that no other transaction updates this record for either commit or rollback. Moreover, I think only transaction ID would be mandatory here since it is decided that this record must be rolled back (no chance of role forwarding), and the transaction ID always be updated if someone recovers.

komamitsu: Ah, sorry for the silly question. I had a wrong idea that `ID` meant a key not a transaction ID at that moment. I sometimes confuse them...

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
jnmt: @brfrn169 Thanks for the additional info. But I guess it only checks to prevent from being created both xxx and before_xxx as normal columns. We can still create before_xxx column itself...

brfrn169: @jnmt Yes, the following bug fix allows users to create `before_` prefixed columns:

https://github.com/scalar-labs/scalardb/pull/844

feeblefakie (commented):
Left a few questions, so PTAL when you get a chance.

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
jnmt: Thanks for seeking clarification. Simply put, the record cannot be rollbacked correctly since the recovery process falls in the deletion path for the initial record case here.
https://github.com/scalar-labs/scalardb/blob/b3cf9c9735c534a37404a6366734710d0c79c3fc/core/src/main/java/com/scalar/db/transaction/consensuscommit/RollbackMutationComposer.java#L67-L69

core/src/main/java/com/scalar/db/transaction/consensuscommit/RollbackMutationComposer.java
```diff
@@ -55,17 +57,18 @@ public void add(Operation base, @Nullable TransactionResult result) throws Execu
       // the record was not prepared (yet) by this transaction or has already been rollback deleted
       return;
     }
-    if (!latest.getId().equals(id)) {
+    if (latest.getId() == null || !latest.getId().equals(id)) {
```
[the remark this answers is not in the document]
jnmt: No, I think this check is necessary for the following corner case.

TX1: prepare record A to update deemed-commit-state record (i.e., TxID = null -> TX1)
TX1: abort and start rollback (record A is still in prepare state)
TX2: read record A, see prepare state, try to rollback and succeed (TxID = TX1 -> null)
TX1: get the latest state and see TxID = null in this add method

jnmt:
@Torch3333 Thank you for the feedback! Fixed in b3cf9c9. PTAL if you have time.

Torch3333 (approved):
LGTM, thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
feeblefakie: Ah, right. Thank you!
I think it would be great if you could make the comment more specific about the case.
(i.e., adding the reason why we want to use 0 instead of null for before_version.)

core/src/main/java/com/scalar/db/transaction/consensuscommit/RollbackMutationComposer.java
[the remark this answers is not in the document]
feeblefakie: Ah, right, now I remember it. Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
jnmt: Sure, I will try to revise it with a more specific reason.

jnmt: @feeblefakie Fixed in d4566f0. PTAL when you get a chance.

komamitsu:
@jnmt (cc: @brfrn169) Just noticed this PR doesn't have integration test cases for NULL status record. Do you think we should add some integration tests? I think it would require non-straightforward way to add NULL status records (directly using the native JDBC?) at arrangement phase. So I'm not sure those integration tests should be added in this PR or in another PR, though.

jnmt:
> @jnmt (cc: @brfrn169) Just noticed this PR doesn't have integration test cases for NULL status record. Do you think we should add some integration tests? I think it would require non-straightforward way to add NULL status records (directly using the native JDBC?) at arrangement phase. So I'm not sure those integration tests should be added in this PR or in another PR, though.

Thanks for the question. Yes, we should add some integration tests since this is a kind of major change in the protocol behavior. We are also thinking about applying the Jepsen tests for it and going to discuss it at some point in time (next or next next week). But, in any case, the integration tests will be added by another PR.

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitNullMetadataIntegrationTestBase.java
jnmt, line 1: I added integration tests for null-tx-metadata handling (incl. rollback/roll-forward) referencing `ConsensusCommitSpecificIntegrationTestBase`, but skipped the following cases.
1) Tests that do not use existing data (e.g., insert a new record and do something)
2) Tests for conflict / write skew situations

jnmt (commented):
Sorry for re-reviewing, but I updated a little for a better understanding of the logic and added integration tests based on the feedback in db51113. So, it would be great if you could take a look again when you get a chance.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionResult.java
```diff
@@ -129,4 +136,23 @@ public long getCommittedAt() {
...
+  public boolean isDeemedAsCommitted() {
+    return getId() == null;
+  }
+
+  public boolean hasBeforeImage() {
+    // We need to check not only before_id but also before_version to determine if the record has
+    // the before image or not since we set before_version to 0 for the prepared record when
+    // updating the record deemed as the committed state (cf. PrepareMutationComposer).
+    return !getBeforeIdColumn().hasNullValue() || !getBeforeVersionColumn().hasNullValue();
+  }
```
brfrn169, line 149: 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/RollbackMutationComposer.java
brfrn169: I think adding a comment for this would be helpful.

brfrn169 (approved):
Left several minor comments but LGTM other than that. Thank you!

brfrn169 (approved):
Left several minor comments but LGTM other than that. Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionResult.java
```diff
@@ -106,12 +108,17 @@ public Map<String, Column<?>> getColumns() {
...
   }
 
+  @Nullable
   public String getId() {
     return getText(Attribute.ID);
   }
 
   public TransactionState getState() {
-    return TransactionState.getInstance(getInt(Attribute.STATE));
+    int state = getInt(Attribute.STATE);
```
feeblefakie, line 117: Although `getInt` returns 0 for a null value, 
but should it be slightly more appropriate/readable to use `isNull` since it actually checks if the value is null?

feeblefakie (commented):
Overall looking good to me!
Left one minor comment.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionResult.java
```diff
@@ -129,4 +136,23 @@ public long getCommittedAt() {
...
+    return !getBeforeIdColumn().hasNullValue() || !getBeforeVersionColumn().hasNullValue();
+  }
+
+  public TextColumn getBeforeIdColumn() {
+    return (TextColumn) result.getColumns().get(Attribute.BEFORE_ID);
+  }
+
+  public IntColumn getBeforeVersionColumn() {
+    return (IntColumn) result.getColumns().get(Attribute.BEFORE_VERSION);
+  }
```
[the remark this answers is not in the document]
jnmt: Good catch! Fixed in d5ff7af.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionResult.java
jnmt: That's right. Thanks. Fixed in d5ff7af.

feeblefakie (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM! 👍

Torch3333 (approved):
LGTM, thank you!

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitNullMetadataIntegrationTestBase.java
```diff
@@ -0,0 +1,1502 @@
...
+  private ConsensusCommitAdmin consensusCommitAdmin;
+  private String namespace1;
+  private String namespace2;
+  private ParallelExecutor parallelExecutor;
+
+  private ConsensusCommitManager manager;
+  private DistributedStorage storage;
+  private Coordinator coordinator;
+  private RecoveryHandler recovery;
+  private CommitHandler commit;
```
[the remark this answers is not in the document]
jnmt: Good catch! Thank you! Fixed on 3dd3072.

PR #842, opened 2023-04-19
Add operation interfaces for upcoming relational abstraction support

jnmt:
This PR introduces operation interfaces to re-use the same logic for the new relational abstraction operation (i.e., `RelationalScan` with arbitrary conditions, and it's introduced by another PR) while keeping backward compatibility.

Also, operations in the relational abstraction don't have to have partition keys and clustering keys. So, I'm adding a more basic operation called `BaseOperation` than the current `Operation` class.

In short, what we would like to achieve here is that both `Scan` and `RelationalScan` implements `ScanInterface` but 
those two have different origins, i.e., `Operation` (with keys) and `BaseOperation` (without keys).

### DIFF
--- a/core/src/main/java/com/scalar/db/api/Operation.java
+++ b/core/src/main/java/com/scalar/db/api/Operation.java
@@ -32,65 +32,6 @@ public abstract class Operation extends BaseOperation {
   }
 
   /**
-   * Returns the namespace for this operation
-   *
-   * @return an {@code Optional} with the returned namespace
-   */
-  @Override
-  @Nonnull
-  public Optional<String> forNamespace() {
-    return super.forNamespace();
-  }
-
-  /**
-   * Returns the table name for this operation
-   *
-   * @return an {@code Optional} with the returned table name
-   */
-  @Override
-  @Nonnull
-  public Optional<String> forTable() {
-    return super.forTable();
-  }
-
-  /**
-   * Returns the full table name with the full namespace for this operation
-   *
-   * @return an {@code Optional} with the returned the full table name
-   */
-  @Override
-  @Nonnull
-  public Optional<String> forFullTableName() {
-    return super.forFullTableName();
-  }
-
-  /**
-   * Sets the specified target namespace for this operation
-   *
-   * @param namespace target namespace for this operation
-   * @return this object
-   * @deprecated As of release 3.6.0. Will be removed in release 5.0.0
-   */
-  @Deprecated
-  @Override
-  public Operation forNamespace(String namespace) {
-    return (Operation) super.forNamespace(namespace);
-  }
-
-  /**
-   * Sets the specified target table for this operation
-   *
-   * @param tableName target table name for this operation
-   * @return this object
-   * @deprecated As of release 3.6.0. Will be removed in release 5.0.0
-   */
-  @Deprecated
-  @Override
-  public Operation forTable(String tableName) {
-    return (Operation) super.forTable(tableName);
-  }
-
-  /**
    * Returns the partition key
    *
    * @return the partition {@code Key}
@@ -140,8 +81,7 @@ public abstract class Operation extends BaseOperation {
    * <ul>
    *   <li>both super class instances are equal and
    *   <li>it is also an {@code Operation} and
-   *   <li>both instances have the same partition key, clustering key, namespace, table name and
-   *       consistency
+   *   <li>both instances have the same partition key and clustering key
    * </ul>
    *
    * @param o an object to be tested for equality
--- a/core/src/main/java/com/scalar/db/api/SelectionInterface.java
+++ b/core/src/main/java/com/scalar/db/api/SelectionInterface.java
@@ -1,0 +1,14 @@
+package com.scalar.db.api;
+
+import java.util.List;
+
+/** An interface for selection operations. */
+public interface SelectionInterface extends BaseOperationInterface {
+
+  /**
+   * Returns the table name for this operation
+   *
+   * @return a {@code List} of column names to be projected
+   */
+  List<String> getProjections();
+}
--- a/core/src/main/java/com/scalar/db/api/SelectionInterface.java
+++ b/core/src/main/java/com/scalar/db/api/SelectionInterface.java
@@ -6,9 +6,9 @@ import java.util.List;
 public interface SelectionInterface extends BaseOperationInterface {
 
   /**
-   * Returns the table name for this operation
+   * Returns the list of column names to be projected for this operation
    *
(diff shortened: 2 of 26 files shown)


core/src/main/java/com/scalar/db/api/SelectionInterface.java
```diff
@@ -0,0 +1,14 @@
+package com.scalar.db.api;
+
+import java.util.List;
+
+/** An interface for selection operations. */
+public interface SelectionInterface extends BaseOperationInterface {
+
+  /**
+   * Returns the table name for this operation
```
komamitsu, line 9: "column names"?

core/src/main/java/com/scalar/db/api/Operation.java
```diff
@@ -9,72 +9,59 @@
...
   /**
    * Returns the namespace for this operation
    *
    * @return an {@code Optional} with the returned namespace
    */
+  @Override
   @Nonnull
   public Optional<String> forNamespace() {
-    return namespace;
+    return super.forNamespace();
```
komamitsu, line 42: I think you can remove these verbose overridden methods as the superclass's methods are called

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -93,7 +93,7 @@ public Optional<Result> get(Get get) throws ExecutionException {
 
   @Override
   public Scanner scan(Scan scan) throws ExecutionException {
-    scan = copyAndSetTargetToIfNot(scan);
+    scan = (Scan) copyAndSetTargetToIfNot(scan);
```
komamitsu, line 96: I guess that the caller of `copyAndSetTargetToIfNot(ScanInterface)` expects the same return type as the parameter type (e.g. expects `Scan` when passing `Scan` instance.) If so, how about keeping the existing `Scan copyAndSetTargetToIfNot(Scan)` and adding `RelationalScan copyAndSetTargetToIfNot(RelationalScan)` later so that we can remove this class cast?

jnmt: I'm not sure why I chose the cast rather than adding the relational one, but now I think your suggestion is far better. I will fix it (including other comments). Thank you!

core/src/main/java/com/scalar/db/api/SelectionInterface.java
jnmt: Good catch! Thank you! Fixed in d6da904.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
jnmt: Fixed in d6da904. Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -177,8 +178,8 @@ private TableMetadata getTableMetadata(Scan scan) throws ExecutionException {
     }
   }
 
-  public Optional<List<Key>> get(Scan scan) {
-    if (isWriteSetOverlappedWith(scan)) {
+  public Optional<List<Key>> get(ScanInterface scan) {
+    if (scan instanceof Scan && isWriteSetOverlappedWith((Scan) scan)) {
```
komamitsu, line 182: This check (or an equivalent) won't be needed for RelationalScan?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -318,14 +319,14 @@ void toSerializableWithExtraRead(DistributedStorage storage)
...
-    for (Map.Entry<Scan, List<Key>> entry : scanSet.entrySet()) {
+    for (Map.Entry<ScanInterface, List<Key>> entry : scanSet.entrySet()) {
       tasks.add(
           () -> {
             Map<Key, TransactionResult> currentReadMap = new HashMap<>();
             Set<Key> validatedReadSet = new HashSet<>();
             Scanner scanner = null;
             try {
-              Scan scan = entry.getKey();
+              Scan scan = (Scan) entry.getKey();
```
komamitsu, line 329: What will this code be like when RelationalScan is introduced?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: Good question. A check for preventing scan-after-write is necessary for RelationalScan too. But it's done at a bit different timing, and the PR will be proposed later.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: It's like the following, and `rebuildScan` will create another Scan operation using a builder pattern.

```java
              ScanInterface scan = entry.getKey();
              if (scan instanceof RelationalScan) {
                scanner =
                    storage.scan(
                        rebuildRelationalScan((RelationalScan) scan, getTableMetadata(scan)));
              } else {
                scanner = storage.scan(rebuildScan((Scan) scan, getTableMetadata(scan)));
              }
```

komamitsu: Thanks! It would be great if we have 2 overloaded `rebuildScan(Scan)` and `rebuildScan(RelationalScan)` as we can simply write it like `scanner = storage.scan(rebuildScan(scan, getTableMetadata(scan)))`

jnmt: Even if we have 2 overloaded `rebuildScan`, I think we still need a cast and dispatch here since `scan` here is `ScanInterface`. Am I understanding correctly?

komamitsu (approved):
LGTM! 👍

jnmt:
As a result of a careful discussion with Toshi, we decided to explore another option, extending `Scan` instead of introducing the interface changes. This is mainly because we have several similar interface and abstraction issues (e.g., ScanAll), and they cannot be solved without breaking backward compatibilities if we try to keep smart and elegant interface designs (e.g., not like using XyzInterface). So, anyway, let me make this PR back to the draft (maybe I will create another PR).

@komamitsu Sorry for taking your time for the review.

jnmt:
I will close this draft PR since I submit another PR (#889) with a new solution.

PR #843, opened 2023-04-20
Enable dependabot

Torch3333:
This enables Dependabot to : 
- For Docker images, update the internal JRE container
- For the Gradle build, update dependencies and plugin to the latest non-major version. This is to avoid introducing incompatible changes.
- For Github Actions workflows, update all the actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,48 @@
+version: 2
+registries:
+  ghcr:
+    type: docker-registry
+    url: https://ghcr.io
+    username: ${{secrets.CR_USER}}
+    password: ${{secrets.CR_PAT}}
+    replaces-base: true
+updates:
+  # For the Schema Loader Docker image, only update the scalar-labs/jre8 container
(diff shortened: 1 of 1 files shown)


komamitsu (approved):
LGTM! Great. I didn't know such various resource types are supported in Dependabot.

.github/dependabot.yml
```diff
@@ -0,0 +1,48 @@
...
+  # For Gradle, update dependencies and plugins to the latest non-major version
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    reviewers:
+      - "scalar-labs/scalardb"
+    ignore:
+      - dependency-name: "*"
+        update-types: [ "version-update:semver-major" ]
```
brfrn169, line 41: It looks like this `ignore` config can be configured for individual dependencies, which should be useful.

brfrn169 (approved):
LGTM! Thank you!

PR #851, opened 2023-04-20
Add decommissioning state to ScalarDB Server

brfrn169:
This PR adds a decommissioning state to ScalarDB Server to achieve graceful shutdown. Please take a look!

### DIFF
--- a/server/src/main/java/com/scalar/db/server/ScalarDbServer.java
+++ b/server/src/main/java/com/scalar/db/server/ScalarDbServer.java
@@ -1,6 +1,7 @@
 package com.scalar.db.server;
 
 import com.google.common.base.Strings;
+import com.google.common.util.concurrent.Uninterruptibles;
 import com.scalar.db.api.DistributedStorage;
 import com.scalar.db.api.DistributedStorageAdmin;
 import com.scalar.db.api.DistributedTransactionAdmin;
@@ -27,7 +28,8 @@ import picocli.CommandLine.Command;
 @Command(name = "scalardb-server", description = "Starts ScalarDB Server.")
 public class ScalarDbServer implements Callable<Integer> {
   private static final Logger logger = LoggerFactory.getLogger(ScalarDbServer.class);
-  private static final long MAX_WAIT_TIME_MILLIS = 60000; // 60 seconds
+
+  private static final long MAX_SHUTDOWN_WAIT_TIME_MILLIS = 60000;
 
   @CommandLine.Option(
       names = {"--properties", "--config"},
@@ -38,6 +40,7 @@ public class ScalarDbServer implements Callable<Integer> {
 
   private ServerConfig config;
   private Server server;
+  private final HealthService healthService = new HealthService();
 
   private DistributedStorage storage;
   private DistributedStorageAdmin storageAdmin;
@@ -89,7 +92,7 @@ public class ScalarDbServer implements Callable<Integer> {
                     transactionManager, tableMetadataManager, gateKeeper, metrics))
             .addService(new DistributedTransactionAdminService(transactionAdmin, metrics))
             .addService(new AdminService(gateKeeper))
-            .addService(new HealthService())
+            .addService(healthService)
             .addService(ProtoReflectionService.newInstance());
 
     // Two-phase commit for JDBC is not supported for now
@@ -120,10 +123,18 @@ public class ScalarDbServer implements Callable<Integer> {
         .addShutdownHook(
             new Thread(
                 () -> {
-                  logger.info("Signal received. Shutting down the server ...");
-                  shutdown(MAX_WAIT_TIME_MILLIS, TimeUnit.MILLISECONDS);
-                  logger.info("The server shut down.");
+                  logger.info("Signal received. Decommissioning ...");
+                  decommission();
+                  logger.info("Decommissioned. Shutting down.");
+                  shutdown(MAX_SHUTDOWN_WAIT_TIME_MILLIS, TimeUnit.MILLISECONDS);
+                  logger.info("ScalarDB Server shut down.");
                 }));
+  }
+
+  public void decommission() {
+    healthService.decommission();
+    Uninterruptibles.sleepUninterruptibly(
+        config.getDecommissioningDurationSecs(), TimeUnit.SECONDS);
   }
 
   public void shutdown() {
--- a/docs/scalardb-server.md
+++ b/docs/scalardb-server.md
@@ -43,7 +43,7 @@ It contains two sections: ScalarDB Server configurations and underlying storage/
 # ScalarDB Server configurations
 #
 
-# Port number of ScalarDB Server. 60051 by default
+# Port number of ScalarDB Server. 60051 by default.
 scalar.db.server.port=60051
 
 # Prometheus exporter port. Use 8080 if this is not given. Prometheus exporter will not be started if a negative number is given.
@@ -54,6 +54,9 @@ scalar.db.server.grpc.max_inbound_message_size=
 
 # The maximum size of metadata allowed to be received. If not specified, use the gRPC default value.
 scalar.db.server.grpc.max_inbound_metadata_size=
+
+# The decommissioning duration in seconds. 30 seconds by default.                 
+scalar.db.server.decommissioning_duration_secs=30
 
 #
 # Underlying storage/database configurations
--- a/server/src/main/java/com/scalar/db/server/HealthService.java
+++ b/server/src/main/java/com/scalar/db/server/HealthService.java
@@ -4,17 +4,26 @@ import io.grpc.health.v1.HealthCheckRequest;
 import io.grpc.health.v1.HealthCheckResponse;
 import io.grpc.health.v1.HealthGrpc;
 import io.grpc.stub.StreamObserver;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.concurrent.ThreadSafe;
 
 @ThreadSafe
 public class HealthService extends HealthGrpc.HealthImplBase {
 
+  private final AtomicReference<HealthCheckResponse.ServingStatus> servingStatus =
+      new AtomicReference<>(HealthCheckResponse.ServingStatus.SERVING);
+
   @Override
   public void check(
       HealthCheckRequest request, StreamObserver<HealthCheckResponse> responseObserver) {
     HealthCheckResponse.Builder builder = HealthCheckResponse.newBuilder();
-    builder.setStatus(HealthCheckResponse.ServingStatus.SERVING);
+    builder.setStatus(servingStatus.get());
     responseObserver.onNext(builder.build());
     responseObserver.onCompleted();
+  }
+
+  /** Decommissions the cluster node from the cluster */
+  public void decommission() {
+    servingStatus.set(HealthCheckResponse.ServingStatus.NOT_SERVING);
   }
 }
--- a/server/src/main/java/com/scalar/db/server/ServerConfig.java
+++ b/server/src/main/java/com/scalar/db/server/ServerConfig.java
@@ -26,8 +26,14 @@ public class ServerConfig {
   public static final String GRPC_MAX_INBOUND_METADATA_SIZE =
       PREFIX + "grpc.max_inbound_metadata_size";
 
+  /** The decommissioning duration in seconds. */
+  public static final String DECOMMISSIONING_DURATION_SECS =
+      PREFIX + "decommissioning_duration_secs";
+
   public static final int DEFAULT_PORT = 60051;
   public static final int DEFAULT_PROMETHEUS_EXPORTER_PORT = 8080;
+
+  public static final int DEFAULT_DECOMMISSIONING_DURATION_SECS = 30;
 
   private final Properties props;
   private int port;
@@ -35,6 +41,8 @@ public class ServerConfig {
 
   @Nullable private Integer grpcMaxInboundMessageSize;
   @Nullable private Integer grpcMaxInboundMetadataSize;
+
+  private int decommissioningDurationSecs;
 
   public ServerConfig(File propertiesFile) throws IOException {
     try (FileInputStream stream = new FileInputStream(propertiesFile)) {
@@ -73,6 +81,10 @@ public class ServerConfig {
 
     grpcMaxInboundMessageSize = getInt(getProperties(), GRPC_MAX_INBOUND_MESSAGE_SIZE, null);
     grpcMaxInboundMetadataSize = getInt(getProperties(), GRPC_MAX_INBOUND_METADATA_SIZE, null);
+
+    decommissioningDurationSecs =
+        getInt(
+            getProperties(), DECOMMISSIONING_DURATION_SECS, DEFAULT_DECOMMISSIONING_DURATION_SECS);
   }
 
   public int getPort() {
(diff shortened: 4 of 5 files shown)


server/src/main/java/com/scalar/db/server/ScalarDbServer.java
```diff
@@ -120,12 +123,20 @@ public void addShutdownHook() {
         .addShutdownHook(
             new Thread(
                 () -> {
-                  logger.info("Signal received. Shutting down the server ...");
-                  shutdown(MAX_WAIT_TIME_MILLIS, TimeUnit.MILLISECONDS);
-                  logger.info("The server shut down.");
+                  logger.info("Signal received. Decommissioning ...");
+                  decommission();
```
brfrn169, line 127: Before shutting down ScalarDB Server, change the healthService response to `NOT_SERVING` and sleep for some time. This way, we can wait for no new requests to come in.

komamitsu:
@brfrn169 In the current implementation, `io.grpc.Server#shutdown()` which rejects new requests and `io.grpc.Server#awaitTermination()` are called. So, it looks to work to me, but probably I'm missing something. Can you share what issue we're going to solve?

kota2and3kan:
@komamitsu (@brfrn169 ) 
The summary of issues that we are going to resolve is the following.

In the current deployment on the Kubernetes environment, we use Envoy with ScalarDB Server, there is a possibility that the clients receive a `5xx` error when we do the rolling update of ScalarDB Server.

And, we already fixed a similar issue in the ScalarDB Cluster. So, I will explain the behavior of ScalarDB Cluster first. After that, I will explain the behavior (what we are going to resolve) in the ScalarDB Server.

## ScalarDB Cluster (already fixed some issues)

When we shutdown (send `SIGTERM` to) ScalarDB Cluster, it works as follows.

1. Returns `NOT_SERVING` and waiting 30sec by default.
2. Envoy detects `NOT_SERVING` (i.e., not healthy) in its health check feature.
3. Envoy removes the ScalarDB Cluster node from the load balancing pool (i.e., Envoy doesn't send new requests to ScalarDB Cluster node that we will shutdown).
4. (After waiting 30sec) ScalarDB Cluster starts shutdown process.
5. ScalarDB Cluster waits for all ongoing Tx will complete.
6. After all ongoing Tx is completed, ScalarDB Cluster stops.

Note: This fix was done by this PR.
https://github.com/scalar-labs/scalardb-cluster/pull/44

By the above process, ScalarDB Cluster can guarantee that `Don't receive a new request after the shutdown process starts`. So, we can avoid receiving `5xx` errors from the perspective of the client when we do the rolling update.

## ScalarDB Server (what we are going to fix)

When we shutdown (send `SIGTERM` to) ScalarDB Server, it works as follows.

1. ScalarDB Server starts shutdown process immediately. After this process, a new request receives a `5xx` error.
2. ScalarDB Server waits for all ongoing Tx will complete.
3. Endpoint resource (i.e., service discovery information for Envoy) updated by Kubernetes.
4. Envoy removes the ScalarDB Server from the load balancing pool based on the update of the Endpoint resource.
5. After all ongoing Tx is completed, ScalarDB Cluster stops.

Note: `2` and `3` runs in parallel.

Since there is a time lag between `1.` and `4.`, there is a possibility that the Envoy sends a new request to the ScalarDB Server that is in stopping progress. In this case, the ScalarDB Server returns a `5xx` error to the client.

To fix this issue, we update ScalarDB Server to implement the same `graceful shutdown` process as ScalarDB Cluster. After this update, the Envoy can detect and treat ScalarDB Server's shutdown process properly.

brfrn169:
@kota2and3kan Thank you for your explanation!

komamitsu:
@kota2and3kan Thanks for the detailed background! It's very helpful for me.

Just out of curiosity, Kubernetes (and/or Envoy?) doesn't have a rolling update orchestration feature like the following?
1. Envoy marks one of the pods  as `disabled` (or just removes it from available pod list?)
2. the target pods is restarted
3. Envoy marks the restarted pod as `enabled` (or just adds it from available pod list?)

AWS's CodeDeploy has this kind of features and I just though it would be great if k8s supports that.

komamitsu (approved):
LGTM! 👍

kota2and3kan:
@komamitsu  
> Just out of curiosity, Kubernetes (and/or Envoy?) doesn't have a rolling update orchestration feature like the following?

In conclusion, Kubernetes has a rolling update feature and Envoy has health check/service discovery features.

However, since Envoy is one of the application pods (i.e., it is not a system pod) from the perspective of Kubernetes, I think there is no feature that Kubernetes and Envoy work on cooperating on the Kubernetes side.

> 1. Envoy marks one of the pods  as `disabled` (or just removes it from available pod list?)
> 2. the target pods is restarted
> 3. Envoy marks the restarted pod as `enabled` (or just adds it from the available pod list?)

In your question, `1.` and `3.` are the Envoy's behavior (from the perspective of Kubernetes, it is an application's behavior). And, `2` is a Kubernetes' behavior. 

Since Envoy is one of the application pods, Kubernetes cannot control Envoy's behavior. 

Also, Envoy is a generic proxy (it is not a dedicated proxy for Kubernetes). So, I think Envoy cannot control Kubernetes' pod deletion behavior.

Therefore, we have to achieve a graceful shutdown by combining several functions of Kubernetes and Envoy.

---

For your reference, I will explain Kubernetes behavior and Envoy behavior respectively. Also, I explain the challenges of ScalarDB Server.

## Kubernetes' feature

As a Kubernetes feature, there is a [`Service`](https://kubernetes.io/docs/concepts/services-networking/service/) resource. You can use this resource for service discovery (you can use it like a Load Balancer). And, this `Service` resource uses the pod list in the [`Endpoint`](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoints-v1/) resource to decide which pod it sends a request to.

For example, if there are three pods, `Endpoint` has information (IP address of pods) as follows.

* Pods (3 pods)
  ```console
  $ kubectl get pod
  NAME                                     READY   STATUS    RESTARTS   AGE
  scalardb-cluster-node-76699489b4-fnz97   1/1     Running   0          55s
  scalardb-cluster-node-76699489b4-j4tqh   1/1     Running   0          55s
  scalardb-cluster-node-76699489b4-mtzxx   1/1     Running   0          55s
  ```
* Service (you can use this resource for service discovery)
  ```console
  $ kubectl get svc scalardb-cluster-metrics
  NAME                       TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
  scalardb-cluster-metrics   ClusterIP   10.111.18.11   <none>        9080/TCP   2m2s
  ```
* Endpoint (manage list of pods)
  ```console
  $ kubectl get endpoints scalardb-cluster-metrics
  NAME                       ENDPOINTS                                               AGE
  scalardb-cluster-metrics   10.244.1.151:9080,10.244.1.152:9080,10.244.1.153:9080   2m24s
  ```

As you can see, the endpoint resource has a pod list that includes 3 IP addresses. So, if you access the Service `scalardb-cluster-metrics`, you can access one of the three pods. The `Service` resource sends your request to one of the pods like a load balancer.

And, if we delete the pod, Kubernetes updates this IP address list in the `Endpoint` resource. For example, if I delete one pod, we can see the `Endpoint` resource (IP address list) is updated by Kubernetes as follows. (In this test case, a new pod is automatically created by the `Deployment` resource when I delete one pod manually.)

* Initial state of pods (there are three pods)
  ```console
  $ kubectl get pod
  NAME                                     READY   STATUS    RESTARTS   AGE
  scalardb-cluster-node-8485886d5c-ggx99   1/1     Running   0          2m52s
  scalardb-cluster-node-8485886d5c-r29hm   1/1     Running   0          2m31s
  scalardb-cluster-node-8485886d5c-svwm2   1/1     Running   0          2m10s
  ```
* Initial state of Endpoint (there are three IP addresses of pods)
  ```console
  $ kubectl get endpoints scalardb-cluster-metrics
  NAME                       ENDPOINTS                                               AGE
  scalardb-cluster-metrics   10.244.1.160:9080,10.244.1.161:9080,10.244.1.162:9080   17m
  ```
* Delete one pod manually
  ```console
  $ kubectl delete pod scalardb-cluster-node-8485886d5c-ggx99
  pod "scalardb-cluster-node-8485886d5c-ggx99" deleted
  ```
* Pod status when the pod deletion is in progress (One pod is `Terminating` status and a new pod is `ContainerCreating` status)
  ```console
  $ kubectl get pod
  NAME                                     READY   STATUS              RESTARTS   AGE
  scalardb-cluster-node-8485886d5c-ggx99   1/1     Terminating         0          3m10s
  scalardb-cluster-node-8485886d5c-j8qjq   0/1     ContainerCreating   0          0s
  scalardb-cluster-node-8485886d5c-r29hm   1/1     Running             0          2m49s
  scalardb-cluster-node-8485886d5c-svwm2   1/1     Running             0          2m28s
  ```
* Endpoint status when the pod deletion is in progress
  ```console
  $ kubectl get endpoints scalardb-cluster-metrics
  NAME                       ENDPOINTS                             AGE
  scalardb-cluster-metrics   10.244.1.161:9080,10.244.1.162:9080   18m
  ```
  * As you can see, the `Endpoint` resource (a list of pod IPs) has only two IP addresses. The IP of the deleted pod is automatically removed from this list by Kubernetes.
* Pod status after the new pod created
  ```console
  $ kubectl get pod
  NAME                                     READY   STATUS    RESTARTS   AGE
  scalardb-cluster-node-8485886d5c-j8qjq   1/1     Running   0          41s
  scalardb-cluster-node-8485886d5c-r29hm   1/1     Running   0          3m30s
  scalardb-cluster-node-8485886d5c-svwm2   1/1     Running   0          3m9s
  ```
* Endpoint status after the new pod created
  ```console
  $ kubectl get endpoints scalardb-cluster-metrics
  NAME                       ENDPOINTS                                               AGE
  scalardb-cluster-metrics   10.244.1.161:9080,10.244.1.162:9080,10.244.1.163:9080   18m
  ```
  * As you can see, the `Endpoint` resource (a list of pod IPs) has three IP addresses. The IP of the created new pod is automatically added to this list by Kubernetes.

In the Kubernetes environment, the restarting pod means `delete the pod and create a new pod`. When we restart the pod, Kubernetes updates the `Endpoint` (IP address list) automatically like the above.

Note: You can see the description of this behavior in the official document [Termination of Pods](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination). Also, you can see the details on pod termination behavior in [this article](https://qiita.com/superbrothers/items/3ac78daba3560ea406b2).

This is a behavior of Kubernetes' `Service` resource and `Endpoint` resource which are native features of Kubernetes. So, basically, Kubernetes can perform the rolling update by updating the `Endpoint` (a list of the pod for load balancing) based on the pod status.

However, Kubernetes is an asynchronous system. So, pod deletion and updating `Endpoint` are done  asynchronously. In other words, Kubernetes doesn't guarantee the order of each operation. So, there is a possibility that the `Service` resource sends new requests to the pod that is in the deleting status. 

To reduce this issue, users/applications can sleep before starting the shutdown process (i.e., wait for updating `Endpoint` by sleep). You can do this using Kubernetes' feature [preStop hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks) or implement sleep in your application.

Note: You can see the details on this workaround in [this article](https://zenn.dev/hhiroshell/articles/kubernetes-graceful-shutdown) and [this article](https://zenn.dev/hhiroshell/articles/kubernetes-graceful-shutdown-experiment).

## ScalarDB Server specific things

As I mentioned above, Kubernetes updates a pod list automatically and can perform the rolling update. So, we can do the rolling update by using the `Service` resource as a load balancer in the Kubernetes environment.

However, we cannot use the `Service` resource directly as a load balancer for ScalarDB Server.

Since the `Service` resource is an L4 load balancer, it cannot treat L7 connections (e.g., state of gRPC) properly. And, ScalarDB Server uses the `Bidirectional streaming RPC` of gRPC.

So, we have to use some load balancer that can treat L7 (`Bidirectional streaming RPC` of gPRC) properly for connections between ScalarDB Server and clients. This is why we use Envoy as a load balancer for ScalarDB Server. Envoy can treat L7 (gRPC) connections.

## Envoy's feature

Envoy has a [Health checking](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/health_checking) feature for upstream servers. If the health check failed, Envoy doesn't send a new request to the failed upstream server.

Also, regarding maintaining the upstream server list, Envoy can detect pod deletion based on the service discovery. If the service discovery information is changed, Envoy updates its upstream server list based on the service discovery. For example, in the Kubernetes environment, the service discovery means a `Service` resource. So, Envoy can detect pod deletion in the Kubernetes environment as follows.

1. Envoy maintains its upstream server list based on the `Service` (strictly `Endpoint`) resource.
2. One pod is deleted.
3. Kubernetes removes the pod from the `Endpoint` resource. In other words, the `Service` resource (service discovery for Envoy) is updated.
4. Envoy detects the update of the `Service` resource.
5. Envoy updates its upstream server list based on the updated `Service` resource of Kubernetes.

You can see the details of the behavior (service discovery + health check behavior) in the official document [On eventually consistent service discovery](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#on-eventually-consistent-service-discovery).

## Issues of ScalarDB Server

As I mentioned, Envoy can detect pod deletion and update the upstream server list. However, in both `health check` and `service discovery` cases, there is a time lag between `pod deletion` and `Envoy detects pod deletion`.

It is difficult to remove this time lag completely, because the health check runs every few seconds, and updating the Endpoint resource process is done asynchronously (basically, Kubernetes is an asynchronous system.)

So, there is a possibility that the Envoy sends a new request to the deletion status pod. This is an issue that we want to resolve.

## How to resolve the issue

To resolve ScalarDB Server's issue that I mentioned above, we have to achieve a graceful shutdown by combining the following functions.

* Endpoint updates (Kubernetes' feature)
* Health check (Envoy's feature)
* Service discovery (Envoy's feature)
* gRPC Health Check (Implemented in ScalarDB Server)
* Sleep (Implemented in ScalarDB Server)

By using these functions, we can achieve the following behavior (i.e., graceful shutdown) for the ScalarDB Server.

1. Delete/Stop ScalarDB Server pod.
2. Kubernetes remove the deleted pod IP from the `Endpoint` resource.
3. Kubernetes send `SIGTERM` to the ScalarDB Server process in the pod.
4. After receiving `SIGTERM`, ScalarDB Server returns `NOT_SERVING` as a result of gPRC Health Check and waits for 30sec.
5. Envoy detects `NOT_SERVING` of the pod. After that, Envoy doesn't send new requests to the pod.
6. Envoy detects service discovery's (i.e., `Endpoint` resource's) change and removes the pod from its upstream server list.
7. After waiting for 30sec, ScalarDB Server starts its actual shutdown process.
8. ScalarDB Server completes all ongoing Tx. After that, ScalarDB Server stops.

Note: From `2.` to `6.` operations run in parallel.

kota2and3kan (approved):
LGTM! Thank you!

komamitsu:
@kota2and3kan Thanks for the great explanation!

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #854, opened 2023-04-25
Apply workaround on CI for Dependabot

Torch3333:
**Issue** 
Dependabot is failing to parse the current CI file even though it is valid by Github actions standards.
image

**Fix**
This is a known issue (cf. https://github.com/dependabot/dependabot-core/issues/5453) for value starting with the semicolon `:` character that the Dependabot parser does not seem to recognize as a string. A workaround is to quote the problematic part to clear up the ambiguity.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,13 +79,13 @@ jobs:
       - name: Dockerfile Lint for ScalarDB Server
         uses: gradle/gradle-build-action@v2
         with:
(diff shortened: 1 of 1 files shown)


feeblefakie (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM! Thank you!

brfrn169 (approved):
LGTM! Thank you!

PR #865, opened 2023-04-27
[Dependabot] Ignore patch versions for Aws sdk

Torch3333:
**Context** 
Since Aws sdk patch versions are released every few days, Dependabot will end up creating PR to update its version every week, which is not justified.

**Changes**
Dependabot won't update Aws sdk dependencies for patch versions anymore. Eventually, Aws sdk will only be updated when minor versions are released.


brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #873, opened 2023-05-15
Add validation for consensus commit mutation operation

Torch3333:
**Context**
There is no validation to ensure Put or Delete operations do not mutate consensus commit metadata columns. Since these columns are only for internal use, users should not be allowed to modify them.

**Changes**
A validation step is added to make sure Put or Delete operations do not mutate consensus commit metadata columns. Additionally, setting operation conditions (see https://github.com/scalar-labs/scalardb/pull/837 ) targeting consensus commit metadata columns are also forbidden.

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
@@ -12,6 +12,7 @@ import com.scalar.db.api.Result;
 import com.scalar.db.api.Scan;
 import com.scalar.db.api.Selection;
 import com.scalar.db.common.AbstractDistributedTransaction;
+import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.exception.transaction.CommitException;
 import com.scalar.db.exception.transaction.CrudException;
 import com.scalar.db.exception.transaction.UnknownTransactionStatusException;
@@ -41,13 +42,19 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
   private final CrudHandler crud;
   private final CommitHandler commit;
   private final RecoveryHandler recovery;
+  private final ConsensusCommitMutationOperationChecker mutationOperationChecker;
   private Runnable beforeRecoveryHook;
 
   @SuppressFBWarnings("EI_EXPOSE_REP2")
-  public ConsensusCommit(CrudHandler crud, CommitHandler commit, RecoveryHandler recovery) {
+  public ConsensusCommit(
+      CrudHandler crud,
+      CommitHandler commit,
+      RecoveryHandler recovery,
+      ConsensusCommitMutationOperationChecker mutationOperationChecker) {
     this.crud = checkNotNull(crud);
     this.commit = checkNotNull(commit);
     this.recovery = checkNotNull(recovery);
+    this.mutationOperationChecker = mutationOperationChecker;
     this.beforeRecoveryHook = () -> {};
   }
 
@@ -79,40 +86,45 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
   }
 
   @Override
-  public void put(Put put) {
+  public void put(Put put) throws CrudException {
     put = copyAndSetTargetToIfNot(put);
+    checkMutation(put);
     crud.put(put);
   }
 
   @Override
-  public void put(List<Put> puts) {
+  public void put(List<Put> puts) throws CrudException {
     checkArgument(puts.size() != 0);
-    puts.forEach(this::put);
+    for (Put p : puts) {
+      put(p);
+    }
   }
 
   @Override
-  public void delete(Delete delete) {
+  public void delete(Delete delete) throws CrudException {
     delete = copyAndSetTargetToIfNot(delete);
+    checkMutation(delete);
     crud.delete(delete);
   }
 
   @Override
-  public void delete(List<Delete> deletes) {
+  public void delete(List<Delete> deletes) throws CrudException {
     checkArgument(deletes.size() != 0);
-    deletes.forEach(this::delete);
+    for (Delete d : deletes) {
+      delete(d);
+    }
   }
 
   @Override
-  public void mutate(List<? extends Mutation> mutations) {
+  public void mutate(List<? extends Mutation> mutations) throws CrudException {
     checkArgument(mutations.size() != 0);
-    mutations.forEach(
-        m -> {
-          if (m instanceof Put) {
-            put((Put) m);
-          } else if (m instanceof Delete) {
-            delete((Delete) m);
-          }
-        });
+    for (Mutation m : mutations) {
+      if (m instanceof Put) {
+        put((Put) m);
+      } else if (m instanceof Delete) {
+        delete((Delete) m);
+      }
+    }
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
```diff
@@ -0,0 +1,96 @@
...
+    TransactionTableMetadata metadata =
+        transactionTableMetadataManager.getTransactionTableMetadata(operation);
+    if (metadata == null) {
+      throw new IllegalArgumentException(
+          "The specified table is not found: " + operation.forFullTableName().get());
+    }
+    return metadata;
+  }
+
+  public void check(Mutation mutation) throws ExecutionException {
```
brfrn169, line 39: I think we can throw `IllegalArgumentException` instead of `ExecutionException` for the operation check.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -79,40 +86,45 @@ public List<Result> scan(Scan scan) throws CrudException {
...
+    checkMutation(put);
     crud.put(put);
   }
 
   @Override
-  public void put(List<Put> puts) {
+  public void put(List<Put> puts) throws CrudException {
     checkArgument(puts.size() != 0);
-    puts.forEach(this::put);
+    for (Put p : puts) {
```
komamitsu, line 98: [just out of curiosity] Is this change for suppressing warning or something?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private void check(Put put) throws ExecutionException {
+    put.getCondition()
+        .ifPresent(
+            condition -> {
+              if (!(condition instanceof PutIf
+                  || condition instanceof PutIfNotExists
+                  || condition instanceof PutIfExists)) {
+                throw new IllegalArgumentException("The condition is not allowed on Put operation");
```
komamitsu, line 54: I guess a user would be interested in the type of the wrong `condition` when receiving this exception. How about showing it in the error message?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
Torch3333: Just to be sure we are on the same page. It is not documented but `check` only throws an `ExceptionException` if the table metadata fails to be retrieved. https://github.com/scalar-labs/scalardb/blob/0212c8b2d8bf773600383687e59f386ce5260faf/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java#L30-L31
For validation errors, `IllegalArgumentException` is thrown. 
Do you think we should catch the `ExecutionException` and rethrow it as an `IllegalArgumentException`?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
Torch3333: `this::put` throws the checked exception `CrudException` but the Consumer functional interface does not throw any checked exception.
```
public void forEach(java.util.function.Consumer<? super T> action)
```
Our only option is to use a non-lambda for each if we want to propagate the `CrudException`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
Torch3333: Yes, that is a good point, thank you. I will add it.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
brfrn169: Ah sorry, I misunderstood. I thought `ExecutionException` was also thrown for validation errors. Please ignore the comment.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
Torch3333: Revised in https://github.com/scalar-labs/scalardb/pull/873/commits/4e05db8de6de4d94bc81de22ad3c3b40bfd7a321

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
Torch3333: I added Javadoc since it's difficult to understand why the `ExecutionException` is thrown.
https://github.com/scalar-labs/scalardb/pull/873/commits/4e05db8de6de4d94bc81de22ad3c3b40bfd7a321

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -180,7 +180,9 @@ DistributedTransaction begin(String txId, Isolation isolation, SerializableStrat
         new Snapshot(txId, isolation, strategy, tableMetadataManager, parallelExecutor);
     CrudHandler crud =
         new CrudHandler(storage, snapshot, tableMetadataManager, isIncludeMetadataEnabled);
-    ConsensusCommit consensus = new ConsensusCommit(crud, commit, recovery);
+    ConsensusCommitMutationOperationChecker operationChecker =
+        new ConsensusCommitMutationOperationChecker(tableMetadataManager);
```
brfrn169, line 184: Instead of creating a `ConsensusCommitMutationOperationChecker` instance each time, can we hold it as an instance field of `ConsensusCommitManager` and pass it to `ConsensusCommit` instances?

brfrn169 (commented):
Left one minor comment. Other than that, LGTM. Please take a look when you have time!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
```diff
@@ -0,0 +1,108 @@
...
+    } else if (mutation instanceof Delete) {
+      check((Delete) mutation);
+    }
+  }
+
+  private void check(Put put) throws ExecutionException {
+    put.getCondition()
+        .ifPresent(
+            condition -> {
+              if (!(condition instanceof PutIf
```
feeblefakie, line 58: Just quick confirmation.
This check will be deleted in another PR, which makes conditions usable, right?

feeblefakie (approved):
LGTM (except for what Toshi pointed out)! Thank you!
I also left a question for just confirmation.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
Torch3333: That is a good point, thank you. 
My bad, I did it for the 2pc transaction manager but not for this one. 
Revised in https://github.com/scalar-labs/scalardb/pull/873/commits/de06ddcebf8e54237bba0b7cfbc61935d00c8035

brfrn169 (approved):
LGTM! Thank you!

PR #889, opened 2023-05-24
Add an interface to add conditions in ScanAll

jnmt:
This PR adds an interface to add arbitrary conditions in ScanAll, i.e., to support a relational scan for JDBC databases. We decided to add them to the existing ScanAll instead of adding a new RelationalScan to keep the backward compatibility and simplicity of the current class inheritance. Actual relational scan processing will be introduced by another PR.

By this PR, users can add a kind of "where()" clause after calling "all()" in a Scan builder. We currently support only a disjunctive normal form (DNF) or a conjunctive normal form (CNF), like the following patterns. 

- .where( conditionalExpression )
- .where( conditionalExpression | andConditionSet ) [ .or( conditionalExpression | andConditionSet ) ] ...
- .where( conditionalExpression | orConditionSet ) [ .and( conditionalExpression | orConditionSet ) ] ...

Users can use `ConditionSetBuilder` to build `AndConditionSet` and `OrCondtionSet`, which is a set of conjunctive (a && b && ...) or disjunctive (a || b || ...) conditions. Note that users need to call `clearConditions()` before updating a where clause in an existing ScallAll object.

See also inline comments for details.


core/src/main/java/com/scalar/db/api/Scan.java
```diff
@@ -34,6 +37,7 @@ public class Scan extends Selection {
   private Optional<Key> endClusteringKey;
   private boolean endInclusive;
   private int limit;
+  private final Set<Conjunction> conjunctions;
```
jnmt, line 40: Internally, we use a disjunctive set of conjunctions (i.e., as disjunctive normal form) to manage whole conditions.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -548,4 +847,277 @@ public Scan build() {
...
+    orderings.forEach(scan::withOrdering);
+
+    if (condition != null) {
+      assert conjunctions.isEmpty() && disjunctions.isEmpty();
+      scan.withConjunctions(ImmutableSet.of(Scan.Conjunction.of(condition)));
+    } else if (conjunctions.isEmpty()) {
+      scan.withConjunctions(
+          Sets.cartesianProduct(disjunctions).stream()
+              .map(Scan.Conjunction::of)
+              .collect(Collectors.toSet()));
```
jnmt, line 1108: When building a where clause with ".and().and()...", the builder holds a (conjunctive) set of disjunctions, and its conjunction should be empty. In this case, we convert CNF to DNF according to the distributive law.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -548,4 +847,277 @@ public Scan build() {
...
+    } else {
+      scan.withConjunctions(conjunctions);
```
jnmt, line 1110: In contrast, when building a where clause with ".or().or()...", the builder naturally holds conditions as a DNF. So, it just adds them.

core/src/main/java/com/scalar/db/api/ScanAll.java
```diff
@@ -88,15 +88,18 @@ public ScanAll withEnd(Key clusteringKey, boolean inclusive) {
...
+   * @param ordering a scan ordering
+   * @return this object
+   * @deprecated As of release 3.6.0. Will be removed in release 5.0.0. Use the setter method of the
+   *     Scan builder instead; to create a Scan builder, use {@link Scan#newBuilder()}
    */
   @Deprecated
   @Override
   public ScanAll withOrdering(Ordering ordering) {
-    throw new UnsupportedOperationException();
+    return (ScanAll) super.withOrdering(ordering);
```
jnmt, line 102: Unlike a current ScanAll, a relational scan accept orderings. So, now we need to implement this, although we would like to use this internally for the builder and users do not have to use this.

komamitsu: It's not guaranteed that `super.withOrdering()` returns ScanAll in terms of the type signature and it seems an unsafe downcast to me. I'm wondering if any type check is needed (or applying Generics type parameter, but it takes effort...) ?

core/src/main/java/com/scalar/db/api/Scan.java
```diff
@@ -457,4 +484,316 @@ public enum Order {
...
+    }
+
+    @Override
+    public String toString() {
+      return conditions.toString();
+    }
+  }
+
+  @Immutable
+  public static class AndConditionSet {
```
komamitsu, line 571: Sorry, `Conjunction` and `AndConditionSet` sound similar to me. Could you add a comment to clarify what they are ?

core/src/main/java/com/scalar/db/api/ScanAll.java
jnmt: I thought it was safe because the caller object of `super.withOrdering()` here is always `ScanAll` (and we are already doing the same in other similar methods).
@brfrn169 What do you think?

core/src/main/java/com/scalar/db/api/Scan.java
jnmt: Good point. Essentially, both are the same. But I intentionally used the different classes for the internal representation (`Conjunction`) and user API (`AndConditionSet`) for ease of understanding since the conjunction would be unfamiliar for many users. From the developer's perspective, I just prefer the term "conjunction" since it's short and technically correct to use many places internal modules, including gRPC proto. Do you think "internal use" or something like that is enough here? Or should we use `AndConditionSet` also for the internal representation?

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -511,13 +785,34 @@ private void checkNotScanWithIndexOrScanAll() {
...
+      if (!isScanAll) {
+        throw new UnsupportedOperationException(
+            "This operation is supported only when scanning all the records of a database.");
+      }
+    }
+
+    private void checkConditionsEmpty() {
+      if (!conjunctions.isEmpty()) {
+        throw new UnsupportedOperationException(
+            "This operation is supported only when no conditions are specified at all.");
```
brfrn169, line 812: Instead of `UnsupportedOperationException`, I think `IllegalStateException` is proper in this case. And I think we can suggest using `clearConditions()` to clear the conditions in the message:
```suggestion
        throw new IllegalStateException(
            "This operation is supported only when no conditions are specified at all. "
                + "Please use clearConditions() to clear all the conditions.");
```

core/src/main/java/com/scalar/db/api/Scan.java
brfrn169: So maybe we should move `AndConditionSet` and `OrConditionSet` and their builder classes into `ScanBuilder`?

brfrn169 (commented):
I reviewed it briefly and left some comments. As we should examine it thoroughly, let me review it again later.

core/src/main/java/com/scalar/db/api/Scan.java
komamitsu: I see. Thanks for the explanation. Either option, adding a comment that describes the relationship between `Conjunction` and `AndConditionSet` and the usages or using `AndConditionSet` for internal use, sounds fine with me.

core/src/main/java/com/scalar/db/api/Scan.java
```diff
@@ -457,4 +484,316 @@ public enum Order {
       DESC,
     }
   }
+
+  @Immutable
+  public static class Conjunction {
+    private final Set<ConditionalExpression> conditions;
```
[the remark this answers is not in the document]
komamitsu: Agreed. It's a good improvement in terms of type consistency.

core/src/main/java/com/scalar/db/api/ScanAll.java
komamitsu: Yeah, it seems there is no reasonable solution to it. I think we can go with the implementation as is.

I wish Java had a special generics type `THIS` or something which represents the concrete class type itself so that we could write the method like `public THIS withOrdering()`...

jnmt: @brfrn169 @komamitsu Thank you! I left this as is.

core/src/main/java/com/scalar/db/api/Scan.java
jnmt: @komamitsu @brfrn169 Thanks for the comments. I will move both sets to the builder and describe the relationship and usage.

core/src/main/java/com/scalar/db/api/Scan.java
```diff
@@ -291,6 +297,24 @@ public Scan withProjections(Collection<String> projections) {
     return (Scan) super.withProjections(projections);
   }
 
+  Scan withConjunctions(Collection<Conjunction> conjunctions) {
+    this.conjunctions.addAll(conjunctions);
+    return this;
+  }
```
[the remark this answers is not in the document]
jnmt: I misunderstand something, but this method is package private. Is "deprecated" still necessary? I imagined current public withFooBar() mutation methods would be changed to package private ones like this function in v5.0.0.

core/src/main/java/com/scalar/db/api/Scan.java
```diff
@@ -457,4 +484,316 @@ public enum Order {
...
+      this.conditions = conditions;
+    }
+
+    /**
+     * Returns the set of {@code ConditionalExpression} which this conjunction is composed of.
+     *
+     * @return set of {@code ConditionalExpression} which this conjunction is composed of
+     */
+    public Set<ConditionalExpression> getConditions() {
+      return ImmutableSet.copyOf(conditions);
```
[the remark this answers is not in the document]
jnmt: Thank you for pointing it out. Fixed in d3bb194.

core/src/main/java/com/scalar/db/api/Scan.java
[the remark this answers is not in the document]
jnmt: Thank you for pointing it out. Fixed in d3bb194.

core/src/main/java/com/scalar/db/api/Scan.java
```diff
@@ -457,4 +484,316 @@ public enum Order {
...
+    public static BuildableOrConditionSet orConditionSet(Set<ConditionalExpression> conditions) {
+      return new BuildableOrConditionSet(conditions);
+    }
+  }
+
+  public static class AndOrConditionSetBuilder {
+
+    private final Set<ConditionalExpression> conditions;
+
+    AndOrConditionSetBuilder(Set<ConditionalExpression> conditions) {
```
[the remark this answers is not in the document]
jnmt: Good catch, thank you! Fixed in d3bb194.

core/src/test/java/com/scalar/db/api/ScanTest.java
```diff
@@ -45,6 +53,79 @@ private Scan prepareAnotherScan() {
...
+  //            ConditionBuilder.column(ANY_NAME_1).isEqualToText(ANY_TEXT_2)));
+  //  }
+  //
+  //  private Set<Conjunction> prepareConjunctionsWithDifferentOrder() {
+  //    return ImmutableSet.of(
+  //        RelationalScan.Conjunction.conjunction(
+  //            ConditionBuilder.column(ANY_NAME_1).isEqualToText(ANY_TEXT_2)),
+  //        RelationalScan.Conjunction.conjunction(
+  //            ConditionBuilder.column(ANY_NAME_1).isEqualToText(ANY_TEXT_1)));
+  //  }
```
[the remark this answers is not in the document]
jnmt: Oh, sorry. Fixed in d3bb194.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -321,11 +355,218 @@ public Scan build() {
...
+      this.tableName = tableName;
+      this.condition = null;
+    }
+
+    private BuildableScanAllWithWhere(
+        String namespaceName, String tableName, ConditionalExpression condition) {
+      this.namespaceName = namespaceName;
+      this.tableName = tableName;
+      this.condition = condition;
+    }
```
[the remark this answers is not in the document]
jnmt: Thank you for the advice. Fixed in d3bb194.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
jnmt: Exactly. Fixed in d3bb194.

jnmt:
@brfrn169 @komamitsu Thank you for the valuable feedback. I fixed all other than [this](https://github.com/scalar-labs/scalardb/pull/889#discussion_r1206164054). PTAL again when you get a chance.

core/src/main/java/com/scalar/db/api/Scan.java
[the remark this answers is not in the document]
jnmt: @brfrn169 Just a confirmation. Will those mutation methods themselves be deleted in v5.0.0? In addition, for example, `Scan` will have `ScanBuilder`, and the builder can directly access the private member of `Scan` without using `withFooBar()`. That's why we should make it deprecated even if a package private method, is my guess correct?

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -321,11 +355,216 @@ public Scan build() {
...
+      implements Consistency<BuildableScanAllWithWhere>,
+          Projection<BuildableScanAllWithWhere>,
+          Ordering<BuildableScanAllWithWhere>,
+          Limit<BuildableScanAllWithWhere> {
+
+    protected final String namespaceName;
+    protected final String tableName;
+    @Nullable protected final ConditionalExpression condition;
+    protected final Set<Scan.Conjunction> conjunctions = new HashSet<>();
+    protected final List<Set<ConditionalExpression>> disjunctions = new ArrayList<>();
```
komamitsu, line 469: Sorry for this delayed question. Why the types of `conjunctions` and `disjunctions` are different (`disjunctions` needs an order? `disjunctions` should be an internal class as well as `conjunctions`?)

jnmt: Thanks for the question. Actually, disjunctions don't need the order information, but `Sets.cartesianProduct()`, which is used for converting a CNF, requires a list.

komamitsu: I see. If the reason of having different types for `conjunctions` and `disjunctions` is only for `Sets.cartesianProduct()`, we can write like this?

```
  Sets.cartesianProduct(
      disjunctions.stream().map(Disjunction::getConditions).collect(Collectors.toList()))
      .stream()
      .map(Conjunction::of)
      .collect(Collectors.toSet()));
```

It would be great if the types of `conjunctions` and `disjunctions` are the same-level abstraction because I guess the readers of this code expect that.

jnmt: Yeah, that's better for readers. Thank you. I think we can do just like the following. BTW, I'm not going to introduce `Scan.Disjunction` since the CNF and disjunctions are only used in the `ScanBuilder` internally at least at this moment. Even so, is it weird and should we have it?

```java
      scan.withConjunctions(
          Sets.cartesianProduct(new ArrayList<>(disjunctions)).stream()
              .map(Scan.Conjunction::of)
              .collect(Collectors.toSet()));
```

jnmt: > BTW, I'm not going to introduce Scan.Disjunction since the CNF and disjunctions are only used in the ScanBuilder internally at least at this moment. Even so, is it weird and should we have it?

@komamitsu Or, how about using `Set<Set<ConditionalExpression>>` for conjunctions when building and converting them at the last, similar to the disjunctions? What do you think?

komamitsu: > how about using Set<Set<ConditionalExpression>> for conjunctions

It means like this?

```
    protected final Set<Set<ConditionalExpression>> conjunctions = new HashSet<>();
    protected final Set<Set<ConditionalExpression>> disjunctions = new HashSet<>();
```

If so, completely agreed.

core/src/main/java/com/scalar/db/api/Scan.java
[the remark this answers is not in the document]
brfrn169: Understood. Since the method in question is package-private and not public, there's no need to remove it or mark it as deprecated. Let's keep it as it is. If further changes or considerations are needed in the future, we can revisit this decision.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
jnmt: Yeah, that's what I meant. Fixed in a64862d. PTAL when you get a chance.

jnmt:
Dear all,
Sorry, I missed some cases for specifying orderings without conditions, updating a namespace and table after the where clause, and clearing the namespace after the where clause. Fixed in 2c305c9. PTAL, including those.

brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -548,4 +873,538 @@ public Scan build() {
...
+    ScanAll scan = new ScanAll();
+    scan.forNamespace(namespaceName).forTable(tableName).withLimit(limit);
+    orderings.forEach(scan::withOrdering);
+
+    if (condition != null) {
+      assert conjunctions.isEmpty() && disjunctions.isEmpty();
+      scan.withConjunctions(ImmutableSet.of(Scan.Conjunction.of(condition)));
+    } else if (conjunctions.isEmpty()) {
+      scan.withConjunctions(
+          Sets.cartesianProduct(new ArrayList<>(disjunctions)).stream()
```
komamitsu, line 1393: (Sorry for the delayed comment. I thought I put the following comment here, but it's gone...)

This seems to generate O x A conditions (O sets of A conditions), where O is the number of ORs and A is the number of ANDs. I'm not sure how these sets of conditions would be used in next PRs, but I guess ScalarDB fetches all entries and then filter the fetched records using these conditions set.

(This would be a silly question..., but) it's not possible to pushdown original conditions basically as-is to underlying RDBMS?

jnmt: Thank you for the question. It will be used in something like the following way. We don't filter results using the conditions (at least for JDBC storage) in the ScalarDB layer.

```java
  private String relationalConditionSqlString() {
   (snip)
    List<String> conjunctionList = new ArrayList<>();
    conjunctions.forEach(
        conjunction -> {
          List<String> conditions = new ArrayList<>();
          conjunction
              .getConditions()
              .forEach(
                  condition ->
                      conditions.add(condition.getColumn().getName() + convert(condition.getOperator())));
          conjunctionList.add(String.join(" AND ", conditions));
        });
    return " WHERE " + String.join(" OR ", conjunctionList);
  }
```

> (This would be a silly question..., but) it's not possible to pushdown original conditions basically as-is to underlying RDBMS?

Do you mean pushing down both DNF and CNF as is? It's technically possible, and it sometimes would perform better, but I think it trades off a development effort and maintainability for two internal representations.

komamitsu: Thanks for the answer. It sounds good!

What I meant was...
converting the following Scan
```
    Scan.newBuilder()
        .namespace("my_ns")
        .table("my_tbl")
        .all()
        .where(ConditionBuilder.column("c1").isEqualToInt(10))
        .and(ConditionBuilder.column("c2").isEqualToBigInt(20))
        .and(
            ConditionSetBuilder.condition(
                ConditionBuilder.column("c3").isEqualToInt(30))
                .or(ConditionBuilder.column("c4").isEqualToInt(40))
                .build())
           :
        .build();
```
into the following SQL
```
SELECT ...
FROM my_ns.my_tbl
WHERE c1 = 10 AND c2 = 20 AND (c3 = 30 OR c4 = 40)
```
although I'm not sure it works well 😅

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/ScanBuilder.java
jnmt: I see. That's pushing down a CNF as is. Yeah, as I said, technically, it works if we make `Scan` always have either conjunctions or disjunctions. But my concern is that we probably need to do the same for the gRPC proto and take care of which one should be used in many places. Anyway, this is kind of an internal part; we can revisit this in the future. Any additional discussions are welcome. Thanks for the comment.

feeblefakie (approved):
LGTM! Thank you!
(Very sorry for the late review 🙇 )

feeblefakie (approved):
LGTM! Thank you!
(Very sorry for the late review 🙇 )

PR #897, opened 2023-06-06
Improve documents for Handle Exceptions

brfrn169:
Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
@@ -56,9 +56,11 @@ public interface DistributedTransactionManager {
    * Begins a new transaction.
    *
    * @return {@link DistributedTransaction}
-   * @throws TransactionNotFoundException if beginning the transaction fails, but it's retryable.
-   *     You can retry the transaction by calling this method again.
-   * @throws TransactionException if beginning the transaction fails
+   * @throws TransactionNotFoundException if the transaction fails to begin due to transient faults.
+   *     You can retry the transaction
+   * @throws TransactionException if the transaction fails to begin due to transient or nontransient
+   *     faults. You can try retrying the transaction, but you may not be able to start the
+   *     transaction due to nontransient faults
    */
   DistributedTransaction begin() throws TransactionNotFoundException, TransactionException;
 
@@ -69,9 +71,11 @@ public interface DistributedTransactionManager {
    *
    * @param txId an user-provided unique transaction ID
    * @return {@link DistributedTransaction}
-   * @throws TransactionNotFoundException if beginning the transaction fails, but it's retryable.
-   *     You can retry the transaction by calling this method again.
-   * @throws TransactionException if beginning the transaction fails
+   * @throws TransactionNotFoundException if the transaction fails to begin due to transient faults.
+   *     You can retry the transaction
+   * @throws TransactionException if the transaction fails to begin due to transient or nontransient
+   *     faults. You can try retrying the transaction, but you may not be able to start the
+   *     transaction due to nontransient faults
    */
   DistributedTransaction begin(String txId)
       throws TransactionNotFoundException, TransactionException;
@@ -80,9 +84,11 @@ public interface DistributedTransactionManager {
    * Starts a new transaction. This method is an alias of {@link #begin()}.
    *
    * @return {@link DistributedTransaction}
-   * @throws TransactionNotFoundException if beginning the transaction fails, but it's retryable.
-   *     You can retry the transaction by calling this method again.
-   * @throws TransactionException if starting the transaction fails
+   * @throws TransactionNotFoundException if the transaction fails to start due to transient faults.
+   *     You can retry the transaction
+   * @throws TransactionException if the transaction fails to start due to transient or nontransient
+   *     faults. You can try retrying the transaction, but you may not be able to start the
+   *     transaction due to nontransient faults
    */
   default DistributedTransaction start() throws TransactionNotFoundException, TransactionException {
     return begin();
@@ -94,9 +100,11 @@ public interface DistributedTransactionManager {
    *
    * @param txId an user-provided unique transaction ID
    * @return {@link DistributedTransaction}
-   * @throws TransactionNotFoundException if beginning the transaction fails, but it's retryable.
-   *     You can retry the transaction by calling this method again.
-   * @throws TransactionException if starting the transaction fails
+   * @throws TransactionNotFoundException if the transaction fails to start due to transient faults.
+   *     You can retry the transaction
+   * @throws TransactionException if the transaction fails to start due to transient or nontransient
+   *     faults. You can try retrying the transaction, but you may not be able to start the
+   *     transaction due to nontransient faults
    */
   default DistributedTransaction start(String txId)
       throws TransactionNotFoundException, TransactionException {
--- a/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
@@ -56,9 +56,11 @@ public interface DistributedTransactionManager {
    * Begins a new transaction.
    *
    * @return {@link DistributedTransaction}
+   * @throws TransactionNotFoundException if beginning the transaction fails, but it's retryable.
+   *     You can retry the transaction by calling this method again.
    * @throws TransactionException if beginning the transaction fails
    */
-  DistributedTransaction begin() throws TransactionException;
+  DistributedTransaction begin() throws TransactionNotFoundException, TransactionException;
 
   /**
    * Begins a new transaction with the specified transaction ID. It is users' responsibility to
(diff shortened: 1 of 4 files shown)


docs/api-guide.md
```diff
@@ -735,6 +735,10 @@ public class Sample {
       DistributedTransaction tx;
       try {
         tx = transactionManager.begin();
+      } catch (TransactionNotFoundException e) {
+        // If beginning a transaction failed, but it's retryable, so that you can retry the
+        // transaction
```
Torch3333, line 740:
```suggestion
        // If beginning a transaction failed, but it's retryable, so you can retry the
        // transaction
```

docs/two-phase-commit-transactions.md
```diff
@@ -222,6 +222,10 @@ public class Sample {
       TwoPhaseCommitTransaction tx;
       try {
         tx = transactionManager.begin();
+      } catch (TransactionNotFoundException e) {
+        // If beginning a transaction failed, but it's retryable, so that you can retry the
+        // transaction
```
Torch3333, line 227:
```suggestion
        // If beginning a transaction failed, but it's retryable, so you can retry the
        // transaction
```

docs/api-guide.md
```diff
@@ -781,6 +785,10 @@ public class Sample {
 }
 ```
 
+The `begin()` API could throw `TransactionException` and `TransactionNotFoundException`.
+If you catch `TransactionException`, it indicates some failure (e.g., database failure and network error) happens during a transaction, so you should cancel the transaction or retry the transaction after the failure/error is fixed.
+If you catch `TransactionNotFoundException`, it indicates the transaction is not found, but it's retryable, so that you can retry the transaction.
```
Torch3333, line 790:
```suggestion
If you catch `TransactionNotFoundException`, it indicates the transaction is not found, but it's retryable, so you can retry the transaction.
```

docs/api-guide.md
```diff
@@ -794,6 +802,12 @@ In such a case, you need to check if the transaction is committed successfully o
 How to identify a transaction status is delegated to users.
 You may want to create a transaction status table and update it transactionally with other application data so that you can get the status of a transaction from the status table.
 
+Please note that if you begin a transaction with specifying a transaction ID, you need to use a different transaction ID when you retry the transaction.
```
Torch3333, line 805:
```suggestion
Please note that if you begin a transaction by specifying a transaction ID, you must use a different ID when you retry the transaction.
```

docs/api-guide.md
```diff
@@ -794,6 +802,12 @@ In such a case, you need to check if the transaction is committed successfully o
...
+
+Although it's not illustrated in the sample code, the `resume()` API could also throw a `TransactionNotFoundException`.
```
Torch3333, line 807:
```suggestion
Although not illustrated in the sample code, the `resume()` API could also throw a `TransactionNotFoundException`.
```

docs/two-phase-commit-transactions.md
```diff
@@ -279,6 +283,13 @@ public class Sample {
 }
 ```
 
+The `begin()` API could throw `TransactionException` and `TransactionNotFoundException`.
+If you catch `TransactionException`, it indicates some failure (e.g., database failure and network error) happens during a transaction, so you should cancel the transaction or retry the transaction after the failure/error is fixed.
+If you catch `TransactionNotFoundException`, it indicates the transaction is not found, but it's retryable, so that you can retry the transaction.
+
+Although it's not illustrated in the sample code, the `join()` API could also throw a `TransactionException` and `TransactionNotFoundException`.
```
Torch3333, line 290:
```suggestion
Although not illustrated in the sample code, the `join()` API could also throw a `TransactionException` and `TransactionNotFoundException`.
```

docs/two-phase-commit-transactions.md
```diff
@@ -300,6 +311,12 @@ In such a case, you need to check if the transaction is committed successfully o
 How to identify a transaction status is delegated to users.
 You may want to create a transaction status table and update it transactionally with other application data so that you can get the status of a transaction from the status table.
 
+Please note that if you begin a transaction with specifying a transaction ID, you need to use a different transaction ID when you retry the transaction.
+
+Although it's not illustrated in the sample code, the `resume()` API could also throw a `TransactionNotFoundException`.
```
Torch3333, line 316:
```suggestion
Although not illustrated in the sample code, the `resume()` API could also throw a `TransactionNotFoundException`.
```

docs/two-phase-commit-transactions.md
```diff
@@ -300,6 +311,12 @@ In such a case, you need to check if the transaction is committed successfully o
 How to identify a transaction status is delegated to users.
 You may want to create a transaction status table and update it transactionally with other application data so that you can get the status of a transaction from the status table.
 
+Please note that if you begin a transaction with specifying a transaction ID, you need to use a different transaction ID when you retry the transaction.
```
Torch3333, line 314:
```suggestion
Please note that if you begin a transaction by specifying a transaction ID, you must use a different ID when you retry the transaction.
```

Torch3333 (commented):
Overall, it looks good, thank you. I left some suggestions.

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
```diff
@@ -56,9 +56,11 @@ public interface DistributedTransactionManager {
    * Begins a new transaction.
    *
    * @return {@link DistributedTransaction}
+   * @throws TransactionNotFoundException if beginning the transaction fails, but it's retryable.
```
feeblefakie, line 59:
```suggestion
   * @throws TransactionNotFoundException if the transaction fails to start.
```
I removed `but it's retryable` because I feel it's a bit redundant with the next sentence.
Ditto for the others.

docs/api-guide.md
```diff
@@ -781,6 +785,10 @@ public class Sample {
 }
 ```
 
+The `begin()` API could throw `TransactionException` and `TransactionNotFoundException`.
+If you catch `TransactionException`, it indicates some failure (e.g., database failure and network error) happens during a transaction, so you should cancel the transaction or retry the transaction after the failure/error is fixed.
```
feeblefakie, line 789:
```suggestion
If you catch `TransactionException`, it indicates some failure (e.g., database failure and network error) happens during the transaction, so you should cancel the transaction or retry the transaction after the failure or error is fixed.
```

feeblefakie (commented):
Overall, looking good.
Left some minor suggestions. PTAL!

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
brfrn169: I think removing `but it's retryable` may make the statement sound unsure about the difference from the `TransactionException`. So, how about removing the next sentence?

Torch3333 (approved):
LGTM, thank you!

docs/api-guide.md
```diff
@@ -781,19 +785,29 @@ public class Sample {
 }
 ```
 
+The `begin()` API could throw `TransactionException` and `TransactionNotFoundException`.
+If you catch `TransactionException`, it indicates some failure (e.g., database failure and network error) happens during the transaction, so you should cancel the transaction or retry the transaction after the failure or error is fixed.
+If you catch `TransactionNotFoundException`, it indicates the transaction is not found, but it's retryable, so you can retry the transaction.
```
komamitsu, line 790: > it indicates the transaction is not found

[minor] Is it possible to describe possible causes of the exception?

komamitsu (approved):
LGTM! 👍

docs/api-guide.md
brfrn169: Actually, this doesn't happens with ScalarDB, but It can happen with ScalarDB Cluster. So, I can't describe possible causes here... :(

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
feeblefakie: OK, then, how about something like this to make things clear?
```
@throws TransactionNotFoundException if the transaction fails to start due to xxx. You can retry the transaction.
@throws TransactionException if the transaction fails to start due to yyy. You cannot retry the transaction.
```

Can we say something for `xxx` and `yyy`?

brfrn169: Thank you for the suggestion.

Actually, it's hard to say concrete reasons for `TransactionNotFoundException` because it's only occurs in ScalarDB Cluster.

So, how about some thing like the following?

```
@throws TransactionNotFoundException if the transaction fails to start due to retryable reasons. You can retry the transaction.
@throws TransactionException if the transaction fails to start.
```

feeblefakie: How about this?
```
@throws TransactionNotFoundException if the transaction fails to start due to transient faults. You can retry the transaction.
@throws TransactionException if the transaction fails to start due to transient or nontransient faults. You can try retrying the transaction, but you may not be able to start the transaction due to nontransient faults.
```

feeblefakie (commented):
Left one suggestion. PTAL!

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
brfrn169: Thank you for the suggestion! Fixed in 59d84339872f8c1859e78f051240c317cde53c55. Please take a look when you have time!

feeblefakie (approved):
LGTM! Thank you!

PR #899, opened 2023-06-07
Support condition for transactional operations

Torch3333:
**Context** 
Setting conditions for Put or Delete in a transaction was not supported. Setting conditions in the operation builder was possible, but they were ignored.

**Changes**
Setting condition is now fully supported on Put and Delete operation for consensus commit (regular and 2pc) or Jdbc transactions. For example :

https://github.com/scalar-labs/scalardb/blob/46441d23a7772be58cd2ad2bfb887b71767f5b31/docs/api-guide.md?plain=1#L622-L640

When the mutation condition is not met, a `UnsatisfiedConditionException` is thrown when the mutation is executed.


brfrn169 (commented):
Overall LGTM. Left one minor comment. Please take a look when you have time!

Since this change affects the core transaction logic, I'll review it again later. Thanks.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -188,17 +193,24 @@ public Optional<List<Key>> get(Scan scan) {
...
+        conditionalMutationValidator.validateConditionIsSatisfied(entry.getValue(), result);
+      }
       composer.add(entry.getValue(), result);
     }
     for (Entry<Key, Delete> entry : deleteSet.entrySet()) {
       TransactionResult result =
           readSet.containsKey(entry.getKey()) ? readSet.get(entry.getKey()).orElse(null) : null;
+      if (composer instanceof PrepareMutationComposer) {
+        conditionalMutationValidator.validateConditionIsSatisfied(entry.getValue(), result);
+      }
```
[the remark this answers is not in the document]
brfrn169: Maybe, we can add a new method to validate conditions in mutations in the write set and the delete set to the `Snapshot` class, and we can call it in `CommitHandler.prepare()`. What do you think?

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -81,14 +89,17 @@ public List<Result> scan(Scan scan) throws CrudException {
...
-    // Ignore the condition in the put
-    if (put.getCondition().isPresent()) {
-      logger.warn("ignoring the condition of the mutation: {}", put);
-      put.withCondition(null);
-    }
-
     try {
-      jdbcService.put(put, connection);
+      if (!jdbcService.put(put, connection)) {
+        // To match the Consensus Commit transaction behavior, instead of throwing an exception here
```
komamitsu, line 94: > To match the Consensus Commit transaction behavior

Could you elaborate this? It would be great if you tell me an example case that violates the Consensus Commit transaction behavior.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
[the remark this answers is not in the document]
Torch3333: Thank you for the suggestion, I agree. 
I did as you say and created the `Snapshot.validateConditionalMutations()` method that is executed from the `CommitHandler.prepare()`
cf. https://github.com/scalar-labs/scalardb/pull/899/commits/afc2fbec6c4d7c64cf7617d94892a84cfce0e297

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
```diff
@@ -52,51 +55,54 @@ public void check(Mutation mutation) throws ExecutionException {
...
+        || condition instanceof PutIfNotExists
+        || condition instanceof PutIfExists)) {
+      throw new IllegalArgumentException(
+          "A "
+              + condition.getClass().getSimpleName()
+              + " condition is not allowed on Put operation");
+    }
+    checkConditionIsNotTargetingMetadataColumns(condition, metadata);
+    ConditionChecker conditionChecker = conditionCheckerFactory.create(metadata.getTableMetadata());
+    conditionChecker.check(condition, true);
```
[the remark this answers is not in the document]
Torch3333: I hesitated to rely on the `@VisibleForTesting` pattern, but since the `ConditionCheckerFactory` is only useful for the `ConsensusCommitMutationOperationChecker`, I think you are right. Thank you.
see https://github.com/scalar-labs/scalardb/pull/899/commits/afc2fbec6c4d7c64cf7617d94892a84cfce0e297

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
```diff
@@ -0,0 +1,161 @@
...
+      case IS_NULL:
+        return existingRecordColumn.hasNullValue();
+      case IS_NOT_NULL:
+        return !existingRecordColumn.hasNullValue();
+      case EQ:
+        return Ordering.natural().compare(existingRecordColumn, conditionalExpressionColumn) == 0;
+      case NE:
+        return Ordering.natural().compare(existingRecordColumn, conditionalExpressionColumn) != 0;
+        // For 'greater than' and 'less than' types of conditions and when the existing record is
+        // null, we consider the condition to be unsatisfied. This mimic the behavior as if
```
komamitsu, line 143:
```suggestion
        // null, we consider the condition to be unsatisfied. This mimics the behavior as if
```

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
Torch3333: For Consensus Commit transaction, the mutational operation condition are evaluated in `tx.commit()`
Though, for JdbcTransaction, the mutational operation conditions are evaluated directly in the `tx.put()` or `tx.delete()`, so before the transaction is committed. 

In the code above, when `jdbcService.put(put, connection)` result is false, it means the `put` was not applied because its condition is not validated. So technically, we could directly throw an exception when `tx.put()` is executed. But we choose not to do that because it means the user would have to handle the "mutation condition unsatisfied exception" differently when using Jdbc Transaction or Consensus Commit Transaction.

komamitsu: Thanks for the explanation! Understood.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -163,6 +178,58 @@ public void commit() throws CommitException, UnknownTransactionStatusException {
...
+    }
+  }
+
+  private void throwUnsatisfiedConditionException(Mutation mutation)
+      throws CommitUnsatisfiedConditionException {
+    assert mutation.getCondition().isPresent();
+
+    // Build the exception message
+    MutationCondition condition = mutation.getCondition().get();
+    String conditionColumns = "";
```
komamitsu, line 207: How about using StringBuilder instead to avoid instantiating temporary objects?

komamitsu: See also https://dev.to/this-is-learning/performance-benchmarking-string-and-string-builder-3bid

komamitsu:
> For regular transaction, when a mutation operation condition is not met, committing the transaction will throw a CommitUnsatisfiedConditionException.
> For 2pc transaction, when a mutation operation condition is not met, preparing the transaction will throw a PreparationUnsatisfiedConditionException.

@Torch3333 What is the main benefit of having the two separate exceptions? The background of this question is that I'm wondering if it's an option to merge them to just a UnsatisfiedConditionException.

komamitsu (approved):
LGTM! 👍

Torch3333:
@komamitsu 
Thanks for the question; I should have detailed the reasoning beforehand.

`transaction.commit()` and `twoPCTransaction.prepare()` do not throw a type of exception that is common between them. 
`commit` throws `CommitException` and `UnknownTransactionStatusException` while `prepare` throws `PreparationException`. 
So if we wanted to have a shared `UnsatisfiedConditionException` for these two methods, we would have needed to add this new exception type to the currently thrown exception.
We did not think it was worth adding a common exception type considering the trouble it would cause our users to have to handle a new checked exception type in their existing codebase.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
Torch3333: Thanks, see https://github.com/scalar-labs/scalardb/pull/899/commits/562ca5f66b2d024fad62e89b3b958fcd36c116d5

docs/api-guide.md
```diff
@@ -760,8 +836,8 @@ public class Sample {
         } catch (RollbackException ex) {
           // Rolling back the transaction failed. You can log it here
         }
-      } catch (CrudException | CommitException e) {
-        // If you catch CrudException or CommitException, it indicates some failure happens, so you
+      } catch (CrudException | CommitUnsatisfiedConditionException | CommitException e) {
```
brfrn169, line 839: The handling of `CommitUnsatisfiedConditionException` is very dependent on the application use cases. Therefore, we should use a separate `catch` block for this exception, as follows:

```suggestion
      } catch (CommitUnsatisfiedConditionException e) {
        // If you catch CommitUnsatisfiedConditionException, ...
      } catch (CrudException | CommitException e) {
```

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
Torch3333: That is a good point, thank you. 
see https://github.com/scalar-labs/scalardb/pull/899/commits/494137ed46c6343b4b8d223130dfd5b76a2b8f24

docs/two-phase-commit-transactions.md
```diff
@@ -257,8 +257,8 @@ public class Sample {
         } catch (RollbackException ex) {
           // Rolling back the transaction failed. You can log it here
         }
-      } catch (CrudException | PreparationException | ValidationException | CommitException e) {
-        // If you catch CrudException or PreparationException or ValidationException or
+      } catch (CrudException | PreparationUnsatisfiedConditionException | PreparationException | ValidationException | CommitException e) {
```
brfrn169, line 260: Ditto.
```suggestion
      } catch (PreparationUnsatisfiedConditionException e) {
        // If you catch PreparationUnsatisfiedConditionException, ...
      } catch (CrudException | PreparationException | ValidationException | CommitException e) {
```

server/src/test/java/com/scalar/db/server/DistributedTransactionServiceTest.java
```diff
@@ -836,6 +839,44 @@ public void abort_ManagerThrowsTransactionException_ShouldThrowInternalError()
     assertThat(exceptionCaptor.getValue().getStatus().getCode()).isEqualTo(Code.INTERNAL);
   }
 
+  @Test
+  public void
+      commit_TransactionThrowsCommitUnsatisfiedConditionException_ShouldRespondWithUnsatisfiedConditionError()
+          throws TransactionException {
```
brfrn169, line 845: Nit. We should add the same test for `TwoPhaseCommitTransactionService`?

brfrn169 (commented):
Left several minor comments. Please take a look when you have time!

komamitsu (approved):
👍

docs/api-guide.md
```diff
@@ -583,6 +583,47 @@ Put put =
         .build();
 ```
 
+##### Put with a condition
+
```
feeblefakie, line 587: I think it might be helpful to add something like this here.

You can write arbitrary conditions (e.g., a bank account balance must be equal to or more than zero) that you require a transaction to meet before being committed by having some checking logic in the transaction. Alternatively, you can also write simple conditions in a mutation operation, such as Put and Delete.

(If this is added, I think we need to update the next sentence a bit.)

feeblefakie (commented):
I made some comments and suggestions.
PTAL!
Since it's the core part, I will take another look later.

docs/api-guide.md
Torch3333: Thank you for the suggestion,  
To avoid redundancy in the explanation, I will merge the `Put with a condition` and `Delete with a condition` paragraphs.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  /**
+   * This validates the condition for the Put operation is satisfied.
+   *
+   * @param put a Put operation
+   * @param existingRecord the current value of the record targeted by the mutation, if any
+   * @throws PreparationUnsatisfiedConditionException if the condition is not satisfied
+   */
+  public void validateConditionIsSatisfied(Put put, @Nullable TransactionResult existingRecord)
```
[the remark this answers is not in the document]
Torch3333: I see, thank you.
I used a verbose name for this method to be clear that this method is not about checking the condition is not contextually correct (e.g. the column used in the condition matches the table metadata)
So, I will use your second suggestion `checkIfConditionIsSatisfied`

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
```diff
@@ -0,0 +1,161 @@
...
+@ThreadSafe
+public class ConditionalMutationValidator {
+  private final String transactionId;
+
+  public ConditionalMutationValidator(String transactionId) {
+    this.transactionId = transactionId;
+  }
+
+  /**
+   * This validates the condition for the Put operation is satisfied.
```
feeblefakie, line 34:
```suggestion
   * This checks if the condition of the specified Put operation is satisfied for the specified record.
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
```diff
@@ -0,0 +1,161 @@
...
+      if (recordExists) {
+        throwWhenRecordExists(condition);
+      }
+    } else {
+      throw new AssertionError();
+    }
+  }
+
+  /**
+   * This validates the condition for the Delete operation is satisfied.
```
feeblefakie, line 67: Ditto.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
```diff
@@ -0,0 +1,161 @@
...
+import com.scalar.db.api.PutIfExists;
+import com.scalar.db.api.PutIfNotExists;
+import com.scalar.db.exception.transaction.PreparationUnsatisfiedConditionException;
+import com.scalar.db.io.Column;
+import java.util.List;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * This class verifies the mutation condition for Put and Delete operation are met by asserting the
```
feeblefakie, line 21: How about something like this?

```
This class checks if a record satisfies the conditions of Put and Delete operations that mutate the record.
```

I feel the sentence `It does not leverage the underlying storage to validate the condition.` is a bit confusing.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
```diff
@@ -0,0 +1,161 @@
...
+ * condition on the existing record (if any) targeted by the mutation. It does not leverage the
+ * underlying storage to validate the condition.
+ */
+@ThreadSafe
+public class ConditionalMutationValidator {
```
feeblefakie, line 26:
```suggestion
public class MutationConditionsValidator {
```
I think it validates conditions (not mutations), so should it be like this?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -85,9 +94,11 @@ public void prepare(Snapshot snapshot) throws PreparationException {
   }
 
   private void prepareRecords(Snapshot snapshot)
-      throws ExecutionException, PreparationConflictException {
+      throws ExecutionException, PreparationConflictException,
+          PreparationUnsatisfiedConditionException {
     PrepareMutationComposer composer =
         new PrepareMutationComposer(snapshot.getId(), tableMetadataManager);
+    snapshot.validateConditionalMutations();
```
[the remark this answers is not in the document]
feeblefakie: Moreover, I'm also wondering if validating conditions should be `Snapshot`'s responsibility or not.

@brfrn169 What do you think?

feeblefakie (commented):
Left some additional comments.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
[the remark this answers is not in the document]
feeblefakie: Sorry, I should have elaborated more.
I meant, can we check the conditions of mutation operations when calling the operations, not when committing 
 the transaction? (so that we can do an early abort)
For example, it seems like we can check the conditions of `Put` when calling `CrudHandler.put`.
I might miss something, though.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
Torch3333: This sounds good, thank you.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
Torch3333: Indeed, thank you.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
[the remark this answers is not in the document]
brfrn169: > For example, it seems like we can check the conditions of Put when calling CrudHandler.put.
@feeblefakie Ah, indeed, it seems we could do that.

@Torch3333 What do you think?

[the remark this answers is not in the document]
Torch3333: @brfrn169 
After discussing with @feeblefakie this morning, we concluded it is possible to move the condition check to `CrudHandler.put` and `CrudHandler.delete`.
Regarding the throw exception when the condition is not satisfied, we can then have a single exception type for regular and 2pc transaction interface that is called `UnsatisfiedConditionException` (extending `CrudException`) when executing `transaction.put()` or `transaction.delete()`. 
Overall, it will simplify the implementation.

docs/api-guide.md
Torch3333: Understood, I put the new `UnsatisfiedConditionException` in a separate catch block. Thank you.
Fixed in https://github.com/scalar-labs/scalardb/pull/899/commits/46441d23a7772be58cd2ad2bfb887b71767f5b31

docs/two-phase-commit-transactions.md
Torch3333: Likewise, thank you. 
Fixed in https://github.com/scalar-labs/scalardb/pull/899/commits/46441d23a7772be58cd2ad2bfb887b71767f5b31

server/src/test/java/com/scalar/db/server/DistributedTransactionServiceTest.java
Torch3333: Good catch, thanks. I added unit test for it in TwoPhaseCommitTransactionService and on the client side too.

docs/api-guide.md
Torch3333: Fixed in https://github.com/scalar-labs/scalardb/pull/899/commits/46441d23a7772be58cd2ad2bfb887b71767f5b31

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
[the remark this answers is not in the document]
Torch3333: I renamed the method to `MutationConditionsValidator.checkIfConditionIsSatisfied` in https://github.com/scalar-labs/scalardb/pull/899/commits/46441d23a7772be58cd2ad2bfb887b71767f5b31

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
Torch3333: Thank you. It's fixed in https://github.com/scalar-labs/scalardb/pull/899/commits/46441d23a7772be58cd2ad2bfb887b71767f5b31

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
Torch3333: Thank you. It's fixed in https://github.com/scalar-labs/scalardb/pull/899/commits/46441d23a7772be58cd2ad2bfb887b71767f5b31

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
Torch3333: Thank you. It's fixed in https://github.com/scalar-labs/scalardb/pull/899/commits/46441d23a7772be58cd2ad2bfb887b71767f5b31

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConditionalMutationValidator.java
Torch3333: It's fixed in https://github.com/scalar-labs/scalardb/pull/899/commits/46441d23a7772be58cd2ad2bfb887b71767f5b31

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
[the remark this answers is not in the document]
Torch3333: I revised the PR as described above. 
Thank you for the suggestion.

Torch3333:
@josh-wong I finished revising the documentation, so you can have a look when you get the chance. Thank you.

brfrn169 (approved):
LGTM! Thank you!

docs/api-guide.md
```diff
@@ -607,6 +607,80 @@ Delete delete =
 transaction.delete(delete);
 ```
 
+#### Put and Delete with a condition
+You can write arbitrary conditions (e.g., a bank account balance must be equal to or more than zero) that you
+require a transaction to meet before being committed by having some checking logic in the transaction.
+Alternatively, you can also write simple conditions in a mutation operation, such as Put and Delete.
```
josh-wong, line 613:
```suggestion
You can write arbitrary conditions (e.g., a bank account balance must be equal to or more than zero) that you require a transaction to meet before being committed by having logic that checks the conditions in the transaction.
Alternatively, you can write simple conditions in a mutation operation, such as Put and Delete.
```

docs/api-guide.md
```diff
@@ -607,6 +607,80 @@ Delete delete =
...
+
+When a Put or Delete operation includes a condition, it is only executed if the specified condition is met.
```
josh-wong, line 615:
```suggestion
When a Put or Delete operation includes a condition, the operation is executed only if the specified condition is met.
```

docs/api-guide.md
```diff
@@ -607,6 +607,80 @@ Delete delete =
...
+        .floatValue("c4", 1.23F)
+        .doubleValue("c5", 4.56)
+        .condition(condition) // condition
+        .build();
+
+// Execute the Put operation
+transaction.put(put);
+```
+
+Other than the `putIf` condition, you can specify the `putIfExists` and `putIfNotExists` conditions as follows:
```
josh-wong, line 643: Does this part mean "Instead of using the `putIf` condition"?

```suggestion
Instead of using the `putIf` condition, you can specify the `putIfExists` and `putIfNotExists` conditions as follows:
```

Or does this part mean "In addition to using the `putIf` condition"?

```suggestion
In addition to using the `putIf` condition, you can specify the `putIfExists` and `putIfNotExists` conditions as follows:
```

docs/api-guide.md
```diff
@@ -607,6 +607,80 @@ Delete delete =
...
+        .partitionKey(partitionKey)
+        .clusteringKey(clusteringKey)
+        .condition(condition)  // condition
+        .build();
+
+// Execute the Delete operation
+transaction.delete(delete);
+```
+
+Other than the `deleteIf` condition, you can specify the `deleteIfExists` condition as follows:
```
josh-wong, line 677: Does this part mean "Instead of using the `deleteIf` condition"?

```suggestion
Instead of using the `deleteIf` condition, you can specify the `deleteIfExists` condition as follows:
```

Or does this part mean "In addition to using the `deleteIf` condition"?

```suggestion
In addition to using the `deleteIf` condition, you can specify the `deleteIfExists` condition as follows:
```

docs/api-guide.md
```diff
@@ -672,9 +746,7 @@ Scan scanUsingSpecifiedNamespace =
 
 #### Notes
 
-- All the builders of the CRUD operations can specify consistency with the `consistency()` methods, but it's ignored, and the `LINEARIZABLE` consistency level is always used in transactions.
-- Also, the builders of the mutation operations (Put and Delete operations) can specify a condition with the `condition()` methods, but it's ignored, too. 
-Please program such conditions in a transaction if you want to implement conditional mutation.
+All the builders of the CRUD operations can specify consistency with the `consistency()` methods, but it's ignored, and the `LINEARIZABLE` consistency level is always used in transactions.
```
josh-wong, line 749: I separated this sentence into two for readability.

```suggestion
Although all the builders of the CRUD operations can specify consistency by using the `consistency()` methods, those methods are ignored. Instead, the `LINEARIZABLE` consistency level is always used in transactions.
```

josh-wong (changes requested):
I added a few comments and suggested revisions. PTAL!

docs/api-guide.md
Torch3333: Same thinking as for the Put operation here. Either a `deleteIf` or a `deleteIfExists` condition can be specified for a Delete operation so the `In addition to using...` suggestion is more appropriate, isn't it?

docs/api-guide.md
josh-wong: @Torch3333 In that case, yes, let's use my second suggestion "In addition ... ."

docs/api-guide.md
josh-wong: @Torch3333 In that case, yes, let's use my second suggestion "In addition to using ... " here as well.

Torch3333:
@josh-wong Thank you for the suggestions. I applied them all.

josh-wong (approved):
LGTM! Thank you!🙇🏻‍♂️

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -519,6 +519,10 @@ public boolean isValidationRequired() {
     return isExtraReadEnabled();
   }
 
+  public Optional<TransactionResult> getFromReadSet(Key key) {
```
feeblefakie, line 522: Super minor, but I think it's better to be placed below `containsKeyInReadSet` just to cluster similar methods.

feeblefakie (approved):
I left one very minor comment, but LGTM! Thank you!

josh-wong (approved):
LGTM! Thank you!🙇🏻‍♂️

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
Torch3333: Indeed, thank you. I updated it in https://github.com/scalar-labs/scalardb/pull/899/commits/c61274026128d88a30b053d4722613253bc33d10

brfrn169 (approved):

PR #900, opened 2023-06-07
Add relational scan for JDBC storage and consensus commit transaction

jnmt:
This PR adds the relational scan feature for JDBC storage. For transactions, it only supports the consensus-commit transaction. JDBC and gRPC transactions will be introduced by other PRs.

### DIFF
--- a/core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
+++ b/core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
@@ -1,11 +1,14 @@
 package com.scalar.db.common.checker;
 
+import com.scalar.db.api.ConditionalExpression;
+import com.scalar.db.api.ConditionalExpression.Operator;
 import com.scalar.db.api.Delete;
 import com.scalar.db.api.Get;
 import com.scalar.db.api.Mutation;
 import com.scalar.db.api.Operation;
 import com.scalar.db.api.Put;
 import com.scalar.db.api.Scan;
+import com.scalar.db.api.Scan.Conjunction;
 import com.scalar.db.api.ScanAll;
 import com.scalar.db.api.Selection;
 import com.scalar.db.api.TableMetadata;
@@ -126,6 +129,10 @@ public class OperationChecker {
     if (scanAll.getLimit() < 0) {
       throw new IllegalArgumentException("The limit cannot be negative. Operation: " + scanAll);
     }
+
+    checkOrderings(scanAll, metadata);
+
+    checkConjunctions(scanAll, metadata);
   }
 
   private void checkProjections(Selection selection, TableMetadata metadata) {
@@ -215,6 +222,40 @@ public class OperationChecker {
       } else {
         if (reverse != rightOrder) {
           throw new IllegalArgumentException(message.get());
+        }
+      }
+    }
+  }
+
+  private void checkOrderings(ScanAll scan, TableMetadata metadata) {
+    for (Scan.Ordering ordering : scan.getOrderings()) {
+      if (!metadata.getColumnNames().contains(ordering.getColumnName())) {
+        throw new IllegalArgumentException(
+            "The specified ordering column is not found. Invalid ordering: "
+                + ordering
+                + ", Operation: "
+                + scan);
+      }
+    }
+  }
+
+  private void checkConjunctions(ScanAll scan, TableMetadata metadata) {
+    for (Conjunction conjunction : scan.getConjunctions()) {
+      for (ConditionalExpression condition : conjunction.getConditions()) {
+        boolean isValid;
+        if (condition.getOperator() == Operator.IS_NULL
+            || condition.getOperator() == Operator.IS_NOT_NULL) {
+          // the value must be null if the operator is 'is null' or `is not null`
+          isValid =
+              new ColumnChecker(metadata, false, true, false, false).check(condition.getColumn());
+        } else {
+          // Otherwise, the value must not be null
+          isValid =
+              new ColumnChecker(metadata, true, false, false, false).check(condition.getColumn());
+        }
+        if (!isValid) {
+          throw new IllegalArgumentException(
+              "The condition is not properly specified. Operation: " + scan);
         }
       }
     }
--- a/core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
@@ -19,6 +19,7 @@ import com.scalar.db.common.TableMetadataManager;
 import com.scalar.db.common.checker.OperationChecker;
 import com.scalar.db.config.DatabaseConfig;
 import com.scalar.db.exception.storage.ExecutionException;
+import com.scalar.db.util.ScalarDbUtils;
 import java.util.List;
 import java.util.Optional;
 import javax.annotation.Nonnull;
@@ -63,6 +64,21 @@ public class Cassandra extends AbstractDistributedStorage {
     operationChecker = new OperationChecker(metadataManager);
   }
 
+  Cassandra(
+      DatabaseConfig config,
+      ClusterManager clusterManager,
+      StatementHandlerManager handlers,
+      BatchHandler batch,
+      TableMetadataManager metadataManager,
+      OperationChecker operationChecker) {
+    super(config);
+    this.clusterManager = clusterManager;
+    this.handlers = handlers;
+    this.batch = batch;
+    this.metadataManager = metadataManager;
+    this.operationChecker = operationChecker;
+  }
+
   @Override
   @Nonnull
   public Optional<Result> get(Get get) throws ExecutionException {
@@ -88,6 +104,11 @@ public class Cassandra extends AbstractDistributedStorage {
   public Scanner scan(Scan scan) throws ExecutionException {
     scan = copyAndSetTargetToIfNot(scan);
     operationChecker.check(scan);
+
+    if (ScalarDbUtils.isRelational(scan)) {
+      throw new UnsupportedOperationException(
+          "scanning all records with orderings or conditions is not supported in Cassandra");
+    }
 
     ResultSet results = handlers.select().handle(scan);
 
--- a/core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
@@ -5,6 +5,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import com.datastax.driver.core.ResultSet;
 import com.datastax.driver.core.Row;
 import com.datastax.driver.core.Session;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Inject;
 import com.scalar.db.api.Delete;
 import com.scalar.db.api.DistributedStorage;
@@ -64,6 +65,7 @@ public class Cassandra extends AbstractDistributedStorage {
     operationChecker = new OperationChecker(metadataManager);
   }
 
+  @VisibleForTesting
   Cassandra(
       DatabaseConfig config,
       ClusterManager clusterManager,
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
@@ -18,6 +18,7 @@ import com.scalar.db.storage.jdbc.query.DeleteQuery;
 import com.scalar.db.storage.jdbc.query.QueryBuilder;
 import com.scalar.db.storage.jdbc.query.SelectQuery;
 import com.scalar.db.storage.jdbc.query.UpsertQuery;
+import com.scalar.db.util.ScalarDbUtils;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -88,10 +89,17 @@ public class JdbcService {
 
     TableMetadata tableMetadata = tableMetadataManager.getTableMetadata(scan);
 
-    SelectQuery selectQuery =
-        scan instanceof ScanAll
-            ? buildSelectQueryForScanAll((ScanAll) scan, tableMetadata)
-            : buildSelectQueryForScan(scan, tableMetadata);
+    SelectQuery selectQuery;
+    if (scan instanceof ScanAll) {
+      ScanAll scanAll = (ScanAll) scan;
+      if (ScalarDbUtils.isRelational(scan)) {
+        selectQuery = buildSelectQueryForRelationalScan(scanAll, tableMetadata);
+      } else {
+        selectQuery = buildSelectQueryForScanAll(scanAll, tableMetadata);
+      }
+    } else {
+      selectQuery = buildSelectQueryForScan(scan, tableMetadata);
+    }
     PreparedStatement preparedStatement = connection.prepareStatement(selectQuery.sql());
     selectQuery.bind(preparedStatement);
     ResultSet resultSet = preparedStatement.executeQuery();
@@ -146,6 +154,16 @@ public class JdbcService {
         .select(scanAll.getProjections())
         .from(scanAll.forNamespace().get(), scanAll.forTable().get(), tableMetadata)
         .limit(scanAll.getLimit())
+        .build();
+  }
+
+  private SelectQuery buildSelectQueryForRelationalScan(ScanAll scan, TableMetadata tableMetadata) {
+    return queryBuilder
+        .select(scan.getProjections())
+        .from(scan.forNamespace().get(), scan.forTable().get(), tableMetadata)
+        .where(scan.getConjunctions())
+        .orderBy(scan.getOrderings())
+        .limit(scan.getLimit())
         .build();
   }
 
--- a/core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
@@ -62,11 +62,10 @@ public class SimpleSelectQuery implements SelectQuery {
   @Override
   public String sql() {
     StringBuilder builder =
-        new StringBuilder(
-            "SELECT "
-                + projectionSqlString()
-                + " FROM "
-                + rdbEngine.encloseFullTableName(schema, table));
+        new StringBuilder("SELECT ")
+            .append(projectionSqlString())
+            .append(" FROM ")
+            .append(rdbEngine.encloseFullTableName(schema, table));
     if (isRelationalQuery) {
       // for relational abstraction
       builder.append(relationalConditionSqlString());
--- a/core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
@@ -61,24 +61,26 @@ public class SimpleSelectQuery implements SelectQuery {
 
   @Override
   public String sql() {
-    String sql =
-        "SELECT "
-            + projectionSqlString()
-            + " FROM "
-            + rdbEngine.encloseFullTableName(schema, table);
+    StringBuilder builder =
+        new StringBuilder(
+            "SELECT "
+                + projectionSqlString()
+                + " FROM "
+                + rdbEngine.encloseFullTableName(schema, table));
     if (isRelationalQuery) {
       // for relational abstraction
-      sql += relationalConditionSqlString();
-      sql += relationalOrderBySqlString();
+      builder.append(relationalConditionSqlString());
+      builder.append(relationalOrderBySqlString());
     } else {
       // for multi-dimensional map abstraction
       if (isConditionalQuery) {
-        sql += " WHERE " + conditionSqlString();
+        builder.append(" WHERE ");
+        builder.append(conditionSqlString());
       }
-      sql += orderBySqlString();
+      builder.append(orderBySqlString());
     }
 
-    return sql;
+    return builder.toString();
   }
 
   private String projectionSqlString() {
@@ -108,19 +110,17 @@ public class SimpleSelectQuery implements SelectQuery {
       return "";
     }
 
-    List<String> conjunctionList = new ArrayList<>();
-    conjunctions.forEach(
-        conjunction -> {
-          List<String> conditions = new ArrayList<>();
-          conjunction
-              .getConditions()
-              .forEach(
-                  condition ->
-                      conditions.add(
-                          rdbEngine.enclose(condition.getColumn().getName())
-                              + convert(condition.getOperator())));
-          conjunctionList.add(String.join(" AND ", conditions));
-        });
+    List<String> conjunctionList =
+        conjunctions.stream()
+            .map(
+                conjunction ->
+                    conjunction.getConditions().stream()
+                        .map(
+                            condition ->
+                                rdbEngine.enclose(condition.getColumn().getName())
+                                    + convert(condition.getOperator()))
+                        .collect(Collectors.joining(" AND ")))
+            .collect(Collectors.toList());
 
     return " WHERE " + String.join(" OR ", conjunctionList);
   }
@@ -144,7 +144,7 @@ public class SimpleSelectQuery implements SelectQuery {
       case IS_NOT_NULL:
         return " IS NOT NULL";
       default:
-        throw new IllegalArgumentException("unknown operator");
+        throw new IllegalArgumentException("unknown operator: " + operator);
     }
   }
 
@@ -187,7 +187,7 @@ public class SimpleSelectQuery implements SelectQuery {
   }
 
   private String relationalOrderBySqlString() {
-    if (orderings.size() == 0) {
+    if (orderings.isEmpty()) {
       return "";
     }
 
@@ -236,8 +236,8 @@ public class SimpleSelectQuery implements SelectQuery {
       for (ConditionalExpression condition : conjunction.getConditions()) {
         if (!condition.getColumn().hasNullValue()) {
           condition.getColumn().accept(binder);
+          binder.throwSQLExceptionIfOccurred();
         }
-        binder.throwSQLExceptionIfOccurred();
       }
     }
   }
--- a/core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
@@ -1,6 +1,8 @@
 package com.scalar.db.storage.jdbc.query;
 
+import com.scalar.db.api.ConditionalExpression;
 import com.scalar.db.api.Scan;
+import com.scalar.db.api.Scan.Conjunction;
 import com.scalar.db.api.TableMetadata;
 import com.scalar.db.io.Column;
 import com.scalar.db.io.Key;
@@ -10,6 +12,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -32,6 +35,8 @@ public class SimpleSelectQuery implements SelectQuery {
   private final boolean isRangeQuery;
   private final Optional<String> indexedColumn;
   private final boolean isConditionalQuery;
+  private final boolean isRelationalQuery;
+  private final Set<Conjunction> conjunctions;
 
   SimpleSelectQuery(Builder builder) {
     projections = builder.projections;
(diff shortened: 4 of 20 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
```diff
@@ -88,10 +89,17 @@ public Scanner getScanner(Scan scan, Connection connection)
...
 
-    SelectQuery selectQuery =
-        scan instanceof ScanAll
-            ? buildSelectQueryForScanAll((ScanAll) scan, tableMetadata)
-            : buildSelectQueryForScan(scan, tableMetadata);
+    SelectQuery selectQuery;
+    if (scan instanceof ScanAll) {
+      ScanAll scanAll = (ScanAll) scan;
+      if (ScalarDbUtils.isRelational(scan)) {
+        selectQuery = buildSelectQueryForRelationalScan(scanAll, tableMetadata);
```
jnmt, line 96: One of the major changes is starting from here. Since we decide to use `ScanAll` for relational scan, we do not have to change the interface of `DistributedStorage`, but just have to change query building for conditions and orderings.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -73,6 +73,20 @@ private Optional<Result> createGetResult(Snapshot.Key key, List<String> projecti
...
   public List<Result> scan(Scan scan) throws CrudException {
+    List<Result> results = scanInternal(scan);
+
+    if (ScalarDbUtils.isRelational(scan)) {
+      // We verify if this scan does not overlap previous writes using the results obtained from
+      // the actual scan since the condition (i.e., where clause) is arbitrary in the relational
+      // scan, and thus, the write command may not have columns used in the condition, which are
+      // necessary to determine overlaps.
+      snapshot.verify(scan);
+    }
```
jnmt, line 84: The other major change is starting from here. This prevents scan-after-write with a bit different approach (i.e., use actual scan results) in existing scans since we need to handle the case that the user's `Put` does not have columns required to check if they match the `Scan` conditions.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -178,7 +181,9 @@ private TableMetadata getTableMetadata(Scan scan) throws ExecutionException {
   }
 
   public Optional<List<Key>> get(Scan scan) {
-    if (isWriteSetOverlappedWith(scan)) {
+    if (!ScalarDbUtils.isRelational(scan) && isWriteSetOverlappedWith(scan)) {
+      // Verify overlaps only for non-relational scan since we do it later for relational scan
+      // (right before returning the results).
       throw new IllegalArgumentException("reading already written data is not allowed");
     }
```
brfrn169, line 188: I think we can move this overlap check logic into the newly added `verify()` method for consistency and readability. This would execute this check logic after the actual read, but I don't think this is a problem. What do you think?

core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
```diff
@@ -88,10 +89,17 @@ public Scanner getScanner(Scan scan, Connection connection)
 
     TableMetadata tableMetadata = tableMetadataManager.getTableMetadata(scan);
 
-    SelectQuery selectQuery =
-        scan instanceof ScanAll
-            ? buildSelectQueryForScanAll((ScanAll) scan, tableMetadata)
-            : buildSelectQueryForScan(scan, tableMetadata);
+    SelectQuery selectQuery;
```
brfrn169, line 92: I think we need the same change for the `scan()` method to support the relational scan for JDBC transactions. Do you plan to do this in another PR?

brfrn169 (commented):

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: Let me confirm a bit. Do you mean that we can do the overlap check after the actual scan for both cases, regardless of whether it is relational or not? If yes, the answer is yes; we can do it without losing correctness. My only concern was unnecessary additional latency, but I think the readability would have more benefits here if we assume skilled users usually know it is prohibited and can avoid it.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
jnmt: Yes, I will add it in another PR, including tests for it.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: >  Do you mean that we can do the overlap check after the actual scan for both cases, regardless of whether it is relational or not?

Yes, correct.

> My only concern was unnecessary additional latency, but I think the readability would have more benefits here if we assume skilled users usually know it is prohibited and can avoid it.

The `IllegalArgumentException` issue is basically a bug, so maybe I don't think we need to worry about the additional latency. So in this case, I think we can prioritize the readability.

jnmt: OK, I will revise it. Thanks for the clarification and advice.

core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
```diff
@@ -220,6 +227,40 @@ private void checkOrderings(Scan scan, TableMetadata metadata) {
...
+    }
+  }
+
+  private void checkConjunctions(ScanAll scan, TableMetadata metadata) {
+    for (Conjunction conjunction : scan.getConjunctions()) {
+      for (ConditionalExpression condition : conjunction.getConditions()) {
+        boolean isValid;
+        if (condition.getOperator() == Operator.IS_NULL
+            || condition.getOperator() == Operator.IS_NOT_NULL) {
+          // the value must be null if the operator is 'is null' or `is not null`
```
komamitsu, line 248: Sorry, I don't get this. Why `ColumnChecker.requireNull` is used even for `IS_NOT_NULL`?

jnmt: It's just because we assume the conditional expression for `IS_NOT_NULL` is built with a null column, e.g., like [this](https://github.com/scalar-labs/scalardb/blob/e4ca22cd7acba6b0ba429a6f9c1ac27cfcdd339c/core/src/main/java/com/scalar/db/api/ConditionBuilder.java#L675-L677). I added this check referring to [this](https://github.com/scalar-labs/scalardb/blob/e4ca22cd7acba6b0ba429a6f9c1ac27cfcdd339c/core/src/main/java/com/scalar/db/common/checker/ConditionChecker.java#L69-L74). (But, I couldn't reuse itself since the check condition is a bit different.)

komamitsu: Ah, thank. I had a wrong idea. I thought the check was for filtering actual scanned value, but any scan hasn't started yet here. This check is kind of pre-checking if the conditions are valid or not, and actual filtering is performed in the underlying relational database (e.g, SimpleSelectQuery)

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -59,10 +66,17 @@ public String sql() {
             + projectionSqlString()
             + " FROM "
             + rdbEngine.encloseFullTableName(schema, table);
-    if (isConditionalQuery) {
-      sql += " WHERE " + conditionSqlString();
+    if (isRelationalQuery) {
+      // for relational abstraction
+      sql += relationalConditionSqlString();
```
komamitsu, line 71: How about using StringBuilder instead to avoid instantiating temporary objects? Using StringBuilder is basically better in terms of performance and memory footprint https://dev.to/this-is-learning/performance-benchmarking-string-and-string-builder-3bid

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -89,6 +103,51 @@ private String conditionSqlString() {
...
+      case GT:
+        return ">?";
+      case GTE:
+        return ">=?";
+      case LT:
+        return "<?";
+      case LTE:
+        return "<=?";
+      case IS_NULL:
+        return " IS NULL";
```
komamitsu, line 143: [minor] The above condition strings don't have leading space, but these 2 ones have. How about making the all strings consistent if possible?

jnmt: Thank you for the suggestion. For "is null" and "is not null", we cannot remove the space since they must be separated from the column name. In contrast, we end up creating inconsistency with other places like `conditionSqlString()` if we add spaces for other conditions. So, can we keep it as is?

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -89,6 +103,51 @@ private String conditionSqlString() {
     return String.join(" AND ", conditions);
   }
 
+  private String relationalConditionSqlString() {
+    if (conjunctions.isEmpty()) {
+      return "";
+    }
+
+    List<String> conjunctionList = new ArrayList<>();
```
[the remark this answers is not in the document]
jnmt: Cool, thank you. I don't think your idea loose readability very much. So, let's do it.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
komamitsu: Sounds reasonable! I think we can go with it as is.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -89,6 +103,51 @@ private String conditionSqlString() {
...
+      case LT:
+        return "<?";
+      case LTE:
+        return "<=?";
+      case IS_NULL:
+        return " IS NULL";
+      case IS_NOT_NULL:
+        return " IS NOT NULL";
+      default:
+        throw new IllegalArgumentException("unknown operator");
```
komamitsu, line 147: How about showing the unexpected operator in the error message? I don't think it would be useful for now but for the future when we add another operator.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -127,6 +186,17 @@ private String orderBySqlString() {
             .collect(Collectors.joining(","));
   }
 
+  private String relationalOrderBySqlString() {
+    if (orderings.size() == 0) {
```
komamitsu, line 190: [minor] I think `if (orderings.isEmpty())` would be slightly clearer.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -265,6 +277,68 @@ private boolean isWriteSetOverlappedWith(Scan scan) {
     return false;
   }
 
+  private boolean isWriteSetOverlappedWithRelational(Scan scan) {
+    for (Map.Entry<Key, Put> entry : writeSet.entrySet()) {
+      if (scanSet.get(scan).contains(entry.getKey())) {
+        return true;
+      }
+
+      // the following check prevents scan-after-write even if the update record potentially matches
```
komamitsu, line 286: Could you elaborate this? It would be great if you give me an example problematic case.

Torch3333 (commented):
It looks good thanks, I just have a minor comment.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: We need to consider three cases here to prevent scan-after-write.
1) A put operation overlaps the scan range regardless of the update (put) results.
2) A put operation does not overlap the scan range as a result of the update.
3) A put operation overlaps the scan range as a result of the update.

We can check cases 1 and 2 with L. 282-284. (Note that case 2 does not have the overlap, but we intentionally prohibit it due to the consistency and simplicity of snapshot management.) For case 3, we need to evaluate the operator to check for overlap since the put operation is only in the read-write set and not committed to the database. Let's see the following examples.

```sql
-- Assume the following table.
-- tbl (key int, val int)
-- 1 | 2 
-- 2 | 3

-- Case 2
update tbl set val = 4 where key = 1;
select * from tbl where val < 3;
-- We can find the overlap using the scan results.

-- Case 3
update tbl set val = 2 where key = 2;
select * from tbl where val < 3;
-- We cannot find the overlap since the database is not updated yet.
```

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -161,5 +231,14 @@ public void bind(PreparedStatement preparedStatement) throws SQLException {
       endColumn.get().accept(binder);
       binder.throwSQLExceptionIfOccurred();
     }
+
+    for (Conjunction conjunction : conjunctions) {
+      for (ConditionalExpression condition : conjunction.getConditions()) {
+        if (!condition.getColumn().hasNullValue()) {
+          condition.getColumn().accept(binder);
+        }
+        binder.throwSQLExceptionIfOccurred();
```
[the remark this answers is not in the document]
jnmt: Good catch, thank you! As you mentioned, it should be inside the block as other similar parts do.

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
```diff
@@ -63,6 +64,21 @@ public Cassandra(DatabaseConfig config) {
     operationChecker = new OperationChecker(metadataManager);
   }
 
+  Cassandra(
```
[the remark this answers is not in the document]
jnmt: Thank you for the suggestion. Fixed in 31d2d2b.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: Fixed in 31d2d2b.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
jnmt: Fixed in 1490ed6.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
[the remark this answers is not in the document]
jnmt: Fixed in 1490ed6.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
jnmt: Fixed in 1490ed6.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
jnmt: Fixed in 1490ed6.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -265,6 +277,68 @@ private boolean isWriteSetOverlappedWith(Scan scan) {
...
+      // the scan that can be overlooked above check.
+      Put put = entry.getValue();
+      if (!put.forNamespace().equals(scan.forNamespace())
+          || !put.forTable().equals(scan.forTable())) {
+        continue;
+      }
+
+      if (scan.getConjunctions().size() == 0) {
```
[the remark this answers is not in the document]
jnmt: Fixed in 1490ed6.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -265,6 +277,68 @@ private boolean isWriteSetOverlappedWith(Scan scan) {
...
+      case GT:
+        return column.compareTo((Column<T>) condition.getColumn()) > 0;
+      case GTE:
+        return column.compareTo((Column<T>) condition.getColumn()) >= 0;
+      case LT:
+        return column.compareTo((Column<T>) condition.getColumn()) < 0;
+      case LTE:
+        return column.compareTo((Column<T>) condition.getColumn()) <= 0;
+      default:
+        throw new IllegalArgumentException("unknown operator");
```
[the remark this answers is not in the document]
jnmt: Fixed in 1490ed6.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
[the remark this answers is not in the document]
jnmt: Fixed in 1490ed6.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
komamitsu: Thanks a lot for the explanation!!

komamitsu (approved):
LGTM! 👍

jnmt:
@brfrn169 @komamitsu @Torch3333 Thank you for your review. I fixed some issues based on your comment. PTAL when you have time.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
komamitsu: @jnmt This is just a suggestion and doesn't prevent this PR from being merged. The above explanation was really helpful for me to understand why we need to consider updated values. How about putting it as a comment so that new developers and 2y later me who has a bad memory can maintain the code?

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -50,21 +55,32 @@ public class SimpleSelectQuery implements SelectQuery {
...
-            + " FROM "
-            + rdbEngine.encloseFullTableName(schema, table);
-    if (isConditionalQuery) {
-      sql += " WHERE " + conditionSqlString();
+    StringBuilder builder =
+        new StringBuilder(
+            "SELECT "
+                + projectionSqlString()
+                + " FROM "
+                + rdbEngine.encloseFullTableName(schema, table));
```
Torch3333, line 69:
```suggestion
        new StringBuilder( "SELECT ")
        .append(projectionSqlString())
        .append(" FROM ")
        .append(rdbEngine.encloseFullTableName(schema, table));
```
We could use the builder for this concatenation.

komamitsu: JFYI : String concatenation in a single line is implicitly converted into StringBuilder as a part of optimization in recent Java versions.

```java
public class Test {
  public static void main(String[] args) {
    String s = "AAAA " + System.currentTimeMillis() + " BBBB";
  }
}
```

For instance, this code is compiled into the following class file

```
$ javap -c Test
Warning: Binary file Test contains org.example.Test
Compiled from "Test.java"
public class org.example.Test {
  public org.example.Test();
    Code:
       0: aload_0
       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
       4: return

  public static void main(java.lang.String[]);
    Code:
       0: new           #2                  // class java/lang/StringBuilder
       3: dup
       4: invokespecial #3                  // Method java/lang/StringBuilder."<init>":()V
       7: ldc           #4                  // String AAAA
       9: invokevirtual #5                  // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      12: invokestatic  #6                  // Method java/lang/System.currentTimeMillis:()J
      15: invokevirtual #7                  // Method java/lang/StringBuilder.append:(J)Ljava/lang/StringBuilder;
      18: ldc           #8                  // String  BBBB
      20: invokevirtual #5                  // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      23: invokevirtual #9                  // Method java/lang/StringBuilder.toString:()Ljava/lang/String;
      26: astore_1
      27: return
}
```

And you can see the optimization. But it's not easy to always confirm the optimization is performed, so explicitly using StringBuilder may be a good option.

Torch3333: I see. Thank you for going into detail about this. I did not know about this tool, it is quite handy.
 
My concern was that with an example that somewhat mirrors the current code.
```java
 public static void main(String[] args) {
    StringBuilder builder = new StringBuilder("AAAA " + System.currentTimeMillis() + " BBBB");
    if (System.currentTimeMillis() % 2 == 0) {
      builder.append("CCCC");
    } else {
      builder.append("DDDD");
    }
    String s = builder.toString();
  }
```
A first StringBuilder is created to concatenate `"AAAA "+ System.currentTimeMillis()+" BBBB"` and then a second is created to append the string in the `if` condition instead of reusing the first one. 
```
public class com.scalar.db.api.StrBuild {
  public com.scalar.db.api.StrBuild();
    Code:
       0: aload_0
       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
       4: return

  public static void main(java.lang.String[]);
    Code:
       0: new           #2                  // class java/lang/StringBuilder
       3: dup
       4: new           #2                  // class java/lang/StringBuilder
       7: dup
       8: invokespecial #3                  // Method java/lang/StringBuilder."<init>":()V
      11: ldc           #4                  // String AAAA
      13: invokevirtual #5                  // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      16: invokestatic  #6                  // Method java/lang/System.currentTimeMillis:()J
      19: invokevirtual #7                  // Method java/lang/StringBuilder.append:(J)Ljava/lang/StringBuilder;
      22: ldc           #8                  // String  BBBB
      24: invokevirtual #5                  // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      27: invokevirtual #9                  // Method java/lang/StringBuilder.toString:()Ljava/lang/String;
      30: invokespecial #10                 // Method java/lang/StringBuilder."<init>":(Ljava/lang/String;)V
      33: astore_1
      34: invokestatic  #6                  // Method java/lang/System.currentTimeMillis:()J
      37: ldc2_w        #11                 // long 2l
      40: lrem
      41: lconst_0
      42: lcmp
      43: ifne          56
      46: aload_1
      47: ldc           #13                 // String CCCC
      49: invokevirtual #5                  // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      52: pop
      53: goto          63
      56: aload_1
      57: ldc           #14                 // String DDDD
      59: invokevirtual #5                  // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      62: pop
      63: aload_1
      64: invokevirtual #9                  // Method java/lang/StringBuilder.toString:()Ljava/lang/String;
      67: astore_2
      68: return
}
```
As you say, a more recent compiler might optimize that case better, but since we refactored to use a StringBuilder on purpose in this method, we could use it for all of the concatenations.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: Yeah, I guess I will also forget it. It's a bit too much, but I added the detailed comments in 4b2fbe5.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
jnmt: Thank you for the information. I agree it's better to use append here. Fixed in 4b2fbe5.

jnmt:
SQLite integration tests sometimes failed because I didn't add ordering, and there was no guarantee to preserve the expected order. I fixed it in 7163241 and b9ea15e. Also, I noticed there is a bug in the builder part. I will fix it in another PR.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
komamitsu: 💯

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169, line 188: Since it's moved to `verify()`, we can delete this, right?

brfrn169 (approved):
Left one comment. Other than that, LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -178,15 +181,19 @@ private TableMetadata getTableMetadata(Scan scan) throws ExecutionException {
...
-    }
     if (scanSet.containsKey(scan)) {
       return Optional.ofNullable(scanSet.get(scan));
     }
     return Optional.empty();
   }
 
+  public void verify(Scan scan) {
+    if ((ScalarDbUtils.isRelational(scan) && isWriteSetOverlappedWithRelational(scan))
+        || (!ScalarDbUtils.isRelational(scan) && isWriteSetOverlappedWith(scan))) {
```
brfrn169, line 192: Sorry, one more point. To avoid duplicate calculations, I think we should do the following:
```suggestion
    boolean isRelational = ScalarDbUtils.isRelational(scan);
    if ((isRelational && isWriteSetOverlappedWithRelational(scan))
        || (!isRelational) && isWriteSetOverlappedWith(scan))) {
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: Sorry for that. I removed it (and fix some related tests) in 2bdbae4.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: Right, I will fix it. Thank you!

jnmt: @brfrn169 Fixed in 0c90275. PTAL if you have time.

feeblefakie (commented):
Very sorry for the late review.
I left several questions for clarification. PTAL!

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -89,6 +104,49 @@ private String conditionSqlString() {
...
+                conjunction ->
+                    conjunction.getConditions().stream()
+                        .map(
+                            condition ->
+                                rdbEngine.enclose(condition.getColumn().getName())
+                                    + convert(condition.getOperator()))
+                        .collect(Collectors.joining(" AND ")))
+            .collect(Collectors.toList());
+
+    return " WHERE " + String.join(" OR ", conjunctionList);
```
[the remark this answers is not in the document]
jnmt: We are using DNF for the internal representation and building queries. So, it should be fine here.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -73,6 +73,20 @@ private Optional<Result> createGetResult(Snapshot.Key key, List<String> projecti
   }
 
   public List<Result> scan(Scan scan) throws CrudException {
+    List<Result> results = scanInternal(scan);
+
+    // We verify if this scan does not overlap previous writes after the actual scan. For a
+    // relational scan, this must be done here, using the obtained results. This is because the
```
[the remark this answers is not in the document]
jnmt: > Or you meant scanSet is updated by scanInternal and is used in the check logic (isWriteSetOverlappedWithRelational)?

Right, we use the obtained results put in the `scanSet` by `scanInternal`.

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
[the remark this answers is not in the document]
jnmt: For clarification, I added some comments for the `Scan` class in 19f4ada.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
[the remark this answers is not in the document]
jnmt: "using the obtained results" was a bit confusing since the check only uses keys in the scan set (not returned results above). I changed the comment a little in 19f4ada.

feeblefakie (approved):
LGTM! Thank you!

PR #901, opened 2023-06-07
Add a redirect from the docs site hosted in this repository to the new docs site

josh-wong:
This PR adds files related to Jekyll (static site generator) to add the following functionality:

- Automatically redirect visitors from the previous ScalarDB docs site (hosted in this repository) to the [new ScalarDB docs site](https://scalardb.scalar-labs.com/docs).
- Track visits to the previous ScalarDB docs site (hosted in this repository) so that we can decommission it in the future (after confirming minimal or no impact on visitors).

I have tested the redirect (with different URLs) in my personal repository. In my test, I confirmed the redirect works as expected. Regardless of the page the visitor lands on on the previous docs site, they will be redirected to the new ScalarDB docs home page.


Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

brfrn169:
@josh-wong I have a question. Users accessing any file in the `docs` directory are all redirected to the top page of the new ScalarDB docs site, right? If so, I feel that may create a poor user experience.

josh-wong:
> @josh-wong I have a question. Users accessing any file in the `docs` directory are all redirected to the top page of the new ScalarDB docs site, right? If so, I feel that may create a poor user experience.

@brfrn169 In this case, yes. We might be able to redirect visitors to the page they expect to land on on the new docs site, but I believe doing so would require us to add front matter to each of the Markdown files. Let me do a bit more research to see if and how this is possible.

brfrn169:
> > @josh-wong I have a question. Users accessing any file in the `docs` directory are all redirected to the top page of the new ScalarDB docs site, right? If so, I feel that may create a poor user experience.
> 
> @brfrn169 In this case, yes. We might be able to redirect visitors to the page they expect to land on on the new docs site, but I believe doing so would require us to add front matter to each of the Markdown files. Let me do a bit more research to see if and how this is possible.

Thank you! If it's hard to implement, I'm fine with that.

_config.yml
```diff
@@ -0,0 +1,9 @@
+# Note: This file is used only for redirecting visitors from the old ScalarDB docs site hosted in this repository to the new ScalarDB docs site. This file can be deleted after the old ScalarDB docs site hosted in this repository is no longer deployed.
+google_analytics: "G-Q4TKS77KCP"
+
+defaults:
+  - scope:
+      path: "docs" # Specifies where the docs are located.
+      type: "default" # Defines the type for docs.
+    values:
+      layout: default # Defines the template type used for docs (templates are located in the "_layouts" folder).
```
komamitsu, line 9: This mark means "No newline at end of file." It's recommended to add a newline at the end of file https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

josh-wong: Thank you for catching that! I've added a newline and mentioned it in the style guide I am working on.

komamitsu (approved):
LGTM! 👍

josh-wong:
> > > @josh-wong I have a question. Users accessing any file in the `docs` directory are all redirected to the top page of the new ScalarDB docs site, right? If so, I feel that may create a poor user experience.
> > 
> > 
> > @brfrn169 In this case, yes. We might be able to redirect visitors to the page they expect to land on on the new docs site, but I believe doing so would require us to add front matter to each of the Markdown files. Let me do a bit more research to see if and how this is possible.
> 
> Thank you! If it's hard to implement, I'm fine with that.

After finding a way to do redirect page by page, I don't think the method we have available to achieve this is beneficial.

1) We would need to manually add each old page to a list and then specify which new page the visitor should be directed to. For example:

```yml
"old/page1" : /new/page1
"old/page2" : /new/page2
"old/page3" : /new/page3
```

This would be a fairly time-consuming process and leave significant room for error.

2) This method of redirecting requires manually building the site and then deploying the site from the `site` folder instead of `root`. We would need to change where the site is deployed from in the GitHub, but this may cause problems for some sites that direct users to source code deployed through GitHub Pages. 

Because of the above two points, I think we should avoid implementing a page-by-page redirect when redirecting from the old docs site to the new docs site.

brfrn169:
@josh-wong Thank you for the investigation! Understood. So let’s proceed as it is. Thanks.

brfrn169 (approved):
LGTM! Thank you!

PR #906, opened 2023-06-14
Fix the wrong argument validations in Auto PR workflow

komamitsu:
### Context

https://github.com/scalar-labs/scalardb/actions/runs/5262633849/jobs/9512108324 failed due to the argument validation error of `create_pull_requests` script. But the number of parameters passed to the script was 10, which is enough for the validation. The cause was the script compares "10" and "5" as strings.

### Changes

This PR revises the validation to treat the values as integers.

### DIFF
--- a/ci/auto-pr/create_pull_requests
+++ b/ci/auto-pr/create_pull_requests
@@ -2,7 +2,7 @@
 
 set -e -o pipefail; [[ -n "$DEBUG" ]] && set -x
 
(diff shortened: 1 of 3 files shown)


Torch3333 (approved):
LGTM, thank you!

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #919, opened 2023-06-28
Should call abort() when aborting transaction in transaction wrapper classes

brfrn169:
Currently, some transaction wrapper classes do not implement the `abort()` method. As a result, when a transaction is aborted, the `abort()` method of the wrapped transaction class is not called; instead, the `rollback()` method is called. To address this issue, this PR implements the `abort()` method in the transaction wrapper classes.

Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionManager.java
@@ -10,6 +10,7 @@ import com.scalar.db.api.Put;
 import com.scalar.db.api.Result;
 import com.scalar.db.api.Scan;
 import com.scalar.db.config.DatabaseConfig;
+import com.scalar.db.exception.transaction.AbortException;
 import com.scalar.db.exception.transaction.CommitException;
 import com.scalar.db.exception.transaction.CrudException;
 import com.scalar.db.exception.transaction.RollbackException;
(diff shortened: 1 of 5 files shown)


komamitsu (commented):
The changes look good to me! But no unit tests are needed?

brfrn169:
> The changes look good to me! But no unit tests are needed?

@komamitsu Ideally, yes, we should add unit tests. But for now, we don't have the unit test code of `ActiveTransactionManagedDistributedTransactionManager` and `ActiveTransactionManagedTwoPhaseCommitTransactionManager` actually.

core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionManager.java
```diff
@@ -183,6 +184,18 @@ public void rollback() throws RollbackException {
       }
     }
 
+    @Override
+    public void abort() throws AbortException {
+      if (status == Status.COMMITTED || status == Status.ROLLED_BACK) {
+        throw new IllegalStateException("The transaction has already been committed or aborted");
```
komamitsu, line 190: Just noticed users would want to know the actual state in the log message?

brfrn169: That's correct. Since we need to improve the messages in other parts, let's do that in another PR. Thank for pointing it out!

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

brfrn169:
@Torch3333 It's merged but please take a look when you get a chance!

Torch3333 (commented):
LGTM, thank you!

PR #922, opened 2023-06-29
Backport to branch(3.8) : Should call abort() when aborting transaction in transaction wrapper classes

feeblefakie:
Backport of https://github.com/scalar-labs/scalardb/pull/919


komamitsu:
Current git-cherry-pick based PR creation basically doesn't work expectedly...

komamitsu:
> Current git-cherry-pick based PR creation basically doesn't work expectedly...

Ah, maybe no in this case. This seems to have come from the previous error of creating a PR for `3.7`. I'll fix it.

PR #930, opened 2023-07-05
Revise Javadoc for exceptions

brfrn169:
This PR revises the Javadoc for the exceptions to make them more precise. Please take a look!

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedTransaction.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedTransaction.java
@@ -72,8 +72,8 @@ public interface DistributedTransaction extends TransactionCrudOperable {
    * @throws CommitConflictException if the transaction fails to commit due to transient faults
    *     (e.g., a conflict error). You can retry the transaction from the beginning
    * @throws CommitException if the transaction fails to commit due to transient or nontransient
-   *     faults. You can try retrying the transaction from the beginning, but you may not be able to
-   *     retry the transaction due to nontransient faults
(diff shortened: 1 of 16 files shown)


core/src/main/java/com/scalar/db/api/DistributedTransaction.java
```diff
@@ -69,24 +69,28 @@ public interface DistributedTransaction extends TransactionCrudOperable {
...
    * Commits a transaction.
    *
-   * @throws CommitConflictException if a transaction conflict occurs, You can retry the transaction
-   *     from the beginning in this case
-   * @throws CommitException if the operation fails
+   * @throws CommitConflictException if the transaction fails to commit due to transient faults
+   *     (e.g., a conflict error). You can retry the transaction from the beginning
+   * @throws CommitException if the transaction fails to commit due to transient or nontransient
+   *     faults. You can try retrying the transaction from the beginning, but you may not be able to
+   *     retry the transaction due to nontransient faults
```
komamitsu, line 76: > but you may not be able to retry the transaction

It sounds to me users can't even retry the transaction, but I think you mean a retry may fail.

How about `but the transaction may still fail if the cause is nontranient` or something?

brfrn169: Thank you! That's correct.

Fixed in 3ceb430edf10a8abf2309ae0c64c9bc3b390b86c. Please take a look when you have time!

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

PR #931, opened 2023-07-06
Add table importing feature

jnmt:
This PR adds a table-importing feature to enable users to import their existing databases to ScalarDB. Sorry for the big PR.

You can import your existing databases using schema-loader with the `--import` option. For the `--import` option, you need to specify a schema JSON file with `-f`. The grammar of the schema file is basically the same as the usual loading, but it only requires a namespace, table, and transactional flag. Other items are ignored, and you don't have to specify columns since the import function automatically reads table metadata and recognizes columns.

Note that we do not support SQLite at this moment.  Rough policies of supported data types are as follows.

- We do not support data types that are larger or have higher precision than any of the ScalarDB data types.
  - The only exception is BIGINT. Our BIGINT is a little smaller than a regular 8-byte BIGINT due to the DynamoDB support, but I left BIGINT supported with a warning because we plan to remove this limitation (w/o DynamoDB).
- We notify users when we use a ScalarDB data type for an underlying data type that is smaller or has lower precision than ScalarDB's one.
- NUMERIC and DECIMAL are currently not supported.
- See inline comments for all supported (mapped data types) and unsupported data types, respectively.

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
@@ -1,0 +1,426 @@
+package com.scalar.db.storage.jdbc;
+
+import com.scalar.db.api.TableMetadata;
+import com.scalar.db.config.DatabaseConfig;
+import com.scalar.db.io.DataType;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.stream.Collectors;
+import org.apache.commons.dbcp2.BasicDataSource;
+
+public class JdbcAdminImportTestUtils {
+  static final List<String> UNSUPPORTED_DATA_TYPES_MYSQL =
+      Arrays.asList(
+          "BIGINT UNSIGNED",
+          "BIT(8)",
+          "DATE",
+          "DATETIME",
+          "DECIMAL(8,2)",
+          "ENUM('a','b')",
+          "SET('a','b')",
+          "GEOMETRY",
+          "JSON",
+          "NUMERIC",
+          "TIME",
+          "TIMESTAMP",
+          "YEAR");
+  static final List<String> UNSUPPORTED_DATA_TYPES_PGSQL =
+      Arrays.asList(
+          "bigserial",
+          "bit(8)",
+          "bit varying(8)",
+          "box",
+          "cidr",
+          "circle",
+          "date",
+          "inet",
+          "interval",
+          "json",
+          "jsonb",
+          "line",
+          "lseg",
+          "macaddr",
+          "macaddr8",
+          "money",
+          "numeric(8,2)",
+          "path",
+          "pg_lsn",
+          "pg_snapshot", // after v14
+          "point",
+          "polygon",
+          "smallserial",
+          "time",
+          "time with time zone",
+          "timestamp",
+          "timestamp with time zone",
+          "tsquery",
+          "tsvector",
+          "txid_snapshot",
+          "uuid",
+          "xml");
+  static final List<String> UNSUPPORTED_DATA_TYPES_ORACLE =
+      Arrays.asList(
+          "BFILE",
+          "DATE",
+          "FLOAT(54)",
+          "INT",
+          "INTERVAL YEAR(3) TO MONTH",
+          "INTERVAL DAY(2) TO SECOND",
+          "JSON",
+          "NUMBER(16,0)",
+          "ROWID",
+          "TIMESTAMP",
+          "TIMESTAMP WITH TIME ZONE",
+          "TIMESTAMP WITH LOCAL TIME ZONE",
+          "UROWID");
+  static final List<String> UNSUPPORTED_DATA_TYPES_MSSQL =
+      Arrays.asList(
+          "date",
+          "datetime",
+          "datetime2",
+          "datetimeoffset",
+          "decimal(8,2)",
+          "hierarchyid",
+          "money",
+          "numeric(8,2)",
+          "rowversion",
+          "smalldatetime",
+          "smallmoney",
+          "sql_variant",
+          "time",
+          "uniqueidentifier",
+          "xml");
+
+  private final JdbcConfig config;
+  private final RdbEngineStrategy rdbEngine;
+
+  public JdbcAdminImportTestUtils(Properties properties) {
+    config = new JdbcConfig(new DatabaseConfig(properties));
+    rdbEngine = RdbEngineFactory.create(config);
+  }
+
+  public Map<String, TableMetadata> createExistingDatabaseWithAllDataTypes(String namespace)
+      throws SQLException {
+    Map<String, TableMetadata> results = new HashMap<>();
+    List<String> sqls = new ArrayList<>();
+    LinkedHashMap<String, String> goodTableColumns;
+    TableMetadata goodTableMetadata;
+    Map<String, String> badTables;
+    if (rdbEngine instanceof RdbEngineMysql) {
+      goodTableColumns = prepareColumnsForMysql();
+      goodTableMetadata = prepareTableMetadataForMysql();
+      badTables = prepareCreateNonImportableTableSql(namespace, UNSUPPORTED_DATA_TYPES_MYSQL);
+    } else if (rdbEngine instanceof RdbEnginePostgresql) {
+      goodTableColumns = prepareColumnsForPostgresql();
+      goodTableMetadata = prepareTableMetadataForPostgresql();
+      badTables = prepareCreateNonImportableTableSql(namespace, UNSUPPORTED_DATA_TYPES_PGSQL);
+    } else if (rdbEngine instanceof RdbEngineOracle) {
+      goodTableColumns = prepareColumnsForOracle();
+      goodTableMetadata = prepareTableMetadataForOracle();
+      badTables = prepareCreateNonImportableTableSql(namespace, UNSUPPORTED_DATA_TYPES_ORACLE);
+
+      // LONG columns must be tested with separated tables since they cannot be coexisted
+      TableMetadata longRawMetadata = prepareTableMetadataForOracleForLongRaw();
+      sqls.add(
+          prepareCreateTableSql(
+              namespace,
+              "good_table_long_raw",
+              prepareColumnsForOracleLongRaw(),
+              longRawMetadata.getPartitionKeyNames()));
+      results.put("good_table_long_raw", longRawMetadata);
+    } else if (rdbEngine instanceof RdbEngineSqlServer) {
+      goodTableColumns = prepareColumnsForSqlServer();
+      goodTableMetadata = prepareTableMetadataForSqlServer();
+      badTables = prepareCreateNonImportableTableSql(namespace, UNSUPPORTED_DATA_TYPES_MSSQL);
+    } else {
+      throw new RuntimeException();
+    }
+
+    // table with all supported columns
+    sqls.add(
+        prepareCreateTableSql(
+            namespace, "good_table", goodTableColumns, goodTableMetadata.getPartitionKeyNames()));
+    results.put("good_table", goodTableMetadata);
+
+    // tables with an unsupported column
+    badTables.forEach(
+        (table, sql) -> {
+          sqls.add(sql);
+          results.put(table, null);
+        });
+
+    createExistingDatabase(namespace);
+    execute(sqls.toArray(new String[0]));
+    return results;
+  }
+
+  public void createExistingDatabase(String namespace) throws SQLException {
+    execute(rdbEngine.createNamespaceSqls(rdbEngine.enclose(namespace)));
+  }
+
+  public void dropTable(String namespace, String table) throws SQLException {
+    String dropTable = "DROP TABLE " + rdbEngine.encloseFullTableName(namespace, table);
+    execute(dropTable);
+  }
+
+  public void execute(String sql) throws SQLException {
+    try (BasicDataSource dataSource = JdbcUtils.initDataSourceForAdmin(config, rdbEngine);
+        Connection connection = dataSource.getConnection()) {
+      JdbcAdmin.execute(connection, sql);
+    }
+  }
+
+  public void execute(String[] sql) throws SQLException {
+    try (BasicDataSource dataSource = JdbcUtils.initDataSourceForAdmin(config, rdbEngine);
+        Connection connection = dataSource.getConnection()) {
+      JdbcAdmin.execute(connection, sql);
+    }
+  }
+
+  private LinkedHashMap<String, String> prepareColumnsForMysql() {
+    LinkedHashMap<String, String> columns = new LinkedHashMap<>();
+    columns.put("pk1", "INT");
+    columns.put("pk2", "INT");
+    columns.put("col01", "BOOLEAN");
+    columns.put("col02", "INT");
+    columns.put("col03", "INT UNSIGNED");
+    columns.put("col04", "TINYINT");
+    columns.put("col05", "SMALLINT");
+    columns.put("col06", "MEDIUMINT");
+    columns.put("col07", "BIGINT");
+    columns.put("col08", "FLOAT");
+    columns.put("col09", "DOUBLE");
+    columns.put("col10", "CHAR(8)");
+    columns.put("col11", "VARCHAR(512)");
+    columns.put("col12", "TEXT");
+    columns.put("col13", "TINYTEXT");
+    columns.put("col14", "MEDIUMTEXT");
+    columns.put("col15", "LONGTEXT");
+    columns.put("col16", "VARBINARY(1024)");
+    columns.put("col17", "BLOB");
+    columns.put("col18", "TINYBLOB");
+    columns.put("col19", "MEDIUMBLOB");
+    columns.put("col20", "LONGBLOB");
+    columns.put("col21", "BINARY(255)");
+    return columns;
+  }
+
+  private TableMetadata prepareTableMetadataForMysql() {
+    return TableMetadata.newBuilder()
+        .addColumn("pk1", DataType.INT)
+        .addColumn("pk2", DataType.INT)
+        .addColumn("col01", DataType.BOOLEAN)
+        .addColumn("col02", DataType.INT)
+        .addColumn("col03", DataType.BIGINT)
+        .addColumn("col04", DataType.INT)
+        .addColumn("col05", DataType.INT)
+        .addColumn("col06", DataType.INT)
+        .addColumn("col07", DataType.BIGINT)
+        .addColumn("col08", DataType.FLOAT)
+        .addColumn("col09", DataType.DOUBLE)
+        .addColumn("col10", DataType.TEXT)
+        .addColumn("col11", DataType.TEXT)
+        .addColumn("col12", DataType.TEXT)
+        .addColumn("col13", DataType.TEXT)
+        .addColumn("col14", DataType.TEXT)
+        .addColumn("col15", DataType.TEXT)
+        .addColumn("col16", DataType.BLOB)
+        .addColumn("col17", DataType.BLOB)
+        .addColumn("col18", DataType.BLOB)
+        .addColumn("col19", DataType.BLOB)
+        .addColumn("col20", DataType.BLOB)
+        .addColumn("col21", DataType.BLOB)
+        .addPartitionKey("pk1")
+        .addPartitionKey("pk2")
+        .build();
+  }
+
+  private LinkedHashMap<String, String> prepareColumnsForPostgresql() {
+    LinkedHashMap<String, String> columns = new LinkedHashMap<>();
+    columns.put("pk1", "integer");
+    columns.put("pk2", "integer");
+    columns.put("col01", "boolean");
+    columns.put("col02", "smallint");
+    columns.put("col03", "integer");
+    columns.put("col04", "bigint");
+    columns.put("col05", "real");
+    columns.put("col06", "double precision");
+    columns.put("col07", "char(8)");
+    columns.put("col08", "varchar(512)");
+    columns.put("col09", "text");
+    columns.put("col10", "bytea");
+    return columns;
+  }
+
+  private TableMetadata prepareTableMetadataForPostgresql() {
+    return TableMetadata.newBuilder()
+        .addColumn("pk1", DataType.INT)
+        .addColumn("pk2", DataType.INT)
+        .addColumn("col01", DataType.BOOLEAN)
+        .addColumn("col02", DataType.INT)
+        .addColumn("col03", DataType.INT)
+        .addColumn("col04", DataType.BIGINT)
+        .addColumn("col05", DataType.FLOAT)
+        .addColumn("col06", DataType.DOUBLE)
+        .addColumn("col07", DataType.TEXT)
+        .addColumn("col08", DataType.TEXT)
+        .addColumn("col09", DataType.TEXT)
+        .addColumn("col10", DataType.BLOB)
+        .addPartitionKey("pk1")
+        .addPartitionKey("pk2")
+        .build();
+  }
+
+  private LinkedHashMap<String, String> prepareColumnsForOracle() {
+    LinkedHashMap<String, String> columns = new LinkedHashMap<>();
+    columns.put("pk1", "CHAR(8)");
+    columns.put("pk2", "CHAR(8)");
+    columns.put("col01", "NUMERIC(15,0)");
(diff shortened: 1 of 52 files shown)


core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -0,0 +1,442 @@
...
+  }
+
+  private TableMetadata prepareTableMetadataForMysql() {
+    TableMetadata.Builder builder =
+        TableMetadata.newBuilder()
+            .addColumn("pk1", DataType.INT)
+            .addColumn("pk2", DataType.INT)
+            .addColumn("col01", DataType.BOOLEAN)
+            .addColumn("col02", DataType.INT)
+            .addColumn("col03", DataType.BIGINT)
```
jnmt, line 236: Refer to the original data types above and how they are expected to be mapped to ScalarDB data types. We probably need a document for supported and unsupported data types.

schema-loader/src/test/java/com/scalar/db/schemaloader/ImportSchemaParserTest.java
```diff
@@ -0,0 +1,68 @@
...
+      throws SchemaLoaderException {
+    // Arrange
+    String serializedSchemaJson =
+        "{"
+            + "  \"sample_db.sample_table1\": {"
+            + "    \"transaction\": true"
+            + "  },"
+            + "  \"sample_db.sample_table2\": {"
+            + "    \"transaction\": false"
+            + "  },"
```
jnmt, line 21: This is an example of a schema JSON file for the table import.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
```diff
@@ -188,6 +193,33 @@ public void addNewColumnToTable(
     admin.addNewColumnToTable(namespace, table, beforeColumnName, columnType);
   }
 
+  @Override
+  public void importTable(String namespace, String table) throws ExecutionException {
```
jnmt, line 197: This is an entry point for a major change in this PR. It first gets JDBC metadata and creates table metadata for ScalarDB in `getImportTableMetadata()`. Then, it adds transaction metadata columns. Finally, it constructs correct ScalarDB metadata using the repair function.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -429,6 +441,62 @@ public TableMetadata getTableMetadata(String namespace, String table) throws Exe
...
+        primaryKeyExists = true;
+        String columnName = resultSet.getString(JDBC_COL_COLUMN_NAME);
+        builder.addPartitionKey(columnName);
+      }
+
+      if (!primaryKeyExists) {
+        throw new IllegalStateException("The table must have a primary key");
+      }
+
+      resultSet = metadata.getColumns(null, namespace, table, "%");
```
komamitsu, line 473: Just a quick question. Clustering keys aren't automatically created in this import feature, right?

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
```diff
@@ -249,4 +250,23 @@ public static boolean isRelational(Scan scan) {
     return scan instanceof ScanAll
         && (!scan.getOrderings().isEmpty() || !scan.getConjunctions().isEmpty());
   }
+
+  public static JDBCType getJdbcType(int sqlTypes) {
+    JDBCType type;
+    switch (sqlTypes) {
+      case 100:
```
komamitsu, line 257: 100 and 101 are BINARY_FLOAT and BINARY_DOUBLE, respectively? How about adding comments about that?

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
```diff
@@ -249,4 +250,23 @@ public static boolean isRelational(Scan scan) {
     return scan instanceof ScanAll
         && (!scan.getOrderings().isEmpty() || !scan.getConjunctions().isEmpty());
   }
+
+  public static JDBCType getJdbcType(int sqlTypes) {
```
komamitsu, line 254: > int sqlTypes

`sqlType` is better?

schema-loader/src/main/java/com/scalar/db/schemaloader/ImportSchemaParser.java
```diff
@@ -0,0 +1,44 @@
...
+  public ImportSchemaParser(String serializedSchemaJson) throws SchemaLoaderException {
+    try {
+      schemaJson = JsonParser.parseString(serializedSchemaJson).getAsJsonObject();
+    } catch (JsonParseException e) {
+      throw new SchemaLoaderException("Parsing the schema JSON failed", e);
+    }
+  }
+
+  public List<ImportTableSchema> parse() throws SchemaLoaderException {
+    ArrayList<ImportTableSchema> tableSchemaList = new ArrayList<>();
```
komamitsu, line 37: [trivial] We don't need to use the concrete class type `ArrayList` as only `List#add` is used here.

```
    List<ImportTableSchema> tableSchemaList = new ArrayList<>();
```

schema-loader/src/main/java/com/scalar/db/schemaloader/ImportTableSchema.java
```diff
@@ -0,0 +1,40 @@
...
+  private final String namespace;
+  private final String tableName;
+  private final boolean isTransactionTable;
+
+  public ImportTableSchema(String tableFullName, JsonObject tableDefinition)
+      throws SchemaLoaderException {
+    String[] fullName = tableFullName.split("\\.", -1);
+    if (fullName.length < 2) {
+      throw new SchemaLoaderException(
+          "Parsing the schema JSON failed. Table full name must contains table name and namespace");
```
komamitsu, line 18: [minor] How about adding `tableFullName` in this error message?

schema-loader/src/main/java/com/scalar/db/schemaloader/ImportTableSchema.java
```diff
@@ -0,0 +1,40 @@
...
+public class ImportTableSchema {
+  private static final String TRANSACTION = "transaction";
+  private final String namespace;
+  private final String tableName;
+  private final boolean isTransactionTable;
+
+  public ImportTableSchema(String tableFullName, JsonObject tableDefinition)
+      throws SchemaLoaderException {
+    String[] fullName = tableFullName.split("\\.", -1);
+    if (fullName.length < 2) {
```
komamitsu, line 16: Is it expected if the length of `fullName` is 3 or more?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
```diff
@@ -202,6 +206,82 @@ public String getDataTypeForKey(DataType dataType) {
...
+        if (columnSize < 53) {
+          logger.info(
+              "data type larger than that of underlying database is assigned: {} to DOUBLE",
+              numericTypeDescription);
+        }
+        return DataType.DOUBLE;
+      case DOUBLE:
+        logger.warn(
+            String.format(
+                "data type that may be smaller than that of underlying database is assigned: %s (Oracle %s to our BIGINT)",
```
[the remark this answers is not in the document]
jnmt: Oops, Oracle's BINARY_DOUBLE is compatible with ScalarDB's double (w.r.t. IEEE 754). So, we don't need to warn in the first place. Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
jnmt: No, they aren't. Because the current table-importing feature targets relational databases only, clustering keys no longer be needed under the relational abstraction.

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
jnmt: Ahh, sqlTypes comes from `java.sql.Types`. Still weird? How about adding usage comments?

schema-loader/src/main/java/com/scalar/db/schemaloader/ImportTableSchema.java
jnmt: Thanks. Should it be `length!= 2` right?

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
komamitsu: Yeah, just looked at `java.sql.Types`. It consists of many `int` constant values (e.g., `public final static int SMALLINT =   5`.) So it's a set of integers (SQL types) in a sense.

`com.scalar.db.util.ScalarDbUtils#getJdbcType` seems to find the corresponding type out of the set of SQL types (`java.sql.Types`) by a single key `int sqlType(s)`. So I think we don't need to treat `int sqlType(s)` as plural.

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -0,0 +1,442 @@
...
+          "SET('a','b')",
+          "GEOMETRY",
+          "JSON", // we remove this for MariaDB because it is an alias of a supported type, LONGTEXT
+          "NUMERIC",
+          "TIME",
+          "TIMESTAMP",
+          "YEAR");
+  static final List<String> UNSUPPORTED_DATA_TYPES_PGSQL =
+      Arrays.asList(
+          "bigserial",
```
komamitsu, line 38: `serial` should be added as well?

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -0,0 +1,442 @@
...
+          "inet",
+          "interval",
+          "json",
+          "jsonb",
+          "line",
+          "lseg",
+          "macaddr",
+          "macaddr8",
+          "money",
+          "numeric(8,2)",
```
komamitsu, line 54: `float4` should be added for https://github.com/scalar-labs/scalardb/pull/931/files#diff-33515a82af5b0cb8f8944cc2237473e9eb8df95bca427831ddfdb72212184009R240 ?

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -0,0 +1,442 @@
...
+          "tsquery",
+          "tsvector",
+          "txid_snapshot",
+          "uuid",
+          "xml");
+  static final List<String> UNSUPPORTED_DATA_TYPES_ORACLE =
+      Arrays.asList(
+          "BFILE",
+          "DATE",
+          "FLOAT(54)",
```
komamitsu, line 74: This is for https://github.com/scalar-labs/scalardb/pull/931/files#diff-de99a8d73f3719cd7ae87026fbfbf7a953774fa4719bfdb18f898b5925e7d484R236? So how about testing with `FLOAT(52)` to test another validation?

komamitsu: Oops. The validation is just for logging. Please ignore it...

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
```diff
@@ -202,6 +206,82 @@ public String getDataTypeForKey(DataType dataType) {
...
+      case VARBINARY:
+        logger.info(
+            "data type larger than that of underlying database is assigned: {} (Oracle {} to BLOB)",
+            columnDescription,
+            typeName);
+        return DataType.BLOB;
+      case LONGVARBINARY:
+        return DataType.BLOB;
+      case BLOB:
+        logger.info(
```
komamitsu, line 274: This should be `warn`?

jnmt: I guess this should be `info`, but the message was wrong. Anyway, good catch! Thank you!

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -0,0 +1,442 @@
...
+      Arrays.asList(
+          "date",
+          "datetime",
+          "datetime2",
+          "datetimeoffset",
+          "decimal(8,2)",
+          "hierarchyid",
+          "money",
+          "numeric(8,2)",
+          "rowversion",
```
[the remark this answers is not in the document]
jnmt: Do you mean why we do not have to add the timestamp data type here? If yes, that is because there is no such data type in the reference. I'm not really sure, but I don't think `rowversion` is an alternative for `timestamp` since `datetime` and `datetime2` are mapped to JDBC `TIMESTAMP`. Note that the document you pointed out also says that:

> The Transact-SQL timestamp data type is different from the timestamp data type defined in the ISO standard.

Anyway, we don't have to have `timestamp` data type here, I think.

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
jnmt: No, we don't have to add it because the JDBC data type for PostrgreSQL `real` (`float4`) is always `REAL`.

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
jnmt: Oh, good catch! Thank you!

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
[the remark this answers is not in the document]
komamitsu: I see. Thanks for the explanation!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
jnmt: Let me take that back. I think I used the "smaller" message since the maximum size is larger than the logical maximum size of ScalarDB BLOB. So, this should be `warn` as you noticed.

https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Data-Types.html#GUID-7B72E154-677A-4342-A1EA-C74C1EA928E6

> A binary large object. Maximum size is (4 gigabytes - 1) * (database block size).

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -429,6 +441,62 @@ public TableMetadata getTableMetadata(String namespace, String table) throws Exe
...
 
+  @Override
+  public TableMetadata getImportTableMetadata(String namespace, String table)
+      throws ExecutionException {
+    TableMetadata.Builder builder = TableMetadata.newBuilder();
+    boolean primaryKeyExists = false;
+
+    if (!rdbEngine.isImportable()) {
+      throw new UnsupportedOperationException(
+          "Importing table is not allowed in this storage: " + rdbEngine);
```
brfrn169, line 452: `RdbEngineStrategy` doesn't override `toString()`, so maybe `rdbEngine.getClass().getName()` is more readable:
```suggestion
          "Importing table is not allowed in this storage: " + rdbEngine.getClass().getName());
```

brfrn169 (commented):
I've completed the first review and left a few comments. Please take a look when you have time! And I'll conduct the second review later. Thanks.

core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -402,4 +402,15 @@ void repairTable(
...
+  /**
+   * Import an existing table that is not managed by ScalarDB.
+   *
+   * @param namespace an existing namespace
+   * @param table an existing table
+   * @throws IllegalArgumentException if the table is already managed by ScalarDB, if the target
+   *     table does not exist, or if the table does not meet the requirement of ScalarDB table
+   * @throws ExecutionException if the operation fails
+   */
+  void importTable(String namespace, String table) throws ExecutionException;
```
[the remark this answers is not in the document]
jnmt: That's a good question. Behind this question, are you wondering if we should only support transaction API for importing? It's a decision for what use cases should be covered. I originally thought covering both use cases (storage and transaction) would be nice. But, now I'm not really sure since the major reason for migrating to ScalarDB would be that users want to use transactions in ScalarDB. I would like to also hear other reviewers' thoughts.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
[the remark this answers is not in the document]
jnmt: Fixed in 0d3a2f8.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSqlServer.java
```diff
@@ -176,6 +180,75 @@ public String getDataTypeForKey(DataType dataType) {
...
+            "data type larger than that of underlying database is assigned: {} ({} to INT)",
+            columnDescription,
+            typeName);
+        return DataType.INT;
+      case INTEGER:
+        return DataType.INT;
+      case BIGINT:
+        logger.warn(
+            String.format(
+                "data type that may be smaller than that of underlying database is assigned: %s (SQL Server %s to our BIGINT)",
```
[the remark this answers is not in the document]
jnmt: Fixed in 0d3a2f8.

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
jnmt: Fixed in 0d3a2f8.

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
jnmt: Fixed in 0d3a2f8.

schema-loader/src/main/java/com/scalar/db/schemaloader/ImportSchemaParser.java
jnmt: Fixed in 0d3a2f8.

schema-loader/src/main/java/com/scalar/db/schemaloader/ImportTableSchema.java
jnmt: Fixed in 0d3a2f8.

schema-loader/src/main/java/com/scalar/db/schemaloader/ImportTableSchema.java
jnmt: Fixed in 0d3a2f8.

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
jnmt: Fixed in 0d3a2f8.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
jnmt: Fixed in 0d3a2f8.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
jnmt: Fixed in 0d3a2f8.

komamitsu (approved):
LGTM! 👍

brfrn169 (commented):
I did the second review and I left some additional comments. Please take a look when you have time!

schema-loader/src/main/java/com/scalar/db/schemaloader/command/SchemaLoaderCommand.java
```diff
@@ -90,6 +90,12 @@ private static class Mode {
                 + "It compares the provided table schema to the existing schema to decide which columns need to be added and which indexes need to be created or deleted",
         defaultValue = "false")
     boolean alterTables;
+
+    @Option(
+        names = {"-I", "--import"},
+        description = "",
```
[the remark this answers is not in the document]
jnmt: Good catch, thank you! I totally forgot to fill it out. Fixed in 1526725.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -28,7 +28,7 @@ public final class ConsensusCommitUtils {
           .put(Attribute.COMMITTED_AT, DataType.BIGINT)
           .build();
 
-  private static final ImmutableMap<String, DataType> TRANSACTION_META_COLUMNS =
+  static final ImmutableMap<String, DataType> TRANSACTION_META_COLUMNS =
```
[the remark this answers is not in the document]
jnmt: Fixed in 1526725.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -127,7 +127,7 @@ public static boolean isTransactionTableMetadata(TableMetadata tableMetadata) {
     return true;
   }
 
-  private static Set<String> getNonPrimaryKeyColumns(TableMetadata tableMetadata) {
+  static Set<String> getNonPrimaryKeyColumns(TableMetadata tableMetadata) {
```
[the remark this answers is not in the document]
jnmt: Fixed in 1526725.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransactionAdmin.java
```diff
@@ -90,6 +91,14 @@ public boolean namespaceExists(String namespace) throws ExecutionException {
     return jdbcAdmin.namespaceExists(namespace);
   }
 
+  @Override
+  public void importTable(String namespace, String table) throws ExecutionException {
+    TableMetadata tableMetadata = jdbcAdmin.getImportTableMetadata(namespace, table);
+
+    // add ScalarDB metadata
+    jdbcAdmin.repairTable(namespace, table, tableMetadata, ImmutableMap.of());
```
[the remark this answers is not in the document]
jnmt: Exactly, we should. Fixed in 1526725.

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
```diff
@@ -249,4 +250,29 @@ public static boolean isRelational(Scan scan) {
...
         && (!scan.getOrderings().isEmpty() || !scan.getConjunctions().isEmpty());
   }
+
+  /**
+   * Get {@code JDBCType} of the specified {@code sqlType}.
+   *
+   * @param sqlType a type defined in {@code java.sql.Types}
+   * @return a JDBCType
+   */
+  public static JDBCType getJdbcType(int sqlType) {
```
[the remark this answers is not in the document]
jnmt: Yeah, it's also a good point. Thank you. Fixed in 1526725.

core/src/main/java/com/scalar/db/api/Admin.java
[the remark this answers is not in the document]
jnmt: @feeblefakie We are discussing whether we should support table importing in both storage and transaction API. What do you think?

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/api/Admin.java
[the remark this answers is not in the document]
feeblefakie: @jnmt I think supporting only transactions seems fine to me.
Can we add support for storage if we really want to do it later on?

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/api/Admin.java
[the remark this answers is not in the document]
feeblefakie: @jnmt I discuss it with Toshi, and we think we can keep it as is.
Later, we might remove the support for storage, though.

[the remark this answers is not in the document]
jnmt: @feeblefakie @Torch3333 Got it! Let's keep it as is. Thank you! (I was just wondering about the revert cost of the current PR.)

brfrn169 (approved):
LGTM! Thank you!

PR #934, opened 2023-07-11
Revise document for handling exceptions for Two-phase Commit Transactions

brfrn169:
This PR revises the document for handling exceptions for Two-phase Commit Transactions. Please take a look!

### DIFF
--- a/docs/two-phase-commit-transactions.md
+++ b/docs/two-phase-commit-transactions.md
@@ -201,6 +201,7 @@ You can call `prepare()`, `commit()`, `rollback()` in the coordinator/participan
 #### Validate the transaction
 
 Depending on the concurrency control protocol, you need to call `validate()` in all the coordinator/participant processes after `prepare()` and before `commit()`:
+
 ```java
 // Prepare phase 1: Prepare the transaction in all the coordinator/participant processes
 tx.prepare();
@@ -282,16 +283,27 @@ try {
 } catch (Exception e) {
   // Rollback the transaction
   if (transaction1 != null) {
-    transaction1.rollback();
+    try {
+      transaction1.rollback();
+    } catch (RollbackException e1) {
+      // Handle the exception
+    }
   }
   if (transaction2 != null) {
+    try {
     transaction2.rollback();
+    } catch (RollbackException e1) {
+      // Handle the exception
+    }
   }
 }
 ```
 
-As described above, for `commit()`, if any one of the coordinator or participant processes succeeds in committing the transaction, you can regard the transaction as committed.
-Also, you can execute `prepare()`, `validate()`, `commit()` in parallel for better performance.
+For simplicity, the above example code doesn't handle the exceptions that can be thrown by the APIs.
+See [Handle exceptions](#handle-exceptions) for more details.
+
+As previously mentioned, for `commit()`, if any one of the coordinator or participant processes succeeds in committing the transaction, you can regard the transaction as committed.
+Also, for better performance, you can execute `prepare()`, `validate()`, `commit()` in parallel, respectively.
 
 ### Resume a transaction
 
@@ -503,10 +515,11 @@ public class Sample {
 
         return;
       } catch (UnknownTransactionStatusException e) {
-        // If you catch `UnknownTransactionStatusException` when committing the transaction, you are
-        // not sure if the transaction succeeds or not. In such a case, you need to check if the
-        // transaction is committed successfully or not and retry it if it failed. How to identify a
-        // transaction status is delegated to users
+        // If you catch `UnknownTransactionStatusException` when committing the transaction, it
+        // indicates that the status of the transaction, whether it has succeeded or not, is
+        // unknown. In such a case, you need to check if the transaction is committed successfully
+        // or not and retry it if it failed. How to identify a transaction status is delegated to
+        // users
         return;
       } catch (TransactionException e) {
         // For other exceptions, you can try retrying the transaction.
--- a/docs/two-phase-commit-transactions.md
+++ b/docs/two-phase-commit-transactions.md
@@ -704,7 +704,7 @@ In such a case, since a transaction processing in Two-phase Commit Transactions 
 
 ![](images/two_phase_commit_load_balancing.png)
 
-There are several approaches to achieve it depending on the protocol between the services. Here, we introduce some approaches for gRPC and HTTP/1.1.
+There are several approaches to achieve it depending on the protocol between the services. The next section introduces some approaches for gRPC and HTTP/1.1.
 
 ### gPRC
 
--- a/docs/two-phase-commit-transactions.md
+++ b/docs/two-phase-commit-transactions.md
@@ -261,7 +261,7 @@ try {
 
   // Commit the transaction. If any of the transactions succeeds to commit, you can regard the
   // transaction as committed
-  AtomicReference<Exception> exception = new AtomicReference<>();
+  AtomicReference<TransactionException> exception = new AtomicReference<>();
   boolean anyMatch =
       Stream.of(transaction1, transaction2)
           .anyMatch(
@@ -269,7 +269,7 @@ try {
                 try {
                   t.commit();
                   return true;
-                } catch (Exception e) {
+                } catch (TransactionException e) {
                   exception.set(e);
                   return false;
                 }
@@ -279,7 +279,7 @@ try {
   if (!anyMatch) {
     throw exception.get();
   }
-} catch (Exception e) {
+} catch (TransactionException e) {
   // Rollback the transaction
   if (transaction1 != null) {
     try {
@@ -682,7 +682,7 @@ This exception indicates that the transaction associated with the specified ID w
 In either case, you can retry the transaction from the beginning since the cause of this exception is basically transient.
 
 In the sample code, for `UnknownTransactionStatusException`, the transaction doesn't retry because the cause of the exception is nontransient.
-In the sample code, for `UnknownTransactionStatusException`, the transaction doesn't retry as mentioned earlier.
+Also, for `UnsatisfiedConditionException`, the transaction doesn't retry because how to handle this exception depends on your application requirements.
 For other exceptions, the transaction tries retrying because the cause of the exception is transient or nontransient.
 If the cause of the exception is transient, the transaction may succeed if you retry it.
 However, if the cause of the exception is nontransient, the transaction may still fail even if you retry it.
--- a/docs/two-phase-commit-transactions.md
+++ b/docs/two-phase-commit-transactions.md
@@ -16,7 +16,7 @@ The transaction manager instances can be in the same process/applications or in 
 For example, if you have transaction manager instances in multiple microservices, you can execute a transaction that spans multiple microservices.
 
 In Two-phase Commit Transactions, there are two roles, a coordinator and a participant, that collaboratively execute a single transaction.
-A coordinator process and participant processes has another transaction manager instance.
+A coordinator process and participant processes all have different transaction manager instances.
 The coordinator process first begins a transaction, and the participant processes join the transaction.
 And after executing CRUD operations, the coordinator process and the participant processes execute the two-phase commit protocol to commit the transaction.
 
@@ -189,7 +189,7 @@ try {
 }
 ```
 
-For `preapre()`, if any one of the coordinator or participant processes fails to prepare the transaction, you need to call `rollback()` (or `abort()`) in all the coordinator/participant processes.
+For `prepare()`, if any one of the coordinator or participant processes fails to prepare the transaction, you need to call `rollback()` (or `abort()`) in all the coordinator/participant processes.
 
 For `commit()`, if any one of the coordinator or participant processes succeeds in committing the transaction, you can consider the transaction as committed.
 In other words, in that situation, you can ignore the errors in the other coordinator/participant processes.
--- a/docs/two-phase-commit-transactions.md
+++ b/docs/two-phase-commit-transactions.md
@@ -10,15 +10,15 @@ This document briefly explains how to execute Two-phase Commit Transactions in S
 ScalarDB transactions normally execute in a single transaction manager instance.
 In that case, you begin a transaction, execute CRUD operations, and commit the transaction in the same transaction manager instance.
 
-In addition to the normal transactions, ScalarDB also supports two-phase commit style transactions called *Two-phase Commit Transactions*.
+In addition to normal transactions, ScalarDB also supports two-phase-commit-style transactions called *Two-phase Commit Transactions*.
 Two-phase Commit Transactions execute a transaction that spans multiple transaction manager instances.
-The transaction manager instances can be in the same process/applications or in different processes/applications.
+The transaction manager instances can be in the same process/application or in different processes/applications.
 For example, if you have transaction manager instances in multiple microservices, you can execute a transaction that spans multiple microservices.
 
 In Two-phase Commit Transactions, there are two roles, a coordinator and a participant, that collaboratively execute a single transaction.
 A coordinator process and participant processes all have different transaction manager instances.
 The coordinator process first begins a transaction, and the participant processes join the transaction.
-And after executing CRUD operations, the coordinator process and the participant processes execute the two-phase commit protocol to commit the transaction.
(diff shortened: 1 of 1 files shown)


docs/two-phase-commit-transactions.md
```diff
@@ -193,178 +215,88 @@ tx.commit();
...
+
+  // Validate the transaction
+  transaction1.validate();
+  transaction2.validate();
+
+  // Commit the transaction. If any of the transactions succeeds to commit, you can regard the
+  // transaction as committed
+  AtomicReference<Exception> exception = new AtomicReference<>();
+  boolean anyMatch =
+      Stream.of(transaction1, transaction2)
```
komamitsu, line 266: With this error handling, I think the commits can be executed in parallel as follows

```java
  Stream.of(transaction1, transaction2)
    .parallel()
    .anyMatch(
        ...
```

docs/two-phase-commit-transactions.md
```diff
@@ -193,178 +215,88 @@ tx.commit();
...
+              });
+
+  // If all the transactions fail to commit, throw the exception and rollback the transaction
+  if (!anyMatch) {
+    throw exception.get();
+  }
+} catch (Exception e) {
+  // Rollback the transaction
+  if (transaction1 != null) {
+    transaction1.rollback();
```
komamitsu, line 285: Wrapping this with try-catch would be safer just in case if `rollback()` could throw any exception.

docs/two-phase-commit-transactions.md
komamitsu: Ah, sorry. I found the parallel version code in the next section below.

This code is the sequential version, right? If so, please ignore my comment. But it might be good to mention sequential execution or something here?

docs/two-phase-commit-transactions.md
```diff
@@ -500,6 +432,287 @@ public class ServiceBImpl implements ServiceB {
...
+    // If any of the transactions failed to commit due to `UnknownTransactionStatusException`, throw
+    // it because you should not retry the transaction in such a case
+    Optional<TransactionException> unknownTransactionStatusException =
+        exceptions.stream().filter(e -> e instanceof UnknownTransactionStatusException).findFirst();
+    if (unknownTransactionStatusException.isPresent()) {
+      throw unknownTransactionStatusException.get();
+    }
+
+    // Otherwise, throw the first exception
+    throw exceptions.get(0);
```
komamitsu, line 616: [minor] This is a very corner case, but this line would throw NPE if `transactions` is empty?

docs/two-phase-commit-transactions.md
```diff
@@ -500,6 +432,287 @@ public class ServiceBImpl implements ServiceB {
...
+If you catch `UnknownTransactionStatusException`, it indicates that the status of the transaction, whether it has succeeded or not, is unknown.
+In such a case, you need to check if the transaction is committed successfully or not and retry it if it fails.
+How to identify a transaction status is delegated to users.
+You may want to create a transaction status table and update it transactionally with other application data so that you can get the status of a transaction from the status table.
+
+Although not illustrated in the sample code, the `resume()` API could also throw a `TransactionNotFoundException`.
+This exception indicates that the transaction associated with the specified ID was not found, and it might have been expired.
+In such cases, you can retry the transaction from the beginning since the cause of this exception is basically transient.
+
+In the sample code, for `UnknownTransactionStatusException`, the transaction doesn't retry because the cause of the exception is nontransient.
```
komamitsu, line 672: > because the cause of the exception is nontransient

Maybe we should also mention that the transaction status is unknown (due to underlying database issue or something) and whether to retry must be decided by the application?

docs/two-phase-commit-transactions.md
```diff
@@ -5,6 +5,21 @@ With Two-phase Commit Transactions, you can execute a transaction that spans mul
...
+ScalarDB transactions normally execute in a single transaction manager instance.
+In that case, you begin a transaction, execute CRUD operations, and commit the transaction in the same transaction manager instance.
+
+In addition to the normal transactions, ScalarDB also supports two-phase commit style transactions called *Two-phase Commit Transactions*.
+Two-phase Commit Transactions execute a transaction that spans multiple transaction manager instances.
+The transaction manager instances can be in the same process/applications or in different processes/applications.
+For example, if you have transaction manager instances in multiple microservices, you can execute a transaction that spans multiple microservices.
+
+In Two-phase Commit Transactions, there are two roles, a coordinator and a participant, that collaboratively execute a single transaction.
+A coordinator process and participant processes has another transaction manager instance.
```
Torch3333, line 19:
```suggestion
A coordinator process and participant processes all have different transaction manager instances.
```

Torch3333 (commented):
It looks good, thanks. I left two suggestions.

docs/two-phase-commit-transactions.md
brfrn169: Fixed in 635b41d1ec91b4be3e98e835bd692a17a29407cb. Thanks.

docs/two-phase-commit-transactions.md
brfrn169: I think that has already been mentioned here:
https://github.com/scalar-labs/scalardb/pull/934/files#diff-2677f880f1a9e512f31cce8e12950863ddffd4be49408d310d72d4d73e6122b9R306

What do you think?

docs/two-phase-commit-transactions.md
brfrn169: Ah, that's a very unexpected usage. We could add validation or an assertion to this method, but I think it might be too verbose since it's a sample code. What do you think?

docs/two-phase-commit-transactions.md
brfrn169: I think that's mentioned here:
https://github.com/scalar-labs/scalardb/pull/934/files#diff-2677f880f1a9e512f31cce8e12950863ddffd4be49408d310d72d4d73e6122b9R676-R679

What do you think?

komamitsu: Yeah, the previous description looks perfect.

But, I have a concern that the readers might forget the previous important explanation about UnknownTransactionStatusException since this line only mentions nontrasient causes.

In my understanding, UnknownTransactionStatusException must not be retried, while other non-transient exceptions can be retried although it may make no sense. Just mentioning nontransient causes seems a bit weak in preventing users from attempting retries.

docs/two-phase-commit-transactions.md
komamitsu: Yeah, I think it's okay to go with current simple one 👍

docs/two-phase-commit-transactions.md
komamitsu: Looks good!

Torch3333 (approved):
LGTM, thank you!

docs/two-phase-commit-transactions.md
```diff
@@ -500,6 +444,288 @@ public class ServiceBImpl implements ServiceB {
...
+  }
+}
+```
+
+The `begin()` API could throw `TransactionException` and `TransactionNotFoundException`.
+If you catch `TransactionException`, it indicates that the transaction has failed to begin due to transient or nontransient faults. You can try retrying the transaction, but you may not be able to begin the transaction due to nontransient faults.
+If you catch `TransactionNotFoundException`, it indicates that the transaction has failed to begin due to transient faults. You can retry the transaction.
+
+The `join()` API could also throw a `TransactionException` and `TransactionNotFoundException`.
+And the way to handle them is the same as the `begin()` API.
```
josh-wong, line 655:
```suggestion
You can handle these exceptions in the same way that you handle the exceptions for the `begin()` API.
```

josh-wong (changes requested):

docs/two-phase-commit-transactions.md
```diff
@@ -500,6 +444,288 @@ public class ServiceBImpl implements ServiceB {
...
+Services using Two-phase Commit Transactions usually execute a transaction by exchanging multiple requests and responses as follows:
+
+![](images/two_phase_commit_sequence_diagram.png)
+
+Also, each service typically has multiple servers (or hosts) for scalability and availability and uses server-side (proxy) or client-side load balancing to distribute requests to the servers.
+In such a case, since a transaction processing in Two-phase Commit Transactions is stateful, requests in a transaction must be routed to the same servers while different transactions need to be distributed to balance the load.
+
+![](images/two_phase_commit_load_balancing.png)
+
+There are several approaches to achieve it depending on the protocol between the services. Here, we introduce some approaches for gRPC and HTTP/1.1.
```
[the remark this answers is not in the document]
brfrn169: `to achieve it` means `requests in a transaction must be routed to the same servers`.

docs/two-phase-commit-transactions.md
brfrn169: That makes sense.

So how about this:
```suggestion
In the sample code, for `UnknownTransactionStatusException`, the transaction doesn't retry as mentioned earlier.
```

komamitsu: Looks good!

komamitsu (approved):
LGTM! 👍

josh-wong (approved):
LGTM! Thank you!🙇‍♂️

docs/two-phase-commit-transactions.md
```diff
@@ -5,6 +5,21 @@ With Two-phase Commit Transactions, you can execute a transaction that spans mul
...
+
+In addition to normal transactions, ScalarDB also supports two-phase-commit-style transactions called *Two-phase Commit Transactions*.
+Two-phase Commit Transactions execute a transaction that spans multiple transaction manager instances.
+The transaction manager instances can be in the same process/application or in different processes/applications.
+For example, if you have transaction manager instances in multiple microservices, you can execute a transaction that spans multiple microservices.
+
+In Two-phase Commit Transactions, there are two roles, a coordinator and a participant, that collaboratively execute a single transaction.
+A coordinator process and participant processes all have different transaction manager instances.
+The coordinator process first begins a transaction, and the participant processes join the transaction.
+After executing CRUD operations, the coordinator process and the participant processes execute the two-phase commit protocol to commit the transaction.
```
feeblefakie, line 21:
```suggestion
After executing CRUD operations, the coordinator process and the participant processes commit the transaction by using the two-phase interface.
```
`the two-phase commit protocol` is not really defined here, so I felt it was difficult to parse.
I also thought it might be misleading since the protocols are both two-phase commits, but the interfaces are different in normal and two-phase commit transactions. so I suggested as above.

feeblefakie (commented):
And, sorry for the late review.
I left some comments and suggestions.

Also, I think the name `Two-phase Commit Transactions` might be misleading since readers would think normal transactions don't employ the two-phase commit protocol.
Let's discuss this offline.

docs/two-phase-commit-transactions.md
```diff
@@ -5,6 +5,21 @@ With Two-phase Commit Transactions, you can execute a transaction that spans mul
 
 This document briefly explains how to execute Two-phase Commit Transactions in ScalarDB.
 
+## Overview
+
+ScalarDB transactions normally execute in a single transaction manager instance.
+In that case, you begin a transaction, execute CRUD operations, and commit the transaction in the same transaction manager instance.
+
+In addition to normal transactions, ScalarDB also supports two-phase-commit-style transactions called *Two-phase Commit Transactions*.
```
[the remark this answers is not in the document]
josh-wong: Sorry, I may have capitalized `Two-phase Commit Transactions` since I thought it was referring to the name of the feature in ScalarDB. If this is a generic mention of `two-phase commit transactions`, let's lowercase it. 

And I would keep `normal transactions` lowercase because, as far as I know, it's not the name of a feature in ScalarDB.

docs/two-phase-commit-transactions.md
```diff
@@ -500,6 +443,288 @@ public class ServiceBImpl implements ServiceB {
...
+For example, if you use [Envoy](https://www.envoyproxy.io/), you can use session affinity (sticky session) for gRPC.
+Alternatively, you can use [bidirectional streaming RPC in gRPC](https://grpc.io/docs/what-is-grpc/core-concepts/#bidirectional-streaming-rpc) since the L7 load balancer distributes requests in the same stream to the same server.
+
+### HTTP/1.1
+
+Typically, you use a server-side (proxy) load balancer with HTTP/1.1.
+When using an L3/L4 load balancer, you can use the same HTTP connection to send requests in a transaction, which guarantees the requests go to the same server.
+When using an L7 load balancer, since requests in the same HTTP connection don't necessarily go to the same server, you need to use cookies or similar method to route requests to the correct server.
+You can use session affinity (sticky session) in that case.
+
```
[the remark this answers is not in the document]
brfrn169: Yes, we should. Let's do that in another PR. Thanks.

docs/two-phase-commit-transactions.md
```diff
@@ -5,6 +5,21 @@ With Two-phase Commit Transactions, you can execute a transaction that spans mul
 
 This document briefly explains how to execute Two-phase Commit Transactions in ScalarDB.
 
+## Overview
+
+ScalarDB transactions normally execute in a single transaction manager instance.
```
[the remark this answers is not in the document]
brfrn169: I applied the suggestion in 0db8458fcbd8fe14999231b2da301bff4fd0602b.

> Also, it might be helpful to have a diagram like this.
> https://speakerdeck.com/scalar/scalar-db-universal-transaction-manager?slide=27

Regarding this, let's do in another PR.

docs/two-phase-commit-transactions.md
```diff
@@ -500,6 +443,288 @@ public class ServiceBImpl implements ServiceB {
...
 
+### Handle exceptions
+
+In the previous section, you saw [how to execute a transaction with multiple transaction manager instances](#execute-a-transaction-with-multiple-transaction-manager-instances). However, you may also need to handle exceptions properly.
+This section describes how to handle exceptions in Two-phase Commit Transactions.
+
+Two-phase commit transactions are basically executed by multiple processes/applications (a coordinator and participants).
+However, in this example code, you can see multiple transaction managers (`transactionManager1` and `transactionManager2`) in a single process.
+
+The following example code shows how to handle exceptions in Two-phase Commit Transactions:
```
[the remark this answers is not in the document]
brfrn169: Fixed in 272d88b46c4f21fc6da8001800596115f754d637. Can you please check it?

brfrn169:
@feeblefakie As discussed, let's rephrase the terms `Two-phase Commit Transactions` to `(transactions with) Two-phase Commit interface` and `normal transactions` to `(transactions with) One-phase Commit interface` in a separate PR. Thanks.

josh-wong:
@brfrn169 Just one note about the following:

> @feeblefakie As discussed, let's rephrase the terms `Two-phase Commit Transactions` to `(transactions with) Two-phase Commit interface` and `normal transactions` to `(transactions with) One-phase Commit interface` in a separate PR. Thanks.

If `Two-phase Commit interface` and `One-phase Commit interface` aren't proper names (in ScalarDB), we should lowercase them. Most mentions of these are lowercased in Google search results, so it doesn't seem to be a term that is often capitalized.

komamitsu:
> normal transactions to (transactions with) One-phase Commit interface

I might be missing something and sorry in advance if this is a silly comment.

If `normal transactions` means `DistributedTransaction` in a sense, it sounds a bit weird since ScalarDB's normal transactions are internally 2PC.

komamitsu:
Ah, I missed https://github.com/scalar-labs/scalardb/pull/934#discussion_r1266409101. Got the point.

(`single service` and `multi service` transactions or something might sound a bit less weird to me...?)

brfrn169:
> If Two-phase Commit interface and One-phase Commit interface aren't proper names (in ScalarDB), we should lowercase them. Most mentions of these are lowercased in Google search results, so it doesn't seem to be a term that is often capitalized.

@josh-wong Thank you for the input. Let's proceed with the lowercase notation.

brfrn169:
@komamitsu Yes, we should clarify that while ScalarDB uses a variant of the two-phase commit protocol (Consensus Commit), it offers two types of interfaces: `one-phase commit interface` and `two-phase commit interface`.

feeblefakie (approved):
LGTM! Thank you!

PR #984, opened 2023-07-27
Enable to specify set of conditional expression set in ScanAll

jnmt:
This PR enables specifying a set of conditional expression sets in ScanAll. The main reason to add this interface is to easily build a where clause without calling builder methods several times (especially when we build ScanAll for an internal purpose, e.g., in ScalarDB SQL).


core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -349,6 +351,12 @@ public BuildableScanAllWithOngoingWhere where(ConditionalExpression condition) {
       return new BuildableScanAllWithOngoingWhere(this, condition);
     }
 
+    @Override
+    public BuildableScanAllWithWhere where(Set<Set<ConditionalExpression>> conditions) {
```
jnmt, line 355: This interface is mainly for internal purposes, but users can see the interface and might use it. So, I didn't use `Conjunction` here.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/api/OperationBuilder.java
```diff
@@ -432,6 +433,16 @@ interface WhereOr<T> {
...
   }
 
+  interface WhereConditions<T> {
+    /**
+     * Appends the specified conditions.
+     *
+     * @param conditions an or-wise set of and-wise conditions.
+     * @return the operation builder
+     */
+    T where(Set<Set<ConditionalExpression>> conditions);
```
brfrn169, line 443: Can you please provide some examples to help us understand when this would be useful? I'm curious as to why the existing methods don't suffice.

jnmt: @brfrn169 Thank you for the question. When we have a set of conjunctions, we need to call `where()` and `or()` several times and change the behavior depending on the number of conjunctions. I just wanted to avoid managing an intermediate builder like the following.

```java
BuildableScanAll buildableScanAll =
    Scan.newBuilder()
        .namespace(statement.namespaceName)
        .table(statement.tableName)
        .all();

BuildableScanAllWithOngoingWhereOr ongoing = null;
for (Conjunction conjunction : conjunctions) {
  // snip conversion of conjunction in the SQL layer -> conditional expression
  if (ongoing == null) {
     ongoing = builder.where(ConditionSetBuilder.andConditionSet(conditions).build());
  } else {
     ongoing.or(ConditionSetBuilder.andConditionSet(conditions).build());
  }
}
```

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/OperationBuilder.java
jnmt: @brfrn169 That's the point that bothered me (since it's somewhat imbalanced in and/or set, how about adding a message like "internal purpose", and so on and so on...). Anyway, after reconsidering, I think your idea is better. Fixed in acd3be2. Thank you!

core/src/main/java/com/scalar/db/api/OperationBuilder.java
```diff
@@ -416,20 +417,28 @@ interface WhereAnd<T> {
     /**
      * Appends the specified set of or-wise conditions.
      *
-     * @param conditions a set of conditions
+     * @param orConditionSet a set of or-wise conditions
      * @return the operation builder
      */
-    T where(ScanBuilder.OrConditionSet conditions);
+    T where(ScanBuilder.OrConditionSet orConditionSet);
```
brfrn169, line 423: BTW, can we also add `where(Set<ScanBuilder.OrConditionSet > andConditionSets)` to `WhereAnd<T>` as well, it could cause signature conflicts though?

jnmt: Unfortunately, we cannot add it since it causes the signature conflict, as you guess.

brfrn169: Okay. In that case, how about changing the method name to address the signature conflict as follows?

```java
  interface WhereAnd<T> {
    T where(ScanBuilder.OrConditionSet orConditionSet);

    T whereAnd(Set<ScanBuilder.OrConditionSet> orConditionSets);
  }

  interface WhereOr<T> {
    T where(ScanBuilder.AndConditionSet andConditionSet);

    T whereOr(Set<ScanBuilder.AndConditionSet> andConditionSets);
  }
```

jnmt: Thanks for the solution. Umm, it's a bit unstylish, but right, consistency is more important here, I think.

jnmt: Fixed in 1b9ae6a. PTAL when you have time!

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #986, opened 2023-07-31
Find the latest release version in `vuln-check` GHA workflow

komamitsu:
## Context

We've released https://github.com/scalar-labs/scalardb/blob/master/.github/workflows/vuln-check.yaml and https://github.com/scalar-labs/scalardb/blob/master/.github/workflows/scheduled-vuln-check.yaml for scheduled vulnerability checking. But in `scheduled-vuln-check.yaml`, we need to update all the target-ref(s) that contain patch version when releasing a new patch version. It would be great if we make those workflows automatically find the latest patch version releases. For instance, it finds `v3.9.2` when receiving `v3.9`.

## Changes

This PR introduces
- `find-latest-release` input parameter in `vuln-check.yaml`
  - This is enabled only when to find the latest release name. It shouldn't be enabled when testing `master` branch and so on
- `prepare-target-ref` step in  `vuln-check.yaml`
  - This finds the latest release name using newly added scripts if `find-latest-release` is enabled
- New scripts in `ci/vuln-check`
  - fetch_gh_releases: Get all release names
  - find_latest_release: Find the latest release name out of release names passed via argument


Torch3333 (approved):
LGTM, thank you!

brfrn169 (approved):
LGTM! Thank you!

PR #989, opened 2023-08-03
Add Namespaces management

Torch3333:
**Context**
ScalarDB did not manage namespaces so we could not retrieve the names of existing namespaces. We could only retrieve the tables of a given namespace.

**Changes**
ScalarDB is now managing namespace metadata in addition to the table metadata. At the moment, the namespace metadata is only the namespace name. With the Admin API `admin.getNamespaceNames()`, the list of namespaces names created through ScalarDB can be retrieved. 
To manage the created namespace metadata, a new table called `namespace` is added to the same namespace which contains the metadata table. 

The namespace name containing namespace and table metadata can be set via the configuration : 
- Cassandra : created the new `scalar.db.cassandra.metadata.keyspace` configuration.
- CosmosDB: we deprecated the current `scalar.db.cosmos.table_metadata.database` to rename it to `scalar.db.cosmos.metadata.database`
- DynamoDB: we deprecated the existing `scalar.db.dynamo.table_metadata.namespace` to rename it to `scalar.db.dynamo.metadata.namespace`
- JDBC: we deprecated the existing `scalar.db.jdbc.table_metadata.schema` to rename it to `scalar.db.jdbc.metadata.schema`

**Future related changes**
We will also add two related features to the Admin API and the Schema Loader tool : 
- `upgrade` allows the users to update their existing environment to be compatible with the new Scalar DB version using the namespace table
- `repair the namespace` allows the users to repair the namespaces metadata. If for some reason, a namespace exists in the underlying storage, but its metadata is missing from the namespaces metadata managed by ScalarDB (because the namespaces table was deleted or its data is corrupted), this method will re-add the namespace to the ScalarDB namespaces metadata.


core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -413,4 +413,12 @@ void addNewColumnToTable(String namespace, String table, String columnName, Data
...
    */
   void importTable(String namespace, String table) throws ExecutionException;
+
+  /**
+   * Returns the names of the existing namespaces created through Scalar DB.
+   *
+   * @return a set of namespaces names, an empty set if no namespaces exist
+   * @throws ExecutionException if the operation fails
+   */
+  Set<String> getNamespaceNames() throws ExecutionException;
```
Torch3333, line 423: The new `admin.getNamespaceNames()` interface.

docs/api-guide.md
```diff
@@ -203,6 +203,23 @@ boolean ifExists = true;
...
+```
+
+### Get the tables of a namespace
+
+You can get the tables of a namespace as follows:
+
+```java
+// Get the tables of the namespace "ns"
+Set<String> tables = admin.getNamespaceTableNames("ns");
+```
```
Torch3333, line 221: Add the doc for `admin.getNamespaceTableNames("ns")` since it was missing.

core/src/main/java/com/scalar/db/common/CheckedDistributedStorageAdmin.java
```diff
@@ -14,26 +14,15 @@ public class CheckedDistributedStorageAdmin implements DistributedStorageAdmin {
 
   private final DistributedStorageAdmin admin;
 
-  /**
-   * Whether to check if the namespace exists or not. Set false when the storage does not support
-   * namespaces.
-   */
-  private final boolean checkNamespace;
```
Torch3333, line 21: All storages support namespace now so it is not needed anymore.

core/src/test/java/com/scalar/db/storage/cassandra/CassandraAdminTestBase.java
```diff
@@ -44,20 +53,42 @@
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
-public class CassandraAdminTest {
+/**
+ * Abstraction that defines unit tests for the {@link CassandraAdmin}. The class purpose is to be
+ * able to run the {@link CassandraAdmin} unit tests with different values for the {@link
+ * CassandraAdmin}, notably {@link CassandraAdmin#METADATA_KEYSPACE}.
+ */
+public abstract class CassandraAdminTestBase {
```
Torch3333, line 61: Abstracted the unit test for Cassandra to be able to test with different values for the metadata keyspace.

docs/api-guide.md
```diff
@@ -203,6 +203,23 @@ boolean ifExists = true;
 admin.dropNamespace("ns", ifExists);
 ```
 
+### Get the namespaces
+
+You can get the existing namespaces as follows:
+
+```java
+Set<String> namespaces = admin.getNamespaceNames();
+```
```
Torch3333, line 212: Added the new API to the documentation.

docs/configurations.md
```diff
@@ -49,34 +49,35 @@ The following describes the configurations available for each storage.
 
 - For Cassandra, the following configurations are available:
```
Torch3333, line 51: Updated the documentation for the configuration

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -466,25 +484,29 @@ private ThroughputProperties calculateThroughput(Map<String, String> options) {
...
+          .readItem(namespace, new PartitionKey(namespace), CosmosNamespace.class);
+      return true;
+    } catch (RuntimeException e) {
+      if (e instanceof CosmosException
+          && ((CosmosException) e).getStatusCode() == CosmosErrorCode.NOT_FOUND.get()) {
+        return false;
+      }
+      throw new ExecutionException("Checking if the namespace exists failed", e);
+    }
   }
```
Torch3333, line 498: For each admin, the implementation of `admin.namespaceExists()` is now uniformized and checks if the namespace is contained in the namespace metadata table.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -98,14 +138,30 @@ public void dropTable(String namespace, String table) throws ExecutionException
 
   @Override
   public void dropNamespace(String namespace) throws ExecutionException {
-    String dropKeyspace = SchemaBuilder.dropKeyspace(quoteIfNecessary(namespace)).getQueryString();
     try {
-      clusterManager.getSession().execute(dropKeyspace);
+      dropKeyspace(namespace);
+      deleteFromKeyspacesTable(namespace);
+      dropKeyspacesTableIfEmpty();
```
komamitsu, line 144: Does this mean the keyspace for metadata is removed if there is no namespace? If so, I think we can keep the namespace for metadata in this case.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosNamespace.java
```diff
@@ -0,0 +1,46 @@
...
+public class CosmosNamespace {
+  private String id;
+
+  public CosmosNamespace() {}
+
+  public CosmosNamespace(String id) {
+    this.id = id;
+  }
+
+  public void setId(String id) {
```
komamitsu, line 16: This method doesn't seem to be used. How about putting `final` keyword on `id`?

core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosAdminTestUtils.java
```diff
@@ -34,26 +36,22 @@ public CosmosAdminTestUtils(Properties properties) {
...
-            .getTableMetadataDatabase()
+            .getMetadataDatabase()
             .orElse(CosmosAdmin.METADATA_DATABASE);
   }
 
   @Override
   public void dropMetadataTable() {
-    client.getDatabase(metadataDatabase).delete();
-    try {
-      client.getDatabase(metadataDatabase).read();
```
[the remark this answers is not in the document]
komamitsu: Ah, probably the behavior should be guaranteed by Cosmos DB side not by us. I think we can go without the subsequent `read()`.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
komamitsu: I saw similar codes in other classes and I guess you implemented this for consistency. If so, let's go with the consistent behavior for now.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -214,9 +218,31 @@ private ApplicationAutoScalingClient createApplicationAutoScalingClient(DynamoCo
...
-    // separator.
+  public void createNamespace(String nonPrefixedNamespace, Map<String, String> options)
+      throws ExecutionException {
+    Namespace namespace = Namespace.of(namespacePrefix, nonPrefixedNamespace);
+    try {
+      boolean noBackup = Boolean.parseBoolean(options.getOrDefault(NO_BACKUP, DEFAULT_NO_BACKUP));
+      createNamespacesTableIfNotExists(noBackup);
+      insertIntoNamespaceTable(namespace);
+    } catch (ExecutionException e) {
+      throw new ExecutionException("creating the namespace " + namespace + " failed", e);
```
komamitsu, line 229:
```suggestion
      throw new ExecutionException("Creating the namespace " + namespace + " failed", e);
```

core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosAdminTestUtils.java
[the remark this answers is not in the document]
Torch3333: Yes, that is my thinking as well. If something goes wrong in the `delete()` operation, then an exception should be thrown.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -1219,6 +1287,73 @@ public void importTable(String namespace, String table) {
...
+                      .build())
+              .provisionedThroughput(
+                  ProvisionedThroughput.builder()
+                      .readCapacityUnits(METADATA_TABLES_REQUEST_UNIT)
+                      .writeCapacityUnits(METADATA_TABLES_REQUEST_UNIT)
+                      .build())
+              .tableName(ScalarDbUtils.getFullTableName(metadataNamespace, NAMESPACES_TABLE))
+              .build());
+    } catch (Exception e) {
+      throw new ExecutionException("Creating the namespaces table failed", e);
```
komamitsu, line 1348: How about adding the namespace name in this error message just in case?

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -36,17 +40,24 @@ public class CassandraAdmin implements DistributedStorageAdmin {
...
   @VisibleForTesting static final String INDEX_NAME_PREFIX = "index";
-
   private final ClusterManager clusterManager;
+  private final String metadataKeyspace;
 
   @Inject
   public CassandraAdmin(DatabaseConfig config) {
     clusterManager = new ClusterManager(config);
+    CassandraConfig cassandraConfig = new CassandraConfig(config);
+    metadataKeyspace = cassandraConfig.getMetadataKeyspace().orElse(METADATA_KEYSPACE);
```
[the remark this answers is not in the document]
Torch3333: Good point, thanks!

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -60,7 +71,20 @@ public void createTable(
   @Override
   public void createNamespace(String namespace, Map<String, String> options)
       throws ExecutionException {
-    CreateKeyspace query = SchemaBuilder.createKeyspace(quoteIfNecessary(namespace));
+    try {
+      createKeyspace(namespace, options);
+      createMetadataKeyspaceIfNotExists();
+      createKeyspacesTableIfNotExists();
+      insertIntoKeyspacesTable(namespace);
+    } catch (IllegalArgumentException e) {
```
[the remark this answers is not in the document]
Torch3333: I probably wanted to bubble up the IllegalArgumentException instead of an `ExecutionException` for a deliberate reason, but I can't remember nor find the reason again. 
Thanks, I will remove it as it does not seem important anyway.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
Torch3333: > Does this mean the keyspace for metadata is removed if there is no namespace? If so, I think we can keep the namespace for metadata in this case.
If the user has no table and namespace in the storage anymore, I think it's more kind to the user to delete any trace of ScalarDB left by removing the metadata namespace and their tables. 
What is your concern about deleting the namespace for metadata?

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -466,25 +484,29 @@ private ThroughputProperties calculateThroughput(Map<String, String> options) {
...
+    try {
+      getNamespacesContainer()
+          .readItem(namespace, new PartitionKey(namespace), CosmosNamespace.class);
+      return true;
+    } catch (RuntimeException e) {
+      if (e instanceof CosmosException
+          && ((CosmosException) e).getStatusCode() == CosmosErrorCode.NOT_FOUND.get()) {
+        return false;
+      }
+      throw new ExecutionException("Checking if the namespace exists failed", e);
```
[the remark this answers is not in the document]
Torch3333: Good point, thank you. 
Any checked exception thrown by the `[Cosmos/Cassandra/Dynamoc/Jdbc]Admin` methods will be caught and rethrown by the `CheckedDistributedStorageAdmin` wrapper. All rethrown exception messages will contain the namespace or table name which failed. 
https://github.com/scalar-labs/scalardb/blob/ea4c543465ca430dcbfa0eb627f8036779049351/core/src/main/java/com/scalar/db/common/CheckedDistributedStorageAdmin.java#L202-L208

Considering that, adding the namespace in the error message is not necessary. What do you think?

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
komamitsu: Your idea sounds reasonable.

My comment above came from the following thoughts:
1. It increases the possibility of race conditions issue like this
   a. Request A is processing `dropNamespace` and just confirmed `KEYSPACES_TABLE` is empty in `dropKeyspacesTableIfEmpty`
   b. Request B is processing `createNamespace` and just created `metadataKeyspace` right after `a`
   c. Request A removes `metadataKeyspace`
   d. Request B fails to execute `insertIntoKeyspacesTable` due to removed `metadataKeyspace`
2. The checking if `KEYSPACES_TABLE` is empty is executed on every table deletion. It introduces additional duration especially in integration tests that may remove tables frequently
3. I don't think users think of an empty metadata table

`1` may not be matter since Request B can retry? And also `2` isn't critical issue. I think we can go with current behavior as my opinion above isn't strong.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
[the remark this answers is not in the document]
komamitsu: Thanks for checking that. 

> Any checked exception thrown by the [Cosmos/Cassandra/Dynamoc/Jdbc]Admin methods will be caught and rethrown by the CheckedDistributedStorageAdmin wrapper

It sounds okay for now. But `CosmosAdmin#namespaceExists` is a public method and it's not guaranteed that only `CheckedDistributedStorageAdmin.namespaceExists` calls `CosmosAdmin#namespaceExists` even in the future. 

This might be a matter of taste. I'm a bit paranoid about avoiding lack of information and prefer to add information just in case even if some is duplicated in log message. But either is fine with me based on your above check 💪

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -824,4 +848,87 @@ private String enclose(String name) {
...
+    }
+  }
+
+  private void insertIntoNamespacesTable(Connection connection, String namespaceName)
+      throws SQLException {
+    String insertStatement =
+        "INSERT INTO "
+            + encloseFullTableName(metadataSchema, NAMESPACES_TABLE)
+            + " VALUES ('"
+            + namespaceName
```
komamitsu, line 901: How about using prepared statement? Strictly speaking, this code **itself** seems to allow SQL injection. So either using prepared statement or some more SQL escape is needed, ideally.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
Torch3333: I see. Let's keep it like this.
 About 1, the Admin API is not designed to be thread-safe, so I don't think we should worry about this scenario.
 About 2, indeed, it adds an overhead when deleting table and namespace. In this case, we prioritize usability over adding a way for the user to delete the metadata namespace expressly.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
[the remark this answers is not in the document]
Torch3333: Right, that's a sensible concern. 
Almost all methods of the Admins are concerned with this, so I will revise everything in a separate PR.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -67,22 +69,27 @@ public JdbcAdmin(DatabaseConfig databaseConfig) {
     JdbcConfig config = new JdbcConfig(databaseConfig);
     rdbEngine = RdbEngineFactory.create(config);
     dataSource = JdbcUtils.initDataSourceForAdmin(config, rdbEngine);
-    metadataSchema = config.getTableMetadataSchema().orElse(METADATA_SCHEMA);
+    metadataSchema = config.getMetadataSchema().orElse(METADATA_SCHEMA);
```
komamitsu, line 72: This string is inserted into SQL statements. How about validating this to ensure it doesn't contain any quote or something?

For instance, malicious users could potentially pass `scalar.db.jdbc.metadata.schema=pg_user"; create table foobar (id int); --`.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -53,7 +53,8 @@ public class CosmosAdmin implements DistributedStorageAdmin {
   public static final String DEFAULT_NO_SCALING = "false";
```
brfrn169, line 53: It looks like `CosmosAdmin.databaseExists()` is no longer in use. Please remove it.

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
```diff
@@ -200,6 +207,48 @@ public void importTable(String namespace, String table) throws ExecutionExceptio
...
+  @Override
+  public Set<String> getNamespaceNames() throws ExecutionException {
+    // Only return existing namespaces that are listed in the namespace mapping configuration or
+    // when they belong to the default storage
+    //
+    // For example, if the storages contain the following namespaces :
+    // - mysql : mysqlStorageAdmin.getNamespaceNames() = [ns1, ns2]
+    // - cassandra : cassandraStorageAdmin.getNamespaceNames() = [ns3]
+    // - cosmos : cosmosStorageAdmin.getNamespaceNames() = [ns4, ns5]
+    // And the default storage is cassandra :
```
komamitsu, line 219:
```suggestion
    // And the default storage is cosmos :
```

brfrn169 (commented):
Left several comments. Actually, I haven't finished this review yet. I'll continue with it later. Thanks.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -74,13 +98,29 @@ public void createNamespace(String namespace, Map<String, String> options)
...
+
+    clusterManager.getSession().execute(queryString);
+  }
+
+  private void createMetadataKeyspaceIfNotExists() {
+    CreateKeyspace query =
+        SchemaBuilder.createKeyspace(quoteIfNecessary(metadataKeyspace)).ifNotExists();
+    Map<String, Object> replicationOptions = new HashMap<>();
+    replicationOptions.put("class", ReplicationStrategy.SIMPLE_STRATEGY.toString());
+    replicationOptions.put("replication_factor", "1");
```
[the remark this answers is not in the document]
Torch3333: If the replication factor should be configurable, should the replication strategy be too?

The way replication factor and strategy can be configured by the Schema Loader tool is very rough at the moment 
- only one strategy/replication factor can be set for all the tables in the schema file
- the network topology strategy cannot be configured properly. The datacenter names and a custom replication factor for each cannot be configured. (cf. the [code](https://github.com/scalar-labs/scalardb/blob/ea4c543465ca430dcbfa0eb627f8036779049351/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java#L70C1-L76))

So, if we want to configure the replication factor and strategy properly with the Schema Loader/Admin API, I think we need to rework this feature.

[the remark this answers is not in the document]
brfrn169: I just remembered that we've already had this discussion.

> only one common strategy/replication factor can be set for all the keyspaces in the schema file

Actually, we have the option to configure the strategy/replication factor for each table as follows:
https://github.com/scalar-labs/scalardb/blob/08dd2117cc088a079a8b2197a686b17229757825/schema-loader/sample/schema_sample.json#L23-L24

But, in fact, the configuration is for a keyspace. Thus, being able to configure it for each table seems inconsistent.
And, as you mentioned, we can't properly configure the network topology strategy for now. So, I agree that we need to rework this feature in the future.

Given the current state, would it make sense to apply the same logic (using `options` to determine the replication factor and strategy) from `createKeyspace()` to the creation of the namespace meta table in `createMetadataKeyspaceIfNotExists()` with the same limitation?

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
Torch3333: Suppose we sanitize the metadata schema config; doing the same for the keyspace, table, and index creation input parameters makes sense right?
Checking if the input does not contain illegal characters by ourselves seems to be a bad practice as the chance of doing something wrong or too unsafe are high. If we do it, relying on a library seems wiser but I don't know if that exists.
On the other hand, I am not sure if sanitizing the input for the Admin API is the responsibility of ScalarDB. I would think it is more the responsibility of the application using ScalarDB.

docs/schema-loader.md
brfrn169, line 1: Do we really need the changes of this file?

schema-loader/src/test/java/com/scalar/db/schemaloader/SchemaOperatorTest.java
```diff
@@ -440,20 +440,6 @@ public void alterTables_WithChangesOnTwoTables_ShouldCallAdminCorrectly()
     verifyNoMoreInteractions(storageAdmin);
   }
 
-  @Test
-  public void constructor_ShouldNotCreateAdmins() {
```
brfrn169, line 444: Why did you delete this test? Any changes related to this test?

brfrn169 (commented):
Left additional comments. Please take a look when you have time!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
komamitsu: Oh, I was confused, sorry. `scalar.db.jdbc.metadata.schema` comes from the configuration file and I don't think we need to consider it seriously like input parameters.

As for whether to validate all inputs, it depends. I experienced such a strict validation rule in some enterprise projects. It takes some time and effort, but it reduces uncertainty of systems. I think it's kinda tradeoff.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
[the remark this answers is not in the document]
Torch3333: > Given the current state, would it make sense to apply the same logic (using options to determine the replication factor and strategy) from createKeyspace() to the creation of the namespace meta table in createMetadataKeyspaceIfNotExists() with the same limitation?

Yes, it does. I will do that in this PR. 

Concerning updating the default replication factor to 3, I will do it in a separate PR since I need to update multiple integration test classes to overwrite the replication factor back to 1.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -36,17 +40,24 @@ public class CassandraAdmin implements DistributedStorageAdmin {
   public static final String REPLICATION_STRATEGY = "replication-strategy";
   public static final String COMPACTION_STRATEGY = "compaction-strategy";
   public static final String REPLICATION_FACTOR = "replication-factor";
+  public static final String METADATA_KEYSPACE = "scalardb";
+  public static final String KEYSPACES_TABLE = "keyspaces";
+  public static final String KEYSPACES_NAME_COL = "name";
```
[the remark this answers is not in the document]
Torch3333: Indeed, thanks!

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -571,9 +593,48 @@ public void importTable(String namespace, String table) {
...
+        ThroughputProperties.createManualThroughput(Integer.parseInt(DEFAULT_REQUEST_UNIT));
+    client.createDatabaseIfNotExists(metadataDatabase, manualThroughput);
+    client.getDatabase(metadataDatabase).createContainerIfNotExists(NAMESPACES_CONTAINER, "/id");
+  }
+
+  private CosmosContainer getNamespacesContainer() {
+    return client.getDatabase(metadataDatabase).getContainer(NAMESPACES_CONTAINER);
+  }
+
+  private boolean tableMetadataContainerExists() throws RuntimeException {
```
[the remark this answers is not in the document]
Torch3333: Thank you, I revised it in https://github.com/scalar-labs/scalardb/pull/989/commits/fc2def1faecf44e9de2ea88c07560db7693cb92a

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -571,9 +593,48 @@ public void importTable(String namespace, String table) {
...
+    return containerExists(metadataDatabase, TABLE_METADATA_CONTAINER);
+  }
+
+  private boolean namespacesContainerExists() throws RuntimeException {
```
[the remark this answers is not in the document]
Torch3333: Thank you, I revised it in https://github.com/scalar-labs/scalardb/pull/989/commits/fc2def1faecf44e9de2ea88c07560db7693cb92a

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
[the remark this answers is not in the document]
Torch3333: Thank you, I revised it in https://github.com/scalar-labs/scalardb/pull/989/commits/fc2def1faecf44e9de2ea88c07560db7693cb92a#diff-8075a0901e2d7e2e76931750324c27882e3bb8b70d312af111f1331b25363493

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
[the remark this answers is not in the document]
Torch3333: Thanks to a unit test I had written, I found out why I had this specific handling for the `IllegalArgumentException`. 
If the provided replication strategy in the creation options is not recognized, the underlying method `ReplicationStrategy.fromString()` will throw an `IllegalArgumentException`. 
Eventually, keeping it like this is better in my opinion. What do you think?

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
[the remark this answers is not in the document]
Torch3333: > Given the current state, would it make sense to apply the same logic (using options to determine the replication factor and strategy) from createKeyspace() to the creation of the namespace meta table in createMetadataKeyspaceIfNotExists() with the same limitation?

I revised it in https://github.com/scalar-labs/scalardb/pull/989/commits/fc2def1faecf44e9de2ea88c07560db7693cb92a

core/src/main/java/com/scalar/db/storage/cosmos/CosmosNamespace.java
Torch3333: Since this object by the Cosmos SDK to unmarshal the query result of the namespaces table using Jackson. I thought this setter method was used to initialize the `id` attribute but eventually, making this object immutable works fine. Having the default constructor is needed though.
Thank you. cf. https://github.com/scalar-labs/scalardb/pull/989/commits/fc2def1faecf44e9de2ea88c07560db7693cb92a#diff-8538e3bb29b97980fc88f869c47afc4c1215b0608f81fe0143652964cbeef9fa

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
Torch3333: I will revise this in another PR. 
cf. https://github.com/scalar-labs/scalardb/pull/989#discussion_r1286624632

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
Torch3333: Thank you, I switch to prepared statements for DML statements. 
I revised it https://github.com/scalar-labs/scalardb/pull/989#discussion_r1286624632

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
Torch3333: Thank you, I revised it in https://github.com/scalar-labs/scalardb/pull/989#discussion_r1286624632

core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosAdminTestUtils.java
```diff
@@ -1,20 +1,22 @@
-package com.scalar.db.storage.cosmos;
+package com.scalar.db.util;
```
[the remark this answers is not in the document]
Torch3333: Indeed, thank you. 
I revised it in https://github.com/scalar-labs/scalardb/pull/989/commits/7e670ed77b26389fd51c5e250f6c874bd8553770

core/src/integration-test/java/com/scalar/db/transaction/jdbc/JdbcTransactionAdminIntegrationTest.java
```diff
@@ -28,75 +27,8 @@ private boolean isSqlite() {
     return JdbcEnv.isSqlite();
   }
```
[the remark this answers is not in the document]
Torch3333: Indeed, thank you. 
I revised it in https://github.com/scalar-labs/scalardb/pull/989/commits/7e670ed77b26389fd51c5e250f6c874bd8553770

docs/schema-loader.md
Torch3333: No, we don't. 
Thank you, I revised it in https://github.com/scalar-labs/scalardb/pull/989/commits/7e670ed77b26389fd51c5e250f6c874bd8553770

core/src/main/java/com/scalar/db/storage/cosmos/CosmosNamespace.java
brfrn169: Oh I didn't know that we don't need setter methods for this kind of classes. So we also don't need setter methods in `CosmosTableMetadata` and `Record`?

brfrn169 (approved):
Left a comment. But LGTM, Thank you!

core/src/main/java/com/scalar/db/storage/cosmos/CosmosNamespace.java
Torch3333: @brfrn169 Good idea, let me try.

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
```diff
@@ -200,6 +207,48 @@ public void importTable(String namespace, String table) throws ExecutionExceptio
...
+    // And the default storage is cosmos :
+    // - scalar.db.multi_storage.default_storage=cosmos
+    // And the namespace mapping set in the configuration is :
+    // - scalar.db.multi_storage.namespace_mapping=ns1:mysql,ns2:cassandra,ns3:cassandra
+    //
+    // Then multiStorageAdmin.getNamespaceNames() = [ns1, ns3, ns4, ns5]
+    // The reasoning is:
+    // - ns1 is present in the mysql storage and listed in the mapping belonging to mysql
+    //     => returned
+    // - ns2 is present in the mysql storage but listed in the mapping belonging to cassandra
```
feeblefakie, line 228: Sorry, I don't fully get this. 
Is this a misconfiguration by users? (ns2 is defined in MySQL but the namespace mapping says it's mapped to Cassandra?)
It would be great if you could elaborate on this.

feeblefakie (commented):
Left one question.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosNamespace.java
Torch3333: @brfrn169 Removing the setters also works for `CosmosTableMetadata` and `Record`. I will push the changes in a separate PR.

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
Torch3333: > Is this a misconfiguration by users? (ns2 is defined in MySQL but the namespace mapping says it's mapped to Cassandra?)

It can be a misconfiguration or deliberate to do some testing (even though it's probably not safe to do so). 

This is a corner case that is probably unlikely to happen, but I included it in this explanation to highlight only namespaces strictly defined in the configuration are returned by this method.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -739,8 +770,43 @@ private void waitForTableDeletion(Namespace namespace, String tableName)
...
+      throw new ExecutionException(
+          "Dropping the namespace "
+              + Namespace.of(namespacePrefix, nonPrefixedNamespace)
+              + " failed",
+          e);
+    }
+  }
+
+  private void dropNamespacesTableIfEmpty() throws ExecutionException {
+    String NamespaceTableFullName =
```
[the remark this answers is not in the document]
Torch3333: Thank you, I revised it in https://github.com/scalar-labs/scalardb/pull/989/commits/38bd49de551b714efb84e8f9298484649e8578db

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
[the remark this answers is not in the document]
komamitsu: If ExecutionException shouldn't be thrown when a provided replication strategy isn't recognized, we should leave it as is. IMO, we may want to remove this `catch (IllegalArgumentException e)` otherwise. But this discussion won't block this PR from being merged and it's up to you.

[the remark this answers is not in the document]
Torch3333: Ok.

> If ExecutionException shouldn't be thrown when a provided replication strategy isn't recognized, we should leave it as is
Yes, I think it is more appropriate that an IllegalArgumentException is thrown from this method in that case.

komamitsu (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
feeblefakie: Let me clarify.
So, there are two apps that share the same ScalarDB Cluster.
One app (app1) defines ns1 and ns2 in its config, and the other (app2) defines ns3 and ns4 in its config.
But, the actual database defines ns1, ns2, ns3, and ns4.
In this case, app1 returns ns1 and ns2, and app2 returns ns3 and ns4 by `getNamespaceNames`?
I'm not sure if it is an expected behavior.

feeblefakie:
@brfrn169 @Torch3333  It's OK to merge, but I left one additional comment.
https://github.com/scalar-labs/scalardb/pull/989#discussion_r1299966755

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
Torch3333: > In this case, app1 returns ns1 and ns2, and app2 returns ns3 and ns4 by getNamespaceNames?

Correct. Yes, that is the expected behavior.

PR #1012, opened 2023-08-16
Create supported storages compatibility check workflow

Torch3333:
This adds **temporarily** a workflow only triggerable by dispatch event to check the compatibility of ScalarDB for all the supported database versions. (cf. the [compatibility matrix](https://scalardb.scalar-labs.com/docs/latest/scalardb-supported-databases/#compatibility-matrix)). This workflow is an enhanced copy of the CI workflow, which runs integration tests for all the versions of supported databases. 
I will use this workflow to test the compatibility of recent ScalarDB releases. I already detected some [failures](https://github.com/scalar-labs/scalardb/actions/runs/5873822757/job/15927627618) on the master branch, so I suspect older releases present the same failures (when applicable) too.

Once fixes are merged on the master and release branches, I will delete this workflow and port the changes to the CI workflow on the master and release branches.


.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+          arguments: integrationTestCassandra
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: cassandra_3.11_integration_test_reports
+          path: core/build/reports/tests/integrationTestCassandra
+
+  cosmos-integration-test:
```
brfrn169, line 183: Maybe we don't need this integration test for Cosmos since we already check it in the usual integration tests.

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+        run : find /home/azureuser/.gradle/daemon -name "*.out.log" -type f -mtime +3 -exec rm -vf {} +
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: cosmos_integration_test_reports
+          path: core/build/reports/tests/integrationTestCosmos
+
+  dynamo-integration-test:
```
brfrn169, line 217: Ditto. But I think we should upgrade the docker image version.

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+          arguments: integrationTestJdbc -Dscalardb.jdbc.url=jdbc:sqlserver://localhost:1433;encrypt=true;trustServerCertificate=true -Dscalardb.jdbc.username=sa -Dscalardb.jdbc.password=SqlServer22
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: jdbc_sqlserver_2022_integration_test_reports
+          path: core/build/reports/tests/integrationTestJdbc
+
+  jdbc-sqlite-integration-test:
```
brfrn169, line 600: Ditto.

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+          arguments: integrationTestJdbc -Dscalardb.jdbc.mariadb=true
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: jdbc_mariadb_integration_test_reports
+          path: core/build/reports/tests/integrationTestJdbc
+
+  multi-storage-integration-test:
```
brfrn169, line 659: Ditto.

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+          arguments: integrationTestMultiStorage
+
+      - name: Upload Gradle test reports
+        uses: actions/upload-artifact@v3
+        if : always()
+        with:
+          name: multi_storage_integration_test_reports
+          path: core/build/reports/tests/integrationTestMultiStorage
+
+  scalardb-server-integration-test:
```
brfrn169, line 698: Ditto.

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+          arguments: integrationTestJdbc -Dscalardb.jdbc.url=jdbc:sqlite:integration.sqlite3?busy_timeout=50000
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: jdbc_sqlite_integration_test_reports
+          path: core/build/reports/tests/integrationTestJdbc
+
+  jdbc-mariadb-integration-test:
```
brfrn169, line 627: @feeblefakie If we officially support MariaDB, we need to decide support versions. What do you think?

brfrn169 (commented):

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: Right, I will delete all the storage versions that are already listed in the CI.

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+        with:
+          name: cassandra_2.2_integration_test_reports
+          path: core/build/reports/tests/integrationTestCassandra
+
+  cassandra-3-0-integration-test:
+    runs-on: ubuntu-latest
+
+    services:
+      cassandra:
+        image: cassandra:3.0.29
```
komamitsu, line 122: How about specifying just `cassandra:3.0` so that we don't need to update the patch version frequently?

Torch3333: That is a good point, thank you. I will update it.

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: @feeblefakie (cc @brfrn169)
After discussing with @brfrn169, MariaDB only supports the major versions 10 and 11 presently, so we are considering supporting them both.
cf. the list of MariaDB version and their EOL https://endoflife.date/mariadb

I will revise the workflow to test with the latest `10.x` and `11.x` versions, but please let us know if you have any concerns.

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+          arguments: integrationTestJdbc -Dscalardb.jdbc.url=jdbc:postgresql://localhost:5432/ -Dscalardb.jdbc.username=postgres -Dscalardb.jdbc.password=postgres
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: jdbc_postgresql_13_integration_test_reports
+          path: core/build/reports/tests/integrationTestJdbc
+
+  jdbc-postgresql-14-integration-test:
```
brfrn169, line 377: It looks like PostgreSQL 15 is already released:
https://www.postgresql.org/support/versioning/

Can you please add this version to this CI?

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+          arguments: integrationTestCassandra
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: cassandra_3.0_integration_test_reports
+          path: core/build/reports/tests/integrationTestCassandra
+
+  cassandra-3-11-integration-test:
```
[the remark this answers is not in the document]
brfrn169: Currently, we don't support Cassandra 4.0 and 4.1.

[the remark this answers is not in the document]
komamitsu: I see. Just out of curiosity, is there any technical issue to prevent us from supporting the versions?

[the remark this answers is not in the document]
brfrn169: Yes, in fact, there is no backward compatibility between the APIs in Cassandra 3 and 4. To support Cassandra 4, we'll need to make some modifications to the existing code.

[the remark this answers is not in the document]
komamitsu: Thanks! Understood.

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: I kept this job as we are testing with the latest image version.

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: I removed this job.

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: I removed this job.

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: I removed this job.

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: I tested MariaDB 11 and MariaDB 10, and ScalarDB is not compatible with all versions :
- 10.0 to 10.3: ❌ it fails  
- 10.4 to 10.11: ✅ it works  
- 11.0: ❌ it fails  
- 11.1-rc: ✅ which is currently only available as a release candidate version, it is working. I suppose a production version will be released soon. 

For these reasons, we decided to only mention in the supported storage for now MariaDB 10 but not MariaDB 11. 
Once MariaDB 11.1 is released officially, I think we can add it to the list of supported versions.

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: I updated all the images in the workflow to use the latest major version besides Cassandra and Mysql, which use the latest minor version.

.github/workflows/supported_storages_compatibility_check.yaml
Torch3333: I added it.

.github/workflows/supported_storages_compatibility_check.yaml
```diff
@@ -0,0 +1,728 @@
...
+          arguments: integrationTestJdbc -Dscalardb.jdbc.url=jdbc:oracle:thin:@${{ env.oracle_db_ip }}:1521/XEPDB1 -Dscalardb.jdbc.username=SYSTEM -Dscalardb.jdbc.password=oracle
+
+      - name: Upload Gradle test reports
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: jdbc_oracle_18_integration_test_reports
+          path: core/build/reports/tests/integrationTestJdbc
+
+  jdbc-oracle-21-integration-test:
```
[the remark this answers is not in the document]
Torch3333: I added it. 
For this version, I generated an image that has already prebuilt the database. This means that compared to the mint image which needs to build the database (which takes 5-10min) booting up the image now only takes a few seconds (thanks to this readme https://github.com/oracle/docker-images/tree/main/OracleDatabase/SingleInstance/samples/prebuiltdb). The drawback is that the prebuilt image size weight 15GB instead of 6GB.

The goal is to see if the Oracle integration test will take less time or not to finish by using the prebuilt image. The image will boot faster but will take more time to download eventually.

Torch3333:
@brfrn169 @komamitsu 
Thank for the comments. The latest commit https://github.com/scalar-labs/scalardb/pull/1012/commits/d1efc0a43151306462bbf0ea9d4f56dd6ff7aa96 contains all the revisions. 
Please have a look again.

.github/workflows/supported_storages_compatibility_check.yaml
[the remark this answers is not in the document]
Torch3333: Besides, I added a job for the latest Mysql version: 8.1

komamitsu (approved):
LGTM! Thank you!

brfrn169 (approved):
LGTM! Thank you!

PR #1036, opened 2023-08-25
Fix how the indexed column is retrieved for Cassandra 2.2

Torch3333:
This fixes a problem that occurs only on Cassandra 2.2 when parsing table metadata.
Please take a look at the inline comments for more details.

Also, I ran the [compatibility check workflow](https://github.com/scalar-labs/scalardb/actions/runs/5996720560) to verify the integration test pass with all Cassandra versions. The Oracle and PostgreSQL failures will be resolved with this [fix](https://github.com/scalar-labs/scalardb/pull/1037).

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
@@ -6,6 +6,7 @@ import static com.scalar.db.util.ScalarDbUtils.getFullTableName;
 import com.datastax.driver.core.ClusteringOrder;
 import com.datastax.driver.core.ColumnMetadata;
 import com.datastax.driver.core.KeyspaceMetadata;
+import com.datastax.driver.core.ParseUtils;
 import com.datastax.driver.core.ResultSet;
 import com.datastax.driver.core.Row;
 import com.datastax.driver.core.querybuilder.QueryBuilder;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -229,7 +230,16 @@ private TableMetadata createTableMetadata(com.datastax.driver.core.TableMetadata
...
     }
-    metadata.getIndexes().forEach(i -> builder.addSecondaryIndex(i.getTarget()));
+    metadata
+        .getIndexes()
+        .forEach(
+            i -> {
+              // On Cassandra 2.2 only, the indexed column will be in a quoted format so we need to
+              // unquote it. For this purpose, we use the Cassandra SDK `ParseUtils.unDoubleQuote()`
+              // which reverses the effect of `Metadata.quoteIfNecessary()`
+              String indexedColumn = ParseUtils.unDoubleQuote(i.getTarget());
```
Torch3333, line 240: Suppose the indexed column is quoted because it contains non-alphanumeric characters or reserved keywords. In that case, the value returned by `i.getTarget()` will be : 
- with Cassandra 2.2 : the quoted column name such as `"select_col"`
- with Cassandra 3.X : the unquoted column name such as `select_col`
Since we use the unquoted column name in the ScalarDB `TableMetadata`, we unquote it by using a method of the Cassandra SDK.

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
Torch3333: Indeed, thank you. I am ok with phasing out Cassandra 2.2.
Also, Cassandra 3 EOL is planned for December this year so maybe we should consider supporting Cassandra 4 with ScalarDB 4.0.

feeblefakie: @brfrn169 Do you see any issues?
If you don't, let's make it unsupported.

brfrn169: @feeblefakie @Torch3333 Agreed. let's make Cassandra 2.2 unsupported. Thanks.

Torch3333: Understood. 
I will close this PR then.

PR #1037, opened 2023-08-25
Fix to handle version-specific integration tests

jnmt:
This PR introduces version-specific integration tests for the table importing feature to fix issues found in the supported storage compatibility tests. Although I also checked the new supported version (Oracle 23, MySQL 8.1, PostgreSQL 15, SQL Server 2022), I don't think there are updates related to supported data types for importing. You can see the latest results of compatibility tests [here](https://github.com/scalar-labs/scalardb/actions/runs/5974810663). Note that the failure in Cassandra's tests is not related to this PR.

By the way, I apparently found a bug that might happen when importing a table with minor data types in SQL Server, regardless of the versions. So, I will fix it in another PR.


.github/workflows/ci.yaml
```diff
@@ -240,7 +240,7 @@ jobs:
       - name: Setup and execute Gradle 'integrationTestJdbc' task
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: integrationTestJdbc -Dscalardb.jdbc.url=jdbc:postgresql://localhost:5432/ -Dscalardb.jdbc.username=postgres -Dscalardb.jdbc.password=postgres
+          arguments: integrationTestJdbc -Dscalardb.jdbc.url=jdbc:postgresql://localhost:5432/ -Dscalardb.jdbc.username=postgres -Dscalardb.jdbc.password=postgres -Dscalardb.jdbc.database.version.major=14
```
jnmt, line 243: When you don't specify versions, it will work with the version = max integer. It can work without specifying the version in most cases if there is no critical change around the version in the database. But, specifying version is now recommended for PostgreSQL and Oracle just in case (and might also be for MySQL and SQL Server in the future).

Torch3333: Instead of having to set the database's major version as an external parameter, it seems we can retrieve this data using the JDBC driver with 
```
dataSource.getConnection().getMetaData().getDatabaseProductVersion()
```
cf. the [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getDatabaseMajorVersion--) 
What do you think ?

jnmt: @Torch3333 Thank you! I didn't know it. Let me check how it works.

jnmt: @Torch3333 I could get the expected versions using the JDBC driver except for MariaDB. Although MariaDB returns a bit confusing name and versions due to (maybe) the use of MySQL driver, currently, we don't care about versions for compatibility tests with MySQL/MariaDB. So, I will go with your suggestion.

FYI: the returned versions from each database are as follows.

```sh
# Name: ProductName, Version: ProductVersion, Major: MajorVersion, Minor: MinorVersion
Name: MySQL, Version: 8.1.0, Major: 8, Minor: 1
Name: MySQL, Version: 5.5.5-10.11.5-MariaDB-1:10.11.5+maria~ubu2204, Major: 5, Minor: 5
Name: PostgreSQL, Version: 14.9 Major: 14 Minor: 9
Name: Oracle, Version: Oracle Database 18c Express Edition Release 18.0.0.0.0 - Production Version 18.4.0.0.0, Major: 18, Minor: 0
Name: Microsoft SQL Server, Version: 16.00.4065, Major: 16 Minor: 0
```

jnmt: Fixed in 5feddc5.

Torch3333 (approved):
LGTM, thank you.

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -435,4 +393,165 @@ private String prepareCreateTableSql(
...
+  }
+
+  private int getMajorVersion() {
+    try (BasicDataSource dataSource = JdbcUtils.initDataSourceForAdmin(config, rdbEngine);
+        Connection connection = dataSource.getConnection()) {
+      return connection.getMetaData().getDatabaseMajorVersion();
+    } catch (SQLException e) {
+      throw new RuntimeException("Get database major version failed");
+    }
+  }
```
brfrn169, line 556: 👍

brfrn169 (approved):
LGTM! Thank you!

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
brfrn169, line 556: 👍

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -435,4 +393,165 @@ private String prepareCreateTableSql(
...
+    return results;
+  }
+
+  private boolean isMariaDB() {
+    try (BasicDataSource dataSource = JdbcUtils.initDataSourceForAdmin(config, rdbEngine);
+        Connection connection = dataSource.getConnection()) {
+      String version = connection.getMetaData().getDatabaseProductVersion();
+      return version.contains("MariaDB");
+    } catch (SQLException e) {
+      throw new RuntimeException("Get database major version failed");
```
Torch3333, line 545:
```suggestion
      throw new RuntimeException("Get database product version failed");
```
I missed that.

feeblefakie (approved):
LGTM! Thank you!

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
jnmt: Good catch! Thank you! Fixed in 7cb850a.

komamitsu (approved):
LGTM! 👍

PR #1045, opened 2023-08-30
Fix ScanBuilder bug

brfrn169:
I encountered a bug where ScanBuilder doesn't create a proper `ScanAll` instance when we call `projection()`/`projections()`/`ordering()`/`orderings()`/`limit()`/`consistency()` before calling `where()`. I'll add inline comments for the details. Please take a look!


jnmt (approved):
LGTM! Thank you for the fix. It's my mad. I should have added more tests and noticed it when fixing a similar bug before.

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #1046, opened 2023-08-30
Add like operator

jnmt:
This PR adds the like operator for the relational scan. In the LIKE conditional expression, you can specify an SQL-LIKE-like pattern and an escape character (optional). The following specification of wild cards and escape characters is basically the same as PostgreSQL and MySQL.

- `_` matches any single character
- `%` matches any sequence of zero or more characters
- `\` works as the escape character by default
- Escape character can be disabled by specifying an empty escape character


core/src/main/java/com/scalar/db/api/LikeExpression.java
```diff
@@ -0,0 +1,152 @@
...
+  public String getEscape() {
+    return escape;
+  }
+
+  /**
+   * Returns true if the LIKE pattern matches a given {@code String}.
+   *
+   * @return true if the LIKE pattern matches a given {@code String}
+   */
+  public boolean isMatchedWith(String value) {
```
jnmt, line 107: We need our own matcher in the ScalarDB layer to avoid scan-after-write properly in `Snapshot`. The matcher uses regular expressions converted from "like" patterns.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -309,4 +310,25 @@ public Driver getDriver() {
       throw new RuntimeException(e);
     }
   }
+
+  @Override
+  public String getPattern(LikeExpression likeExpression) {
+    String escape = likeExpression.getEscape();
+    String pattern = likeExpression.getTextValue();
+    if (escape.isEmpty()) {
+      // MySQL accepts an empty escape character to disable the escape function, but MariaDB ignores
```
jnmt, line 319: We need a tricky conversion for MySQL and MariaDB to handle both in the same manner since we don't want to distinguish them at this moment. But, in the future, we probably support the MariaDB driver and introduce `RdbEngineMariadb` or something like that.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSqlServer.java
```diff
@@ -276,4 +277,28 @@ public String computeBooleanValue(boolean value) {
   public Driver getDriver() {
     return new com.microsoft.sqlserver.jdbc.SQLServerDriver();
   }
+
+  @Override
+  public String getPattern(LikeExpression likeExpression) {
+    String escape = likeExpression.getEscape();
+    String pattern = likeExpression.getTextValue();
+    if (escape.isEmpty()) {
+      // Even if users do not want to use escape character in ScalarDB (i.e., specifying "" for
```
jnmt, line 286: We also need special care for SQL Server due to its dialect.

core/src/main/java/com/scalar/db/api/LikeExpression.java
```diff
@@ -0,0 +1,152 @@
...
+import com.scalar.db.io.TextColumn;
+import java.util.Objects;
+import java.util.regex.Pattern;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class LikeExpression extends ConditionalExpression {
+  private static final String DEFAULT_ESCAPE_CHAR = "\\";
+  @Nonnull private final String escape;
```
komamitsu, line 13: [trivial] IIUC, instance variables are regarded as non-null by default and we can remove this `@Nonnull`...?

core/src/main/java/com/scalar/db/api/LikeExpression.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /**
+   * Constructs a {@code LikeExpression} with the specified column, operator and escape character.
+   *
+   * @param column a target column used to compare
+   * @param operator an operator used to compare the target column
+   * @param escape an escape character for the like operator
+   */
+  LikeExpression(TextColumn column, Operator operator, String escape) {
```
komamitsu, line 33: It seems `escape` is basically used as a single length string (or empty.) How about simply treating it as a nullable character? It would work as same as empty string if it's null.

```java
  @Nullable private final Character escape;
```

core/src/main/java/com/scalar/db/api/LikeExpression.java
```diff
@@ -0,0 +1,152 @@
...
+  LikeExpression(TextColumn column, Operator operator) {
+    this(column, operator, DEFAULT_ESCAPE_CHAR);
+  }
+
+  /**
+   * Constructs a {@code LikeExpression} with the specified column, operator and escape character.
+   *
+   * @param column a target column used to compare
+   * @param operator an operator used to compare the target column
+   * @param escape an escape character for the like operator
```
komamitsu, line 31: Passing empty string looks to disable escapes. How about describing the behavior?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
komamitsu: This is just an idea, but how about giving up disabling escape on ScalarDB side like MariaDB does? In this way, users would need to use the default escape or specify a single escape character, but can't specify an empty string.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
```diff
@@ -309,4 +310,10 @@ public String computeBooleanValue(boolean value) {
   public Driver getDriver() {
     return new oracle.jdbc.driver.OracleDriver();
   }
+
+  @Override
+  public String getEscape(LikeExpression likeExpression) {
+    String escape = likeExpression.getEscape();
+    return escape.isEmpty() ? null : escape;
```
komamitsu, line 317: Is this also needed in `RdbEnginePostgresql`?

Moving this logic to `com.scalar.db.storage.jdbc.RdbEngineStrategy#getEscape` might be an option.

core/src/main/java/com/scalar/db/api/LikeExpression.java
jnmt: That's a good point. I also considered using `Character` as you suggested. But, in that case, users need to pass `null` to disable the escape character, right? (e.g., `.isLikeText("%xyz", null)`) This is because we want to use the default escape char `\` for `.isLikeText("%xyz")`. I prefer passing an empty string "" rather than null since its semantics is the same as SQL. What do you think?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
jnmt: No, we always need the escape clause for PostgreSQL (if it's empty, we need to return the empty string as is). But for Oracle and SQLite, we need to remove the escape clause to disable the escape character.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
jnmt: Thank you for the idea. That's one of the options I thought. However, all other databases other than MariaDB support disabling the escape, and it's probably convenient for users. So, can we go with it as is?

core/src/main/java/com/scalar/db/api/LikeExpression.java
komamitsu: > users need to pass null to disable the escape character, right? (e.g., .isLikeText("%xyz", null)) 
Yes.

I understand your point to use consistent semantics with SQL. I think I was confused a bit because `com.scalar.db.storage.jdbc.RdbEngineStrategy#getEscape` returns null while `com.scalar.db.api.LikeExpression#getEscape` returns an empty. How about making `RdbEngineStrategy#getEscape` return an empty as well as `LikeExpression#getEscape`?

komamitsu: After reading your comment https://github.com/scalar-labs/scalardb/pull/1046#discussion_r1311219596, I think my suggestion above was wrong... Please ignore it.

komamitsu: > I prefer passing an empty string "" rather than null since its semantics is the same as SQL.
👍 

Maybe https://github.com/scalar-labs/scalardb/pull/1046#discussion_r1311085517 is also related. It would be great if you take care of it.

jnmt: > Maybe https://github.com/scalar-labs/scalardb/pull/1046#discussion_r1311085517 is also related. It would be great if you take care of it.

Sure. I will revise it. Thanks for the feedback.

core/src/main/java/com/scalar/db/api/LikeExpression.java
```diff
@@ -0,0 +1,157 @@
...
+   * literally. If an escape character specified, escaping is done for '_', '%', and the escape
+   * character itself. An invalid pattern will throw {@code IllegalArgumentException}. This method
+   * is implemented referencing the following Spark SQL implementation.
+   * https://github.com/apache/spark/blob/a8eadebd686caa110c4077f4199d11e797146dc5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
+   *
+   * @param likePattern a SQL LIKE pattern to convert
+   * @param escape an escape character.
+   * @return the equivalent Java regular expression of the given pattern
+   */
+  private String convertRegexPatternFrom(String likePattern, @Nullable Character escape) {
```
brfrn169, line 65: Can we move this logic into `Snapshot` since it's only used there?

core/src/main/java/com/scalar/db/api/LikeExpression.java
```diff
@@ -0,0 +1,157 @@
...
+  public String getEscape() {
+    return escape;
+  }
+
+  /**
+   * Returns true if the LIKE pattern matches a given {@code String}.
+   *
+   * @return true if the LIKE pattern matches a given {@code String}
+   */
+  public boolean isMatchedWith(String value) {
```
brfrn169, line 112: Ditto. Can we move this logic into `Snapshot` since it's only used there?

core/src/main/java/com/scalar/db/storage/jdbc/query/PreparedStatementBinder.java
```diff
@@ -133,6 +134,19 @@ public void visit(BlobColumn column) {
     }
   }
 
+  public void bindLikeClause(LikeExpression likeExpression) {
```
brfrn169, line 137: Hmm, we can't use the visitor pattern only for the LIKE operation. Honestly, I don't find it ideal, but I'm out of alternatives, and maybe it's acceptable...

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
```diff
@@ -233,10 +248,13 @@ public void bind(PreparedStatement preparedStatement) throws SQLException {
 
     for (Conjunction conjunction : conjunctions) {
       for (ConditionalExpression condition : conjunction.getConditions()) {
-        if (!condition.getColumn().hasNullValue()) {
+        if ((condition.getOperator().equals(Operator.LIKE)
+            || condition.getOperator().equals(Operator.NOT_LIKE))) {
+          binder.bindLikeClause((LikeExpression) condition);
+        } else if (!condition.getColumn().hasNullValue()) {
```
brfrn169, line 254: This isn't directly related to this PR, but do we need to bind a NULL value when `condition.getColumn().hasNullValue()` returns true, similar to what `PreparedStatementBinder` does?

brfrn169 (commented):

core/src/main/java/com/scalar/db/api/LikeExpression.java
jnmt: The regular expression is only used in `Snapshot`, so technically, yes. But I think we should check if a specified pattern is an expected one because throwing an exception with "pattern must not ..." from `Snapshot` is weird. So, the choices are 1) keeping this as is and 2) only adding the check here. IMHO, 1) is not so bad since the check and conversion are tightly coupled with. What do you think?
p.s. `isMatchedWith` can be moved to `Snapshot` in any case.

core/src/main/java/com/scalar/db/storage/jdbc/query/PreparedStatementBinder.java
jnmt: Yeah, we have the same feelings...

core/src/main/java/com/scalar/db/storage/jdbc/query/SimpleSelectQuery.java
jnmt: We don't need bindings for the columns with null values in conjunctions because the string `IS (NOT) NULL` is used for `Operator.IS(_NOT)_NULL`. But I'm not sure we should also use bind markers for them. Do you have any idea about that?

brfrn169: Yeah, that makes sense. We don't need to bind null values in conjunctions. Thanks.

core/src/main/java/com/scalar/db/api/LikeExpression.java
brfrn169: I think we should choose the `2)` option. We should separate the regular expression thing from `LikeExpression` because the responsibility of the regular expression thing is not for `LikeExpression` but Consensus Commit (e.g., `JdbcTransaction` doesn't use the regular expression). What do you think?

jnmt: > because the responsibility of the regular expression thing is not for LikeExpression but Consensus Commit (e.g., JdbcTransaction doesn't use the regular expression).

Right. Totally agreed. Thanks for pointing it out!

core/src/main/java/com/scalar/db/api/LikeExpression.java
```diff
@@ -0,0 +1,152 @@
...
+   * literally. If an escape character specified, escaping is done for '_', '%', and the escape
+   * character itself. An invalid pattern will throw {@code IllegalArgumentException}. This method
+   * is implemented referencing the following Spark SQL implementation.
+   * https://github.com/apache/spark/blob/a8eadebd686caa110c4077f4199d11e797146dc5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
+   *
+   * @param likePattern a SQL LIKE pattern to convert
+   * @param escape an escape character.
+   * @return the equivalent Java regular expression of the given pattern
+   */
+  private String convertRegexPatternFrom(String likePattern, Character escape) {
```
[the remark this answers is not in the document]
jnmt: Fixed in 9654df0, but removed when fixing others in 0c14b83.

core/src/main/java/com/scalar/db/api/LikeExpression.java
jnmt: Fixed in 9654df0.

core/src/main/java/com/scalar/db/api/LikeExpression.java
jnmt: Fixed in 9654df0.

komamitsu (approved):
LGTM! 👍

core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosConditionalMutationIntegrationTest.java
```diff
@@ -37,6 +37,9 @@ protected Map<String, String> getCreationOptions() {
   protected List<OperatorAndDataType> getOperatorAndDataTypeListForTest() {
```
brfrn169, line 37: This is a minor suggestion. I think we can modify this in the following way, ensuring we won't need any adjustments for similar changes in the future:

```Java
  @Override
  protected List<OperatorAndDataType> getOperatorAndDataTypeListForTest() {
    return super.getOperatorAndDataTypeListForTest().stream()
        .filter(
            operatorAndDataType -> {
              // Cosmos DB only supports the 'equal' and 'not equal' and 'is null' and 'is not null'
              // conditions for BLOB type
              if (operatorAndDataType.getDataType() == DataType.BLOB) {
                return operatorAndDataType.getOperator() == Operator.EQ
                    || operatorAndDataType.getOperator() == Operator.NE
                    || operatorAndDataType.getOperator() == Operator.IS_NULL
                    || operatorAndDataType.getOperator() == Operator.IS_NOT_NULL;
              }

              return true;
            })
        .collect(Collectors.toList());
  }
```

core/src/integration-test/java/com/scalar/db/storage/dynamo/DynamoConditionalMutationIntegrationTest.java
```diff
@@ -27,6 +27,9 @@ protected Map<String, String> getCreationOptions() {
   protected List<OperatorAndDataType> getOperatorAndDataTypeListForTest() {
```
brfrn169, line 27: Ditto.

```java
  @Override
  protected List<OperatorAndDataType> getOperatorAndDataTypeListForTest() {
    return super.getOperatorAndDataTypeListForTest().stream()
        .filter(
            operatorAndDataType -> {
              // DynamoDB only supports the 'equal' and 'not equal' and 'is null' and 'is not null'
              // conditions for BOOLEAN type
              if (operatorAndDataType.getDataType() == DataType.BOOLEAN) {
                return operatorAndDataType.getOperator() == Operator.EQ
                    || operatorAndDataType.getOperator() == Operator.NE
                    || operatorAndDataType.getOperator() == Operator.IS_NULL
                    || operatorAndDataType.getOperator() == Operator.IS_NOT_NULL;
              }

              return true;
            })
        .collect(Collectors.toList());
  }
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -523,6 +530,64 @@ public boolean isValidationRequired() {
...
+   * https://github.com/apache/spark/blob/a8eadebd686caa110c4077f4199d11e797146dc5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
+   *
+   * @param likePattern a SQL LIKE pattern to convert
+   * @param escape an escape character.
+   * @return the equivalent Java regular expression of the given pattern
+   */
+  private String convertRegexPatternFrom(String likePattern, @Nullable Character escape) {
+    if (likePattern == null) {
+      throw new IllegalArgumentException("LIKE pattern must not be null");
+    }
```
brfrn169, line 562: Since the validation for likePattern is already done here, I think using assertions is more appropriate:

```suggestion
    assert likePattern != null : "LIKE pattern must not be null";
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -523,6 +530,64 @@ public boolean isValidationRequired() {
...
+    for (int i = 0; i < chars.length; i++) {
+      char c = chars[i];
+      if (escape != null && c == escape && i + 1 < chars.length) {
+        char nextChar = chars[++i];
+        if (nextChar == '_' || nextChar == '%') {
+          out.append(Pattern.quote(Character.toString(nextChar)));
+        } else if (nextChar == escape) {
+          out.append(Pattern.quote(Character.toString(nextChar)));
+        } else {
+          throw new IllegalArgumentException("LIKE pattern must not include only escape character");
```
brfrn169, line 575: Ditto.
```suggestion
          throw new AssertionError("LIKE pattern must not include only escape character");
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -523,6 +530,64 @@ public boolean isValidationRequired() {
...
+        }
+      } else if (escape != null && c == escape) {
+        throw new IllegalArgumentException("LIKE pattern must not end with escape character");
```
brfrn169, line 578:
```suggestion
        throw new AssertionError("LIKE pattern must not end with escape character");
```

brfrn169 (approved):
Left several minor suggestions, but LGTM! Thank you!

jnmt:
@brfrn169 Thank you for the additional feedback! I fixed all in 4aabb08.

feeblefakie (approved):
LGTM! Thank you!

core/src/test/java/com/scalar/db/storage/jdbc/query/QueryBuilderTest.java
```diff
@@ -481,6 +481,204 @@ public void selectQueryWithConjunctionsTest(RdbEngine rdbEngineType) throws SQLE
...
+    verify(preparedStatement).setString(11, "text\\%");
+    verify(preparedStatement).setString(12, "");
+  }
+
+  @ParameterizedTest
+  @EnumSource(
+      value = RdbEngine.class,
+      names = {"ORACLE", "SQLITE"},
+      mode = EnumSource.Mode.INCLUDE)
+  public void selectQueryWithLikeConditionsTestForOracle(RdbEngine rdbEngineType)
```
Torch3333, line 589:
```suggestion
  @ParameterizedTest
  @EnumSource(
      value = RdbEngine.class,
      names = {"ORACLE", "SQLITE"},
      mode = EnumSource.Mode.INCLUDE)
  public void selectQueryWithLikeConditionsTestForOracleAndSqlLite(RdbEngine rdbEngineType)
```
Since the test targets also SqlLite ?

Torch3333 (approved):
Besides a minor suggestion, LGTM, thank you!

core/src/test/java/com/scalar/db/storage/jdbc/query/QueryBuilderTest.java
jnmt: Right. Applied it. Thank you!

PR #1055, opened 2023-09-01
Fix to handle supported data types properly when importing tables in MySQL and SQL Server

jnmt:
This PR fixes handling supported data types correctly. Without this fix, users can accidentally import some minor unsupported data types like "geometry" as the blob type in ScalarDB.

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
@@ -235,9 +235,8 @@ class RdbEngineMysql implements RdbEngineStrategy {
       case CHAR:
       case VARCHAR:
       case LONGVARCHAR:
-        if (typeName.equalsIgnoreCase("ENUM")
-            || typeName.equalsIgnoreCase("SET")
-            || typeName.equalsIgnoreCase("JSON")) {
+        if (!typeName.toUpperCase().endsWith("CHAR") && !typeName.toUpperCase().endsWith("TEXT")) {
+          // to exclude ENUM, SET, JSON, etc.
           throw new IllegalArgumentException(
               String.format("Data type %s is unsupported: %s", typeName, columnDescription));
         }
--- a/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSqlServer.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSqlServer.java
@@ -223,14 +223,16 @@ class RdbEngineSqlServer implements RdbEngineStrategy {
         return DataType.TEXT;
       case BINARY:
       case VARBINARY:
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSqlServer.java
```diff
@@ -223,14 +223,16 @@ public DataType getDataTypeForScalarDb(
         return DataType.TEXT;
       case BINARY:
       case VARBINARY:
-        if (typeName.equalsIgnoreCase("timestamp") || typeName.equalsIgnoreCase("hierarchyid")) {
+        if (!typeName.equalsIgnoreCase("binary") && !typeName.equalsIgnoreCase("varbinary")) {
```
jnmt, line 226: This is the main part of this fix. A black-list-based exclusion sometimes makes us miss the unsupported data types. So, we only accept the expected ones.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -235,9 +235,8 @@ public DataType getDataTypeForScalarDb(
       case CHAR:
       case VARCHAR:
       case LONGVARCHAR:
-        if (typeName.equalsIgnoreCase("ENUM")
-            || typeName.equalsIgnoreCase("SET")
-            || typeName.equalsIgnoreCase("JSON")) {
+        if (!typeName.toUpperCase().endsWith("CHAR") && !typeName.toUpperCase().endsWith("TEXT")) {
+          // to exclude ENUM, SET, JSON, etc.
```
jnmt, line 239: Changed to a more white-list-based check for the same reason in the SQL Server case below.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSqlServer.java
```diff
@@ -223,14 +223,16 @@ public DataType getDataTypeForScalarDb(
...
           throw new IllegalArgumentException(
               String.format("Data type %s is unsupported: %s", typeName, columnDescription));
         }
-        logger.info(
-            "Data type larger than that of underlying database is assigned: {} ({} to BLOB)",
-            columnDescription,
-            typeName);
+        if (columnSize < Integer.MAX_VALUE) {
```
jnmt, line 230: I missed that SQL server users can specify `varbinary(max)` for the binary data larger than 8000 bytes (`varbinary(8000)`), and it can hold the same size as ScalarDB BLOB. In that case, the message is unnecessary.

brfrn169 (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #1069, opened 2023-09-04
Fix flaky relational scan integration tests

jnmt:
This PR fixes a flaky test issue when running relational scan integration tests in parallel. For relational scan integration tests, we use a different table schema with regular integration tests. So, when running those two integration tests,  they rarely misidentify their tables (and tests fail) since the namespace and table name are the same. I faced it when testing them (inherited ones) for ScalarDB cluster. This PR fixes the issue.


brfrn169 (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM, thank you!

Torch3333 (approved):
LGTM, thank you!

PR #1082, opened 2023-09-12
Revise configuration-related docs

josh-wong:
## Description

This PR revises configuration-related documents.

## Related issues and/or PRs

N/A

## Changes made

- Revised content.
- Added notice blocks for noteworthy content.
- Added content tabs for databases (in `configurations.md`).
- Added Mermaid code for some diagrams (in `configurations.md`).

## Branches this PR applies to

> **Note**
> 
> The `configurations.md` doc includes mentions of ScalarDB Server, which I don't think we should include in branches `3.10` and `3.9`. We'll need to make a separate PR to address the differences between including ScalarDB Server in `3.8` to `3.5` and not including ScalarDB Server in `3.9`, `3.10`, and `master`.

- ~~`master`~~
- `3.8`
- `3.7`
- `3.6`
- `3.5`

## Testing done

I ran our docs site locally with these changes and confirmed that these revised docs, including content tabs, notice blocks, images, Mermaid diagrams, and formatting, appeared as expected. Attached are three screenshots for each doc in this PR.

The following screenshot shows **configurations.md**:

configurations

The following screenshot shows **two-phase-commit-transactions.md**:

two-phase-commit-transactions

The following screenshot shows **multi-storage-transactions.md**:

multi-storage-transactions

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] The documentation has been updated to reflect the changes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

Please see my comments in the revised files.

PR #1084, opened 2023-09-15
Revise the contents of the Java API guide

josh-wong:
## Description

This PR revises the contents of the ScalarDB Java API Guide, mostly from an editorial perspective.

## Related issues and/or PRs

Related to #1082 (some similarities in content)

## Changes made

- Revised content.
- Added notice blocks for noteworthy content.
- Added content tabs for databases.

## Branches this PR applies to

- `master`

> **Note**
> 
> After this PR is approved and merged, I will create another PR for branches `3.10`, `3.9`, `3.8`, `3.7`, and `3.6` that removes the **Get existing namespaces** section, which doesn't apply to those branches.

## Testing done

I ran our docs site locally with these changes and confirmed that these revised docs, including content tabs, notice blocks, and formatting, appeared as expected. Attached is a screenshot of the `api-guide.md` doc as it appears on the docs site.

The following screenshot shows `api-guide.md`:

API guide

## Checklist

Please see my comments in the revised files.

### DIFF
--- a/core/src/main/java/com/scalar/db/api/Admin.java
+++ b/core/src/main/java/com/scalar/db/api/Admin.java
@@ -378,19 +378,21 @@ public interface Admin {
    * Add a new column to an existing table. The new column cannot be a partition or clustering key.
    * <br>
    * <br>
-   * <strong>Attention:</strong> You should carefully consider adding a new column to a table because the
-   * execution time may vary greatly depending on the underlying storage. Please plan accordingly and
-   * consider the following, especially if the database runs in production:
+   * <strong>Attention:</strong> You should carefully consider adding a new column to a table
+   * because the execution time may vary greatly depending on the underlying storage. Please plan
+   * accordingly and consider the following, especially if the database runs in production:
    *
    * <ul>
-   *   <li><strong>For Cosmos DB for noSQL and DynamoDB:</strong> Adding a column is almost instantaneous as the
-   *       table schema is not modified. Only the table metadata stored in a separate table is updated.
-   *   <li><strong>For Cassandra:</strong> Adding a column will only update the schema metadata and will not 
-   *       modify the existing schema records. The cluster topology is the main factor for the execution time.
-   *       Changes to the schema metadata are shared to each cluster node via a gossip protocol. Because of this,
-   *       the larger the cluster, the longer it will take for all nodes to be updated.
-   *   <li><strong>For relational databases (MySQL, Oracle, etc.):</strong> Adding a column shouldn't take a long
-   *       time to execute.
+   *   <li><strong>For Cosmos DB for noSQL and DynamoDB:</strong> Adding a column is almost
+   *       instantaneous as the table schema is not modified. Only the table metadata stored in a
+   *       separate table is updated.
+   *   <li><strong>For Cassandra:</strong> Adding a column will only update the schema metadata and
+   *       will not modify the existing schema records. The cluster topology is the main factor for
+   *       the execution time. Changes to the schema metadata are shared to each cluster node via a
+   *       gossip protocol. Because of this, the larger the cluster, the longer it will take for all
+   *       nodes to be updated.
+   *   <li><strong>For relational databases (MySQL, Oracle, etc.):</strong> Adding a column
+   *       shouldn't take a long time to execute.
    * </ul>
    *
    * @param namespace the table namespace
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -250,45 +250,47 @@ You can get table metadata as follows:
 TableMetadata tableMetadata = admin.getTableMetadata("ns", "tbl");
 ```
 
-### Specify operations for coordinator tables
+### Specify operations for Coordinator tables
 
-Depending on the transaction manager type that you are using, you might need to create coordinator tables to execute transactions. In addition to creating tables, you can truncate and drop coordinator tables.
+Coordinator tables are used by the [Transactional API](#transactional-api) to track the statuses of transactions.
 
-#### Create coordinator tables
+When using a transaction manager, you must create Coordinator tables to execute transactions. In addition to creating tables, you can truncate and drop Coordinator tables.
 
-You can create coordinator tables as follows:
+#### Create Coordinator tables
+
+You can create Coordinator tables as follows:
 
 ```java
-// Create coordinator tables.
+// Create Coordinator tables.
 admin.createCoordinatorTables();
 
-// Create coordinator tables only if they do not already exist.
+// Create Coordinator tables only if they do not already exist.
 boolean ifNotExist = true;
 admin.createCoordinatorTables(ifNotExist);
 
-// Create coordinator tables with options.
+// Create Coordinator tables with options.
 Map<String, String> options = ...;
 admin.createCoordinatorTables(options);
 ```
 
-#### Truncate coordinator tables
+#### Truncate Coordinator tables
 
-You can truncate coordinator tables as follows:
+You can truncate Coordinator tables as follows:
 
 ```java
-// Truncate coordinator tables.
+// Truncate Coordinator tables.
 admin.truncateCoordinatorTables();
 ```
 
-#### Drop coordinator tables
+#### Drop Coordinator tables
 
-You can drop coordinator tables as follows:
+You can drop Coordinator tables as follows:
 
 ```java
-// Drop coordinator tables.
+// Drop Coordinator tables.
 admin.dropCoordinatorTables();
 
-// Drop coordinator tables if they exist.
+// Drop Coordinator tables if they exist.
 boolean ifExist = true;
 admin.dropCoordinatorTables(ifExist);
 ```
@@ -811,7 +813,7 @@ transaction.mutate(Arrays.asList(put, delete));
 
 #### Default namespace for CRUD operations
 
-A default namespace for all CRUD operations can be set by using a property in the ScalarDB configuration. If you would like to use this setting with ScalarDB Server, you will need to configure the setting on the client-side configuration.
+A default namespace for all CRUD operations can be set by using a property in the ScalarDB configuration.
 
 ```properties
 scalar.db.default_namespace_name=<NAMESPACE_NAME>
@@ -939,7 +941,7 @@ public class Sample {
 
         // For `CrudConflictException`, `CommitConflictException`, and `TransactionNotFoundException`,
         // you can basically retry the transaction. However, for the other exceptions, the transaction
-        // may still fail if the cause of the exception is non-transient. In such a case, you will 
+        // will still fail if the cause of the exception is non-transient. In such a case, you will 
         // exhaust the number of retries and throw the last exception.
 
         if (transaction != null) {
@@ -964,6 +966,8 @@ The `begin()` API could throw `TransactionException` or `TransactionNotFoundExce
 
 - If you catch `TransactionException`, this exception indicates that the transaction has failed to begin due to transient or non-transient faults. You can try retrying the transaction, but you may not be able to begin the transaction due to non-transient faults.
 - If you catch `TransactionNotFoundException`, this exception indicates that the transaction has failed to begin due to transient faults. In this case, you can retry the transaction.
+
+The `join()` API could also throw `TransactionNotFoundException`. You can handle this exception in the same way that you handle the exceptions for the `begin()` API.
 
 ### `CrudException` and `CrudConflictException`
 
@@ -992,7 +996,7 @@ How to identify a transaction status is delegated to users. You may want to crea
 
 Although not illustrated in the example code, the `resume()` API could also throw `TransactionNotFoundException`. This exception indicates that the transaction associated with the specified ID was not found and/or the transaction might have expired. In either case, you can retry the transaction from the beginning since the cause of this exception is basically transient.
 
-In the sample code, for `UnknownTransactionStatusException`, the transaction is not retried because the cause of the exception is non-transient. For other exceptions, the transaction is retried because the cause of the exception is transient or non-transient. If the cause of the exception is transient, the transaction may succeed if you retry it. However, if the cause of the exception is non-transient, the transaction may still fail even if you retry it. In such a case, you will exhaust the number of retries.
+In the sample code, for `UnknownTransactionStatusException`, the transaction is not retried because the cause of the exception is non-transient. For other exceptions, the transaction is retried because the cause of the exception is transient or non-transient. If the cause of the exception is transient, the transaction may succeed if you retry it. However, if the cause of the exception is non-transient, the transaction will still fail even if you retry it. In such a case, you will exhaust the number of retries.
 
 {% capture notice--info %}
 **Note**
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -552,7 +552,7 @@ Optional<Result> result = transaction.get(get);
 {% capture notice--info %}
 **Note**
 
-If the result has more than one record, the `transaction.get()` will throw an exception. If you want to handle multiple results, see [Execute `Scan` by using a secondary index](#execute-scan-by-using-a-secondary-index).
+If the result has more than one record, `transaction.get()` will throw an exception. If you want to handle multiple results, see [Execute `Scan` by using a secondary index](#execute-scan-by-using-a-secondary-index).
 
 {% endcapture %}
 
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -1,6 +1,6 @@
 # ScalarDB Java API Guide
 
-The ScalarDB Java API is mainly composed of the Administrative API and Transactional API. This guide briefly explains what kinds of APIs exist, how to use them, and related topics like handling exceptions.
+The ScalarDB Java API is mainly composed of the Administrative API and Transactional API. This guide briefly explains what kinds of APIs exist, how to use them, and related topics like how to handle exceptions.
 
 ## Administrative API
 
@@ -164,8 +164,8 @@ You can add a new, non-partition key column to a table as follows:
 admin.addNewColumnToTable("ns", "tbl", "c6", DataType.INT)
 ```
 
-{% capture notice--info %}
-**Note**
+{% capture notice--warning %}
+**Attention**
 
 You should carefully consider adding a new column to a table because the execution time may vary greatly depending on the underlying storage. Please plan accordingly and consider the following, especially if the database runs in production:
 
@@ -174,7 +174,7 @@ You should carefully consider adding a new column to a table because the executi
 - **For relational databases (MySQL, Oracle, etc.):** Adding a column may take a very long time to execute. In addition, a table lock may occur.
 {% endcapture %}
 
-<div class="notice--info">{{ notice--info | markdownify }}</div>
+<div class="notice--warning">{{ notice--warning | markdownify }}</div>
 
 ### Truncate a table
 
@@ -351,12 +351,16 @@ DistributedTransaction transaction = transactionManager.start("<TRANSACTION_ID>"
 {% capture notice--info %}
 **Note**
 
-You must guarantee the uniqueness of the transaction ID in this case.
+Specifying a transaction ID is useful when you want to link external systems to ScalarDB. Otherwise, you should use the `start()` method.
+
+When you specify a transaction ID, make sure you specify a unique ID (for example, UUID v4) throughout the system since ScalarDB depends on the uniqueness of transaction IDs for correctness.
 {% endcapture %}
 
 <div class="notice--info">{{ notice--info | markdownify }}</div>
 
 ### Join a transaction
+
+Joining a transaction is particularly useful in a stateful application where a transaction spans multiple client requests. In such a scenario, the application can start a transaction during the first client request. Then, in subsequent client requests, the application can join the ongoing transaction by using the `join()` method.
 
 You can join an ongoing transaction that has already begun by specifying the transaction ID as follows:
 
@@ -365,9 +369,21 @@ You can join an ongoing transaction that has already begun by specifying the tra
 DistributedTransaction transaction = transactionManager.join("<TRANSACTION_ID>");
 ```
 
-Joining a transaction is particularly useful in a stateful application where a transaction spans multiple client requests. In such a scenario, the application can start a transaction during the first client request. Then, in subsequent client requests, the application can join the ongoing transaction by using the `join()` method.
+{% capture notice--info %}
+**Note**
+
+You must guarantee the uniqueness of the transaction ID in this case. To get the transaction ID with `getId()`, you can specify the following:
+
+```java
+tx.getId();
+```
+{% endcapture %}
+
+<div class="notice--info">{{ notice--info | markdownify }}</div>
 
 ### Resume a transaction
+
+Resuming a transaction is particularly useful in a stateful application where a transaction spans multiple client requests. In such a scenario, the application can start a transaction during the first client request. Then, in subsequent client requests, the application can resume the ongoing transaction by using the `resume()` method.
 
 You can resume an ongoing transaction that you have already begun by specifying a transaction ID as follows:
 
@@ -376,7 +392,17 @@ You can resume an ongoing transaction that you have already begun by specifying 
 DistributedTransaction transaction = transactionManager.resume("<TRANSACTION_ID>");
 ```
 
-Resuming a transaction is particularly useful in a stateful application where a transaction spans multiple client requests. In such a scenario, the application can start a transaction during the first client request. Then, in subsequent client requests, the application can resume the ongoing transaction by using the `resume()` method.
+{% capture notice--info %}
+**Note**
+
+You must guarantee the uniqueness of the transaction ID in this case. To get the transaction ID with `getId()`, you can specify the following:
+
+```java
+tx.getId();
+```
+{% endcapture %}
+
+<div class="notice--info">{{ notice--info | markdownify }}</div>
 
 ### Implement CRUD operations
 
@@ -880,7 +906,7 @@ If you don't handle exceptions properly, you may face anomalies or data inconsis
 
 <div class="notice--warning">{{ notice--warning | markdownify }}</div>
 
-The following example code shows how to handle exceptions:
+The following sample code shows how to handle exceptions:
 
 ```java
 public class Sample {
@@ -994,16 +1020,14 @@ How to identify a transaction status is delegated to users. You may want to crea
 
 ### Notes about some exceptions
 
-Although not illustrated in the example code, the `resume()` API could also throw `TransactionNotFoundException`. This exception indicates that the transaction associated with the specified ID was not found and/or the transaction might have expired. In either case, you can retry the transaction from the beginning since the cause of this exception is basically transient.
+Although not illustrated in the sample code, the `resume()` API could also throw `TransactionNotFoundException`. This exception indicates that the transaction associated with the specified ID was not found and/or the transaction might have expired. In either case, you can retry the transaction from the beginning since the cause of this exception is basically transient.
 
 In the sample code, for `UnknownTransactionStatusException`, the transaction is not retried because the cause of the exception is non-transient. For other exceptions, the transaction is retried because the cause of the exception is transient or non-transient. If the cause of the exception is transient, the transaction may succeed if you retry it. However, if the cause of the exception is non-transient, the transaction will still fail even if you retry it. In such a case, you will exhaust the number of retries.
 
 {% capture notice--info %}
 **Note**
 
-If you begin a transaction by specifying a transaction ID, you must use a different ID when you retry the transaction.
-
-In addition, in the sample code, the transaction is retried three times maximum and sleeps for 100 milliseconds before it is retried. But you can choose a retry policy, such as exponential backoff, according to your application requirements.
+In the sample code, the transaction is retried three times maximum and sleeps for 100 milliseconds before it is retried. But you can choose a retry policy, such as exponential backoff, according to your application requirements.
 {% endcapture %}
 
 <div class="notice--info">{{ notice--info | markdownify }}</div>
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -395,7 +395,7 @@ DistributedTransaction transaction = transactionManager.resume("<TRANSACTION_ID>
 {% capture notice--info %}
 **Note**
 
-You must guarantee the uniqueness of the transaction ID in this case. To get the transaction ID with `getId()`, you can specify the following:
+To get the transaction ID with `getId()`, you can specify the following:
 
 ```java
 tx.getId();
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -63,6 +63,7 @@ Select your database to see the options available:
   <button class="tablinks" onclick="openTab(event, 'Cassandra', 'tabset-1')" id="defaultOpen-1">Cassandra</button>
   <button class="tablinks" onclick="openTab(event, 'Cosmos_DB_for_NoSQL', 'tabset-1')">Cosmos DB for NoSQL</button>
   <button class="tablinks" onclick="openTab(event, 'DynamoDB', 'tabset-1')">DynamoDB</button>
+  <button class="tablinks" onclick="openTab(event, 'JDBC_databases', 'tabset-1')">JDBC databases</button>
 </div>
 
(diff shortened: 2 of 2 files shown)


docs/api-guide.md
```diff
@@ -32,60 +36,71 @@ admin.close();
...
 Map<String, String> options = ...;
 admin.createNamespace("ns", options);
 ```
 
-#### Creation Options
+#### Creation options
 
-In the creation operations (creating a namespace, creating a table, etc.), you can specify options that are maps of option names and values (`Map<String, String>`).
-With the options, we can set storage adapter specific configurations.
+In the creation operations, like creating a namespace and creating a table, you can specify options that are maps of option names and values (`Map<String, String>`). By using the options, you can set storage adapter–specific configurations.
```
josh-wong, line 57: Currently, we have options for Cassandra, Cosmos DB for NoSQL, and DynamoDB. Do we need to add options for JDBC databases?

docs/api-guide.md
```diff
@@ -216,177 +231,191 @@ Set<String> namespaces = admin.getNamespaceNames();
...
+```java
+// Start a transaction with specifying a transaction ID.
+DistributedTransaction transaction = transactionManager.start("<TRANSACTION_ID>");
 ```
 
-Note that you must guarantee uniqueness of the transaction ID in this case.
+{% capture notice--info %}
+**Note**
+
+You must guarantee the uniqueness of the transaction ID in this case.
```
josh-wong, line 346: Should we mention how to get the transaction ID like we do in [`two-phase-commmit-transactions.md` (commit in PR #1082)](https://github.com/scalar-labs/scalardb/blob/f4360781940685d5c9bc86daa425fca870ae7f2f/docs/two-phase-commit-transactions.md?plain=1#L93)?

````suggestion
You must guarantee the uniqueness of the transaction ID in this case. To get the transaction ID with `getId()`, you can specify the following:

```java
tx.getId();
```
````

File: docs/api-guide.md [hunk 3]
```diff
@@ -417,62 +446,61 @@ Get get =
...
+And if you need to check if a value of a column is null, you can use the `isNull("<COLUMN_NAME>")` method.
 
 ``` java
 // Check if a value of a column is null
-boolean isNull = result.isNull("<column name>");
+boolean isNull = result.isNull("<COLUMN_NAME>");
 ```
 
-Please see also [Javadoc of `Result`](https://javadoc.io/static/com.scalar-labs/scalardb/3.6.0/com/scalar/db/api/Result.html) for more details.
+For more details, see the Javadoc for [`Result`](https://javadoc.io/static/com.scalar-labs/scalardb/3.10.1/com/scalar/db/api/Result.html).
```
josh-wong, line 494: The link in the following sentence must be updated when versions are updated and should be specific to the branch that this doc is in. For example, since this doc exists in branch `3.6`, this link should point to the latest version of the doc for ScalarDB 3.6 (https://javadoc.io/doc/com.scalar-labs/scalardb/3.6.7/index.html)

I couldn't find a way around needing to change this link, so any other user-friendly suggestions are welcome!

docs/api-guide.md
```diff
@@ -746,13 +803,12 @@ Delete delete =
 transaction.mutate(Arrays.asList(put, delete));
 ```
 
-#### Use a default namespace for CRUD operations
+#### Default namespace for CRUD operations
 
-A default namespace for all the CRUD operations can be set with a property of the ScalarDB configuration.
-If you would like to use this setting with ScalarDB server, it needs to be set on the client-side configuration.
+A default namespace for all CRUD operations can be set by using a property in the ScalarDB configuration. If you would like to use this setting with ScalarDB Server, you will need to configure the setting on the client-side configuration.
```
josh-wong, line 808: I think we need to change this sentence for ScalarDB 3.9 and later docs since ScalarDB Server has been deprecated in ScalarDB 3.9 and later. 

In 3.9 and later docs, can we delete this sentence? Or should we reword it to mention ScalarDB Cluster instead?

docs/api-guide.md
```diff
@@ -696,28 +753,28 @@ Delete delete =
...
 ```java
-// Build a deleteIfExists condition
+// Build a `deleteIfExists` condition
 MutationCondition deleteIfExistsCondition = ConditionBuilder.deleteIfExists();
 ```
 
 #### Mutate operation
 
-Mutate is an operation to execute multiple mutations (Put and Delete operations).
+Mutate is an operation to execute multiple mutations (`Put` and `Delete` operations).
```
josh-wong, line 772: Should the part in parentheses say "for example"?

```suggestion
Mutate is an operation to execute multiple mutations (for example, `Put` and `Delete` operations).
```

Or, are there only two operations that can be used for mutations here?

docs/api-guide.md
```diff
@@ -894,60 +944,59 @@ public class Sample {
...
-The `begin()` API could throw `TransactionException` or `TransactionNotFoundException`.
-If you catch `TransactionException`, it indicates that the transaction has failed to begin due to transient or nontransient faults. You can try retrying the transaction, but you may not be able to begin the transaction due to nontransient faults.
-If you catch `TransactionNotFoundException`, it indicates that the transaction has failed to begin due to transient faults. You can retry the transaction.
+### `TransactionException` and `TransactionNotFoundException`
+
+The `begin()` API could throw `TransactionException` or `TransactionNotFoundException`:
+
+- If you catch `TransactionException`, this exception indicates that the transaction has failed to begin due to transient or non-transient faults. You can try retrying the transaction, but you may not be able to begin the transaction due to non-transient faults.
+- If you catch `TransactionNotFoundException`, this exception indicates that the transaction has failed to begin due to transient faults. In this case, you can retry the transaction.
+
```
josh-wong, line 953: Should we mention how the `join()` API could also throw exceptions like we do in [`two-phase-commmit-transactions.md`](https://github.com/scalar-labs/scalardb/blob/f4360781940685d5c9bc86daa425fca870ae7f2f/docs/two-phase-commit-transactions.md?plain=1#L648C14-L648C14)?

> **Note**
> 
> **Reference:** The commit above is from PR #1082.

```suggestion

The `join()` API could also throw `TransactionException` or `TransactionNotFoundException`. You can handle these exceptions in the same way that you handle the exceptions for the `begin()` API.

```

docs/api-guide.md
komamitsu: How about https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html ?

docs/api-guide.md
komamitsu: > Or, are there only two operations that can be used for mutations here?

I think so. Other operations are read operations (`Selection`) `Get` and `Scan`.

docs/api-guide.md
```diff
@@ -894,60 +944,59 @@ public class Sample {
...
-But you can choose a retry policy, such as exponential backoff, according to your application requirements.
+### Notes about some exceptions
 
-## Transactional operations for Two-phase Commit Transaction
+Although not illustrated in the example code, the `resume()` API could also throw `TransactionNotFoundException`. This exception indicates that the transaction associated with the specified ID was not found and/or the transaction might have expired. In either case, you can retry the transaction from the beginning since the cause of this exception is basically transient.
 
-Please see [Two-phase Commit Transactions](two-phase-commit-transactions.md).
+In the sample code, for `UnknownTransactionStatusException`, the transaction is not retried because the cause of the exception is non-transient. For other exceptions, the transaction is retried because the cause of the exception is transient or non-transient. If the cause of the exception is transient, the transaction may succeed if you retry it. However, if the cause of the exception is non-transient, the transaction may still fail even if you retry it. In such a case, you will exhaust the number of retries.
```
komamitsu, line 981: The reason why `UnknownTransactionStatusException` shouldn't be retried without checking if the transaction succeeded is to avoid potential duplicated operations (e.g., multiple records creation), IIUC. So, this description sounds a bit weird to me.  @brfrn169 What do you think?

docs/api-guide.md
```diff
@@ -124,43 +137,45 @@ admin.createTable("ns", "tbl", tableMetadata, options);
...
-- For relational databases (MySQL, Oracle, etc.): it may take a very long time to execute and a 
-      table-lock may be performed.
+{% capture notice--info %}
+**Note**
+
+You should carefully consider adding a new column to a table because the execution time may vary greatly depending on the underlying storage. Please plan accordingly and consider the following, especially if the database runs in production:
+
+- **For Cosmos DB for NoSQL and DynamoDB:** Adding a column is almost instantaneous as the table schema is not modified. Only the table metadata stored in a separate table is updated.
+- **For Cassandra:** Adding a column will only update the schema metadata and will not modify the existing schema records. The cluster topology is the main factor for the execution time. Changes to the schema metadata are shared to each cluster node via a gossip protocol. Because of this, the larger the cluster, the longer it will take for all nodes to be updated.
+- **For relational databases (MySQL, Oracle, etc.):** Adding a column may take a very long time to execute. In addition, a table lock may occur.
```
choplin, line 168: This is not a direct comment on this PR, but is this the case if a newly added column is set to NULLable and defaults to NULL? I suppose underlying RDBMSs normally take not so long to add a NULL column.

docs/api-guide.md
```diff
@@ -216,177 +231,191 @@ Set<String> namespaces = admin.getNamespaceNames();
...
+You can get table metadata as follows:
 
 ```java
-// Get a table metadata of "ns.tbl"
+// Get the table metadata for "ns.tbl".
 TableMetadata tableMetadata = admin.getTableMetadata("ns", "tbl");
 ```
 
-### Operations for Coordinator tables
+### Specify operations for coordinator tables
```
choplin, line 247: I guess it would be helpful to explain what the coordinator table is here. Users would not know the coordinator table, and it could be hard for users to consider whether they should create a coordinator table or not.

docs/api-guide.md
```diff
@@ -216,177 +231,191 @@ Set<String> namespaces = admin.getNamespaceNames();
...
 
-Depending on the transaction manager type, you need to create coordinator tables to execute transactions. 
-The following items describe the operations for the coordinator table.
+Depending on the transaction manager type that you are using, you might need to create coordinator tables to execute transactions. In addition to creating tables, you can truncate and drop coordinator tables.
 
-#### Create Coordinator tables
+#### Create coordinator tables
```
choplin, line 251: I guess it should be capitalized because Coordinator is a kind of proper noun in ScalarDB. But, we should discuss it with @brfrn169 and @feeblefakie.

docs/api-guide.md
choplin: If the passed transaction ID is unique, it probably works properly regardless of whether it was obtained by calling `getID` on the existing transaction. For example, passing a randomly generated UUID would be fine as well.

But, I think it would not be a good document for users since there are few use cases with such a randomly generated ID. Also, uniqueness is an internal matter and wouldn't involve users.

IMO, we should note first what is a use-case for passing a transaction ID when starting a new transaction, then explain how to do it as we do in the section for `join` or `resume`

If there isn't a particular use case, we may delete the section.

docs/api-guide.md
```diff
@@ -773,117 +829,111 @@ Scan scanUsingSpecifiedNamespace =
...
-Please see [Handle exceptions](#handle-exceptions) for the details of how to handle exceptions in ScalarDB.
+For details about how to handle exceptions in ScalarDB, see [How to handle exceptions](#how-to-handle-exceptions).
 
-## Handle exceptions
+## How to handle exceptions
 
-Handling exceptions correctly in ScalarDB is very important.
-If you mishandle exceptions, your data could become inconsistent.
-This document explains how to handle exceptions properly in ScalarDB.
+When executing a transaction, you will also need to handle exceptions properly. If you don't handle exceptions properly, you may face anomalies or data inconsistency.
```
choplin, line 865: IMO, since anomalies or data inconsistency, are critical issues for ScalarDB users, it would be worth emphasizing those risks. (with `notice` section?)

choplin (commented):
Nice work! It looks much better, especially the section explaining exception handling. 

I left several comments, PTAL!

docs/api-guide.md
josh-wong: > How about https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html ?

@komamitsu Thank you for your input. Initially, I wanted to avoid using a link to the latest version of the Javadoc, but maybe we can compromise by guiding the reader as follows. What do you think?

```suggestion
For more details, see the `Result` page in the [Javadoc](https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html) of the version of ScalarDB that you're using.
```

docs/api-guide.md
josh-wong: > > Or, are there only two operations that can be used for mutations here?
> 
> I think so. Other operations are read operations (`Selection`) `Get` and `Scan`.

Got it. Thanks! I'll leave this part as is.

docs/api-guide.md
josh-wong: I see. Thank you! I've added a tab for JDBC databases that states no options are available.

docs/api-guide.md
```diff
@@ -32,60 +36,71 @@ admin.close();
...
+| no-backup  | Disable continuous backup for DynamoDB. | false   |
+| ru         | Base resource unit.                     | 10      |
+
+</div>
+</div>
 
 ### Create a table
 
-Next, we will discuss table creation.
+When creating a table, you should define the table metadata and then create the table.
```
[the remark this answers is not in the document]
josh-wong: > IMO, table metadata is not a commonly-defined word, so it would be better to explain what table metadata is in ScalarDB here.

@choplin Good idea! How about we use the following definition?:

> **Note**
> 
> The following is an excerpt from the **Table metadata** section in the blog post [ScalarDB Admin API & Schema Loader](https://medium.com/scalar-engineering/scalardb-admin-api-schema-loader-b0ad3809d4ea).

```suggestion
When creating a table, you should define the table metadata and then create the table.

Table metadata is ScalarDB-specific internal metadata for managing the schema of the user’s table. ScalarDB checks if the targeted metadata for the table exists and verifies that the name, type, and function of each column match the content in the `scalardb.metadata` database.
```

docs/api-guide.md
```diff
@@ -169,10 +184,10 @@ admin.truncateTable("ns", "tbl");
 You can drop a secondary index as follows:
 
 ```java
-// Drop a secondary index on a column "c5" of a table "ns.tbl". It will throw an exception if the secondary index does not exist
+// Drop the secondary index on column "c5" from table "ns.tbl". If a secondary index does not exist, an exception will be thrown.
```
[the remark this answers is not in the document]
josh-wong: Ah, good catch! I've changed this to "the secondary index."

docs/api-guide.md
komamitsu: Seems less maintenance! 💯

josh-wong: Cool, I'll follow this method when referencing other Javadocs in other docs👍

docs/api-guide.md
josh-wong: Great idea! I've added an "Attention" block since this is important for users to know. The following is how the "Attention" block will look on the docs site.

how-to-handle-exceptions-attention-block

Some other docs have the same sentence, so I'll make the same change in those docs.

docs/api-guide.md
josh-wong: Good point🤔

I'm not exactly sure which transaction managers (other than [Consensus Commit](https://scalardb.scalar-labs.com/docs/latest/configurations/#basic-configurations)) need to have coordinator tables. Would the following be accurate? Or are there any other transaction managers we should include? (cc: @brfrn169)

> **Note**
> 
> The following is an excerpt from the "Overview of the Admin API" section in the blog post [ScalarDB Admin API & Schema Loader](https://medium.com/scalar-engineering/scalardb-admin-api-schema-loader-b0ad3809d4ea).

```suggestion
### Specify operations for coordinator tables

If you are using the Consensus Commit transaction manager, you must create coordinator tables to execute transactions. Coordinator tables are used by the [Transactional API](#transactional-api) to track the statuses of transactions.
```

Torch3333 (commented):
Overall, it looks good, thank you. I have two minor comments.

docs/api-guide.md
```diff
@@ -32,60 +36,77 @@ admin.close();
...
-In the creation operations (creating a namespace, creating a table, etc.), you can specify options that are maps of option names and values (`Map<String, String>`).
-With the options, we can set storage adapter specific configurations.
+<div id="Cassandra" class="tabcontent" markdown="1">
 
-Currently, we can set the following options for the storage adapters:
+| Name                 | Description                                                                            | Default          |
+|----------------------|----------------------------------------------------------------------------------------|------------------|
+| replication-strategy | Cassandra replication strategy. Must be `SimpleStrategy` or `NetworkTopologyStrategy`. | `SimpleStrategy` |
+| compaction-strategy  | Cassandra compaction strategy, Must be `LCS`, `STCS` or `TWCS`.                        | `STCS`           |
+| replication-factor   | Cassandra replication factor.                                                          | 1                |
```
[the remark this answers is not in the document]
josh-wong: Thank you for updating that!

When backporting to 3.X branches, I'll be sure to change `replication-factor` to `1`.

docs/api-guide.md
```diff
@@ -543,20 +581,26 @@ Scan scan =
...
+{% capture notice--info %}
+**Note**
+
+You can't specify clustering-key boundaries and orderings in `Scan` by using a secondary index.
+{% endcapture %}
+
+<div class="notice--info">{{ notice--info | markdownify }}</div>
 
-##### Scan without a partition key to retrieve all the records of a table
+##### Execute a `Scan` with specifying a partition key to retrieve all the records of a table
```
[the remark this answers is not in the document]
josh-wong: Good catch!

docs/api-guide.md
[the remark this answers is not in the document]
choplin: Looks great! Thanks!

Torch3333 (approved):
LGTM, thank you!

docs/api-guide.md
choplin: I think so, but I'm not 100% sure. I would ask @brfrn169 to check this.

docs/api-guide.md
brfrn169: We can delete this sentence. Thanks.

docs/api-guide.md
brfrn169: I don't think adding a NULL column takes a long time.

@Torch3333 Do you have any comment ton this?

docs/api-guide.md
brfrn169: > The reason why UnknownTransactionStatusException shouldn't be retried without checking if the transaction succeeded by the application is to avoid potential duplicated operations (e.g., multiple records creation), IIUC.

This is correct.

However, I'm actually not sure if "non-transient" is the most appropriate term to use in this context.

@josh-wong What do you think?

docs/api-guide.md
brfrn169: Actually, `DistributedTransactionManager.join()` only throws `TransactionNotFoundException`. Can you please mention this?

docs/api-guide.md
Torch3333: Not really; I don't know the answer to that.

josh-wong: Thanks for bringing this up. 

Since it sounds like RDBMSs normally don't take long to add a NULL column, how about we change this sentence as follows?:

```suggestion
- **For relational databases (MySQL, Oracle, etc.):** Adding a column shouldn't take a long time to execute. 
```

Or, do we still need to mention `In addition, a table lock may occur`?

docs/api-guide.md
josh-wong: Thank you for bringing this up!

After discussing this with @feeblefakie and @brfrn169, we agreed that Coordinator is a component in ScalarDB and should therefore be capitalized.

docs/api-guide.md
josh-wong: Thank you for the explanation. I think we have some other documents that include a note similar to this one.

@brfrn169 Are there any particular use cases that we should mention? Or are you OK with deleting this note? If we should delete this note, should we remove it from the other documents too?

docs/api-guide.md
josh-wong: Understood. 

I've added a sentence about that below this bulleted list. I've also updated the sentence in [two-phase-commit-transactions.md](https://github.com/scalar-labs/scalardb/blob/32a6191ff24f391175cf78217afc967086f80558/docs/two-phase-commit-transactions.md?plain=1#L654). PTAL!

docs/api-guide.md
josh-wong: @komamitsu Thank you for the explanation!

@brfrn169 Thank you for confirming! I agree, I don't think "non-transient" is the right word here. How about if we change the first sentence as follows?:

```suggestion
In the sample code, for `UnknownTransactionStatusException`, the transaction is not retried because the application must check if the transaction was successful to avoid potential duplicate operations. For other exceptions, the transaction is retried because the cause of the exception is transient or non-transient. If the cause of the exception is transient, the transaction may succeed if you retry it. However, if the cause of the exception is non-transient, the transaction will still fail even if you retry it. In such a case, you will exhaust the number of retries.
```

komamitsu (approved):
LGTM, thank you!

docs/api-guide.md
josh-wong: As discussed with @brfrn169, all transaction managers need Coordinator tables. I've updated this sentence to reflect that. PTAL!

docs/api-guide.md
```diff
@@ -894,60 +960,61 @@ public class Sample {
...
 
-This configuration is only available to troubleshoot Consensus Commit transactions. By adding the following configuration, `Get` and `Scan` operations results will contain [transaction metadata](schema-loader.md#internal-metadata-for-consensus-commit).
-To see the transaction metadata columns details for a given table, you can use the `DistributedTransactionAdmin.getTableMetadata()` method which will return the table metadata augmented with the transaction metadata columns. 
-All in all, using this configuration can be useful to investigate transaction related issues.
+{% capture notice--info %}
+**Note**
+
+If you begin a transaction by specifying a transaction ID, you must use a different ID when you retry the transaction.
```
feeblefakie, line 1004: I think we should probably say this in the `begin` section (as I commented) instead of here.

feeblefakie (commented):
Left several suggestions and comments. PTAL!

docs/api-guide.md
brfrn169: @josh-wong Sounds good. Thanks!

brfrn169: @josh-wong Can you please apply the same change to the Two-phase commit transaction document?
https://github.com/scalar-labs/scalardb/pull/1082/files#diff-2677f880f1a9e512f31cce8e12950863ddffd4be49408d310d72d4d73e6122b9R697

docs/api-guide.md
```diff
@@ -216,177 +237,193 @@ Set<String> namespaces = admin.getNamespaceNames();
...
 ```java
-// Get a table metadata of "ns.tbl"
+// Get the table metadata for "ns.tbl".
 TableMetadata tableMetadata = admin.getTableMetadata("ns", "tbl");
 ```
 
-### Operations for Coordinator tables
+### Specify operations for Coordinator tables
+
+Coordinator tables are used by the [Transactional API](#transactional-api) to track the statuses of transactions.
```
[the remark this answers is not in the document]
brfrn169: Actually, we use plural forms in APIs, such as `admin.createCoordinatorTables()`, because I anticipated that we would have multiple coordinator tables in the future, like lock tables for `EXTRA_WRITE`, in addition to the state table. For the sake of consistency, should we continue to use plural forms?

docs/api-guide.md
brfrn169: @josh-wong The same sentence is in the following:

https://github.com/scalar-labs/scalardb/blob/docs/revise-api-guide/docs/storage-abstraction.md#add-a-new-column-to-a-table
https://github.com/scalar-labs/scalardb/blob/docs/revise-api-guide/core/src/main/java/com/scalar/db/api/Admin.java#L392-L393

Can you please apply the same change to them?

docs/api-guide.md
brfrn169: @josh-wong We should discuss the `getId()` matter in the `Join a transaction` section:
https://github.com/scalar-labs/scalardb/pull/1084/files#diff-d2b4776d2d9b9950fc1f58e9dd0e38550a526976d80e2232f4cb2dcfb84171b2R359

It might also be worth mentioning in the `Resume a transaction` section:
https://github.com/scalar-labs/scalardb/pull/1084/files#diff-d2b4776d2d9b9950fc1f58e9dd0e38550a526976d80e2232f4cb2dcfb84171b2R370

And, what @feeblefakie suggested should be included in this `Begin or start a transaction` section.

docs/api-guide.md
```diff
@@ -1,29 +1,33 @@
-# Java API Guide
+# ScalarDB Java API Guide
 
-ScalarDB Java API is mainly composed of Administrative API and Transactional API.
-This guide briefly explains what kind of APIs exist and how to use them.
-
-* [Administrative API](#administrative-api)
-* [Transactional API](#transactional-api)
+The ScalarDB Java API is mainly composed of the Administrative API and Transactional API. This guide briefly explains what kinds of APIs exist, how to use them, and related topics like handling exceptions.
```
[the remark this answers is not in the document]
josh-wong: I'm OK with this change👍
(Either is OK grammatically.)

docs/api-guide.md
josh-wong: Sounds good. I'm currently editing `storage-abstraction.md`, so this change will be included when I submit the PR for that doc.

I've made the same changes in `Admin.java` in this PR, so PTAL!

docs/api-guide.md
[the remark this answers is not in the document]
josh-wong: I only saw one Coordinator table when running ScalarDB, but I assumed there were multiple after seeing APIs like `admin.createCoordinatorTables()`, which is why I left `Coordinator tables` plural.

@brfrn169 If we currently only have one type of Coordinator table, I think we should use the singular form to avoid confusion. Is that OK?

docs/api-guide.md
josh-wong: Got it. I've added the note about `getId()` to the `Join a transaction` and `Resume a transaction` sections, and added the part that @feeblefakie suggested to this section (`Begin or start a transaction`).

docs/api-guide.md
josh-wong: Got it. I've removed that sentence from this note since I've added it to the `Begin or start a transaction` section.

docs/api-guide.md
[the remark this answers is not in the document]
feeblefakie: Oh, I didn't know the API uses plural.
Then, I think we should note it since the method name and the actual number of tables are inconsistent; otherwise, it is misleading.

One way is we call it `Coodinator table` and note that the method name is plural for future extension.
The other way is we call it `Coordinator tables` and note that we only have one table at the moment, but it can be extended.

@brfrn169 @josh-wong Thought?

docs/api-guide.md
```diff
@@ -216,177 +237,219 @@ Set<String> namespaces = admin.getNamespaceNames();
...
+```java
+// Start a transaction with specifying a transaction ID.
+DistributedTransaction transaction = transactionManager.start("<TRANSACTION_ID>");
 ```
 
-Note that you must guarantee uniqueness of the transaction ID in this case.
+{% capture notice--info %}
+**Note**
+
+Specifying a transaction ID is useful when you want to link external systems to ScalarDB. Otherwise, you should use the `start()` method.
```
brfrn169, line 354: We should mention both the `begin()` method or the `start()` method:

```suggestion
Specifying a transaction ID is useful when you want to link external systems to ScalarDB. Otherwise, you should use the `begin()` method or the `start()` method.
```

docs/api-guide.md
```diff
@@ -216,177 +237,219 @@ Set<String> namespaces = admin.getNamespaceNames();
...
+
+```java
+// Join a transaction.
+DistributedTransaction transaction = transactionManager.join("<TRANSACTION_ID>");
+```
+
+{% capture notice--info %}
+**Note**
+
+You must guarantee the uniqueness of the transaction ID in this case. To get the transaction ID with `getId()`, you can specify the following:
```
brfrn169, line 375: The first sentence is unnecessary as users don’t need to care about the uniqueness of the transaction ID here; they can simply get the transaction ID with `getId()` and use it.
```suggestion
To get the transaction ID with `getId()`, you can specify the following:
```

docs/api-guide.md
```diff
@@ -216,177 +237,219 @@ Set<String> namespaces = admin.getNamespaceNames();
...
+
+```java
+// Resume a transaction.
+DistributedTransaction transaction = transactionManager.resume("<TRANSACTION_ID>");
+```
+
+{% capture notice--info %}
+**Note**
+
+You must guarantee the uniqueness of the transaction ID in this case. To get the transaction ID with `getId()`, you can specify the following:
```
brfrn169, line 398: Ditto.

```suggestion
To get the transaction ID with `getId()`, you can specify the following:
```

docs/api-guide.md
josh-wong: Got it. Thanks!

docs/api-guide.md
josh-wong: Got it. Thanks!

docs/api-guide.md
josh-wong: Got it. Thanks!

docs/api-guide.md
josh-wong: Done!

**Ref:** https://github.com/scalar-labs/scalardb/blob/334069fe6d64dfb5f4df4aa33ea7de94b28d8383/docs/two-phase-commit-transactions.md?plain=1#L697C4-L697C4

docs/api-guide.md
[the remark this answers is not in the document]
josh-wong: > One way is we call it `Coordinator table` and note that the method name is plural for future extension.

I'd prefer this way to reduce confusion. @brfrn169, is this way OK with you?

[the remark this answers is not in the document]
brfrn169: @josh-wong Yes, that's fine with me too. Thanks.

docs/api-guide.md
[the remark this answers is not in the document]
brfrn169: It looks like the change @Torch3333 suggested is not applied. Can you please apply it?

core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -378,19 +378,19 @@ void repairTable(
    * Add a new column to an existing table. The new column cannot be a partition or clustering key.
    * <br>
    * <br>
-   * <strong>Warning :</strong> this should be executed with significant consideration as the
-   * execution time may vary greatly depending on the underlying storage. Please plan accordingly
-   * especially if the database runs in production:
+   * <strong>Attention:</strong> You should carefully consider adding a new column to a table because the
```
brfrn169, line 381: @josh-wong We need to run the formatter on the Javadoc for this code. May I push a commit for this?

brfrn169: Ran the formatter in 45aa409330c6b2731e347c18b300a744a5c9caee.

docs/api-guide.md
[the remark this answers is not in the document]
josh-wong: Oops, thank you for catching that! I've applied this suggestion.

brfrn169 (approved):
LGTM! Thank you!

docs/api-guide.md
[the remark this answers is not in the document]
josh-wong: @brfrn169 (cc: @feeblefakie) Thank you for confirming! How does the following note sound?:

```suggestion
The Coordinator table is used by the [Transactional API](#transactional-api) to track the statuses of transactions.

{% capture notice--info %}
**Note**

Although only one Coordinator table can be created, the name of the API method is plural for future extension.
{% endcapture %}

<div class="notice--info">{{ notice--info | markdownify }}</div>
```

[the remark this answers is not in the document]
brfrn169: @josh-wong Sounds good. Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #1113, opened 2023-09-24
Split release.yaml to two sub workflows

reddikih:
The changes made by this PR are as follows:

1. Splits `release.yaml` into two sub workflow files: `upload-artifacts.yaml` and `create-release.yaml`.
   - `upload-artifacts.yaml` creates container images, registers them to the container registries, and registers Jar files in the Maven central.
   - `create-release.yaml` generates release notes and uploads release assets.
2. Uses the [release-note-scritp](https://github.com/scalar-labs/actions/tree/add-releasenote-script) to create the main release note body.

By this separation, the release workflow executed  as follows:

`release.yaml  ->  upload-artifacts.yaml  -(if succeeded)->  create-release.yaml`

`create-release.yaml` is invoked only if `upload-artifacts.yaml` succeeds.  If `upload-artifacts.yaml` fails, `create-release.yaml` will not be executed.  In that case, create-release.yaml can be executed manually.

***NOTE:*** I'll create another PR on the workflow to generate merged release notes for ScalarDB.

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -1,0 +1,93 @@
+name: Create release
+
+on:
+  workflow_dispatch:
+  workflow_call:
+    secrets:
+      GH_PROJECT_ACCESS_TOKEN:
+        required: true
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set version
+        id: version
+        run: |
+          VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/v##g")
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Set up JDK 8 and 11 (default 8)
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: |
+            11
+            8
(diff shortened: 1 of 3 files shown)


komamitsu:
> create-release.yaml is invoked only if upload-artifacts.yaml succeeds. If upload-artifacts.yaml fails, create-release.yaml will not be executed. In that case, create-release.yaml can be executed manually.

Sorry, I might be missing something, but I don't get if this is really necessary. I don't come up with a good situation to manually execute `create-release.yaml` after `update-artifacts.yaml` fails. I also think modifying the order of jobs in `release.yaml` may be enough to improve error flow.

> Uses the [release-note-scritp](https://github.com/scalar-labs/actions/tree/add-releasenote-script) to create the main release note body.

This motivation sounds very reasonable 👍. I think there is another option that creates `release-note.yaml` and `release.yaml` calls it around `create_release` job. We can manually call `release-note.yaml` via `workflow_dispatch` and we don't need to install Java 11 in `release.yaml` (`create-release.yaml`). What do you think?

feeblefakie:
@reddikih I will close it for now. Please reopen it once it's ready for review.

PR #1115, opened 2023-09-25
Fix tla+/consensus-commit

komamitsu:
I ran into some errors when I tried to run `tla+/consensus-commit` on TLA+ Toolbox v1.8.0. This PR fixed the following issues:
1. `CC.tla` and `CCSpec.tla` don't have module-end line (`======`) which is required in the specification. See `2.2 The TLA+ Module` in https://lamport.azurewebsites.net/tla/p-manual.pdf for details
2. The README says `Consistent` invariant can be used with `Spec`. But it doesn't work

---

BTW, the following error still happens. I'll look into it later when I have a time

```
The subscript of the next-state relation specified by the specification
does not seem to contain the state variable msgs
The subscript of the next-state relation specified by the specification
does not seem to contain the state variable rPrepared
current state is not a legal state
While working on the initial state:
/\ msgs = null
/\ rState = (r1 :> "initialized" @@ r2 :> "initialized" @@ r3 :> "initialized")
/\ cState = "initialized"
/\ rPrepared = null
```

### DIFF
--- a/tla+/consensus-commit/CC.tla
+++ b/tla+/consensus-commit/CC.tla
@@ -132,3 +132,4 @@ CCS == INSTANCE CCSpec
 THEOREM Spec => CCS!Spec
 \* Put "CCS!Spec" into the Properties section of the model to check it
 
+=============================================================================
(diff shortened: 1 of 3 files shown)


Torch3333 (approved):
LGTM, thank you!

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #1125, opened 2023-10-05
Revise Admin's repair table behavior

Torch3333:
## Description

The Admin API and Schema Loader provide a way to repair a table using `Admin.repairTable()` or the `--repair-all` command respectively. 
Currently, the repair table feature will create the metadata for the table if they do not exist. From now on, it will also create the table if it does not exist.

The reason for this change is that if a table exists but its metadata does not, the user cannot restore the table back to a working state. For example, this failure scenario could happen if creating the metadata for the table fails when `Admin.createTable` is called. 

## Related issues and/or PRs

N/A

## Changes made

The essential changes are in the `CassandraAdmin`, `CosmosAdmin`, `DynamoAdmin`, and `JdbcAdmin`.
In these classes, `admin.createTable()` and `admin.repairTable()` logic is now mostly shared. The difference is that `repairTable()` will create the table, indexes, and metadata **if they do not exist**.

## Checklist

The CI is failing for the Oracle 23 job for a lack of disk space, which is not related to this PR. I hope to fix it before this PR is ready to be merged.

## Release notes

Revise the behavior of the repair table feature to also create the table if it does not exist. In addition to the current behavior of creating metadata if they do not exist

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosAdminTestUtils.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosAdminTestUtils.java
@@ -88,9 +88,8 @@ public class CosmosAdminTestUtils extends AdminTestUtils {
   public boolean tableExists(String namespace, String table) {
     try {
       client.getDatabase(namespace).getContainer(table).read();
-    } catch (RuntimeException e) {
-      if (e instanceof CosmosException
-          && ((CosmosException) e).getStatusCode() == CosmosErrorCode.NOT_FOUND.get()) {
+    } catch (CosmosException e) {
+      if (e.getStatusCode() == CosmosErrorCode.NOT_FOUND.get()) {
         return false;
       }
       throw e;
--- a/core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosAdminTestUtils.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosAdminTestUtils.java
@@ -6,6 +6,7 @@ import com.azure.cosmos.ConsistencyLevel;
 import com.azure.cosmos.CosmosClient;
 import com.azure.cosmos.CosmosClientBuilder;
 import com.azure.cosmos.CosmosContainer;
+import com.azure.cosmos.CosmosException;
 import com.azure.cosmos.CosmosStoredProcedure;
 import com.azure.cosmos.models.CosmosItemRequestOptions;
 import com.azure.cosmos.models.CosmosQueryRequestOptions;
@@ -39,10 +40,7 @@ public class CosmosAdminTestUtils extends AdminTestUtils {
 
   @Override
   public void dropMetadataTable() {
-    client
-        .getDatabase(metadataDatabase)
-        .getContainer(CosmosAdmin.TABLE_METADATA_CONTAINER)
-        .delete();
+    dropTable(metadataDatabase, CosmosAdmin.TABLE_METADATA_CONTAINER);
   }
 
   @Override
@@ -84,5 +82,24 @@ public class CosmosAdminTestUtils extends AdminTestUtils {
         .getContainer(table)
         .getScripts()
         .getStoredProcedure(CosmosAdmin.STORED_PROCEDURE_FILE_NAME);
+  }
+
+  @Override
+  public boolean tableExists(String namespace, String table) {
+    try {
+      client.getDatabase(namespace).getContainer(table).read();
+    } catch (RuntimeException e) {
+      if (e instanceof CosmosException
+          && ((CosmosException) e).getStatusCode() == CosmosErrorCode.NOT_FOUND.get()) {
+        return false;
+      }
+      throw e;
+    }
+    return true;
+  }
+
+  @Override
+  public void dropTable(String namespace, String table) {
+    client.getDatabase(namespace).getContainer(table).delete();
   }
 }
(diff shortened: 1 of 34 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -99,23 +98,26 @@ public void createNamespace(String namespace, Map<String, String> options)
...
       throw new ExecutionException(
           "Creating the " + getFullTableName(namespace, table) + " table failed ", e);
     }
   }
 
-  private void createTableInternal(
-      Connection connection, String schema, String table, TableMetadata metadata)
-      throws SQLException {
+  @VisibleForTesting
+  void createTableInternal(
```
Torch3333, line 111: I made this method package private to avoid replicating unit tests since several public methods use this method.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -133,35 +135,43 @@ private void createTableInternal(
...
       throws SQLException {
     for (String indexedColumn : metadata.getSecondaryIndexNames()) {
-      createIndex(connection, schema, table, indexedColumn);
+      createIndex(connection, schema, table, indexedColumn, ifNotExists);
     }
   }
 
-  private void addTableMetadata(
+  @VisibleForTesting
+  void addTableMetadata(
```
Torch3333, line 160: Likewise here.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -115,4 +119,8 @@ default String getPattern(LikeExpression likeExpression) {
   default @Nullable String getEscape(LikeExpression likeExpression) {
     return likeExpression.getEscape();
   }
+
+  boolean isDuplicateIndexError(SQLException e);
+
+  String tryAddIfNotExistsToCreateIndexSql(String createIndexSql);
```
Torch3333, line 125: I added these two methods to create an index in an "if not exists" pattern.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -330,4 +334,17 @@ public String getEscape(LikeExpression likeExpression) {
...
     return escape.isEmpty() ? "\\" : escape;
   }
+
+  @Override
+  public String tryAddIfNotExistsToCreateIndexSql(String createIndexSql) {
+    return createIndexSql;
+  }
+
+  @Override
+  public boolean isDuplicateIndexError(SQLException e) {
```
komamitsu, line 344: There are some `isXxxxx()` methods above. How about moving this there?

komamitsu (approved):
LGTM! 👍

brfrn169 (commented):
Overall, LGTM. Thanks!

Just one point to consider: I think the scenarios where we would use `repairTable()` are:

- When a table exists, but its metadata doesn't.
- When a table's metadata exists, but the actual table doesn't.

Given this, maybe we could throw an exception in `repairTable()` when both the actual table and its metadata are missing. What are your thoughts?

core/src/integration-test/java/com/scalar/db/storage/cosmos/CosmosAdminTestUtils.java
```diff
@@ -85,4 +83,23 @@ public CosmosStoredProcedure getTableStoredProcedure(String namespace, String ta
         .getScripts()
         .getStoredProcedure(CosmosAdmin.STORED_PROCEDURE_FILE_NAME);
   }
+
+  @Override
+  public boolean tableExists(String namespace, String table) {
+    try {
+      client.getDatabase(namespace).getContainer(table).read();
+    } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
Torch3333: Indeed, thanks. 
https://github.com/scalar-labs/scalardb/pull/1125/commits/feb9fbc07f8a5933166f06d4309948ea9102ced2

core/src/integration-test/java/com/scalar/db/storage/multistorage/MultiStorageAdminTestUtils.java
```diff
@@ -76,4 +81,69 @@ private void execute(String sql) throws SQLException {
       stmt.execute(sql);
     }
   }
+
+  @Override
+  public boolean tableExists(String namespace, String table) throws Exception {
+    boolean existsOnCassandraStorage = clusterManager.getMetadata(namespace, table) != null;
```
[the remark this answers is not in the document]
Torch3333: Indeed, thanks. 
https://github.com/scalar-labs/scalardb/pull/1125/commits/feb9fbc07f8a5933166f06d4309948ea9102ced2

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
Torch3333: Yes, I moved this method with the other similar methods for all the `RdbEngineXXX.java`. Thank you.
https://github.com/scalar-labs/scalardb/pull/1125/commits/feb9fbc07f8a5933166f06d4309948ea9102ced2

Torch3333:
Regarding @brfrn169 question:
> Just one point to consider: I think the scenarios where we would use `repairTable()` are:
> 
>     * When a table exists, but its metadata doesn't.
> 
>     * When a table's metadata exists, but the actual table doesn't.
> 
> 
> Given this, maybe we could throw an exception in `repairTable()` when both the actual table and its metadata are missing. What are your thoughts?

We discussed this offline, so I will share the details here:

`repairTable` will not only repair the table and metadata but also the indexes, scaling configurations (for Cosmos and Dynamo), stored procedure (CosmosDB), etc. If we want to assert that none of these elements requires to be repaired to throw an exception, it will significantly increase the amount of code while also requiring thorough integration testing. 
This feature would be nice to have but since it requires a significant development cost, we decided not to implement it.

feeblefakie (approved):
LGTM! Thank you!

PR #1222, opened 2023-10-23
Support implicit pre-read in Consensus Commit

brfrn169:
## Description

This PR adds support for implicit pre-read in Consensus Commit. With implicit pre-read, if a record is in the write set but not in the read set, ScalarDB will automatically read the record and add it to the read set before preparing records. This allows users to perform mutations (`Put` and `Delete` operations) without reading the records beforehand.

One scenario where the pre-read isn't necessary is when writing a record that doesn't already exist. In such a case, users originally don't need to read the record before writing it. To address this, we can disable implicit pre-read for `Put` operations by calling the `Put.disableImplicitPreRead()` method. When disabling implicit pre-read, ScalarDB will not read the record automatically before preparing it.

## Related issues and/or PRs

N/A

## Changes made

- Execute implicit pre-read right before preparing records
- Also execute implicit pre-read when condition is specified for the `Put` and `Delete` operations
- Added the the `disableImplicitPreRead()` method to the `Put` operation to disable implicit pre-read
- Added a new configuration `scalar.db.consensus_commit.parallel_implicit_pre_read.enabled` that indicates whether executing implicit pre-read in parallel

## Checklist

N/A

## Release notes

Supported implicit pre-read in Consensus Commit. With implicit pre-read, users can perform mutations (`Put` and `Delete` operations) without reading the records beforehand.


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -129,6 +131,19 @@ public void mutate(List<? extends Mutation> mutations) throws CrudException {
...
+      crud.fillReadSetForRecordsFromWriteAndDeleteSetsIfUnread();
+    } catch (CrudConflictException e) {
+      throw new CommitConflictException(
+          "Conflict occurred while reading unread records in the write and delete sets",
+          e,
+          getId());
+    } catch (CrudException e) {
+      throw new CommitException(
+          "Failed to read unread records in the write and delete sets", e, getId());
+    }
```
brfrn169, line 145: Before committing (preparing) a transaction, we will fill the read set with records from the write and delete sets if they are unread.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -159,16 +162,79 @@ private List<Result> createScanResults(Scan scan, List<String> projections, List
...
-    snapshot.put(new Snapshot.Key(put), put);
+  public void put(Put put) throws CrudException {
+    if (put.getCondition().isPresent() && put.isBlind()) {
+      throw new IllegalArgumentException("Blind put cannot have a condition: " + put);
+    }
+
+    Snapshot.Key key = new Snapshot.Key(put);
+
+    if (put.getCondition().isPresent()) {
+      fillReadSetIfUnread(key);
```
brfrn169, line 173: We will also fill the read set if the mutation has a condition.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -159,16 +162,79 @@ private List<Result> createScanResults(Scan scan, List<String> projections, List
...
+    }
+
+    snapshot.put(key, put);
+  }
+
+  public void delete(Delete delete) throws CrudException {
+    Snapshot.Key key = new Snapshot.Key(delete);
+
+    if (delete.getCondition().isPresent()) {
+      fillReadSetIfUnread(key);
```
brfrn169, line 185: Ditto. We will also fill the read set if the mutation has a condition.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -159,16 +162,79 @@ private List<Result> createScanResults(Scan scan, List<String> projections, List
...
+      }
+    }
+    for (Delete delete : snapshot.getDeletesInDeleteSet()) {
+      Snapshot.Key key = new Snapshot.Key(delete);
+      if (!snapshot.containsKeyInReadSet(key)) {
+        tasks.add(() -> fillReadSet(key));
+      }
+    }
+
+    parallelExecutor.fillReadSetForRecordsFromWriteAndDeleteSetsIfUnread(tasks, snapshot.getId());
```
brfrn169, line 237: Use `ParallelExecutor` to fill read set in parallel for better performance.

docs/configurations.md
```diff
@@ -26,27 +26,28 @@ If you don't specify the `scalar.db.transaction_manager` property, `consensus-co
...
+| `scalar.db.consensus_commit.parallel_commit.enabled`        | Whether or not the commit phase is executed in parallel.                                            | `true`                                                            |
+| `scalar.db.consensus_commit.parallel_rollback.enabled`      | Whether or not the rollback phase is executed in parallel.                                          | The value of `scalar.db.consensus_commit.parallel_commit.enabled` |
+| `scalar.db.consensus_commit.async_commit.enabled`           | Whether or not the commit phase is executed asynchronously.                                         | `false`                                                           |
+| `scalar.db.consensus_commit.async_rollback.enabled`         | Whether or not the rollback phase is executed asynchronously.                                       | The value of `scalar.db.consensus_commit.async_commit.enabled`    |
+| `scalar.db.consensus_commit.parallel_fill_read_set.enabled` | Whether or not to fill the read set with records from the write and delete sets if they are unread. | `true`                                                            |
```
brfrn169, line 50: Added a new configuration `scalar.db.consensus_commit.parallel_fill_read_set.enabled`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommit.java
```diff
@@ -131,6 +133,19 @@ public void mutate(List<? extends Mutation> mutations) throws CrudException {
...
+      crud.fillReadSetForRecordsFromWriteAndDeleteSetsIfUnread();
+    } catch (CrudConflictException e) {
+      throw new PreparationConflictException(
+          "Conflict occurred while reading unread records in the write and delete sets",
+          e,
+          getId());
+    } catch (CrudException e) {
+      throw new PreparationException(
+          "Failed to read unread records in the write and delete sets", e, getId());
+    }
```
brfrn169, line 147: Before preparing a transaction, we will fill the read set with records from the write and delete sets if they are unread.

komamitsu:
@brfrn169 It seems you removed the review requests. But you may want us to review the PR?

brfrn169:
> @brfrn169 It seems you removed the review requests. But you may want us to review the PR?

@komamitsu Thank you for the remainder! 🙇  Yeah, I found some mistakes in the PR and removed the review requests. I requested review again. Pleas take a look when you have time!

docs/api-guide.md
```diff
@@ -697,17 +689,28 @@ Put put =
         .build();
 ```
 
-#### `Delete` operation
+##### Blind `Put` operation
 
-`Delete` is an operation to delete a record specified by a primary key.
+If you're certain that a record you're trying to write doesn't exist, you can use a blind `Put` operation for better performance. A blind `Put` is faster than a regular `Put` as it doesn't read the record beforehand. However, if the record does exist, the operation will fail due to a conflict.
```
komamitsu, line 694: IIUC, the "blind write" feature in this PR is basically implicitly filling read-set if needed so that users don't need to manually read the target record in advance. But `com.scalar.db.api.PutBuilder.Buildable#blind()` is to skip the implicit read for insert operations. So it sounds a bit confusing to me. 

`com.scalar.db.api.PutBuilder.Buildable#create()`, `insert()`, `skipImplicitRead()` or something might be better instead of `com.scalar.db.api.PutBuilder.Buildable#blind()`?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
komamitsu, line 185: QQ: Why only if a condition is set?
This might be a silly idea or I might be missing something, but how about always calling `fillReadSetIfUnread()` in `delete()` and `put()` even if no condition is set so that a commit doesn't need to call `fillReadSetForRecordsFromWriteAndDeleteSetsIfUnread()`?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
```diff
@@ -113,9 +123,27 @@ public void rollbackRecords(List<ParallelExecutorTask> tasks, String transaction
...
+      List<ParallelExecutorTask> tasks, String transactionId) throws CrudException {
+    try {
+      executeTasks(
+          tasks,
+          config.isParallelFillReadSetEnabled(),
+          false,
+          true,
+          "fillReadSetForRecordsFromWriteAndDeleteSetsIfUnread",
+          transactionId);
+    } catch (ExecutionException | ValidationConflictException e) {
```
komamitsu, line 143: [minor] The order of exception classes is opposite to the above existing methods

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Indeed, we can fill the read set anytime before preparing records. So, as you suggested, we could always call `fillReadSetIfUnread()` in `delete()` and `put()`. The benefit of calling `fillReadSetForRecordsFromWriteAndDeleteSetsIfUnread()` before preparing records is that we can fill the read set in parallel for better performance. So we basically fill the read set right before preparing records, but for conditional updates, we need to fill the read set at that time.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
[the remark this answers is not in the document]
brfrn169: In this case, it won't cause an issue because the read set is a concurrent map.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
brfrn169: Let me fix this. Thanks.

brfrn169: Ah, the order is consistently `ExecutionException`, `ValidationConflictException`, and `CrudException`. So I don't think we need to fix it. What do you think?

docs/api-guide.md
brfrn169: Good point. In this context, a `blind write` means preparing a record without having an existing record (in the read set). So when you set the `blind` option for the Put operation, we skip the implicit read. What are your thoughts on this?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParallelExecutor.java
komamitsu: Oops. I confused `ExecutionException` with `CrudException`. You're right. Please ignore my comment above.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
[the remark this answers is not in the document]
komamitsu: Oh, I see. The inconsistency of class-level non-thread-safety and method-level thread-safety is a little bit confusing, but I don't come up with a good straightforward solution...

docs/api-guide.md
komamitsu: Sounds good.

But, when I read the description of this PR, I thought "blind writes" meant implicit filling read-set. So, I still feel some inconsistency between the description and `com.scalar.db.api.PutBuilder.Buildable#blind()`.

> This PR adds support for blind writes in Consensus Commit. Specifically, if a record is in the write set but not in the read set, ScalarDB will read the record and add it to the read set before preparing the records internally.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
komamitsu: Understood. Sounds reasonable 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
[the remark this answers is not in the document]
brfrn169: > The inconsistency of class-level non-thread-safety and method-level thread-safety is a little bit confusing

That's true. For now, I left some comments for the methods. Let's refactoring it if we come up with a good solution.

docs/api-guide.md
komamitsu: Thanks for the detailed explanation. I think we are on the same page.

I also agree to the ideas of `user-perspective blind writes` and `transaction-manager-perspective blind writes`.

> Maybe we don't need to modify the option name blind. And should I update the PR description instead?

Maybe it's related to how we want to appeal this feature to users. If we publicly call `user-perspective blind writes` "blind write", the `blind()` option might confuse users. Otherwise, the description and release note (`With this enhancement, users can execute blind writes without the need to read records prior to writing them.`) section should be updated to clearly describe "blind write" means `transaction-manager-perspective blind writes` (plus "We don't need in-advance reads anymore when you write a fixed value!"?).

I want to hear opinion from @feeblefakie as a product owner as well.

brfrn169: I've updated the description (and the release notes) slightly. Maybe we should also update the PR title but I didn't for now. Can you please check the description and let me know if anything still seems confusing?

komamitsu: It looks good to me!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/Put.java
```diff
@@ -750,6 +754,25 @@ public Put withCondition(MutationCondition condition) {
     return (Put) super.withCondition(condition);
   }
 
+  /**
+   * Returns whether this Put is a blind write.
+   *
+   * @return whether this Put is a blind write
+   */
+  public boolean isBlind() {
```
feeblefakie, line 762: I might miss something, so just a question.
Don't we need these methods in `Delete`?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
```diff
@@ -30,6 +30,9 @@ public class ConsensusCommitConfig {
   public static final String ASYNC_COMMIT_ENABLED = PREFIX + "async_commit.enabled";
   public static final String ASYNC_ROLLBACK_ENABLED = PREFIX + "async_rollback.enabled";
 
+  public static final String PARALLEL_FILL_READ_SET_ENABLED =
+      PREFIX + "parallel_fill_read_set.enabled";
```
feeblefakie, line 34: This is a difficult one...
```suggestion
      PREFIX + "parallel_read_for_blind_mutation.enabled";
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -23,30 +23,33 @@
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
-import javax.annotation.concurrent.ThreadSafe;
+import javax.annotation.concurrent.NotThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@ThreadSafe
+@NotThreadSafe
```
feeblefakie, line 30: Just a question. What makes this class not thread-safe?
(maybe it's good to be noted.)

feeblefakie (commented):
Overall, looking good!
Thank you!
(and sorry for the late review.)
Left several comments and suggestions. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: This was originally not thread-safe, but I have just corrected it.

core/src/main/java/com/scalar/db/api/Put.java
brfrn169: I don't think we need the `blind` option for `Delete` because we always need the pre-read for `Delete`. What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -159,18 +162,83 @@ private List<Result> createScanResults(Scan scan, List<String> projections, List
...
+      fillReadSetIfUnread(key);
+      mutationConditionsValidator.checkIfConditionIsSatisfied(
+          delete, snapshot.getFromReadSet(key).orElse(null));
+    }
+
+    snapshot.put(key, delete);
+  }
+
+  @VisibleForTesting
+  void fillReadSetIfUnread(Snapshot.Key key) throws CrudException {
```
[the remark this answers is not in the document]
brfrn169: Let me use `read` because it sounds natural to me. If you want to use `get`, please let me know. Thanks.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -159,18 +162,83 @@ private List<Result> createScanResults(Scan scan, List<String> projections, List
...
+    if (!snapshot.containsKeyInReadSet(key)) {
+      fillReadSet(key);
+    }
+  }
+
+  private void fillReadSet(Snapshot.Key key) throws CrudException {
```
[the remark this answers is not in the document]
brfrn169: I've tried that before, but for some reason, it was unsuccessful. However, I've forgotten the reason, so let me try once more. Thanks.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -23,30 +23,33 @@
...
-@ThreadSafe
+@NotThreadSafe
 public class CrudHandler {
   private static final Logger logger = LoggerFactory.getLogger(CrudHandler.class);
   private final DistributedStorage storage;
   private final Snapshot snapshot;
   private final TransactionTableMetadataManager tableMetadataManager;
   private final boolean isIncludeMetadataEnabled;
   private final MutationConditionsValidator mutationConditionsValidator;
+  private final ParallelExecutor parallelExecutor;
```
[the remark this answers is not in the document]
brfrn169: No, the reason is that it contains `Snapshot` which is not thread-safe.

docs/api-guide.md
```diff
@@ -697,17 +689,28 @@ Put put =
         .build();
 ```
 
-#### `Delete` operation
+##### Blind `Put` operation
```
kota2and3kan, line 692: In my current understanding, this new feature (blind write) seems to work as follows.

1. If `record A` exists and we run `put(record A)` **without** `blind()`, ScalarDB automatically reads `record A` to create the read set, and `put()` (i.e., UPDATE) will succeed.
2. If `record A` exists and we run `put(record A)` **with** `blind()`, ScalarDB returns a conflict error since there is no `record A` information on the read set.
   * This is the same behavior as current (v3) ScalarDB (i.e., same as `record A` exists and we run `put(record A)` without any `get()`/`scan()`, in current (v3) ScalarDB).
3. If `record A` does NOT exist and we run `put(record A)` **with** `blind()`, ScalarDB puts a new record without read operation (there is a performance benefit).
   * This is the same behavior as current (v3) ScalarDB (i.e., same as `record A` does NOT exist and we run `put(record A)`, in current (v3) ScalarDB).
4. If `record A` does NOT exist and we run `put(record A)` **without** `blind()`, ...

---

### Question 1

Regarding `1.`, `2.`, and `3.`, is my understanding correct?

---

### Question 2

What will happen in `4.`?
If it puts (INSERT) `record A` as a new record, what is the difference between `3.` and `4.`? (Does ScalarDB try to read the null/empty record in case of `4.`?)

kota2and3kan (commented):
Thank you for the update!
I reviewed mainly documents.
And, I have two questions.
Please take a look when you have time!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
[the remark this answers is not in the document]
brfrn169: Fixed in ed8cc7ac7d2684b9a70c8d5e28be9225de8a8ac5. Thanks.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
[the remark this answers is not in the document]
brfrn169: Fixed in e5768fdb6c2e4731e828a2a1e8a1164dbd563b7b.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -159,18 +162,83 @@ private List<Result> createScanResults(Scan scan, List<String> projections, List
...
+            .partitionKey(key.getPartitionKey());
+    key.getClusteringKey().ifPresent(buildableGet::clusteringKey);
+    return buildableGet.build();
   }
 
-  public void delete(Delete delete) throws UnsatisfiedConditionException {
-    mutationConditionsValidator.checkIfConditionIsSatisfied(
-        delete, snapshot.getFromReadSet(new Snapshot.Key(delete)).orElse(null));
-    snapshot.put(new Snapshot.Key(delete), delete);
+  public void fillReadSetForRecordsFromWriteAndDeleteSetsIfUnread() throws CrudException {
```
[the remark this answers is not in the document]
brfrn169: Fixed in ed8cc7ac7d2684b9a70c8d5e28be9225de8a8ac5. It's renamed to `executeImplicitPreReadIfEnabled`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
brfrn169: Fixed in ed8cc7ac7d2684b9a70c8d5e28be9225de8a8ac5. It's renamed to `scalar.db.consensus_commit.parallel_implicit_pre_read.enabled`.

brfrn169:
@komamitsu I've discussed with @feeblefakie offline, and we've decided to name this feature `implicit pre-read`. Accordingly, I've updated the method names and the configuration name, and also revised the documentation. Can you please take a look at this PR once again? Sorry for the inconvenience.

docs/api-guide.md
brfrn169: Note that, after some discussion, the `blind()` method is renamed to `disableImplicitPreRead()`.

For question 1, yes, your understanding is correct for `1.`, `2.`, and `3.`.

For question 2, in case of `4.`, ScalarDB tries to read record A, but it doesn't exist. So ScalarDB tries to put a new record without information from any existing record, which is the same behavior of `3.`. Therefore, the difference between `3.` and `4.` is whether or not ScalarDB tries to read the record.

kota2and3kan: I see. I understand!
Thank you for your explanation!

kota2and3kan (approved):
LGTM! Thank you!

docs/api-guide.md
```diff
@@ -697,17 +689,28 @@ Put put =
         .build();
 ```
 
-#### `Delete` operation
+##### Disable implicit pre-read for `Put` operations
 
-`Delete` is an operation to delete a record specified by a primary key.
+If you're certain that a record you're trying to write doesn't exist, you can disable implicit pre-read for the `Put` operation for better performance. A `Put` without implicit pre-read is faster than a regular `Put` as it skips the implicit pre-read. However, if the record does exist, the operation will fail due to a conflict.
```
komamitsu, line 694: > If you're certain that a record you're trying to write doesn't exist, you can disable implicit pre-read for the `Put` operation for better performance.

This describes the implicit pre-read feature can be disabled. But there is no explanation in advance about what "implicit pre-read" is. How about adding a short description about "implicit pre-read" mechanism? It may be helpful for those who are interest in ScalarDB performance, but don't know implicit pre-read happens by default.

brfrn169: Thank you for pointing that out. That's right. I've added a short description about "implicit pre-read" in a7c007146b5501822088b4e950449b551d357dd2. Please take a look! Thanks.

docs/api-guide.md
```diff
@@ -697,17 +689,30 @@ Put put =
         .build();
 ```
 
-#### `Delete` operation
+##### Disable implicit pre-read for `Put` operations
 
-`Delete` is an operation to delete a record specified by a primary key.
+In Consensus Commit, write operations (`Put` and `Delete`) within a transaction require pre-reading the affected records to obtain the necessary information for the writes. If these records are not explicitly read during the transaction, ScalarDB performs an `implicit pre-read` of the records before committing the transaction.
```
feeblefakie, line 694:
```suggestion
In Consensus Commit, an application must read a record before mutating the record with `Put` and `Delete` operations to obtain the latest states of the record in case the record exists. If an application does not read the record explicitly in a transaction, ScalarDB reads the record on behalf of the application before committing the transaction. We call the feature `implicit pre-read`.
```

docs/api-guide.md
```diff
@@ -697,17 +689,30 @@ Put put =
...
 
-{% capture notice--info %}
-**Note**
+If you're certain that a record you're trying to write doesn't exist, you can disable implicit pre-read for the `Put` operation for better performance. A `Put` without implicit pre-read is faster than a regular `Put` as it skips the implicit pre-read. However, if the record does exist, the operation will fail due to a conflict.
```
feeblefakie, line 696:
```suggestion
If you are certain that a record you are trying to mutate does not exist, you can disable implicit pre-read for the `Put` operation for better performance. For example, if you load initial data, you can and should disable it. A `Put` without implicit pre-read is faster than a regular `Put` as it skips an unnecessary read. However, if the record exists, the operation will fail due to a conflict, i.e., ScalarDB thinks another transaction writes the record.
```

docs/storage-abstraction.md
```diff
@@ -680,6 +680,15 @@ Put put =
         .build();
 ```
 
+{% capture notice--info %}
+**Note**
+
+If you specify `disableImplicitPreRead()` or `implicitRreReadEnabled()` in the `Put` operation builder, they will be ignored.
```
feeblefakie, line 686:
```suggestion
If you specify `disableImplicitPreRead()` or `implicitPreReadEnabled()` in the `Put` operation builder, they will be ignored.
```

feeblefakie (commented):
Thank you for the naming change!
Left additional comments. PTAL!

core/src/main/java/com/scalar/db/api/OperationBuilder.java
```diff
@@ -281,6 +281,23 @@ interface ClearValues<T> {
     T clearValue(String columnName);
   }
 
+  interface ImplicitRreReadEnabled<T> {
```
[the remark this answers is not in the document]
brfrn169: Oops... Let me fix it.

core/src/main/java/com/scalar/db/api/OperationBuilder.java
```diff
@@ -281,6 +281,23 @@ interface ClearValues<T> {
...
+     */
+    T disableImplicitPreRead();
+
+    /**
+     * Sets whether implicit pre-read is enabled or not for this put operation.
+     *
+     * @param implicitRreReadEnabled whether implicit pre-read is enabled or not
+     * @return the operation builder
+     */
+    T implicitRreReadEnabled(boolean implicitRreReadEnabled);
```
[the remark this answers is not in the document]
brfrn169: I've added the `implicitPreReadEnabled(boolean implicitPreReadEnabled)` method because it can be useful in some cases.

Let's see the following example:

```java
boolean implicitPreRead = ...;

Buildable buildable =
    Put.newBuilder()
        .namespace("ns")
        .table("tbl")
        .partitionKey(partitionKey)
        .clusteringKey(clusteringKey)
        .floatValue("c4", 1.23F)
        .doubleValue("c5", 4.56);

if (!implicitPreRead) {
  buildable.disableImplicitPreRead();
}

Put put = buildable.build();
```

With the `implicitPreReadEnabled(boolean implicitPreReadEnabled)` method, the above code can be rewritten as follows:

```java
boolean implicitPreRead = ...;

Put put =
    Put.newBuilder()
        .namespace("ns")
        .table("tbl")
        .partitionKey(partitionKey)
        .clusteringKey(clusteringKey)
        .floatValue("c4", 1.23F)
        .doubleValue("c5", 4.56)
        .implicitPreReadEnabled(implicitPreRead)
        .build();
```

So I think we need a method that takes a boolean parameter that toggles the implicit pre-read feature. However, we can discuss the method name (`implicitPreReadEnabled`) further if needed. Thank you.

core/src/main/java/com/scalar/db/api/OperationBuilder.java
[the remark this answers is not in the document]
brfrn169: Fixed in a5bdbe70d84894cf9655a8881184ad2010d6cadb. Thanks.

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

brfrn169:
@josh-wong Can you please take a look at the documentation part of this PR?

docs/api-guide.md
```diff
@@ -654,14 +654,6 @@ You can't specify clustering-key boundaries and orderings in `Scan` without spec
...
 `Put` is an operation to put a record specified by a primary key. The operation behaves as an upsert operation for a record, in which the operation updates the record if the record exists or inserts the record if the record does not exist.
 
-{% capture notice--info %}
-**Note**
-
-When you update an existing record, you need to read the record by using `Get` or `Scan` before using a `Put` operation.
-{% endcapture %}
-
-<div class="notice--info">{{ notice--info | markdownify }}</div>
-
```
josh-wong, line 664: Just to confirm, do we need to remove this note from this doc on the version branches?

josh-wong (changes requested):
Apologies for the delay! I've added a few comments and suggestions. PTAL!🙇🏻‍♂️

docs/api-guide.md
brfrn169: No, the implicit pre-read feature will be supported starting from version `3.11`, so the note is still required for version `3.10` and earlier versions.

docs/api-guide.md
```diff
@@ -697,17 +689,30 @@ Put put =
...
 
-#### `Delete` operation
+##### Disable implicit pre-read for `Put` operations
 
-`Delete` is an operation to delete a record specified by a primary key.
+In Consensus Commit, an application must read a record before mutating the record with `Put` and `Delete` operations to obtain the latest states of the record in case the record exists. If an application does not read the record explicitly in a transaction, ScalarDB reads the record on behalf of the application before committing the transaction. We call the feature `implicit pre-read`.
 
-{% capture notice--info %}
-**Note**
+If you are certain that a record you are trying to mutate does not exist, you can disable implicit pre-read for the `Put` operation for better performance. For example, if you load initial data, you can and should disable it. A `Put` without implicit pre-read is faster than a regular `Put` as it skips an unnecessary read. However, if the record exists, the operation will fail due to a conflict, i.e., ScalarDB thinks another transaction writes the record.
```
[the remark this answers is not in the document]
brfrn169: Maybe it should be `another transaction wrote the record`? Or `another transaction has written the record`?

@feeblefakie What do you think?

docs/api-guide.md
josh-wong: Understood. Thank you!

docs/api-guide.md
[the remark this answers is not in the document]
feeblefakie: @josh-wong @brfrn169 Yes, that's what I meant. Sorry!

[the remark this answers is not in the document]
brfrn169: @josh-wong Applied your suggestions. Can you please take a quick look and approve if you are okay with them? Thanks.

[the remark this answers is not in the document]
josh-wong: @brfrn169 LGTM! Thank you!

josh-wong (approved):
LGTM! Thank you!🙇‍♂️

brfrn169:
@Torch3333 It's been merged, but please take a look when you come back!

brfrn169:
@Torch3333 It's merged but please take a look when you are back!

PR #1231, opened 2023-10-24
Split release.yaml to two sub workflows

reddikih:
## Description

The purpose of this PR are:
- Split `release.yaml` into two sub-workflows to facilitate rerunning the workflow when the workflow fails partially.
- Use [release-note-script](https://github.com/scalar-labs/actions/tree/main/release-note-script) to create the release note body by extracting from the PR description to make it more meaningful to the reader.

## Related issues and/or PRs

N/A

## Changes made

The changes made by this PR are as follows:

1. Splits `release.yaml` into two sub workflow files: `upload-artifacts.yaml` and `create-release.yaml`.
   - `upload-artifacts.yaml` creates container images, registers them to the container registries, and registers Jar files in the Maven central.
   - `create-release.yaml` generates release notes and uploads release assets.
2. Uses the [release-note-scritp](https://github.com/scalar-labs/actions/tree/main/release-note-script) to create the main release note body.

By this separation, the release workflow executed as follows:

`release.yaml  ->  upload-artifacts.yaml  -(if succeeded)->  create-release.yaml`

`create-release.yaml` is invoked only if `upload-artifacts.yaml` succeeds. If `upload-artifacts.yaml` fails, `create-release.yaml` will not be executed. In that case, `create-release.yaml` can be executed manually.

## Checklist

- `secrets.GH_PROJECT_ACCESS_TOKEN` used by `create-release.yaml` is needed to create before running this workflow. It should include `admin:org`, `project`, `repo`, `workflow` permission to access to the organization's projects.
- I'll create another PR on the workflow to generate merged release notes for ScalarDB.

## Release notes

N/A

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -55,9 +55,7 @@ jobs:
         with:
           tag_name: ${{ github.ref }}
           release_name: ${{ github.ref }}
-          # TODO: Remove '#' and enable "body_path: rnbody" when every PR becomes including the release note section.
-          # Until that, the release note body is created by actions/create-release automatically.
-          # body_path: rnbody.md
+          body_path: rnbody.md
           draft: true
           prerelease: false
 
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -1,0 +1,91 @@
+name: Create release
+
+on:
+  workflow_dispatch:
+  workflow_call:
+    secrets:
+      GH_PROJECT_ACCESS_TOKEN:
+        required: true
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set version
+        id: version
+        run: |
+          VERSION=$(echo ${{ github.ref }} | sed -e "s#refs/tags/v##g")
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Set up JDK 8 and 11 (default 8)
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: |
+            11
+            8
+
+      - name: Checkout release note scripts
+        uses: actions/checkout@v4
+        with:
+          repository: scalar-labs/actions
+          token: ${{ secrets.GH_PROJECT_ACCESS_TOKEN }}
+          path: ${{ github.workspace }}
+          sparse-checkout-cone-mode: false
+          sparse-checkout: |
+            release-note-script/src/main/java
+
+      - name: Move scripts to the working directory
+        run: cp ${{ github.workspace }}/release-note-script/src/main/java/* ${{ github.workspace }}
+
+      - name: Create release note body
+        id: rn_body
+        env:
+          GH_TOKEN: ${{ secrets.GH_PROJECT_ACCESS_TOKEN }}
+        run: |
+          $JAVA_HOME_11_X64/bin/java ReleaseNoteCreation.java ${{ github.repository_owner }} ScalarDB ${{ steps.version.outputs.version }} ${{ github.event.repository.name }} > rnbody.md
+          cat rnbody.md # For debug
+
+      - name: Create release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PROJECT_ACCESS_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          # TODO: Remove '#' and enable "body_path: rnbody" when every PR becomes including the release note section.
+          # Until that, the release note body is created by actions/create-release automatically.
+          # body_path: rnbody.md
+          draft: true
+          prerelease: false
+
+      - name: Checkout the current repository
+        uses: actions/checkout@v4
+
+      - name: Build scalardb-server zip
+        run: ./gradlew :server:distZip
+
+      - name: Upload scalardb-server zip
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PROJECT_ACCESS_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: server/build/distributions/scalardb-server-${{ steps.version.outputs.version }}.zip
+          asset_name: scalardb-server-${{ steps.version.outputs.version }}.zip
+          asset_content_type: application/zip
+
+      - name: Build scalardb-schema-loader jar
+        run: ./gradlew :schema-loader:shadowJar
+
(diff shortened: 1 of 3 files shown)


reddikih:
@komamitsu I open this PR again. I copied your comments from the closed PR.

Your comment:

> create-release.yaml is invoked only if upload-artifacts.yaml succeeds. If upload-artifacts.yaml fails, create-release.yaml will not be executed. In that case, create-release.yaml can be executed manually.

Sorry, I might be missing something, but I don't get if this is really necessary. I don't come up with a good situation to manually execute `create-release.yaml` after `update-artifacts.yaml` fails. I also think modifying the order of jobs in `release.yaml` may be enough to improve error flow.

> Uses the [release-note-scritp](https://github.com/scalar-labs/actions/tree/add-releasenote-script) to create the main release note body.

This motivation sounds very reasonable 👍. I think there is another option that creates `release-note.yaml` and `release.yaml` calls it around `create_release` job. We can manually call `release-note.yaml` via `workflow_dispatch` and we don't need to install Java 11 in `release.yaml` (`create-release.yaml`). What do you think?

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Move scripts to the working directory
+        run: cp ${{ github.workspace }}/release-note-script/src/main/java/* ${{ github.workspace }}
+
+      - name: Create release note body
+        id: rn_body
+        env:
+          GH_TOKEN: ${{ secrets.GH_PROJECT_ACCESS_TOKEN }}
+        run: |
+          $JAVA_HOME_11_X64/bin/java ReleaseNoteCreation.java ${{ github.repository_owner }} ScalarDB ${{ steps.version.outputs.version }} ${{ github.event.repository.name }} > rnbody.md
+          cat rnbody.md # For debug
```
brfrn169, line 48: I think we should always print this, so we can delete this comment.
```suggestion
          cat rnbody.md
```

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,91 @@
...
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PROJECT_ACCESS_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          # TODO: Remove '#' and enable "body_path: rnbody" when every PR becomes including the release note section.
+          # Until that, the release note body is created by actions/create-release automatically.
+          # body_path: rnbody.md
```
brfrn169, line 60: Even before every PR becomes including the release note section, I think we can include `body_path` for testing since we can overwrite the generated body. What do you think?

brfrn169 (commented):
Overall, LGTM. Left a couple of comments. Please take a look when you have time!

komamitsu:
@reddikih Thanks for replying to my question! I overlooked the lack of idempotence. All of your consideration seem reasonable.

For `release-note.yaml`, that's just an idea and I don't have a strong opinion on it. Also, we can separate it later if needed.

komamitsu (approved):
LGTM! 👍

.github/workflows/create-release.yaml
reddikih: I see. My concern is that if there are some PRs that don't have the `Release Notes` section in the PR description, it might not be included generated release note body. And in that case, we have to find such PRs. 

If we can find them without too much effort, I agree with you to remove the comment and place the generated release note body to the release note as a test.

Maybe my concern is a minor. I hope to hear your opinion about that.

reddikih:
@komamitsu

> @reddikih Thanks for replying to my question! I overlooked the lack of idempotence. All of your consideration seem reasonable.
> 

Thank you for your questions. And I'm sorry for not explaining my motivations at first.

> For `release-note.yaml`, that's just an idea and I don't have a strong opinion on it. Also, we can separate it later if needed.

Sure, we can do it if needed. Thank you.

.github/workflows/create-release.yaml
reddikih: I see. I removed it in ca24035

.github/workflows/create-release.yaml
reddikih: Ah, either way, as a test, we should check whether all the PRs that should be included in the release notes are included in the generated release notes themselves.

Sorry for the lack of this perspective. I'll remove those comments and include `body_path` for testing.

Please ignore my previous comment🙇‍♂️

reddikih: Done at 1c6536f9

brfrn169 (approved):
LGTM! Thank you!

choplin (commented):
While the contents of the workflows look good to me, I guess the names should be revised because `create-release` uploads the artifact to the created release, and `upload-artifact` builds the docker images as well.

reddikih:
@choplin 
> While the contents of the workflows look good to me, I guess the names should be revised because `create-release` uploads the artifact to the created release, and `upload-artifact` builds the docker images as well.

I see. So, I would like to change the file names as follows:
- `upload-artifacts.yaml` -> `create-and-upload-artifacts.yaml`
- `create-release.yaml` -> `create-release-and-upload-assets.yaml`

How about this? If you have any other suitable names, I would appreciate any advice.

choplin:
Sorry for the confusion. After considering why I felt uncomfortable for a while, I noticed that, what we want to do is to:

1. Create a release on GitHub
2. Push the artifact to Maven Central
3. Push the docker images

And, to do those, we have to:

a. Build the jar files (required for 1. 2. and 3.)
b. Build the docker images (required only for 3.)

`create-release.yaml` contains 1. and a. while `upload-artifacts.yaml` contains 2. 3. a. and b. So, dependencies were not clear to me at a glance. (Also, several dependencies are resolved implicitly within Gradle).

Ideally, IMO, we should define all those jobs as separate jobs and declare the dependencies using the GitHub Actions feature, but it could be too complex. So, I suggest separating the jobs 1. 2. and 3. By doing so, we can make the file names simpler and more specific, `create-release`, `upload-artifact-to-maven-repo`, and `upload-docker-images`. What do you think?

reddikih:
@choplin 
Thank you for clarifying the points.
Totally, I agree with you.

But I'd like to keep this PR as is (or just to do rename the file names).
The reasons why I think so are as follows:
- Since I'd like to keep the behavior of this workflows (`upload-artifacts.yaml` and `create-release`) as the original one (`release.yaml`) as much as possible, I simply split the original workflow into a before and after `Create release` job.
- In this PR, I also added the creation of a release note body using a new release note script, so I wanted to minimize the impact of the changes.

So, I'd like to reorganize these workflows as you suggested after this PR merge (Maybe as a kaizen-day task). I'll also consider the advice from @komamitsu  at that time. 

What do you think?

choplin:
@reddikih Okay! Let's keep this PR as is.

choplin (approved):

PR #1237, opened 2023-10-28
Should not create PR for base branch in Auto-PR

brfrn169:
## Description

Currently, if we create a PR based on a release branch, Auto-PR creates a PR for the base branch in addition to other target branches. Since the target change is already merged to the base branch, it's unnecessary. This PR addresses this issue.

## Related issues and/or PRs

N/A

## Changes made

- Change not to create a PR for a base branch

## Checklist

I've tested it in my private repository.

## Release notes

N/A

### DIFF
--- a/ci/auto-pr/conv_proj_version_to_branch
+++ b/ci/auto-pr/conv_proj_version_to_branch
@@ -12,8 +12,7 @@ versions.inject(Set.new) {|acc, v|
       # e.g. project: "ScalarDB 4.0.0" -> branch: "master"
       #
       # This GitHub project corresponds to `main`/`master` branch.
-      # The target change is already merged to the default branch.
-      # Nonthing to do.
+      acc << "master"
     else
       # e.g. project: "ScalarDB 3.8.0" -> branch: "3"
       #
--- a/.github/workflows/auto-pr.yaml
+++ b/.github/workflows/auto-pr.yaml
@@ -46,6 +46,8 @@ jobs:
           echo -------------
 
           branches=$(ci/auto-pr/conv_proj_version_to_branch $versions)
+          # Remove the base branch from the list because the target change is already merged to the branch.
+          branches=$(echo "$branches" | sed "/^${{ github.base_ref }}$/d")
           echo -------------
           echo "branches: $branches"
           echo -------------


ci/auto-pr/conv_proj_version_to_branch
```diff
@@ -12,8 +12,7 @@ versions.inject(Set.new) {|acc, v|
       # e.g. project: "ScalarDB 4.0.0" -> branch: "master"
       #
       # This GitHub project corresponds to `main`/`master` branch.
-      # The target change is already merged to the default branch.
-      # Nonthing to do.
+      acc << "master"
```
komamitsu, line 15: Sorry, I didn't get why this change is needed. Could you tell me what would happen without the change?

brfrn169: This change actually doesn't affect the behavior. With the change in this PR, we remove the base branch of the original PR from the target branch list. So we no longer need the special care for the `main`/`master` branch in this `conv_proj_version_to_branch` script. What do you think?

komamitsu: I see. My concern is this default branch name is different among ScalarDB projects, and ideally it should be passed by `auto-pr.yaml` that gets the default branch name from `github.event.repository.default_branch`. But it takes some effort and introduces a bit of complexity. So it would be better to make it simple by removing this.

brfrn169: > My concern is this default branch name is different among ScalarDB projects

I thought the same thing, and for now, it seems we'll need to use slightly modified scripts (with different default branch names) in other repositories. In the future, we can unify the default branch name across the repositories. 

Moreover, considering that we create PRs from not just the default branches (`master` or `main`) but also from support/release branches (`3` and `3.x`), the following comment may not always apply:
```bash
      # The target change is already merged to the default branch.
      # Nonthing to do.
```

Given this, I feel the special handling to the `master`/`main` branches strange here. That's why I made the change.

What do you think?

komamitsu: Sounds reasonable to me 👍

komamitsu (approved):
LGTM! Thanks for handling the issue 👍

PR #1260, opened 2023-11-01
Create namespace metadata when importing tables

Torch3333:
## Description

In the table importing feature, we need to create namespace metadata if they do not exist when importing a table. 

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/931

## Changes made

- Updates the `JdbcAdmin` and `ConsensuCommitAdmin` `importTable()` method to create the metadata of the imported table namespace.
- Updates the signature to `admin.importTable(String namespace, String table, Map<String, String> options)`: the `options` parameter is added. This parameter is required to pass specific options when creating namespace metadata for the Cassandra and Dynamo Admin. Even though the import feature is not supported for these storages as of now, we update the signature to be future-proof.

## Checklist

N/A

## Release notes

N/A


komamitsu (approved):
LGTM! 👍

schema-loader/src/main/java/com/scalar/db/schemaloader/SchemaOperator.java
```diff
@@ -413,9 +414,9 @@ public void importTables(List<ImportTableSchema> tableSchemaList) throws SchemaL
       String table = tableSchema.getTable();
       try {
         if (tableSchema.isTransactionTable()) {
-          transactionAdmin.get().importTable(namespace, table);
+          transactionAdmin.get().importTable(namespace, table, Collections.emptyMap());
```
brfrn169, line 417: I think we should add `Map<String, String> options` to `ImportTableSchema`, and pass it to `importTable()` as the third argument. What do you think?

jnmt (approved):
LGTM! Thank you!

schema-loader/src/main/java/com/scalar/db/schemaloader/SchemaOperator.java
Torch3333: Ok. 
Similarly to `TableSchema`, we could define custom options for each table in the schema file **or** pass options globally via CLI argument that will be applied to each table. 
I realized that setting a given option via the CLI **and** also in the schema file will throw an exception because of [this](https://github.com/scalar-labs/scalardb/blob/ee0e9a26f5afcebc532f9d19f3d7afe6308133c5/schema-loader/src/main/java/com/scalar/db/schemaloader/TableSchema.java#L159) (the same option key will be added twice in the list builder and this is not allowed by default). I don't know if this behavior is intended eventually. If an option is set globally as CLI argument and in the schema file, using the value specified in the schema file is more user-friendly in my opinion.

feeblefakie (approved):
LGTM! Thank you!

schema-loader/src/main/java/com/scalar/db/schemaloader/SchemaOperator.java
brfrn169: > If an option is set globally as CLI argument and in the schema file, using the value specified in the schema file is more user-friendly in my opinion.

I agree. That's more user-friendly behavior.

Torch3333: I added the options handling for the Schema Loader `--import` command in https://github.com/scalar-labs/scalardb/pull/1260/commits/b6e16cf10ca9b056f570d5062a9b5e8f53c1cc31 and it follows this principle.

> If an option is set globally as CLI argument and in the schema file, using the value specified in the schema file is more user-friendly in my opinion.

I will update the options handling for the other commands besides `--import` (all using the `TableSchema` class ) in another PR. 
Thanks for your comment and PTAL when you get the chance.

brfrn169 (approved):
LGTM! Thank you!

PR #1294, opened 2023-11-16
Add cross-partition scan options

jnmt:
## Description

This PR introduces cross-partition scan options and the related errors and warnings so that developers can know `ScanAll`'s behavior is different from the regular `Scan` and make sure to use it explicitly (at their own risk for NoSQL databases).

## Related issues and/or PRs

- #900

## Changes made

- Add three options in `DatabaseConfig`
  - cross-partition scan: `ScanAll` without filtering and ordering
  - cross-partition scan with filtering: `ScanAll` with conjunctions (where clause)
  - cross-partition scan with ordering: `ScanAll` with ordering
  - To enable the last two,  you need to enable the first one
  - Why these three? This granularity will help users understand the support features in their databases, especially when we implement filtering for the NoSQL cross-partition scan.
- Add warnings when you enable cross-partition scan with serializable isolation
- Change the following for each storage
  - Add config check for NoSQL databases if cross-partition scan filtering and ordering are both disabled
  - Add scan operation check if cross-partition scan options are appropriately enabled (in `OperationChecker`)
  - Make the storage name public in the config class to refer to it in a single place because some storage names are required for checking the config and warnings

## Checklist

- [ ] I have updated the documentation to reflect the changes. **(Will be done in another PR)**

- We still have a lot of the word `RelationalScan` in tests, but keep them as is for now
- See inline comments for other notes

## Release notes

Added cross-partition scan options


core/src/test/java/com/scalar/db/storage/cassandra/CassandraTest.java
jnmt, line 1: We no longer need this test class because it only tests if the scan is relational or not, and now we test it in `OperationChecker` instead.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcService.java
```diff
@@ -123,18 +122,9 @@ public List<Result> scan(Scan scan, Connection connection)
...
     if (scan instanceof ScanAll) {
-      ScanAll scanAll = (ScanAll) scan;
-      if (ScalarDbUtils.isRelational(scan)) {
-        return buildSelectQueryForRelationalScan(scanAll, tableMetadata);
-      } else {
-        return buildSelectQueryForScanAll(scanAll, tableMetadata);
-      }
-    } else {
-      return buildSelectQueryForScan(scan, tableMetadata);
+      return buildSelectQuery((ScanAll) scan, tableMetadata);
```
jnmt, line 125: After introducing the cross-partition scan concept, I was convinced that we don't have to distinguish `ScanAll` and `RelationalScan` anymore. This is a kind of refactoring, but I included this since it is related to re-organizing relational scan terminology.

core/src/main/java/com/scalar/db/config/DatabaseConfig.java
```diff
@@ -90,6 +98,18 @@ protected void load() {
     activeTransactionManagementExpirationTimeMillis =
         getLong(getProperties(), ACTIVE_TRANSACTION_MANAGEMENT_EXPIRATION_TIME_MILLIS, -1);
     defaultNamespaceName = getString(getProperties(), DEFAULT_NAMESPACE_NAME, null);
+    crossPartitionScanEnabled = getBoolean(getProperties(), CROSS_PARTITION_SCAN, false);
```
jnmt, line 101: The default value is `false` in v4.0.0 or later, but it should be `true` in v3.x for backward compatibility and warned when using it with `SERIALIZABLE`. I will handle this after this PR is merged to the master.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
```diff
@@ -166,4 +172,30 @@ public boolean isIncludeMetadataEnabled() {
...
+      }
+    } else if (!databaseConfig.getStorage().equals(JdbcConfig.STORAGE_NAME)
+        && databaseConfig.isCrossPartitionScanEnabled()) {
+      warnCrossPartitionScan(databaseConfig.getStorage());
+    }
+  }
+
+  private void warnCrossPartitionScan(String storageName) {
+    logger.warn(
+        "Enabling cross-partition scan for {} in production is not recommended "
```
komamitsu, line 196: [trivial] I imagined what the message would be like (e.g., "Enabling cross-partition scan for cosmos in production ..."), and I think adding quotations might be a bit better.
```suggestion
        "Enabling cross-partition scan for '{}' in production is not recommended "
```

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
```diff
@@ -45,6 +44,13 @@ public class Cassandra extends AbstractDistributedStorage {
   @Inject
   public Cassandra(DatabaseConfig config) {
     super(config);
+
+    if (config.isCrossPartitionScanFilteringEnabled()
+        || config.isCrossPartitionScanOrderingEnabled()) {
+      throw new IllegalArgumentException(
+          "Cross-partition scan with filtering or ordering is not supported in Cassandra");
```
komamitsu, line 51: I think this is related to https://github.com/scalar-labs/scalardb/pull/1294/files#diff-660ca76c9b64d94000e0854e2a18fdc5e6a69c028a74b8cacfa4ef733d4604c2L111, but sorry, I forgot why these types of scan were prohibitted. Could you tell me that?

jnmt: Thanks for the question. Basically, it's hard to push down all kinds of filtering and ordering to NoSQL databases. But for filtering, we can support ScalarDB-side filtering in the near future since we already have a logic for evaluating all types of our predicates.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
jnmt: Fixed in [7bffcbb](https://github.com/scalar-labs/scalardb/pull/1294/commits/7bffcbb7a942bdcad7ec4c99cdeffa36378226d2). Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
```diff
@@ -81,12 +83,14 @@ public class JdbcConfig {
   public JdbcConfig(DatabaseConfig databaseConfig) {
     String storage = databaseConfig.getStorage();
     String transactionManager = databaseConfig.getTransactionManager();
-    if (!"jdbc".equals(storage) && !"jdbc".equals(transactionManager)) {
+    if (!storage.equals(STORAGE_NAME) && !transactionManager.equals(STORAGE_NAME)) {
```
brfrn169, line 86: Currently, when we set the `transactionManager` config to `jdbc`, the `storage` config is not required. So the `storage` config can be null. So we need to use the `STORAGE_NAME.equals(...)` style to avoid `NullPointerExceptiom`:
```suggestion
    if (!STORAGE_NAME.equals(storage) && !STORAGE_NAME.equals(transactionManager)) {
```

Or we also use `Objects.equals()` as well. Either is fine for me. Thanks.

brfrn169 (approved):
Left a minor comment, but LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
jnmt: @brfrn169 Thank you for pointing it out. I think `cassandra` is always set to the storage by default when we create `DatabaseConfig`, it will never be null. If I understand it correctly, is the assertion better?

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
brfrn169: Ah yes, you are right. I don't remember why we used that reverse style (`"jdbc".equals(storage)`) originally. It looks like we have default values for `storage` and `transactionManager`, so we don't need to care about `NullPointerException`. So let's keep it as it is. Thanks.

feeblefakie (approved):
LGTM! Thank you!

kota2and3kan (approved):
LGTM! Thank you!
(Sorry for my late review... 🙇‍♀️)

jnmt:
@Torch3333 I will merge this, but PTAL when you are back.

PR #1301, opened 2023-11-21
Update documents for cross-partition scan and import feature

jnmt:
## Description

This PR adds documents related to cross-partition scan (so-called the relational scan before) and the table import feature. It depends on #1294, which is still under review, but PTAL in parallel since reviewing and revising the docs might take a long time.

## Related issues and/or PRs

- #931
- #1294

## Changes made

- Add cross-partition scan documents
- Add table import documents

## Checklist

- [ ] Any dependent changes in other PRs have been merged and published. (#1294 should be merged together)

- This PR basically focuses on the specification of the functions and how to use each function. We will provide a holistic guide for the migration to ScalarDB and its sample in other PRs.

## Release notes

Added documents for cross-partition scan and table import.

### DIFF
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -295,6 +295,25 @@ boolean ifExist = true;
 admin.dropCoordinatorTables(ifExist);
 ```
 
+### Import a table
+
+You can import an existing table to ScalarDB as follows:
+
+```java
+// Import the table "ns.tbl". If the table is already managed by ScalarDB, the target table does not
+// exist, or the table does not meet the requirement of ScalarDB table, an exception will be thrown.
+admin.importTable("ns", "tbl");
+```
+
+{% capture notice--warning %}
+**Attention**
+
+You should carefully plan to import a table to ScalarDB in production because it will add transaction metadata columns to your database tables and the ScalarDB metadata tables. There would also be several differences between your database and ScalarDB and limitations. See also the following document.
+
+- [Importing existing tables to ScalarDB using ScalarDB Schema Loader](./schema-loader-import.md)
+
+{{ notice--warning | markdownify }}
+
 ## Transactional API
 
 This section explains how to execute transactional operations by using the Transactional API in ScalarDB.
@@ -621,9 +640,21 @@ You can't specify clustering-key boundaries and orderings in `Scan` by using a s
 
 <div class="notice--info">{{ notice--info | markdownify }}</div>
 
-##### Execute `Scan` without specifying a partition key to retrieve all the records of a table
+##### Execute cross-partition `Scan` without specifying a partition key to retrieve all the records of a table
 
-You can execute a `Scan` operation without specifying a partition key.
+You can execute a `Scan` operation across all partitions without specifying a partition key by enabling the following property in the ScalarDB configuration.
+
+```properties
+scalar.db.cross_partition_scan.enabled=true
+```
+
+{% capture notice--warning %}
+**Attention**
+
+Except for JDBC databases, we do not recommend enabling cross-partition scan with serializable isolation because it could make the isolation level lower (i.e., snapshot). Use it at your own risk only if the consistency does not matter for your transactions.
+{% endcapture %}
+
+<div class="notice--warning">{{ notice--warning | markdownify }}</div>
 
 Instead of calling the `partitionKey()` method in the builder, you can call the `all()` method to scan a table without specifying a partition key as follows:
 
@@ -645,10 +676,70 @@ List<Result> results = transaction.scan(scan);
 {% capture notice--info %}
 **Note**
 
-You can't specify clustering-key boundaries and orderings in `Scan` without specifying a partition key.
+You can't specify any filtering conditions and orderings in cross-partition `Scan` except for JDBC databases. See the following section to use cross-partition `Scan` with filtering or ordering for JDBC databases.
 {% endcapture %}
 
 <div class="notice--info">{{ notice--info | markdownify }}</div>
+
+##### Execute cross-partition `Scan` with filtering and ordering
+
+By enabling the cross-partition scan option with filtering and ordering for JDBC databases as follows, you can execute a cross-partition `Scan` operation with flexible conditions and orderings.
+
+```properties
+scalar.db.cross_partition_scan.enabled=true
+scalar.db.cross_partition_scan.filtering.enabled=true
+scalar.db.cross_partition_scan.ordering.enabled=true
+```
+
+You can call the `where()` and `ordering()` methods after calling the `all()` method to specify arbitrary conditions and orderings as follows:
+
+```java
+// Create a `Scan` operation with arbitrary conditions and orderings.
+Scan scan =
+    Scan.newBuilder()
+        .namespace("ns")
+        .table("tbl")
+        .all()
+        .where(ConditionBuilder.column("c1").isNotEqualToInt(10))
+        .projections("c1", "c2", "c3", "c4")
+        .orderings(Scan.Ordering.desc("c3"), Scan.Ordering.asc("c4"))
+        .limit(10)
+        .build();
+
+// Execute the `Scan` operation.
+List<Result> results = transaction.scan(scan);
+```
+
+As an argument of the `where()` method, you can specify a condition, an and-wise condition set, or an or-wise condition set. After calling the `where()` method, you can add more conditions or condition sets using the `and()` method or `or()` method as follows:
+
+```java
+// Create a `Scan` operation with condition sets.
+Scan scan =
+    Scan.newBuilder()
+        .namespace("ns")
+        .table("tbl")
+        .all()
+        .where(
+            ConditionSetBuilder.condition(ConditionBuilder.column("c1").isLessThanInt(10))
+                .or(ConditionBuilder.column("c1").isGreaterThanInt(20))
+                .build())
+        .and(
+            ConditionSetBuilder.condition(ConditionBuilder.column("c2").isLikeText("a%"))
+                .or(ConditionBuilder.column("c2").isLikeText("b%"))
+                .build())
+        .limit(10)
+        .build();
+```
+
+{% capture notice--info %}
+**Note**
+
+In the `where()` condition method chain, the conditions must be an and-wise junction of `ConditionalExpression` or `OrConditionSet`  (so-called conjunctive normal form) like the above example or an or-wise junction of `ConditionalExpression` or `AndConditionSet`  (so-called disjunctive normal form).
+{% endcapture %}
+
+<div class="notice--info">{{ notice--info | markdownify }}</div>
+
+For more details of available conditions and condition sets, see the `ConditionBuilder` and `ConditionSetBuilder` page in the [Javadoc](https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html) of the version of ScalarDB that you're using.
 
 #### `Put` operation
 
--- a/docs/schema-loader-import.md
+++ b/docs/schema-loader-import.md
@@ -1,0 +1,271 @@
+# Importing existing tables to ScalarDB using ScalarDB Schema Loader
+
+You might want to use ScalarDB (e.g., for database-spanning transactions) with your existing databases. In that case, you can import those databases under the ScalarDB control using ScalarDB Schema Loader. ScalarDB Schema Loader automatically adds ScalarDB-internal metadata columns in each existing table and metadata tables to enable various ScalarDB functionalities including transaction management across multiple databases.
+
+## Before you begin
+
+{% capture notice--warning %}
+**Attention**
+
+You should carefully plan to import a table to ScalarDB in production because it will add transaction metadata columns to your database tables and the ScalarDB metadata tables. There would also be several differences between your database and ScalarDB and limitations.
+{% endcapture %}
+
+<div class="notice--warning">{{ notice--warning | markdownify }}</div>
+
+### What will be added to your databases
+
+- ScalarDB metadata tables: ScalarDB manages namespace names and table metadata in an namespace (schema or database in underlying databases) called 'scalardb'.
+- Transaction metadata columns: the Consensus Commit transaction manager requires metadata (for example, transaction ID, record version, and transaction status) stored along with the actual records to handle transactions properly. Thus, this tool adds the metadata columns if you use the Consensus Commit transaction manager.
+
+### Requirements
+
+- [JDBC databases](./scalardb-supported-databases.md#jdbc-databases) except for SQLite are importable.
+- Each table must have primary key column(s) (composite primary keys can be available) .
+- Target tables must only have columns with supported data type (see also [here](#data-type-mapping-from-jdbc-databases-to-scalardb)).
+
+### Set up Schema Loader
+
+See the [ScalarDB Schema Loader](./schema-loader.md#set-up-schema-loader) document to set up Schema Loader for importing existing tables.
+
+## Run Schema Loader for importing existing tables
+
+You can import an existing table in JDBC databases to ScalarDB using `--import` option and an import-specific schema file. To import tables, run the following command, replacing the contents in the angle brackets as described:
+
+```console
+$ java -jar scalardb-schema-loader-<VERSION>.jar --config <PATH_TO_SCALARDB_PROPERTIES_FILE> -f <PATH_TO_SCHEMA_FILE> --import
+```
+
+- `<VERSION>`: version of ScalarDB Schema Loader that you set up.
+- `<PATH_TO_SCALARDB_PROPERTIES_FILE>`: path to a property file of ScalarDB. For a sample properties file, see [`database.properties`](https://github.com/scalar-labs/scalardb/blob/master/conf/database.properties).
+- `<PATH_TO_SCHEMA_FILE>`: path to an import schema file. See also a [sample](#sample-import-schema-file) in the next section.
+
+If you use the Consensus Commit transaction manager after importing existing tables, run the following command separately.
+
+```console
+$ java -jar scalardb-schema-loader-<VERSION>.jar --config <PATH_TO_SCALARDB_PROPERTIES_FILE> --coordinator
+```
+
+## Sample import schema file
+
+The following is a sample schema for importing tables. For the sample schema file, see [`import_schema_sample.json`](https://github.com/scalar-labs/scalardb/blob/master/schema-loader/sample/import_schema_sample.json).
+
+```json
+{
+  "sample_namespace1.sample_table1": {
+    "transaction": true
+  },
+  "sample_namespace1.sample_table2": {
+    "transaction": true
+  },
+  "sample_namespace2.sample_table3": {
+    "transaction": false
+  }
+}
+```
+
+The import table schema consists of a namespace name, a table name, and a `transaction` field. The `transaction` field indicates whether the table will be imported for transactions or not. If you set the `transaction` field to `true` or don't specify the `transaction` field, this tool creates a table with transaction metadata if needed. If you set the `transaction` field to `false`, this tool imports a table without adding transaction metadata (that is, for a table with [Storage API](storage-abstraction.md)).
+
+## Data-type mapping from JDBC databases to ScalarDB
+
+The following table shows the supported data types in each JDBC database and their mapping to the ScalarDB data types. Select your database and check if your existing tables are importable or not.
+
+<div id="tabset-1">
+<div class="tab">
+  <button class="tablinks" onclick="openTab(event, 'MySQL', 'tabset-1')" id="defaultOpen-1">MySQL</button>
+  <button class="tablinks" onclick="openTab(event, 'PostgreSQL', 'tabset-1')">PostgreSQL</button>
+  <button class="tablinks" onclick="openTab(event, 'Oracle', 'tabset-1')">Oracle</button>
+  <button class="tablinks" onclick="openTab(event, 'SQLServer', 'tabset-1')">SQL Server</button>
+</div>
+
+<div id="MySQL" class="tabcontent" markdown="1">
+
+| MySQL        | ScalarDB | Notes                 |
+|--------------|----------|-----------------------|
+| bigint       | BIGINT   | [*1](#warn-data-size) |
+| binary       | BLOB     |                       |
+| bit          | BOOLEAN  |                       |
+| blob         | BLOB     | [*2](#warn-data-size) |
+| char         | TEXT     | [*2](#warn-data-size) |
+| double       | DOUBLE   |                       |
+| float        | FLOAT    |                       |
+| int          | INT      |                       |
+| int unsigned | BIGINT   | [*2](#warn-data-size) |
+| integer      | INT      |                       |
+| longblob     | BLOB     |                       |
+| longtext     | TEXT     |                       |
+| mediumblob   | BLOB     | [*2](#warn-data-size) |
+| mediumint    | INT      | [*2](#warn-data-size) |
+| mediumtext   | TEXT     | [*2](#warn-data-size) |
+| smallint     | INT      | [*2](#warn-data-size) |
+| text         | TEXT     | [*2](#warn-data-size) |
+| tinyblob     | BLOB     | [*2](#warn-data-size) |
+| tinyint      | INT      | [*2](#warn-data-size) |
+| tinyint(1)   | BOOLEAN  |                       |
+| tinytext     | TEXT     | [*2](#warn-data-size) |
+| varbinary    | BLOB     | [*2](#warn-data-size) |
+| varchar      | TEXT     | [*2](#warn-data-size) |
+
+Data types not listed in the above are not supported. The typical examples are shown below.
+
+- bigint unsigned
+- bit(n) (n > 1)
+- date
+- datetime
+- decimal
+- enum
+- geometry
+- json
+- numeric
+- set
+- time
+- timestamp
+- year
+
+</div>
+
+<div id="PostgreSQL" class="tabcontent" markdown="1">
+
+| PostgreSQL        | ScalarDB | Notes                 |
+|-------------------|----------|-----------------------|
+| bigint            | BIGINT   | [*1](#warn-data-size) |
+| boolean           | BOOLEAN  |                       |
+| bytea             | BLOB     |                       |
+| character         | TEXT     | [*2](#warn-data-size) |
+| character varying | TEXT     | [*2](#warn-data-size) |
+| double precision  | DOUBLE   |                       |
+| integer           | INT      |                       |
+| real              | FLOAT    |                       |
+| smallint          | INT      | [*2](#warn-data-size) |
+| text              | TEXT     |                       |
+
+Data types not listed in the above are not supported. The typical examples are shown below.
+
+- bigserial
+- bit
+- box
+- cidr
+- circle
+- date
+- inet
+- interval
+- json
+- jsonb
+- line
+- lseg
+- macaddr
+- macaddr8
+- money
+- numeric
+- path
+- pg_lsn
+- pg_snapshot
+- point
+- polygon
+- smallserial
+- serial
+- time
+- timestamp
+- tsquery
+- tsvector
+- txid_snapshot
+- uuid
+- xml
+
+</div>
+
+<div id="Oracle" class="tabcontent" markdown="1">
+
+| Oracle        | ScalarDB        | Notes                 |
+|---------------|-----------------|-----------------------|
+| binary_double | DOUBLE          |                       |
+| binary_float  | FLOAT           |                       |
+| blob          | BLOB            | [*3](#warn-data-size) |
+| char          | TEXT            | [*2](#warn-data-size) |
+| clob          | TEXT            |                       |
+| float         | DOUBLE          | [*4](#warn-data-size) |
+| long          | TEXT            |                       |
+| long raw      | BLOB            |                       |
+| nchar         | TEXT            | [*2](#warn-data-size) |
+| nclob         | TEXT            |                       |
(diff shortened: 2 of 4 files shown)


docs/schema-loader-import.md
```diff
@@ -0,0 +1,271 @@
...
+
+You might want to use ScalarDB (e.g., for database-spanning transactions) with your existing databases. In that case, you can import those databases under the ScalarDB control using ScalarDB Schema Loader. ScalarDB Schema Loader automatically adds ScalarDB-internal metadata columns in each existing table and metadata tables to enable various ScalarDB functionalities including transaction management across multiple databases.
+
+## Before you begin
+
+{% capture notice--warning %}
+**Attention**
+
+You should carefully plan to import a table to ScalarDB in production because it will add transaction metadata columns to your database tables and the ScalarDB metadata tables. There would also be several differences between your database and ScalarDB and limitations.
+{% endcapture %}
```
jnmt, line 11: After adding the holistic migration guide, I will add a reference for it around here.

docs/schema-loader-import.md
```diff
@@ -0,0 +1,271 @@
...
+</div>
+
+</div>
+
+{% capture notice--warning %}
+**Attention**
+
+1. The value range of `BIGINT` in ScalarDB is from -2^53 to 2^53, regardless of the size of `bigint` in the underlying database. Thus, if the data out of this range exists in the imported table, ScalarDB cannot read it.
+2. For certain data types noted above, ScalarDB may map a data type larger than that of the underlying database. In that case, You will see errors when putting a value with a size larger than the size specified in the underlying database.
+3. The maximum size of `BLOB` in ScalarDB is about 2GB (precisely 2^31-1 bytes). In contrast, Oracle `blob` can have (4GB-1)*(number of blocks). Thus, if the data larger than 2GB exists in the imported table, ScalarDB cannot read it.
```
kota2and3kan, line 258: Ditto. Do we observe `null`, some value, or an error?

docs/api-guide.md
```diff
@@ -645,11 +676,71 @@ List<Result> results = transaction.scan(scan);
...
+
+##### Execute cross-partition `Scan` with filtering and ordering
+
+By enabling the cross-partition scan option with filtering and ordering for JDBC databases as follows, you can execute a cross-partition `Scan` operation with flexible conditions and orderings.
+
+```properties
+scalar.db.cross_partition_scan.enabled=true
+scalar.db.cross_partition_scan.filtering.enabled=true
+scalar.db.cross_partition_scan.ordering.enabled=true
+```
```
kota2and3kan, line 692: It seems that the description of this parameter is added to ScalarDB SQL document in PR https://github.com/scalar-labs/scalardb-sql/pull/347.

However, I think it might be better to add descriptions in the `docs/configurations.md` on the ScalarDB (`scalardb` repository) side since we can use related API in the ScalarDB core API.

What do you think?

kota2and3kan (commented):
Thank you for the updates!
I left some suggestions and questions.
Please take a look when you have time!

docs/schema-loader-import.md
```diff
@@ -0,0 +1,271 @@
...
+- xml
+
+</div>
+
+</div>
+
+{% capture notice--warning %}
+**Attention**
+
+1. The value range of `BIGINT` in ScalarDB is from -2^53 to 2^53, regardless of the size of `bigint` in the underlying database. Thus, if the data out of this range exists in the imported table, ScalarDB cannot read it.
```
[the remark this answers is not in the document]
jnmt: Good question. The answer is #3. `IllegalArgumentException` will be thrown with a message like: "Out of range column value for BigInt. value: xxx".

docs/schema-loader-import.md
jnmt: Good question again. The 2GB limit is due to Java's byte array limits. I don't test it, but I guess the Oracle JDBC driver throws an `SQLException`. Or, we might see an OOM error if the heap size is not correctly configured. Apart from that, we might be able to handle such large objects better by using JDBC `Blob getBlob(...)` and offset-based access instead of `byte[] getBytes(...)`.

docs/api-guide.md
jnmt: Definitely, I will revise both.

docs/api-guide.md
```diff
@@ -645,11 +676,71 @@ List<Result> results = transaction.scan(scan);
...
+                .or(ConditionBuilder.column("c2").isLikeText("b%"))
+                .build())
+        .limit(10)
+        .build();
+```
+
+{% capture notice--info %}
+**Note**
+
+In the `where()` condition method chain, the conditions must be an and-wise junction of `ConditionalExpression` or `OrConditionSet`  (so-called conjunctive normal form) like the above example or an or-wise junction of `ConditionalExpression` or `AndConditionSet`  (so-called disjunctive normal form).
```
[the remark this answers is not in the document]
jnmt: Fixed in 6726530.

docs/api-guide.md
jnmt: Fixed in 6726530.

jnmt:
@kota2and3kan Thanks for the feedback! Fixed based on the feedback (though the exception-related part is left as is for now), so PTAL when you get a chance.

docs/schema-loader-import.md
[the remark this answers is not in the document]
kota2and3kan: I see. I understand.
Thank you!

docs/schema-loader-import.md
kota2and3kan: Ah, I see. I understood that it depends on Java side.
Thank you!

docs/api-guide.md
```diff
@@ -621,9 +640,21 @@ You can't specify clustering-key boundaries and orderings in `Scan` by using a s
...
+You can execute a `Scan` operation across all partitions without specifying a partition key by enabling the following property in the ScalarDB configuration.
+
+```properties
+scalar.db.cross_partition_scan.enabled=true
+```
+
+{% capture notice--warning %}
+**Attention**
+
+Except for JDBC databases, we do not recommend enabling cross-partition scan with serializable isolation because it could make the isolation level lower (i.e., snapshot). Use it at your own risk only if the consistency does not matter for your transactions.
```
feeblefakie, line 654:
```suggestion
We do not recommend enabling cross-partition scan with serializable isolation for non-JDBC databases because transactions could be executed with lower isolation (i.e., snapshot). Use it at your own risk only if the consistency does not matter for your transactions.
```

The expression `it could make the isolation level lower ` sounds a bit unclear to me.

BTW, sorry, I don't fully remember the discussion we had before.
So, we decided to only warn in case users use the cross-partition scan with serializable isolation for backward compatibility instead of throwing a runtime exception, don't we?

docs/api-guide.md
```diff
@@ -621,9 +640,21 @@ You can't specify clustering-key boundaries and orderings in `Scan` by using a s
 
 <div class="notice--info">{{ notice--info | markdownify }}</div>
 
-##### Execute `Scan` without specifying a partition key to retrieve all the records of a table
+##### Execute cross-partition `Scan` without specifying a partition key to retrieve all the records of a table
+
+You can execute a `Scan` operation across all partitions without specifying a partition key by enabling the following property in the ScalarDB configuration.
```
feeblefakie, line 645:
```suggestion
You can execute a `Scan` operation across all partitions, which we call cross-partition scan, without specifying a partition key by enabling the following property in the ScalarDB configuration.
```

docs/configurations.md
```diff
@@ -158,6 +158,24 @@ For details about using multiple storages, see [Multi-Storage Transactions](mult
...
 
+## Cross-partition scan configurations
+
+By enabling the cross-partition scan option below, `Scan` operation can retrieve all records across partitions. In addition, you can specify arbitrary conditions and orderings in the cross-partition `Scan` operation by enabling `cross_partition_scan.filtering` and `cross_partition_scan.ordering`, respectively. Currently, the cross-partition scan with filtering and ordering is available only for JDBC databases. Note that `scalar.db.cross_partition_scan.enabled` must be `true` to enable them. See [Java API Guide - Scan operation](./api-guide.md#scan-operation) for how to use the cross-partition scan.
+
+{% capture notice--warning %}
+**Attention**
+
+Except for JDBC databases, we do not recommend enabling cross-partition scan with serializable isolation because it could make the isolation level lower (i.e., snapshot). Use it at your own risk only if the consistency does not matter for your transactions.
```
feeblefakie, line 168: Ditto.

docs/schema-loader-import.md
```diff
@@ -0,0 +1,271 @@
...
+
+{% capture notice--warning %}
+**Attention**
+
+You should carefully plan to import a table to ScalarDB in production because it will add transaction metadata columns to your database tables and the ScalarDB metadata tables. There would also be several differences between your database and ScalarDB and limitations.
+{% endcapture %}
+
+<div class="notice--warning">{{ notice--warning | markdownify }}</div>
+
+### What will be added to your databases
```
feeblefakie, line 15: I think we should talk about how much importing is lightweight, like, the time for importing doesn't take proportionally to the database size and usually takes several seconds.

docs/api-guide.md
```diff
@@ -295,6 +295,25 @@ boolean ifExist = true;
...
 ```
 
+### Import a table
+
+You can import an existing table to ScalarDB as follows:
+
+```java
+// Import the table "ns.tbl". If the table is already managed by ScalarDB, the target table does not
+// exist, or the table does not meet the requirement of ScalarDB table, an exception will be thrown.
+admin.importTable("ns", "tbl");
```
brfrn169, line 305: We have added `Map<String, String> options` as 3rd argument to this method:
```suggestion
admin.importTable("ns", "tbl", options);
```

And we didn't add the argument for the `3` branch. To avoid diverging this doc between `master` and `3`, so we should probably apply the same change for the `3` branch.

docs/schema-loader.md
```diff
@@ -714,3 +722,35 @@ SchemaLoader.repairTables(properties, serializedSchemaJson, tableReparationOptio
...
+  public static int main(String... args) throws SchemaLoaderException {
+    Path configFilePath = Paths.get("database.properties");
+    // "import_sample_schema.json" can be found in the "/sample" directory.
+    Path schemaFilePath = Paths.get("import_sample_schema.json");
+    Map<String, String> tableCreationOptions = new HashMap<>();
+
+    // Import tables.
+    // You can also use a Properties object instead of configFilePath and a serialized-schema JSON
+    // string instead of schemaFilePath.
+    SchemaLoader.load(configFilePath, schemaFilePath, tableCreationOptions, createCoordinatorTables);
```
brfrn169, line 751: `importTables`, right?
```suggestion
    SchemaLoader.importTables(configFilePath, schemaFilePath, tableCreationOptions);
```

brfrn169 (commented):
Overall, LGTM. Left a couple of comments. Please take a look when you have time!

docs/api-guide.md
jnmt: Thanks for the comment. Fixed in e3dd00d.

> BTW, sorry, I don't fully remember the discussion we had before.
So, we decided to only warn in case users use the cross-partition scan with serializable isolation for backward compatibility instead of throwing a runtime exception, don't we?

At least, we didn't choose to completely disable the cross-partition scan with serializable isolation in v4.0. Disabling it is one idea, but I think it might be useful in some cases regardless of backward compatibility; e.g., users want to basically run transactions in a serializable manner but sometimes run read-only cross-partition scans without changing the setting.

@brfrn169 Do you have any idea?

docs/api-guide.md
jnmt: Fixed in e3dd00d.

docs/configurations.md
jnmt: Fixed in e3dd00d.

docs/schema-loader-import.md
jnmt: Fixed in e3dd00d.

docs/api-guide.md
jnmt: Good catch, thank you! Fixed in e3dd00d.

> And we didn't add the argument for the 3 branch. To avoid diverging this doc between master and 3, so we should probably apply the same change for the 3 branch.

I'm OK to use `admin.importTable("ns", "tbl", options)` even in v3.x, but might be confusing for users. If we can just apply it without any other concerns, I can handle it since it's not a big diverge. Should I or not?

docs/schema-loader.md
jnmt: Thank you! Fixed in e3dd00d.

docs/api-guide.md
brfrn169: Actually, the `importTable()` method was already introduced in 3.10, but it was treated as an experiment feature. Therefore, I think we can also add the `options` argument to the `3` branch.

docs/schema-loader.md
brfrn169: One more thing, it looks like the `importTables` method doesn't receive the `createCoordinatorTables` argument.

And maybe `tableCreationOptions` should be renamed to something like `tableImportOptions` option?

docs/api-guide.md
jnmt: Ahh, right. Got it. Thank you!

docs/schema-loader.md
jnmt: Oops, sorry about that. I fixed it throughout the sample in 148d9e2. PTAL!

brfrn169 (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM! Thank you!

docs/api-guide.md
feeblefakie: OK, thank you! So, for now, we need to enable it in 3.x for backward-compatibility, and we haven't decided to do so in 4.x (we need to think what we should do for 4.x). Is my understanding correct?

feeblefakie (approved):
LGTM! Thank you!

docs/api-guide.md
```diff
@@ -621,9 +640,21 @@ You can't specify clustering-key boundaries and orderings in `Scan` by using a s
...
+You can execute a `Scan` operation across all partitions, which we call cross-partition scan, without specifying a partition key by enabling the following property in the ScalarDB configuration.
+
+```properties
+scalar.db.cross_partition_scan.enabled=true
+```
+
+{% capture notice--warning %}
+**Attention**
+
+We do not recommend enabling the cross-partition scan with serializable isolation for non-JDBC databases because transactions could be executed with lower isolation (i.e., snapshot). Use it at your own risk only if the consistency does not matter for your transactions.
```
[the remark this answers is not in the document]
jnmt: Fixed in de406a2. Thank you for the feedback!

docs/api-guide.md
jnmt: Almost, yes. My understanding is that we will keep warning it as the same as v3.x unless we make a decision to stop the feature in v4.x explicitly.

docs/schema-loader-import.md
```diff
@@ -0,0 +1,272 @@
...
+
+- ScalarDB metadata tables: ScalarDB manages namespace names and table metadata in an namespace (schema or database in underlying databases) called 'scalardb'.
+- Transaction metadata columns: the Consensus Commit transaction manager requires metadata (for example, transaction ID, record version, and transaction status) stored along with the actual records to handle transactions properly. Thus, this tool adds the metadata columns if you use the Consensus Commit transaction manager.
+- Keep in mind that the tool only changes database metadata; thus, the processing time does not increase in proportion to the database size and usually takes only several seconds.
+
+### Requirements
+
+- [JDBC databases](./scalardb-supported-databases.md#jdbc-databases) except for SQLite are importable.
+- Each table must have primary key column(s) (composite primary keys can be available) .
+- Target tables must only have columns with supported data type (see also [here](#data-type-mapping-from-jdbc-databases-to-scalardb)).
```
josh-wong, line 25:
```suggestion
- [JDBC databases](./scalardb-supported-databases.md#jdbc-databases), except for SQLite, can be imported.
- Each table must have primary key columns. (Composite primary keys can be available.)
- Target tables must only have columns with supported data types. For details, see [Data-type mapping from JDBC databases to ScalarDB](#data-type-mapping-from-jdbc-databases-to-scalardb)).
```

docs/schema-loader-import.md
```diff
@@ -0,0 +1,272 @@
...
+You should carefully plan to import a table to ScalarDB in production because it will add transaction metadata columns to your database tables and the ScalarDB metadata tables. There would also be several differences between your database and ScalarDB and limitations.
+{% endcapture %}
+
+<div class="notice--warning">{{ notice--warning | markdownify }}</div>
+
+### What will be added to your databases
+
+- ScalarDB metadata tables: ScalarDB manages namespace names and table metadata in an namespace (schema or database in underlying databases) called 'scalardb'.
+- Transaction metadata columns: the Consensus Commit transaction manager requires metadata (for example, transaction ID, record version, and transaction status) stored along with the actual records to handle transactions properly. Thus, this tool adds the metadata columns if you use the Consensus Commit transaction manager.
+- Keep in mind that the tool only changes database metadata; thus, the processing time does not increase in proportion to the database size and usually takes only several seconds.
```
josh-wong, line 19:
```suggestion
- **ScalarDB metadata tables:** ScalarDB manages namespace names and table metadata in a namespace (schema or database in underlying databases) called 'scalardb'.
- **Transaction metadata columns:** The Consensus Commit transaction manager requires metadata (for example, transaction ID, record version, and transaction status) stored along with the actual records to handle transactions properly. Thus, this tool adds the metadata columns if you use the Consensus Commit transaction manager.

{% capture notice--info %}
**Note**

This tool only changes database metadata. Thus, the processing time does not increase in proportion to the database size and usually takes only several seconds.
{% endcapture %}

<div class="notice--info">{{ notice--info | markdownify }}</div>
```

docs/configurations.md
```diff
@@ -158,6 +158,24 @@ For details about using multiple storages, see [Multi-Storage Transactions](mult
 
 For details about client configurations, see the ScalarDB Cluster [client configurations (redirects to the Enterprise docs site)](https://scalardb.scalar-labs.com/docs/latest/scalardb-cluster/developer-guide-for-scalardb-cluster-with-java-api/#client-configurations).
 
+## Cross-partition scan configurations
+
+By enabling the cross-partition scan option below, `Scan` operation can retrieve all records across partitions. In addition, you can specify arbitrary conditions and orderings in the cross-partition `Scan` operation by enabling `cross_partition_scan.filtering` and `cross_partition_scan.ordering`, respectively. Currently, the cross-partition scan with filtering and ordering is available only for JDBC databases. Note that `scalar.db.cross_partition_scan.enabled` must be `true` to enable them. See [Java API Guide - Scan operation](./api-guide.md#scan-operation) for how to use the cross-partition scan.
```
josh-wong, line 163:
```suggestion
By enabling the cross-partition scan option below, the `Scan` operation can retrieve all records across partitions. In addition, you can specify arbitrary conditions and orderings in the cross-partition `Scan` operation by enabling `cross_partition_scan.filtering` and `cross_partition_scan.ordering`, respectively. Currently, the cross-partition scan with filtering and ordering is available only for JDBC databases. To enable filtering and ordering, `scalar.db.cross_partition_scan.enabled` must be set to `true`.

For details on how to use cross-partition scan, see [Scan operation](./api-guide.md#scan-operation).
```

docs/configurations.md
```diff
@@ -158,6 +158,24 @@ For details about using multiple storages, see [Multi-Storage Transactions](mult
...
+
+{% capture notice--warning %}
+**Attention**
+
+We do not recommend enabling the cross-partition scan with `SERIALIAZABLE` isolation level for non-JDBC databases because transactions could be executed with lower isolation level (i.e., `SNAPSHOT`). Use it at your own risk only if the consistency does not matter for your transactions.
```
josh-wong, line 168: This suggestion is similar to the one I left in a related doc that we are updating for ScalarDB SQL.

```suggestion
For non-JDBC databases, we do not recommend enabling cross-partition scan with the `SERIALIAZABLE` isolation level because transactions could be executed at a lower isolation level (for example, `SNAPSHOT`). Use the `SERIALIAZABLE` isolation level at your own risk only if consistency does not matter for your transactions.
```

docs/api-guide.md
```diff
@@ -621,9 +640,21 @@ You can't specify clustering-key boundaries and orderings in `Scan` by using a s
...
+You can execute a `Scan` operation across all partitions, which we call cross-partition scan, without specifying a partition key by enabling the following property in the ScalarDB configuration.
+
+```properties
+scalar.db.cross_partition_scan.enabled=true
+```
+
+{% capture notice--warning %}
+**Attention**
+
+We do not recommend enabling the cross-partition scan with `SERIALIAZABLE` isolation level for non-JDBC databases because transactions could be executed with lower isolation level (i.e., `SNAPSHOT`). Use it at your own risk only if the consistency does not matter for your transactions.
```
josh-wong, line 654: This suggestion is similar to the one I left in a related doc that we are updating for ScalarDB SQL.

```suggestion
For non-JDBC databases, we do not recommend enabling cross-partition scan with the `SERIALIAZABLE` isolation level because transactions could be executed at a lower isolation level (for example, `SNAPSHOT`). Use the `SERIALIAZABLE` isolation level at your own risk only if consistency does not matter for your transactions.
```

docs/configurations.md
```diff
@@ -158,6 +158,24 @@ For details about using multiple storages, see [Multi-Storage Transactions](mult
...
+{% endcapture %}
+
+<div class="notice--warning">{{ notice--warning | markdownify }}</div>
+
+| Name                                               | Description                                   | Default |
+|----------------------------------------------------|-----------------------------------------------|---------|
+| `scalar.db.cross_partition_scan.enabled`           | Enable the cross partition scan.              | `false` |
+| `scalar.db.cross_partition_scan.filtering.enabled` | Enable filtering in the cross partition scan. | `false` |
+| `scalar.db.cross_partition_scan.ordering.enabled`  | Enable ordering in the cross partition scan.  | `false` |
```
josh-wong, line 177:
```suggestion
| `scalar.db.cross_partition_scan.enabled`           | Enable cross-partition scan.                  | `false` |
| `scalar.db.cross_partition_scan.filtering.enabled` | Enable filtering in cross-partition scan.     | `false` |
| `scalar.db.cross_partition_scan.ordering.enabled`  | Enable ordering in cross-partition scan.      | `false` |
```

docs/api-guide.md
```diff
@@ -621,9 +640,21 @@ You can't specify clustering-key boundaries and orderings in `Scan` by using a s
 
 <div class="notice--info">{{ notice--info | markdownify }}</div>
 
-##### Execute `Scan` without specifying a partition key to retrieve all the records of a table
+##### Execute cross-partition `Scan` without specifying a partition key to retrieve all the records of a table
+
+You can execute a `Scan` operation across all partitions, which we call cross-partition scan, without specifying a partition key by enabling the following property in the ScalarDB configuration.
```
josh-wong, line 645:
```suggestion
You can execute a `Scan` operation across all partitions, which we call *cross-partition scan*, without specifying a partition key by enabling the following property in the ScalarDB configuration.
```

Or should the last part be `configuration in the ScalarDB properties file` as follows instead?:

```suggestion
You can execute a `Scan` operation across all partitions, which we call *cross-partition scan*, without specifying a partition key by enabling the following configuration in the ScalarDB properties file.
```

docs/api-guide.md
```diff
@@ -645,11 +676,71 @@ List<Result> results = transaction.scan(scan);
 {% capture notice--info %}
 **Note**
 
-You can't specify clustering-key boundaries and orderings in `Scan` without specifying a partition key.
+You can't specify any filtering conditions and orderings in cross-partition `Scan` except for JDBC databases. See the following section to use cross-partition `Scan` with filtering or ordering for JDBC databases.
```
josh-wong, line 679: Added a link here in case the section gets moved later on and becomes difficult for readers to find.

```suggestion
You can't specify any filtering conditions and orderings in cross-partition `Scan` except for when using JDBC databases. For details on how to use cross-partition `Scan` with filtering or ordering for JDBC databases, see [Execute cross-partition `Scan` with filtering and ordering](#execute-cross-partition-scan-with-filtering-and-ordering).
```

docs/api-guide.md
```diff
@@ -645,11 +676,71 @@ List<Result> results = transaction.scan(scan);
...
+{% endcapture %}
+
+<div class="notice--info">{{ notice--info | markdownify }}</div>
+
+##### Execute cross-partition `Scan` with filtering and ordering
+
+By enabling the cross-partition scan option with filtering and ordering for JDBC databases as follows, you can execute a cross-partition `Scan` operation with flexible conditions and orderings.
```
josh-wong, line 686:
```suggestion
By enabling the cross-partition scan option with filtering and ordering for JDBC databases as follows, you can execute a cross-partition `Scan` operation with flexible conditions and orderings:
```

docs/api-guide.md
```diff
@@ -645,11 +676,71 @@ List<Result> results = transaction.scan(scan);
...
+        .projections("c1", "c2", "c3", "c4")
+        .orderings(Scan.Ordering.desc("c3"), Scan.Ordering.asc("c4"))
+        .limit(10)
+        .build();
+
+// Execute the `Scan` operation.
+List<Result> results = transaction.scan(scan);
+```
+
+As an argument of the `where()` method, you can specify a condition, an and-wise condition set, or an or-wise condition set. After calling the `where()` method, you can add more conditions or condition sets using the `and()` method or `or()` method as follows:
```
josh-wong, line 713:
```suggestion
As an argument of the `where()` method, you can specify a condition, an and-wise condition set, or an or-wise condition set. After calling the `where()` method, you can add more conditions or condition sets by using the `and()` method or `or()` method as follows:
```

docs/api-guide.md
```diff
@@ -645,11 +676,71 @@ List<Result> results = transaction.scan(scan);
...
+                .or(ConditionBuilder.column("c2").isLikeText("b%"))
+                .build())
+        .limit(10)
+        .build();
+```
+
+{% capture notice--info %}
+**Note**
+
+In the `where()` condition method chain, the conditions must be an and-wise junction of `ConditionalExpression` or `OrConditionSet` (so-called conjunctive normal form) like the above example or an or-wise junction of `ConditionalExpression` or `AndConditionSet` (so-called disjunctive normal form).
```
josh-wong, line 737:
```suggestion
In the `where()` condition method chain, the conditions must be an and-wise junction of `ConditionalExpression` or `OrConditionSet` (known as conjunctive normal form) like the above example or an or-wise junction of `ConditionalExpression` or `AndConditionSet` (known as disjunctive normal form).
```

docs/api-guide.md
```diff
@@ -645,11 +676,71 @@ List<Result> results = transaction.scan(scan);
...
 {% endcapture %}
 
 <div class="notice--info">{{ notice--info | markdownify }}</div>
 
+For more details of available conditions and condition sets, see the `ConditionBuilder` and `ConditionSetBuilder` page in the [Javadoc](https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html) of the version of ScalarDB that you're using.
```
josh-wong, line 742:
```suggestion
For more details about available conditions and condition sets, see the `ConditionBuilder` and `ConditionSetBuilder` page in the [Javadoc](https://javadoc.io/doc/com.scalar-labs/scalardb/latest/index.html) of the version of ScalarDB that you're using.
```

josh-wong (changes requested):

docs/schema-loader.md
```diff
@@ -79,6 +79,8 @@ Create/Delete schemas in the storage defined in the config file
       --coordinator   Create/delete/repair Coordinator tables
   -D, --delete-all    Delete tables
   -f, --schema-file=<schemaFile>
+  -I, --import        Import tables : it will import existing non-ScalarDB
+                        tables to ScalarDB.
```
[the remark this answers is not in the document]
jnmt: This change will make the help message in the command inconsistent, and we should improve other commands as well. So let me fix this in another PR.

docs/schema-loader-import.md
```diff
@@ -0,0 +1,272 @@
+# Importing existing tables to ScalarDB using ScalarDB Schema Loader
+
+You might want to use ScalarDB (e.g., for database-spanning transactions) with your existing databases. In that case, you can import those databases under the ScalarDB control using ScalarDB Schema Loader. ScalarDB Schema Loader automatically adds ScalarDB-internal metadata columns in each existing table and metadata tables to enable various ScalarDB functionalities including transaction management across multiple databases.
+
+## Before you begin
+
+{% capture notice--warning %}
+**Attention**
+
+You should carefully plan to import a table to ScalarDB in production because it will add transaction metadata columns to your database tables and the ScalarDB metadata tables. There would also be several differences between your database and ScalarDB and limitations.
```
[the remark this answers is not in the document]
jnmt: @josh-wong Do you mean the following change?

```suggestion
You should carefully plan to import a table to ScalarDB in production because it will add transaction metadata columns to your database tables and the ScalarDB metadata tables. In this case, there would also be several differences between your database and ScalarDB, as well as some limitations.
```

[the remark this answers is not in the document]
josh-wong: Oops! Yes, that is correct. Copy-paste error...😞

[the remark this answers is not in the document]
jnmt: No problem! Thank you!

docs/schema-loader.md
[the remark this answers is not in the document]
josh-wong: Thank you for letting me know! I wondered if that might be the case. It's a minor issues, so no worries about getting to it later🙂

josh-wong (approved):
LGTM! Thank you!🙇‍♂️

PR #1296, opened 2023-11-17
Fix to prevent importing existing storage table

jnmt:
## Description

This PR fixes a bug that unintentionally changes a storage table to a transaction table using the import function. This happens because we used the consensus commit `getTableMetadata()` API to check the existence of tables, but `getTableMetadata()` returns null not only if the table does not exist but if transaction metadata does not exist.

## Related issues and/or PRs

- #931

## Changes made

- Check the table metadata using storage admin API

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

Currently, we don't have an upgrade feature from storage to transaction, but probably we can do it easily using similar logic like import and repair in the future. 

## Release notes

Fixed to prevent importing existing storage table

PR #1309, opened 2023-11-22
Backport to branch(3) : Fix to prevent importing existing storage table

jnmt:
Backport of #1296


jnmt:
@brfrn169 Conflicted due to #1260 since it changed the `importTable()` interface but it's not included in v3.x. We might backport that part only, but I didn't include it for now. PTAL and merge if no problem.

brfrn169 (approved):
LGTM! Thank you for handling this!

> We might backport that part only, but I didn't include it for now. PTAL and merge if no problem.

I believe that's fine for now. We can consider backporting that part to `3` in the future if necessary.

PR #1318, opened 2023-11-23
Add AuthAdmin

brfrn169:
## Description

This PR adds the `AuthAdmin` interface for administrative operations related to ScalarDB Auth (authentication and authorization). As ScalarDB Auth is an enterprise feature, this PR only introduces the interface without any implementations in this repository.

## Related issues and/or PRs

N/A

## Changes made

- Added the `AuthAdmin` interface for administrative operations for ScalarDB
- Modify exceptions to support authentication and authorization errors

## Checklist

- Documentation for this will be added in a separate PR.

## Release notes

N/A


core/src/main/java/com/scalar/db/api/AuthAdmin.java
```diff
@@ -0,0 +1,166 @@
...
+import com.scalar.db.exception.storage.ExecutionException;
+import java.util.EnumSet;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * An interface for administrative operations for authentication and authorization. This interface
+ * is used for creating and dropping users, and granting and revoking privileges.
+ */
+public interface AuthAdmin {
```
brfrn169, line 12: Added `AuthAdmin` for administrative operations for ScalarDB Auth.

core/src/main/java/com/scalar/db/api/DistributedTransactionAdmin.java
```diff
@@ -8,7 +8,7 @@
  * An administrative interface for distributed transaction implementations. The user can execute
  * administrative operations with it like createNamespace/createTable/getTableMetadata.
  */
-public interface DistributedTransactionAdmin extends Admin {
+public interface DistributedTransactionAdmin extends Admin, AuthAdmin {
```
brfrn169, line 11: Made `DistributedTransactionAdmin` extend `AuthAdmin`.

core/src/main/java/com/scalar/db/exception/storage/ExecutionException.java
```diff
@@ -1,12 +1,58 @@
...
+import com.scalar.db.api.AuthAdmin;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
 public class ExecutionException extends Exception {
 
+  private final boolean authenticationError;
+  private final boolean authorizationError;
+  private final boolean superuserRequired;
+  @Nullable private final AuthAdmin.Privilege requiredPrivilege;
```
brfrn169, line 12: Added flags to the exceptions to indicate authentication and authorization errors. Initially, I considered introducing new exceptions for these errors, but to maintain backward compatibility, I opted for this approach.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
```diff
@@ -240,6 +243,58 @@ public void upgrade(Map<String, String> options) throws ExecutionException {
...
+
+  @Override
+  public EnumSet<Privilege> getPrivileges(String username, String namespaceName) {
+    throw new UnsupportedOperationException("Not supported in the community edition");
+  }
+
+  @Override
+  public EnumSet<Privilege> getPrivileges(String username, String namespaceName, String tableName) {
+    throw new UnsupportedOperationException("Not supported in the community edition");
+  }
```
brfrn169, line 296: No implementations are added here as it's an enterprise feature.

core/src/main/java/com/scalar/db/api/AuthAdmin.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * If specified, the user is created as a non-superuser. If neither SUPERUSER nor NO_SUPERUSER
+     * is specified, the user is created as a non-superuser.
+     */
+    NO_SUPERUSER,
+  }
+
+  /** The privileges for ScalarDB operations. */
+  enum Privilege {
```
komamitsu, line 141: This seems like an union of privileges for user (admin) operations, namespace operations and table operations. For instance, GRANT is for user (admin) operations and TRUNCATE is for table operations. Another option is to separate this set of privileges into each type of operations like UserPrivilege, NamespacePrivilege and TablePrivilege. This option requires additional enum definitions, but it might beneficial in the long run to avoid wrong privilege usage (e.g, TRUNCATE for a user). What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
```diff
@@ -240,6 +243,58 @@ public void upgrade(Map<String, String> options) throws ExecutionException {
     admin.upgrade(options);
   }
 
+  @Override
+  public void createUser(String username, @Nullable String password, UserOption... userOption) {
+    throw new UnsupportedOperationException("Not supported in the community edition");
```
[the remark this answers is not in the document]
brfrn169: I had the same thought, but I didn't do that because I was not sure for which methods we should add a default implementation. In other words, should we add default implementations for the existing methods (`createTable()`, `getTableMetadata()`, etc.) as well? What do you think?

core/src/main/java/com/scalar/db/api/AuthAdmin.java
brfrn169: Let me clarify the specification to ensure we are on the same page. 

Currently, we can grant privileges to a user for either a specific namespace or a table. I think we can imagine the SQL `GRANT` statement as follows:

For table privileges:
```sql
GRANT READ, WRITE, CREATE ON [TABLE] TO [USER] WITH GRANT OPTION
```

In this case, the user has `READ`, `WRITE`, `CREATE` and `GRANT` privileges for the specified table. 

- `READ`: The user can read the table.
- `WRITE`: The user can write the table.
- `CREATE`: The user can create indexes on the table.
- `GRANT`: The user can grant privileges on the table to other users.

For namespace privileges:
```sql
GRANT READ, WRITE, CREATE ON NAMESPACE [NAMESPACE] TO [USER] WITH GRANT OPTION
```

In this case, the user has `READ`, `WRITE`, `CREATE` and `GRANT` privileges for all tables in the namespace.

- `READ`: The user can read all tables in the namespace.
- `WRITE`: The user can write all tables in the namespace.
- `CREATE`: The user can create a table in the namespace or create indexes on all tables in the namespace.
- `GRANT`: The user can grant privileges on all tables in the namespace to other users.

Given this, I believe the approach is consistent and there's no need for separate privilege types for tables, namespaces, and users. What are your thoughts?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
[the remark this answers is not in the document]
komamitsu: I think only the methods of `AuthAdmin` should have default methods to throw the exception since those features aren't supported in the community edition. As for the existing methods coming from `DistributedTransactionAdmin`, we should keep them available in the community edition and don't need default impl for them. But I might be missing something on the background of this auth feature.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransactionAdmin.java
```diff
@@ -153,4 +156,56 @@ public void upgrade(Map<String, String> options) throws ExecutionException {
   public void close() {
     jdbcAdmin.close();
   }
+
+  @Override
+  public void createUser(String username, @Nullable String password, UserOption... userOption) {
+    throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
brfrn169: Initially, I didn't add any message here because we were going to remove the JDBC transaction and ScalarDB Server components. However, recent discussions have raised the possibility of retaining the JDBC transaction. Therefore, I will add a message depending on the outcome of this discussion:
https://github.com/scalar-labs/scalardb/pull/1318#discussion_r1403852190

core/src/main/java/com/scalar/db/api/AuthAdmin.java
komamitsu: Thanks for clarifying the usage. I understand how to use GRANT.

GRANT privilege seems a bit different type (more metadata-like?) attribute to me from other privileges (e.g. CREATE). It's like between `UserOption` and `Privilege`? It might fit `PrivilegeOption`.

> In this case, the user has READ, WRITE, CREATE and GRANT privileges for all tables in the namespace.

Understood. So, the enum is like `TablePrivilege`. BTW, we don't need to another type of privileges to allow or prohibit operations for namespaces like `GRANT CREATE ON DATABASE`?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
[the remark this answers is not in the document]
brfrn169: Okay, I think that's reasonable. Let me add default implementations to the methods in `AuthAdmin`. Thanks.

core/src/main/java/com/scalar/db/api/AuthAdmin.java
brfrn169: Ah sorry, the explanation of `GRANT` for namespace privileges was wrong.  

The correct explanation is as follows

- `GRANT`: The user can grant/revoke privileges on all tables in the namespace to/from other users. Also the user can grant/revoke privileges on the namespace to/from other users.

So for example, if a superuser issues the following:

```sql
GRANT... ON NAMESPACE ns TO user1 WITH GRANT OPTION
```

Then `user1` can issue `GRANT... ON NAMESPACE ns to ...` for other user.

What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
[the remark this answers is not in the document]
brfrn169: Fixed in ca46a9d94507f4ba568ad8947f687f1c1f1489b7. Thanks.

core/src/main/java/com/scalar/db/api/AuthAdmin.java
brfrn169: One more thing, if a superuser issues the following `GRANT` command:

```sql
GRANT DROP ON NAMESPACE ns TO user1
```

Then, `user1` can drop the `ns` namespace as well as all tables in the `ns` namespace.

That sounds inconsistent to you?

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/AuthAdmin.java
brfrn169: Thank you for summarizing the discussion. I've fixed the Javadoc slightly.

I think we can support something like the following grammar:
```sql
GRANT { { SELECT | INSERT | UPDATE | DELETE | CREATE | DROP | TRUNCATE | ALTER } [, ...] | ALL [ PRIVILEGES ] }
    ON { [ TABLE ] table_name [, ...] | ALL TABLES IN NAMESPACE namespace_name [, ...] }
    TO user_name [, ...] [ WITH GRANT OPTION ]
```

If a user specified `ON [TABLE] table_name` in the `GRANT` statement, we call the `AuthAdmin.grant(String username, String namespaceName, String tableName, Privilege... privileges)` method. And if a user specified `ALL TABLES IN NAMESPACE namespace_name` in the `GRANT` statement, we call the `AuthAdmin.grant(String username, String namespaceName, Privilege... privileges)` method.

Also, if a user specified `WITH GRANT OPTION` and `ON [TABLE] table_name` in the GRANT statement, the granted user can issue GRANT statements, and they can grant privileges on the specified table to other users.

And if if a user specified `WITH GRANT OPTION` and `ALL TABLES IN NAMESPACE namespace_name` in the `GRANT` statement, they can grant privileges on all tables in the specified namespace to other users.

Note that users can only grant privileges that they have.

Does that make sense?

komamitsu: Yes. The syntax of the GRANT statement and the behavior looks good 👍 

> As for Privilege.GRANT, I still think it might belong to a different layer, but I don't have a so strong opinion on it. Let's proceed with the current design.

This comment only considers the internal design and implementation since I still think `WITH GRANT OPTION` is a sort of option not a normal privilege. The above GRANT statement syntax itself also seems to handle `GRANT` option differently from other privileges (e.g., `UPDATE`). I checked the PostgreSQL source code, and it doesn't handle GRANT OPTION as well as other privileges https://github.com/postgres/postgres/blob/5ad49322e50459444b201e624d5df706751c9d41/src/include/nodes/parsenodes.h#L74-L93. So, IMO, maybe `Privilege.GRANT` should be removed from the enum and handled differently, although it might be a matter of taste.

brfrn169: As discussed offline, we've concluded the behavior mentioned in the following comment:
https://github.com/scalar-labs/scalardb/pull/1318#discussion_r1406941025

I've also revised the Javadoc and revered the method names in ad52e62256244a28fef27e3a21055be40bf7a05b.

@komamitsu Can you please take a quick look? Thanks.

komamitsu: Thanks. Looks good!

core/src/main/java/com/scalar/db/exception/transaction/AbortException.java
```diff
@@ -13,4 +15,17 @@ public AbortException(String message, String transactionId) {
...
+  public AbortException(
+      String message, @Nullable String transactionId, boolean authenticationError) {
+    super(message, transactionId, authenticationError, false, false, null);
+  }
+
+  public AbortException(
+      String message,
+      Throwable cause,
+      @Nullable String transactionId,
+      boolean authenticationError) {
```
feeblefakie, line 28: Instead of adding an argument only for authentication, it might be better to define an error code and add it here or something?
Since we don't define error code yet, I understand it's hard to do so now...

feeblefakie (commented):
I left one comment. PTAL!

core/src/main/java/com/scalar/db/exception/transaction/AbortException.java
brfrn169: Yeah, at the moment, it's hard to do so since we need to define error code for all errors for that. Anyway, I think we can add error codes later on.

feeblefakie (approved):
LGTM! Thank you!

PR #1395, opened 2023-12-20
Enable setting of transaction expiration handler in active transaction management

brfrn169:
## Description

This PR enables the setting of the transaction expiration handler in active transaction management.

## Related issues and/or PRs

N/A

## Changes made

- Enabled the setting of the transaction expiration handler in `ActiveTransactionManagedDistributedTransactionManager` and `ActiveTransactionManagedTwoPhaseCommitTransactionManager`

## Checklist

N/A

## Release notes

N/A


core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
```diff
@@ -40,14 +53,14 @@ public ActiveTransactionManagedDistributedTransactionManager(DatabaseConfig conf
...
-              try {
-                t.rollback();
-              } catch (Exception e) {
-                logger.warn("Rollback failed. transaction ID: {}", id, e);
-              }
+              transactionExpirationHandler.get().accept(id, t);
             });
   }
 
+  public void setTransactionExpirationHandler(BiConsumer<String, DistributedTransaction> handler) {
```
brfrn169, line 60: We will be able to set a transaction expiration handler after this change.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitTransactionManager.java
```diff
@@ -42,14 +55,15 @@ public ActiveTransactionManagedTwoPhaseCommitTransactionManager(DatabaseConfig c
...
-              try {
-                t.rollback();
-              } catch (Exception e) {
-                logger.warn("Rollback failed. transaction ID: {}", id, e);
-              }
+              transactionExpirationHandler.get().accept(id, t);
             });
   }
 
+  public void setTransactionExpirationHandler(
```
brfrn169, line 62: Ditto.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
```diff
@@ -32,6 +34,17 @@ public abstract class ActiveTransactionManagedDistributedTransactionManager
 
   private final ActiveExpiringMap<String, ActiveTransaction> activeTransactions;
 
+  private final AtomicReference<BiConsumer<String, DistributedTransaction>>
```
feeblefakie, line 37: Just a question.
I think we also have this expiration handler in ScalarDB Cluster, but I'm wondering if we need to have it in both sides: ScalarDB core and ScalarDB Cluster.
If the core implements it, the one in ScalarDB Cluster is not necessary?

feeblefakie (approved):
Although I left a question, LGTM! Thank you!

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
komamitsu: @brfrn169 Can you tell me how the custom handler would be like?

brfrn169: I'm going to use it for the ScalarDB Auth thing. Theoretically, we can do anything for expired transactions. For example, we can output detail logs for the expired transactions.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
brfrn169: Sorry, I can't remember this expiration handler in ScalarDB Cluster. Please give me some more information about that?

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
```diff
@@ -32,6 +34,17 @@ public abstract class ActiveTransactionManagedDistributedTransactionManager
...
+      transactionExpirationHandler =
+          new AtomicReference<>(
+              (id, t) -> {
+                try {
+                  t.rollback();
```
komamitsu, line 42: [minor] I think this `rollback` is always necessary and any custom expiration handler must call it. But users of `setTransactionExpirationHandler()` can pass a custom expiration handler that doesn't call `rollback` by mistake. Another option would be to add a callback like this?
```java
private final AtomicReference<BiConsumer<String, DistributedTransaction>> transactionExpirationCallback = new AtomicReference<>((id, t) -> {});
private final BiConsumer<String, DistributedTransaction>
      transactionExpirationHandler =
              (id, t) -> {
                try {
                   transactionExpirationCallback.get().accept(id, t);
                }
                catch (Throwable t) {
                  // Warning
                }
                try {
                  t.rollback();
                } catch (Exception e) {
                  logger.warn("Rollback failed. transaction ID: {}", id, e);
                }
              };
```

komamitsu (approved):
LGTM! 👍 

Left a minor comment, but it doesn't block this PR from being merged.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
brfrn169: Good point. I agree that `rollback` is always necessary, which is why the default handler executes `rollback`. This feature is advanced and is primarily intended for internal use. Additionally, I don't plan to document it. Therefore, if a user sets a custom handler that doesn't rollback the expired transaction, I believe it falls under the user's responsibility. Thanks.

PR #1409, opened 2023-12-25
Add a requirement for privileges of underlying databases

jnmt:
## Description

This PR adds a requirement for privileges of underlying databases.

## Related issues and/or PRs

- scalar-labs/scalardb-sql#392

## Changes made

- Update the requirement document regarding privileges of underlying databases

## Checklist

N/A

## Release notes

Added requirement for privileges of underlying databases.

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -4,6 +4,8 @@ This document explains the requirements and recommendations in the underlying da
 
 ## Common requirements
 
+This section describes common requirements for the underlying databases when using ScalarDB.
+
 ### Privileges to access the underlying databases
 
 ScalarDB operates the underlying databases not only for CRUD operations but also for performing operations like creating or altering schemas, tables, or indexes. Thus, ScalarDB requires a fully privileged account to access the underlying databases.
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -6,7 +6,7 @@ This document explains the requirements and recommendations in the underlying da
 
 ### Privileges to access the underlying databases
 
-ScalarDB operates the underlying databases not only for CRUD operations but also for creating/altering schema, tables, indexes, and so on. Thus, ScalarDB requires a fully privileged account to access the underlying databases.
+ScalarDB operates the underlying databases not only for CRUD operations but also for performing operations like creating or altering schemas, tables, or indexes. Thus, ScalarDB requires a fully privileged account to access the underlying databases.
 
 ## Cassandra or Cassandra-compatible database requirements
 
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -8,7 +8,7 @@ This section describes common requirements for the underlying databases when usi
 
 ### Privileges to access the underlying databases
 
-ScalarDB operates the underlying databases not only for CRUD operations but also for performing operations like creating or altering schemas, tables, or indexes. Thus, ScalarDB requires a fully privileged account to access the underlying databases.
+ScalarDB operates the underlying databases not only for CRUD operations but also for performing operations like creating or altering schemas, tables, or indexes. Thus, ScalarDB basically requires a fully privileged account to access the underlying databases.
 
 ## Cassandra or Cassandra-compatible database requirements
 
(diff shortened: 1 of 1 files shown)


docs/requirements.md
```diff
@@ -2,6 +2,12 @@
 
 This document explains the requirements and recommendations in the underlying databases of ScalarDB to make ScalarDB applications work correctly.
 
+## Common requirements
+
+### Privileges to access the underlying databases
+
+ScalarDB operates the underlying databases not only for CRUD operations but also for creating/altering schema, tables, indexes, and so on. Thus, ScalarDB requires a fully privileged account to access the underlying databases.
```
jnmt, line 9: It would be better to add example privilege settings (e.g., GRANT xxx) in each database for EoU. Also, I think we should add the related tests. I will work on it later.

We might need to have a further discussion about whether the full privilege requirement should be relaxed or not since I heard that a user does not want to give it to ScalarDB. One idea (from Takanori) is to note that "grant the full privilege if you use admin feature / API, otherwise grant x, y, z ...", or something like that. But for now, I think the full privilege is the simple and comprehensive solution (with also considering introducing the Auth feature soon).

kota2and3kan (approved):
LGTM! Thank you!

docs/requirements.md
josh-wong, line 9: I'd like to avoid `and so on` here since the phrase doesn't provide any value to the reader. How about we rephrase this part as suggested below?:

```suggestion
ScalarDB operates the underlying databases not only for CRUD operations but also for performing operations like creating or altering schemas, tables, or indexes. Thus, ScalarDB requires a fully privileged account to access the underlying databases.
```

docs/requirements.md
```diff
@@ -2,6 +2,12 @@
 
 This document explains the requirements and recommendations in the underlying databases of ScalarDB to make ScalarDB applications work correctly.
 
+## Common requirements
+
```
josh-wong, line 6: I'd like to add a sentence after this heading so that the user has some understanding for what this section is about.

The following sentence is pretty basic, but please feel free to add more context if the reader should know more🙇🏻‍♂️

```suggestion
## Common requirements

This section lists common requirements when using ScalarDB.

```

josh-wong (changes requested):
I've added a couple comments and suggestions. PTAL!

docs/requirements.md
jnmt: Thanks for the suggestion. Just a confirmation. Is it OK to say "list" even if we only have a single item (regarding the privileges)?

josh-wong: True. Let's use a more generic phrase, like `describes` (or `contains` or `includes`) instead.

```suggestion
## Common requirements

This section describes common requirements when using ScalarDB.

```

jnmt: Thank you! Fixed in 4412756.

docs/requirements.md
```diff
@@ -2,6 +2,14 @@
...
+This section describes common requirements for the underlying databases when using ScalarDB.
+
+### Privileges to access the underlying databases
+
+ScalarDB operates the underlying databases not only for CRUD operations but also for performing operations like creating or altering schemas, tables, or indexes. Thus, ScalarDB requires a fully privileged account to access the underlying databases.
```
komamitsu, line 11:
```suggestion
ScalarDB operates the underlying databases not only for CRUD operations but also for performing operations like creating or altering schemas, tables, or indexes. Thus, ScalarDB basically requires a fully privileged account to access the underlying databases.
```

I think a bit limited privileges are enough for ScalarDB, although it's not easy to describe the details in this section. Just adding `basically` or something might be an option?

jnmt: Yeah, that's better. Thank you!

komamitsu (approved):
LGTM! Thank you!

josh-wong (approved):
LGTM! Thank you!🙇‍♂️

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

josh-wong:
@jnmt Looks like this PR wasn't automatically backported to version branches. Is it OK if I backport this PR to branches 3.7 to 3.11, which are currently still under Maintenance Support?

PR #1493, opened 2024-02-06
Add error codes to error messages

brfrn169:
## Description

This PR adds error codes to the error messages.

Following this update, error messages will include the component name (`CORE` for this project), a category ID, an error ID, and the message. For instance, if the category ID is `1`, and the error ID is `0011`, the error format will be:

```
CORE-10011: <message>
```

The number following the component name `CORE` is a combination of the category ID and the error ID, where the first digit represents the category ID, and the subsequent digits denote the error ID.

The categories and their corresponding IDs are listed below:

| category name | ID | 
| --- | --- | 
| USER_ERROR | 1 |
| CONCURRENCY_ERROR | 2 |
| INTERNAL_ERROR | 3 |
| UNKNOWN_TRANSACTION_STATUS_ERROR | 4 |

## Related issues and/or PRs

N/A

## Changes made

- Added error codes to the error messages.

## Checklist

N/A

## Release notes

Added error codes to the error messages.


core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
brfrn169, line 1: For now, I've only added error codes to the error messages in `OperationChecker`. If there are no objections, I plan to apply the same changes to other classes.

core/src/main/java/com/scalar/db/common/error/ScalarDbError.java
```diff
@@ -0,0 +1,76 @@
+package com.scalar.db.common.error;
+
+import java.util.Objects;
+
+public interface ScalarDbError {
```
brfrn169, line 5: The error consists of a component name, category, sub-category, code, message, cause, and solution.

core/src/test/java/com/scalar/db/common/error/CoreErrorTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class CoreErrorTest {
+
+  @Test
+  public void checkDuplicateErrorCode() {
+    Assertions.assertThat(Arrays.stream(CoreError.values()).map(CoreError::buildCode))
+        .doesNotHaveDuplicates();
+  }
```
brfrn169, line 15: We can check error code duplication, here.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,215 @@
+package com.scalar.db.common.error;
+
+public enum CoreError implements ScalarDbError {
```
brfrn169, line 3: The errors in ScalarDB Core are defined here. We will add similar definitions for other projects (GraphQL, SQL, and Cluster).

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,215 @@
...
+      SubCategory.ILLEGAL_ARGUMENT,
+      "022",
+      "The clustering key is not properly specified. Operation: %s",
+      "",
+      ""),
+  ;
+
+  private static final String COMPONENT_NAME = "CORE";
+
+  private final Category category;
```
komamitsu, line 157: `SubCategory` contains `Category`, so I think this class doesn't need to have `Category` so that we can simplified `ScalarDbError.validate()`. Maybe only holding `SubCategory` sounds a bit odd. If so, renaming `SubCategory` to `Type` or something might be an option?

core/src/main/java/com/scalar/db/common/error/SubCategory.java
```diff
@@ -0,0 +1,45 @@
...
+  // Sub-categories for user errors
+  //
+  ILLEGAL_ARGUMENT(Category.USER_ERROR, "0"),
+  AUTHENTICATION_ERROR(Category.USER_ERROR, "1"),
+  AUTHORIZATION_ERROR(Category.USER_ERROR, "2"),
+
+  //
+  // Sub-categories for internal errors
+  //
+  UNKNOWN_TRANSACTION_STATUS(Category.INTERNAL_ERROR, "0"),
```
komamitsu, line 23: Probably users want to know whether to retry the failed transaction operations based on the category in error code. There is the separate `RETRYABLE_ERROR` category, so I guess users shouldn't retry for `INTERNAL_ERROR`? I know `UNKNOWN_TRANSACTION_STATUS` shouldn't be retried to avoid duplicated inserts or etc., but I'm wondering what kind of errors would be categorized in `UNKNOWN`.

(This isn't a suggestion from me and I'm just writing down my random thoughts)

Maybe connection failures to the database is categorized in `INTERNAL_ERROR` via `UNKNOWN`. In this case, I think retry would work with some interval and `RETRYABLE_ERROR` may be an option.

But if the connection error occurred since the database was busy, the retry interval should be longer than the retry for transaction conflicts. Considering this case, 2 retryable categories are needed for transaction conflicts and other tentative errors...? 🤔

core/src/main/java/com/scalar/db/common/error/ScalarDbError.java
```diff
@@ -0,0 +1,76 @@
...
+
+    Objects.requireNonNull(message, "The message must not be null.");
+    Objects.requireNonNull(cause, "The cause must not be null.");
+    Objects.requireNonNull(solution, "The solution must not be null.");
+  }
+
+  /**
+   * Builds the error code. The code is built as follows:
+   *
+   * <p>{@code <componentName>-<categoryCode><subCategoryCode><code>}
```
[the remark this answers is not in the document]
brfrn169: Good point. Indeed, that's confusing.

I think we can rename `getCode()` to `getNumber()` of `ScalarDBError`, `Category`, and `SubCategory`. What do you think?

core/src/main/java/com/scalar/db/common/error/CoreError.java
brfrn169: Yes, `SubCategory` includes `Category` as its `parent` to represent the parent-child relationship between `Category` and `SubCategory`. And, in the current implementation, the `parent` is used only for validation purposes.

I think it's natural that the error has a category as well as a sub-category, but the challenge is how to effectively model this relationship in Java.

> Maybe only holding SubCategory sounds a bit odd. If so, renaming SubCategory to Type or something might be an option?

I think only holding SubCategory sounds a bit odd, too. Instead of renaming `SubCategory` to `Type`, maybe we could perhaps a new enum holding both `Category` and `SubCategory` to represent the parent-child relationship. And in this case, we can remove `parent` from `SubCategory`. But I feel that it might be too much, so I'm not sure which is better the original one or this one.

core/src/main/java/com/scalar/db/common/error/SubCategory.java
brfrn169: Firstly, the purpose of this PR is just to add error codes to error messages, so I think we can address error handling as a separate concern. However, it's true that these aspects can't be entirely isolated actually.

I think that when we encounter `RETRYABLE_ERROR`, we can retry the transaction for sure. However, when we face `INTERNAL_ERROR` via `UNKNOWN`, it's uncertain whether we can retry the transaction. In some cases, we may be able to retry the transaction, but in others, we may not. From this perspective, categorizing such errors as `UNKNOWN` seems reasonable.

brfrn169: As discussed, I will rename `RETRYABLE_ERROR` to `TRANSACTION_RETRYABLE_ERROR` to imply that users could retry even for `INTERNAL_ERROR`.

core/src/main/java/com/scalar/db/common/error/ScalarDbError.java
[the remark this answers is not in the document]
brfrn169: Fixed in 60fb835a531e657dc7f1e8485aa2a03650819056. Thanks.

core/src/main/java/com/scalar/db/common/error/SubCategory.java
brfrn169: Fixed in 60fb835a531e657dc7f1e8485aa2a03650819056. Thanks.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,165 @@
...
+      "",
+      ""),
+  OPERATION_CHECK_ERROR_CONDITION(
+      Category.USER_ERROR,
+      "0015",
+      "The condition is not properly specified. Operation: %s",
+      "",
+      ""),
+  OPERATION_CHECK_ERROR_TABLE_NOT_FOUND(
+      Category.USER_ERROR, "0016", "The specified table is not found: %s", "", ""),
```
komamitsu, line 89:
```suggestion
      Category.USER_ERROR, "0016", "The specified table is not found. Table: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
```diff
@@ -303,10 +307,7 @@ private void checkColumnsInPut(Put put, TableMetadata metadata) {
     for (Column<?> column : put.getColumns().values()) {
       if (!new ColumnChecker(metadata, false, false, false, true).check(column)) {
         throw new IllegalArgumentException(
-            "The column value is not properly specified. Invalid column: "
-                + column
-                + ", Operation: "
-                + put);
+            CoreError.OPERATION_CHECK_ERROR_TABLE_NOT_FOUND.buildMessage(column, put));
```
komamitsu, line 310:
```suggestion
            CoreError.OPERATION_CHECK_ERROR_INVALID_COLUMN.buildMessage(column, put));
```

komamitsu: I think the number of arguments was wrong, but the CI didn't fail...?

brfrn169: Oops... Good catch, thank you.

> I think the number of arguments was wrong, but the CI didn't fail...?

Ah, it seems like the error occurs only when the number of arguments is less than the number of the parameters. Umm...

core/src/main/java/com/scalar/db/common/error/CoreError.java
brfrn169, line 1: @josh-wong Could you please check this file when you have time? 🙇

brfrn169:
@feeblefakie @komamitsu I've added the error codes to the error messages in the core module. Please take a look when you have time.

@kota2and3kan I don't think you need to check this PR thoroughly, but could you please skim the error messages when you have time.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -23,7 +23,7 @@ public interface RdbEngineStrategy {
 
   boolean isUndefinedTableError(SQLException e);
 
-  boolean isConflictError(SQLException e);
+  boolean isConflict(SQLException e);
```
brfrn169, line 26: Renamed this method.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -123,20 +129,26 @@ public void commitState(Snapshot snapshot)
           TransactionState state = s.get().getState();
           if (state.equals(TransactionState.ABORTED)) {
             rollbackRecords(snapshot);
-            throw new CommitException(
-                "Committing state in coordinator failed. the transaction is aborted", e, id);
+            throw new CommitConflictException(
```
brfrn169, line 132: I think we should throw `CommitConflictException` instead of `CommitException`.

core/src/main/java/com/scalar/db/common/error/Category.java
```diff
@@ -0,0 +1,24 @@
+package com.scalar.db.common.error;
+
+import java.util.Objects;
+
+public enum Category {
```
feeblefakie, line 5: I think we should say something about retryability in the messages, the code, or the comments.
What do you think?

brfrn169: I think we can add the retryability thing to the `solution` field later. What do you think?

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  GET_BUILD_ERROR_OPERATION_NOT_SUPPORTED_WHEN_GETTING_RECORDS_OF_DATABASE_WITHOUT_USING_INDEX(
+      Category.USER_ERROR,
+      "0031",
+      "This operation is not supported when getting records of a database without using an index",
+      "",
+      ""),
+  GET_BUILD_ERROR_OPERATION_NOT_SUPPORTED_WHEN_GETTING_RECORDS_OF_DATABASE_USING_INDEX(
+      Category.USER_ERROR,
+      "0032",
+      "This operation is not supported when getting records of a database using an index",
```
josh-wong, line 156:
```suggestion
      "This operation is not supported when getting records of a database by using an index",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  SCAN_BUILD_ERROR_OPERATION_NOT_SUPPORTED_WHEN_SCANNING_ALL_RECORDS_OF_DATABASE_OR_SCANNING_RECORDS_OF_DATABASE_USING_INDEX(
+      Category.USER_ERROR,
+      "0033",
+      "This operation is not supported when scanning all the records of a database "
+          + "or scanning records of a database using an index",
```
josh-wong, line 163:
```suggestion
          + "or scanning records of a database by using an index",
```

core/src/main/java/com/scalar/db/common/error/Category.java
feeblefakie: Ah, right. That should be OK.
I'm also wondering if we could have a reusable message (e.g., constant variable) for retryability because, for example, `CONCURRENCY_ERROR` is always retryable, so we should add the same description for retryability to all the concurrency errors' messages.
Maybe, we can think about it later.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  SCAN_BUILD_ERROR_OPERATION_SUPPORTED_ONLY_WHEN_SCANNING_RECORDS_OF_DATABASE_USING_INDEX(
+      Category.USER_ERROR,
+      "0034",
+      "This operation is supported only when scanning records of a database using an index",
```
josh-wong, line 169:
```suggestion
      "This operation is supported only when scanning records of a database by using an index",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  SCAN_BUILD_ERROR_OPERATION_NOT_SUPPORTED_WHEN_SCANNING_RECORDS_OF_DATABASE_USING_INDEX(
+      Category.USER_ERROR,
+      "0035",
+      "This operation is not supported when scanning records of a database using an index",
```
josh-wong, line 175:
```suggestion
      "This operation is not supported when scanning records of a database by using an index",
```

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      Category.USER_ERROR, "0050", "%s is the system namespace name", "", ""),
+  NAMESPACE_ALREADY_EXISTS(
+      Category.USER_ERROR, "0051", "Namespace already exists. Namespace: %s", "", ""),
+  NAMESPACE_NOT_FOUND(
+      Category.USER_ERROR, "0052", "Namespace does not exist. Namespace: %s", "", ""),
+  TABLE_ALREADY_EXISTS(Category.USER_ERROR, "0053", "Table already exists. Table: %s", "", ""),
+  NAMESPACE_NOT_EMPTY(
+      Category.USER_ERROR,
+      "0055",
+      "Namespace is not empty. Namespace: %s, Tables in the namespace: %s",
```
josh-wong, line 249:
```suggestion
      "The namespace is not empty. Namespace: %s; Tables in the namespace: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  COLUMN_NOT_FOUND2(
+      Category.USER_ERROR, "0056", "Column does not exist. Table: %s, Column: %s", "", ""),
```
josh-wong, line 253:
```suggestion
      Category.USER_ERROR, "0056", "The column does not exist. Table: %s; Column: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  INDEX_ALREADY_EXISTS(
+      Category.USER_ERROR, "0057", "Index already exists. Table: %s, Column: %s", "", ""),
```
josh-wong, line 255:
```suggestion
      Category.USER_ERROR, "0057", "The index already exists. Table: %s; Column: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  INDEX_NOT_FOUND(
+      Category.USER_ERROR, "0058", "Index does not exist. Table: %s, Column: %s", "", ""),
```
josh-wong, line 257:
```suggestion
      Category.USER_ERROR, "0058", "The index does not exist. Table: %s; Column: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  COLUMN_ALREADY_EXISTS(
+      Category.USER_ERROR, "0059", "Column already exists. Table: %s, Column: %s", "", ""),
```
josh-wong, line 259:
```suggestion
      Category.USER_ERROR, "0059", "The column already exists. Table: %s; Column: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  REPAIRING_NAMESPACE_FAILED(
+      Category.INTERNAL_ERROR, "0012", "Repairing the namespace failed. Namespace: %s", "", ""),
+  REPAIRING_TABLE_FAILED(
+      Category.INTERNAL_ERROR, "0013", "Repairing the table failed. Table: %s", "", ""),
+  ADDING_NEW_COLUMN_TO_TABLE_FAILED(
+      Category.INTERNAL_ERROR,
+      "0014",
+      "Adding a new column to the table failed. Table: %s, Column: %s, ColumnType: %s",
```
josh-wong, line 619:
```suggestion
      "Adding a new column to the table failed. Table: %s; Column: %s; ColumnType: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "0009",
+      "Checking the namespace existence failed. Namespace: %s",
+      "",
+      ""),
+  CHECKING_TABLE_EXISTENCE_FAILED(
+      Category.INTERNAL_ERROR, "0010", "Checking the table existence failed. Table: %s", "", ""),
+  CHECKING_INDEX_EXISTENCE_FAILED(
+      Category.INTERNAL_ERROR,
+      "0011",
+      "Checking the index existence failed. Table: %s, Column: %s",
```
josh-wong, line 609:
```suggestion
      "Checking the index existence failed. Table: %s; Column: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "0016",
+      "Getting the table metadata of the importing table failed. Table: %s",
+      "",
+      ""),
+  IMPORTING_TABLE_FAILED(
+      Category.INTERNAL_ERROR, "0017", "Importing the table failed. Table: %s", "", ""),
+  ADDING_RAW_COLUMN_TO_TABLE_FAILED(
+      Category.INTERNAL_ERROR,
+      "0018",
+      "Adding the raw column to the table failed. Table: %s, Column: %s, ColumnType: %s",
```
josh-wong, line 635:
```suggestion
      "Adding the raw column to the table failed. Table: %s; Column: %s; ColumnType: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  UPGRADING_SCALAR_DB_ENV_FAILED(
+      Category.INTERNAL_ERROR, "0019", "Upgrading the ScalarDB environment failed", "", ""),
+  CASSANDRA_WRITE_TIMEOUT_WITH_OTHER_WRITE_TYPE_IN_MUTATION(
+      Category.INTERNAL_ERROR,
+      "0020",
+      "Something wrong because it is neither CAS nor SIMPLE",
+      "",
+      ""),
+  CASSANDRA_ERROR_OCCURRED_IN_SELECTION(
+      Category.INTERNAL_ERROR, "0021", "An error occurred in the selection. Detailed: %s", "", ""),
```
josh-wong, line 647:
```suggestion
      Category.INTERNAL_ERROR, "0021", "An error occurred in the selection. Details: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  COSMOS_ERROR_OCCURRED_IN_MUTATION(
+      Category.INTERNAL_ERROR, "0022", "An error occurred in the mutation. Detailed: %s", "", ""),
+  COSMOS_ERROR_OCCURRED_IN_SELECTION(
+      Category.INTERNAL_ERROR, "0023", "An error occurred in the selection. Detailed: %s", "", ""),
```
josh-wong, line 651:
```suggestion
      Category.INTERNAL_ERROR, "0023", "An error occurred in the selection. Details: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  DYNAMO_ERROR_OCCURRED_IN_MUTATION(
+      Category.INTERNAL_ERROR, "0024", "An error occurred in the mutation. Detailed: %s", "", ""),
```
josh-wong, line 653:
```suggestion
      Category.INTERNAL_ERROR, "0024", "An error occurred in the mutation. Details: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  DYNAMO_ERROR_OCCURRED_IN_SELECTION(
+      Category.INTERNAL_ERROR, "0025", "An error occurred in the selection. Detailed: %s", "", ""),
+  JDBC_ERROR_OCCURRED_IN_MUTATION(
+      Category.INTERNAL_ERROR, "0026", "An error occurred in the mutation. Detailed: %s", "", ""),
```
josh-wong, line 657:
```suggestion
      Category.INTERNAL_ERROR, "0026", "An error occurred in the mutation. Details: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  JDBC_ERROR_OCCURRED_IN_SELECTION(
+      Category.INTERNAL_ERROR, "0027", "An error occurred in the selection. Detailed: %s", "", ""),
```
josh-wong, line 659:
```suggestion
      Category.INTERNAL_ERROR, "0027", "An error occurred in the selection. Details: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  //
+  // Errors for the concurrency error category
+  //
+  NO_MUTATION_APPLIED(Category.CONCURRENCY_ERROR, "0000", "No mutation was applied", "", ""),
+  CASSANDRA_LOGGING_FAILED_IN_BATCH(
+      Category.CONCURRENCY_ERROR, "0001", "Logging failed in the batch", "", ""),
+  CASSANDRA_OPERATION_FAILED_IN_BATCH(
+      Category.CONCURRENCY_ERROR, "0002", "Operation failed in the batch with type %s", "", ""),
+  CASSANDRA_ERROR_OCCURRED_IN_BATCH(
+      Category.CONCURRENCY_ERROR, "0003", "An error occurred in the batch. Detailed: %s", "", ""),
```
josh-wong, line 489:
```suggestion
      Category.CONCURRENCY_ERROR, "0003", "An error occurred in the batch. Details: %s", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+  CASSANDRA_ERROR_OCCURRED_IN_MUTATION(
+      Category.CONCURRENCY_ERROR,
+      "0007",
+      "An error occurred in the mutation. Detailed: %s",
+      "",
+      ""),
+  COSMOS_RETRY_WITH_ERROR_OCCURRED_IN_MUTATION(
+      Category.CONCURRENCY_ERROR,
+      "0008",
+      "A RetryWith error occurred in the mutation. Detailed: %s",
```
josh-wong, line 505:
```suggestion
      "A RetryWith error occurred in the mutation. Details: %s",
```

josh-wong (changes requested):

core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
```diff
@@ -46,27 +47,33 @@ public ResultSet handle(Operation operation) throws ExecutionException {
...
       }
       return results;
-
     } catch (WriteTimeoutException e) {
       logger.warn("Write timeout happened during mutate operation", e);
       if (e.getWriteType() == WriteType.CAS) {
         // retry needs to be done if applications need to do the operation exactly
-        throw new RetriableExecutionException("Paxos phase in CAS operation failed", e);
+        throw new RetriableExecutionException(
+            CoreError.CASSANDRA_WRITE_TIMEOUT_IN_PAXOS_PHASE_IN_MUTATION.buildMessage(), e);
```
komamitsu, line 55: This error is in CONCURRENCY_ERROR category. But write timeout sounds to me it occurred since the Cassandra cluster was potentially busy not from internal transactional conflicts. So INTERNAL_ERROR might be safer. What do you think?

kota2and3kan (approved):
Thank you for implementing the error code! It's a great feature! We can increase the supportability so much!
Mainly, I checked `core/src/main/java/com/scalar/db/common/error/CoreError.java`.
It looks good to me! Thank you!

FYI: I will share these error codes with the support team. If the support team has some feedback, I will share it with you later.

core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionManager.java
```diff
@@ -181,7 +180,8 @@ public void rollback() throws RollbackException {
     @Override
     public void abort() throws AbortException {
       if (status == Status.COMMITTED || status == Status.ROLLED_BACK) {
-        throw new IllegalStateException("The transaction has already been committed or aborted");
+        throw new IllegalStateException(
+            CoreError.TRANSACTION_ALREADY_COMMITTED_OR_ABORTED.buildMessage(status));
```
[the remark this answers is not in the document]
brfrn169: Yes, I agree. Let's remove `TRANSACTION_ALREADY_COMMITTED_OR_ABORTED` and use `TRANSACTION_ALREADY_COMMITTED_OR_ROLLED_BACK`. Thank you.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  OPERATION_CHECK_ERROR_INDEX_ORDERING_SPECIFIED(
+      Category.USER_ERROR,
+      "0004",
+      "Orderings cannot be specified when using an index. Operation: %s",
+      "",
+      ""),
+  OPERATION_CHECK_ERROR_LIMIT(
+      Category.USER_ERROR, "0005", "The limit cannot be negative. Operation: %s", "", ""),
```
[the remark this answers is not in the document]
brfrn169: We follow the Google Java Style Guide, which alternates between single-line and multi-line formats based on the line length.

I'm sure how to control this formatting behavior.

@feeblefakie @komamitsu, are you familiar with this?

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  OPERATION_CHECK_ERROR_CLUSTERING_KEY(
+      Category.USER_ERROR,
+      "0021",
+      "The clustering key is not properly specified. Operation: %s",
+      "",
+      ""),
+  NOT_SUPPORTED_IN_COMMUNITY_EDITION(
+      Category.USER_ERROR, "0022", "Not supported in the community edition", "", ""),
```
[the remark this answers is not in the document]
brfrn169: `This feature is not support in the ScalarDB Community edition` sounds better to me. Thank you.

```suggestion
      Category.USER_ERROR, "0022", "This feature is not support in the ScalarDB Community edition", "", ""),
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  TABLE_METADATA_BUILD_ERROR_CLUSTERING_KEY_COLUMN_DEFINITION_NOT_SPECIFIED(
+      Category.USER_ERROR,
+      "0041",
+      "Need to specify the column definition of %s specified as a clustering key",
+      "",
+      ""),
+  TRANSACTION_STATE_INSTANTIATION_ERROR_INVALID_ID(
+      Category.USER_ERROR, "0042", "Invalid id specified. Id: %d", "", ""),
```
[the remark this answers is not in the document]
brfrn169: Yes, should should use `ID`. Thank you.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  INVALID_CONSISTENCY_LEVEL(
+      Category.USER_ERROR,
+      "0079",
+      "The specified consistency level is not supported. Consistency level: %s",
+      "",
+      ""),
+  DYNAMO_ENCODER_0X00_BYTES_NOT_ACCEPTED_IN_BLOB_VALUES_IN_DESC_ORDER(
+      Category.USER_ERROR, "0080", "0x00 bytes not accepted in blob values in DESC order", "", ""),
```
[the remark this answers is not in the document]
brfrn169: Yes, it should be `BLOB`.

brfrn169, line 357:
```suggestion
      Category.USER_ERROR, "0080", "0x00 bytes are not accepted in BLOB values in DESC order", "", ""),
```

core/src/main/java/com/scalar/db/storage/jdbc/RdbEnginePostgresql.java
```diff
@@ -215,15 +215,15 @@ public DataType getDataTypeForScalarDb(
...
+              CoreError.JDBC_IMPORT_DATA_TYPE_WITH_SIZE_NOT_SUPPORTED.buildMessage(
+                  typeName, columnSize, columnDescription));
         }
         return DataType.BOOLEAN;
       case SMALLINT:
         if (typeName.equalsIgnoreCase("smallserial")) {
           throw new IllegalArgumentException(
-              String.format(
-                  "Data type %s(%d) is unsupported: %s", typeName, columnSize, columnDescription));
+              CoreError.JDBC_IMPORT_DATA_TYPE_WITH_SIZE_NOT_SUPPORTED.buildMessage(
```
komamitsu, line 225: These cases don't seem to be related to the column size. So, JDBC_IMPORT_DATA_TYPE_NOT_SUPPORTED is better?

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "0014",
+      "Adding a new column to the table failed. Table: %s, Column: %s, ColumnType: %s",
+      "",
+      ""),
+  GETTING_NAMESPACE_NAMES_FAILED(
+      Category.INTERNAL_ERROR, "0015", "Getting the namespace names failed", "", ""),
+  GETTING_IMPORT_TABLE_METADATA_FAILED(
+      Category.INTERNAL_ERROR,
+      "0016",
+      "Getting the table metadata of the importing table failed. Table: %s",
```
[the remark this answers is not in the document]
brfrn169: I think the latter one is appropriate.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -0,0 +1,767 @@
...
+      "",
+      ""),
+  OPERATION_CHECK_ERROR_CLUSTERING_KEY(
+      Category.USER_ERROR,
+      "0021",
+      "The clustering key is not properly specified. Operation: %s",
+      "",
+      ""),
+  NOT_SUPPORTED_IN_COMMUNITY_EDITION(
+      Category.USER_ERROR, "0022", "This feature is not support in the ScalarDB Community edition", "", ""),
```
josh-wong, line 119: Sorry, I had a misspelling in my previous comment about this message.

```suggestion
      Category.USER_ERROR, "0022", "This feature is not supported in the ScalarDB Community edition", "", ""),
```

core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
brfrn169: We originally throw `RetriableExecutionException` that indicates an retryable error for this case. That's why I categorize it as `CONCURRENCY_ERROR`. What do you think?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEnginePostgresql.java
brfrn169: The `%d` part of the message is the column size.

And, actually, `JDBC_IMPORT_DATA_TYPE_NOT_SUPPORTED` already exists.
https://github.com/scalar-labs/scalardb/pull/1493/files#diff-fd06988f4ced6a0d6d0b1049c69ab07df11c372c3ccb868a78afbe4d24e6f663R420-R421

core/src/main/java/com/scalar/db/storage/cosmos/MutateStatementHandler.java
```diff
@@ -61,11 +64,16 @@ private void throwException(CosmosException exception) throws ExecutionException
     int statusCode = exception.getSubStatusCode();
 
     if (statusCode == CosmosErrorCode.PRECONDITION_FAILED.get()) {
-      throw new NoMutationException("No mutation was applied");
+      throw new NoMutationException(CoreError.NO_MUTATION_APPLIED.buildMessage());
```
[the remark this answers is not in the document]
brfrn169: I pass the exception as the 2nd argument of the `NoMutationException` constructor in 77fd1e633d130596a64fa36a9339b6154195de40. Thanks.

core/src/main/java/com/scalar/db/common/error/Category.java
brfrn169: @feeblefakie 

> I'm also wondering if we could have a reusable message (e.g., constant variable) for retryability because, for example, CONCURRENCY_ERROR is always retryable, so we should add the same description for retryability to all the concurrency errors' messages.
> Maybe, we can think about it later.

Yeah, let's do that as the second step. In any case, we should think about adding the `cause` and `solution` field as the next step. I think we can do that with it. Thank you.

core/src/main/java/com/scalar/db/common/error/CoreError.java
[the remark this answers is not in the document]
josh-wong: Ah, I see. I found the [Column limit](https://google.github.io/styleguide/javaguide.html#s4.4-column-limit) section in the Google Java Style Guide that describes the 100-character limit for Java code. I'll note this for future reference.

brfrn169:
@josh-wong I've applied your suggestion. Could you please double-check when you have time?

josh-wong (approved):
LGTM! Thank you very much for creating these error codes and messages!🙇‍♂️

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionTableMetadataManager.java
```diff
@@ -51,7 +53,8 @@ public Optional<TransactionTableMetadata> load(@Nonnull TableKey key)
   public TransactionTableMetadata getTransactionTableMetadata(Operation operation)
       throws ExecutionException {
     if (!operation.forNamespace().isPresent() || !operation.forTable().isPresent()) {
-      throw new IllegalArgumentException("Operation has no target namespace and table name");
+      throw new IllegalArgumentException(
+          CoreError.OPERATION_DOES_NOT_HAVE_TARGET_NAMESPACE_OR_TABLE_NAME.buildMessage());
```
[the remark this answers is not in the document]
brfrn169: Ah, good catch! Thank you.

Fixed in e97395b860005e05aed01d413edcd44521b5fca0.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
```diff
@@ -57,7 +58,8 @@ private void check(Put put) throws ExecutionException {
     for (String column : put.getContainedColumnNames()) {
       if (metadata.getTransactionMetaColumnNames().contains(column)) {
         throw new IllegalArgumentException(
-            "Mutating transaction metadata columns is not allowed. column: " + column);
+            CoreError.CONSENSUS_COMMIT_MUTATING_TRANSACTION_METADATA_COLUMNS_NOT_ALLOWED
```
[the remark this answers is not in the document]
brfrn169: Fixed in e97395b860005e05aed01d413edcd44521b5fca0. Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEnginePostgresql.java
komamitsu: Yeah, my suggestion is to use JDBC_IMPORT_DATA_TYPE_NOT_SUPPORTED since the column size isn't used for the check like some others.

brfrn169: Ah okay, so you mean that we should change the original error message, right? Yeah, it looks like the original one is wrong.

brfrn169: Fixed in 2a51d6a639c5693dafc77b9b47bf8dcca8c3d784. I made the same change for the `serial` type. Please take a look again!

komamitsu:
@brfrn169 I looked over exceptions thrown in `core` and noticed some exceptions seemed to be thrown with no error code. I listed them as follows. I think I may have wrongly picked some exceptions that don't need to have error codes, but can you take a look at them just in case?

- https://github.com/scalar-labs/scalardb/blob/cddb35cecbafbf4ac22beb9a211f1b3170b66311/core/src/main/java/com/scalar/db/common/ScannerIterator.java#L37
- https://github.com/scalar-labs/scalardb/blob/dc0da88d14c6805f8fe8982caaed586688d3c942/core/src/main/java/com/scalar/db/service/StorageModule.java#L62
  - USER_ERROR?
- https://github.com/scalar-labs/scalardb/blob/dc0da88d14c6805f8fe8982caaed586688d3c942/core/src/main/java/com/scalar/db/service/TransactionModule.java#L73
  - USER_ERROR?
- https://github.com/scalar-labs/scalardb/blob/dc0da88d14c6805f8fe8982caaed586688d3c942/core/src/main/java/com/scalar/db/service/TransactionModule.java#L99
  - USER_ERROR?
- https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java#L544 and some others
  - INTERNAL_ERROR?
  - Other XxxxAdmin have similar ones
- https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/dynamo/BatchHandler.java#L70
  - USER_ERROR?
- https://github.com/scalar-labs/scalardb/blob/5b80478e037278070806a886867ca99f10ff4710/core/src/main/java/com/scalar/db/storage/dynamo/Namespace.java#L10-L15
  - USER_ERROR? or won't happen?
- https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/dynamo/DynamoConfig.java#L38
  - USER_ERROR? or AssertionError?
- https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java#L332
  - INTERNAL_ERROR or USER_ERROR?
- https://github.com/scalar-labs/scalardb/blob/5b80478e037278070806a886867ca99f10ff4710/core/src/main/java/com/scalar/db/storage/rpc/GrpcConfig.java#L33
  - USER_ERROR? or won't happen?
- https://github.com/scalar-labs/scalardb/blob/5b80478e037278070806a886867ca99f10ff4710/core/src/main/java/com/scalar/db/storage/rpc/GrpcConfig.java#L41
  - INVALID_CONTACT_PORT can be used?
- https://github.com/scalar-labs/scalardb/blob/08814500268ddd95274260e62f844153f1ff2315/core/src/main/java/com/scalar/db/util/ProtoUtils.java#L202
  - USER_ERROR?

core/src/main/java/com/scalar/db/transaction/consensuscommit/TransactionTableMetadataManager.java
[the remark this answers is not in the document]
komamitsu: Thanks! (I guess the fix was in https://github.com/scalar-labs/scalardb/pull/1493/commits/b90ae907d82e5faee0341be339a9c8534f8606ff?)

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitMutationOperationChecker.java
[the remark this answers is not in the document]
komamitsu: Thank you! (probably https://github.com/scalar-labs/scalardb/commit/e97395b860005e05aed01d413edcd44521b5fca0, I guess)

brfrn169:
@komamitsu Thank you for your thoroughly checking!

- https://github.com/scalar-labs/scalardb/blob/cddb35cecbafbf4ac22beb9a211f1b3170b66311/core/src/main/java/com/scalar/db/common/ScannerIterator.java#L37
  - I thought `NoSuchElementException` was the Java's Iterator specification, so I didn't add error code for this. What do you think?
- https://github.com/scalar-labs/scalardb/blob/dc0da88d14c6805f8fe8982caaed586688d3c942/core/src/main/java/com/scalar/db/service/StorageModule.java#L62
  - `StorageModule` is deprecated. I don't think we need to add error codes for this class.
- https://github.com/scalar-labs/scalardb/blob/dc0da88d14c6805f8fe8982caaed586688d3c942/core/src/main/java/com/scalar/db/service/TransactionModule.java#L73
  - `TransactionModule` is deprecated. I don't think we need to add error codes for this class.
- https://github.com/scalar-labs/scalardb/blob/dc0da88d14c6805f8fe8982caaed586688d3c942/core/src/main/java/com/scalar/db/service/TransactionModule.java#L99
  - Ditto. `TransactionModule` is deprecated.
- https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java#L544 and some others
  - `XxxxAdmin` is wrapped by `CheckedDistributedStorageAdmin`. And we have added error codes there for the operations of `XxxxAdmin`
- https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/dynamo/BatchHandler.java#L70
  - I missed adding an error code for this. I'll fix this. Thank you.
- https://github.com/scalar-labs/scalardb/blob/5b80478e037278070806a886867ca99f10ff4710/core/src/main/java/com/scalar/db/storage/dynamo/Namespace.java#L10-L15
  - These should not be user-facing error. They should be thrown when we (developers) make a bug or something. That's why I didn't add an error code for it.
- https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/dynamo/DynamoConfig.java#L38
  - This should not be user-facing error.
- https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java#L332
  - I think this won't occur actually. Not sure why `SQLException` is thrown here...
- https://github.com/scalar-labs/scalardb/blob/5b80478e037278070806a886867ca99f10ff4710/core/src/main/java/com/scalar/db/storage/rpc/GrpcConfig.java#L33
  - This should not be user-facing error.
- https://github.com/scalar-labs/scalardb/blob/5b80478e037278070806a886867ca99f10ff4710/core/src/main/java/com/scalar/db/storage/rpc/GrpcConfig.java#L41
  - This is ScalarDB Server thing, which is deprecated.
- https://github.com/scalar-labs/scalardb/blob/08814500268ddd95274260e62f844153f1ff2315/core/src/main/java/com/scalar/db/util/ProtoUtils.java#L202
  - Ditto. This is ScalarDB Server thing.

core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
komamitsu: Thanks for sharing the background. It makes sense.

But WriteTimeoutException seemed to be likely thrown under heavy workload when googling it. So, my concern is still as follows:
1. Cassandra cluster is busy and a write operation is timed out
2. The client throws WriteTimeoutException to ScalarDB
3. ScalarDB wraps it with CONCURRENCY_ERROR
4. The application handles the wrapped exception and retries with a very small wait since it's CONCURRENCY_ERROR
5. The Cassandra cluster gets more busy...

IMO, INTERNAL_ERROR still fits this situation. But I understand the background, so this concern won't prevent this PR from being merged.

brfrn169:
@komamitsu I've fixed the following thing in 117e008c2dacde28bd1ce7d1899427da538f0c20.

https://github.com/scalar-labs/scalardb/blob/19491d2cbde83eae3a5f562c163278ebfdefc4f2/core/src/main/java/com/scalar/db/storage/dynamo/BatchHandler.java#L70

Please take a look again when you have time! Thanks.

komamitsu (approved):
LGTM, thank you! Great job! 👍 

I think there are still a few minor open topics that don't block this PR, but LGTM other than that.

core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
brfrn169: Okay. At this time, this PR doesn't change the original behavior, we can continue this discussion in a separate PR or issue if needed. I'll merge this for now. Thank you for raising this discussion!

brfrn169:
> I see. I think it's okay with the current code as-is unless users can write the following code.
>
>    ScannerIterator scannerIterator = new ScannerIterator(scan);
>    while (true) {
>      System.out.println(scannerIterator.next());
>    }
> In this case, an error code with USER_ERROR category might be helpful, although such a code shouldn't be written.

That's right. As we can add also the message later, let's add it when we're sure it's truly necessary. Thanks.

PR #1607, opened 2024-03-14
Add `util/groupcommit` package in `core` subproject

komamitsu:
## Description

ScalarDB may sometimes need to work with a high latency storage (e.g., multi-region replicated database). In such a case, the throughput of application using ScalarDB would be affected by the high latency transactions. With a group commit mechanism, high latency issue can be hidden with concurrent transactions.

In this PR, we introduce `util/groupcommit` package in `core` subproject which provides a mechanism to support the ScalarDB protocol. The package alone doesn't affect the existing implementations. A following PR will integrate the new package and the existing implementations.

## Related issues and/or PRs

N/A

## Changes made

This PR basically adds the following classes.
image

The `util/groupcommit` package has a concept about keys and they should be described before looking into the classes.

- Keys
  - Child Key : This is passed by a client (e.g., UUID v4 like ScalarDB transaction ID)
  - Parent Key : This is generated by KeyManipulator class which is injected by a client and contains logics to handle Keys
  - Full Key : This consists of Child Key and Parent Key

The main components introduced by this PR are as follows:
- Group
  - A unit of (group) commit
  - Contains slots (see below about slots). The number of slots in a group is decided by the configuration
  - Possible status are as follows:
    - OPEN : Initial status. Accepting new slot reservation as the number of slots isn't fixed yet
    - CLOSED : Not accepting new slot reservations since the number of slots is already fixed. Waiting all the slots are set with values
    - READY : All the slots are set with values. Ready to commit
    - DONE : Group commit is done and all the clients have get the results
  - There are 2 types of Groups
    - NormalGroup
      - Contains multiple slots and commits them at once. Ideally only this Group should be used in terms of performance 
      -  Parent Key is used to identify NormalGroup
    - DelayedGroup
      - Contains only a single slot which is delayed and removed from the NormalGroup
      - Full Key is used to identify DelayedGroup
- Slot
  - A container for a value. Values are supposed to be emitted at once by this group commit mechanism
  - Each Slot has the corresponding Child Key
- GroupManager
  - Manages all the ongoing Groups
  - It internally contains 2 maps
    - NormalGroupMap : Key=Parent-Key, Value=NormalGroup
    - DelayedGroupMap : Key=Full-Key, Value=DelayedGroup
- Worker threads
  - Group-Close Thread : Fixes the number of slot (or close the group) when the Group is timed out
  - Delayed-Slot-Move Thread : Moves delayed Slots from NormalGroupMap to DelayedGroupMap
  - Group-Cleanup Thread : Removes Group entries that are already done from NormalGroupMap or DelayedGroupMap. It's a sort of garbage collector
- GroupCommitter
  - Manages all the above classes
  - Has the following public APIs
    - reserve(Child Key) returns Full Key
      - Allocates a Slot in the current group that is associated with a Full Key
      - The allocated slot status is initially NOT_READY
    - ready(Full Key)
      - Marks the slot associated with the Full Key READY
      - Waits until all the slots in the group are READY
      - Internally calls the `Emittable.emit()` which is injected by the client with either of the following keys once the group gets READY 
        - Parent Key : In most cases, this type of Key is used to emit multiple values in NormalGroup
        - Full Key : This type of  Key is used only when it takes long time to call this API and the delayed slot is separated from the original NormalGroup to a new DelayedGroup to prevent other slots in the Group from being delayed

**Very rough overall architecture of the group commit**
image
Replace `txid` with `Key` when reading the description.

**Lifecycle of Group**
image

## Checklist

The current implementation contains the following 3 TODOs. They'll be handled in other PRs.
- Possible performance improvement with tweaking the scope of locks
- Replace the tentative metrics monitoring with DropWizard's one
- Performance improvement of DelayedSlotMoveWorker
  - The worker basically waits for every item, so the performance improvement is required
  - Introducing a priority queue is an option, which was suggested by @brfrn169 

## Release notes

N/A (This PR itself isn't a user facing change).


core/src/main/java/com/scalar/db/util/groupcommit/Group.java
```diff
@@ -0,0 +1,242 @@
...
+    if (noMoreSlot()) {
+      close();
+    }
+    return slot.fullKey();
+  }
+
+  private void reserveSlot(Slot<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> slot) {
+    Slot<?, ?, ?, ?, ?> oldSlot = slots.put(slot.key(), slot);
+    if (oldSlot != null) {
+      logger.warn("An old slot exist unexpectedly. {}", oldSlot.fullKey());
```
brfrn169, line 89: This shouldn't occur, right? If so, just showing the warning message is okay?

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
```diff
@@ -0,0 +1,190 @@
...
+          "The state of the group has been changed. Retrying. Group:{}, Keys:{}", group, keys);
+    }
+  }
+
+  /**
+   * Removes the slot from the group.
+   *
+   * @param fullKey A full key to specify the slot.
+   */
+  public void remove(FULL_KEY fullKey) {
```
brfrn169, line 168: When is this `remove` method intended to be called?

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Marks the slot associated with the specified key READY and then, waits until the group which
+   * contains the slot is emitted.
+   *
+   * @param fullKey A full key associated with the slot already reserved with {@link
+   *     GroupCommitter#reserve}.
+   * @param value A value to be set to the slot.
+   * @throws GroupCommitException when group commit fails
+   */
+  public void ready(FULL_KEY fullKey, V value) throws GroupCommitException {
```
brfrn169, line 141: What is the purpose of this `value` argument? Maybe we use it for a transaction ID (a child key) for each transaction within a group?

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * @param <PARENT_KEY> A key type to NormalGroup which contains multiple slots and is
+ *     group-committed.
+ * @param <CHILD_KEY> A key type to slot in NormalGroup which can contain a value ready to commit.
+ * @param <FULL_KEY> A key type to DelayedGroup which contains a single slot and is
+ *     singly-committed.
+ * @param <EMIT_KEY> A key type that Emitter can interpret.
+ * @param <V> A value type to be set to a slot.
+ */
+public class GroupCommitter<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> implements Closeable {
```
brfrn169, line 28: I think we can put `@ThreadSafe` or `@NotThreadSafe` for each class.

brfrn169 (commented):
My first review is complete, and I've left several comments. Please take a look when you have time! I will conduct a second review later on. Thank you for your great work!

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
```diff
@@ -0,0 +1,190 @@
...
+        delayedSlotMoveWorker.size(),
+        groupCleanupWorker.size(),
+        groupManager.sizeOfNormalGroupMap(),
+        groupManager.sizeOfDelayedGroupMap());
+  }
+
+  // TODO: This should be replaced by other metrics mechanism.
+  private void startMonitorExecutorService() {
+    Runnable print =
+        () -> logger.info("[MONITOR] Timestamp={}, Metrics={}", Instant.now(), getMetrics());
```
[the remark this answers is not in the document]
komamitsu: Sounds good! I'll update it.

BTW, this will be soon replaced with https://github.com/scalar-labs/scalardb/pull/1614 which uses DropWizard Metrics.

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
komamitsu: Good question.

Let's say a transaction executes some CRUD operations, starts PREPARE, and fails before the group commit for coordinator state. In this case, a group commit slot is allocated for the transaction and needed to be removed by calling this method not to leave garbage slots.

core/src/main/java/com/scalar/db/util/groupcommit/Group.java
komamitsu: Good catch! Right. I'll update it to throw AssertionError or IllegalStateException.

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
komamitsu: In the case of coordinator state commit, `V` is supposed to be `Snapshot` and a passed value will be committed with other snapshots in the same group.

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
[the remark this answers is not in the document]
komamitsu: Fixed in https://github.com/scalar-labs/scalardb/pull/1607/commits/ade2864d14f5902984fb40c553f98380427feef7

core/src/main/java/com/scalar/db/util/groupcommit/Group.java
komamitsu: Fixed in https://github.com/scalar-labs/scalardb/pull/1607/commits/8d96fd484d03b1d127d338ac3d128ad8d6232bbc

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
komamitsu: Updated the comment to make it clearer in https://github.com/scalar-labs/scalardb/pull/1607/commits/fa5e72631b3ca1ef5eee6a47fff13071d9adbd12

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
komamitsu: Updated the comment in https://github.com/scalar-labs/scalardb/pull/1607/commits/4ff928a8cf45c2e137c75fac93608e73f5b60040

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
komamitsu: Fixed in https://github.com/scalar-labs/scalardb/pull/1607/commits/e58191815c85f3d27ed4a4c7e8aa8d36549b9976

core/src/main/java/com/scalar/db/util/groupcommit/Group.java
```diff
@@ -0,0 +1,245 @@
...
+    // Accepting new slot reservation since the number of slots isn't fixed yet.
+    //
+    // Initial status of groups. No group can move back to OPEN.
+    OPEN(false, false, false),
+
+    // Not accepting new slot reservations since the number of slots is already fixed.
+    // Waiting all the slots are set with values.
+    //
+    // Groups with OPEN status can move to CLOSED.
+    CLOSED(true, false, false),
```
feeblefakie, line 35: `CLOSED` corresponds to `OPEN` so it's good, but it sounds like the group has already finished processing.
How about `FIXED`?

core/src/main/java/com/scalar/db/util/groupcommit/GroupManager.java
```diff
@@ -0,0 +1,220 @@
...
+    this.groupCleanupWorker = groupCleanupWorker;
+  }
+
+  // Reserves a new slot in the current NormalGroup. A new NormalGroup will be created and
+  // registered to `normalGroupMap` if the current NormalGroup is already closed.
+  //
+  // If it returns null, the Group is already closed and a retry is needed.
+  @Nullable
+  FULL_KEY reserveNewSlot(CHILD_KEY childKey) {
+    long stamp = lock.writeLock();
```
feeblefakie, line 71: I might miss something, but since we have only one active group at a time, I'm wondering how much it badly affects performance if there are many concurrent transactions.
We could use multiple active groups to mitigate the potential issue, but it might complicate the mechanism.
Any thoughts on this?

feeblefakie (commented):
Thank you for the great improvement and very sorry for the late review.
I took a brief look and left some comments and questions. PTAL!
(Let me check this one more time.)

Let me ask additional questions here.
- We discussed before that we might need to access the coordinator table twice in some recovery cases, but how does the current code handle the case?
- If one transaction in a group fails to commit, will the transaction be handled in a `DelayedGroup`?

core/src/main/java/com/scalar/db/util/groupcommit/Group.java
komamitsu: Thanks for the suggestion! Actually it was first like SIZE-FIXED and I changed it to CLOSED to sound to reject new requests. But as you said, it also sounds things are done. I'll update it to use FIX.

core/src/main/java/com/scalar/db/util/groupcommit/GroupManager.java
komamitsu: Good point. I also thought and tried multiple current (== active) groups before. But, the performance with multiple active groups didn't have a good performance. I think how fast to fill up a group with slots (and values) is a very important factor to move it to the next state (i.e., ready to commit) in the group commit and using multiple active groups doesn't fit.

I benchmarked this `util/groupcommit` pacakge itself and the average throughput was about 65K TPS without any wait while the throughput with some delays to emulate actual latencies of underlying database had far low throughputs. It means current single active group won't be a bottleneck in usual usages of ScalarDB, I think.

komamitsu:
@feeblefakie Thanks for the questions!

> We discussed before that we might need to access the coordinator table twice in some recovery cases, but how does the current code handle the case?

This PR contains doesn't contain any integration with the coordinator state commit, so let me refer to another PR's change (https://github.com/scalar-labs/scalardb/pull/1523/files#diff-8049e66d11b81cdd5a1dc82c4977b5336741bcee24dd7ddd54c6a6940f09497bR83-R89). I tried the idea to put child-tx-id in a clustering key before in order to make it possible to fetch both group-committed-transactions' ID and separately-committed-transactions' ID. But, there wasn't significant performance improvement with YCSB-F benchmark despite it would require cumbersome table schema migrations. So, current implementation is to access the coordinator table twice.

> If one transaction in a group fails to commit, will the transaction be handled in a DelayedGroup?

No. Slots are moved to DelayedGroup only when it's delayed. If a NormalGroup fails to commit due to network or underlying database issues, all the slots in the group fail. BTW, it's about Commit-State phase. If a transaction fails in Prepare or Validate, it'll be just removed from the NormalGroup without being moved to DelayedGroup.

core/src/main/java/com/scalar/db/util/groupcommit/Group.java
komamitsu: Fixed in https://github.com/scalar-labs/scalardb/pull/1607/commits/5000335099a2c15528ef9df89f7ba50e2c4dda82. Used `SIZE_FIXED` to make it clearer.

core/src/main/java/com/scalar/db/util/groupcommit/GroupManager.java
feeblefakie: Thank you for the detailed explanation!
It makes sense. Let's leave it as is.

feeblefakie:
@komamitsu 
> I tried the idea to put child-tx-id in a clustering key before in order to make it possible to fetch both group-committed-transactions' ID and separately-committed-transactions' ID. But, there wasn't significant performance improvement with YCSB-F benchmark despite it would require cumbersome table schema migrations. So, current implementation is to access the coordinator table twice.

Thank you for the explanation.
I think how much performance improves depends on how much the workload triggers conflicts,
so it might be worth trying TPC-C to see how it goes since there are not many conflicts in YCSB.

> No. Slots are moved to DelayedGroup only when it's delayed. If a NormalGroup fails to commit due to network or underlying database issues, all the slots in the group fail. BTW, it's about Commit-State phase. If a transaction fails in Prepare or Validate, it'll be just removed from the NormalGroup without being moved to DelayedGroup.

Thank you!
I got it.

core/src/main/java/com/scalar/db/util/groupcommit/KeyManipulator.java
```diff
@@ -0,0 +1,50 @@
...
+          .add("fullKey", fullKey)
+          .toString();
+    }
+  }
+
+  PARENT_KEY generateParentKey();
+
+  FULL_KEY fullKey(PARENT_KEY parentKey, CHILD_KEY childKey);
+
+  boolean isFullKey(Object obj);
```
brfrn169, line 43: Nit: It looks like it's never used.

core/src/main/java/com/scalar/db/util/groupcommit/ThrowableRunnable.java
```diff
@@ -0,0 +1,6 @@
+package com.scalar.db.util.groupcommit;
+
+@FunctionalInterface
+interface ThrowableRunnable {
```
brfrn169, line 4: We already have `ThrowableRunnable`:
https://github.com/scalar-labs/scalardb/blob/53af316e1badb9460652ef7fcbb54612756edd3b/core/src/main/java/com/scalar/db/util/ThrowableRunnable.java

feeblefakie (commented):
Overall, looking good!
Left one question. PTAL!

komamitsu:
> so it might be worth trying TPC-C to see how it goes since there are not many conflicts in YCSB.

@feeblefakie Yeah, good point. I'll try TPC-C with the clustering key version group commit before submitting a PR for the integration of the group commit with the coordinator state commit.

core/src/main/java/com/scalar/db/util/groupcommit/GroupManager.java
```diff
@@ -0,0 +1,211 @@
...
+  private GroupCleanupWorker<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> groupCleanupWorker;
+
+  // Custom operations injected by the client
+  private final KeyManipulator<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY> keyManipulator;
+  @LazyInit private Emittable<EMIT_KEY, V> emitter;
+
+  private final GroupCommitConfig config;
+
+  GroupManager(
+      String label,
```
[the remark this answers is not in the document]
komamitsu: Fixed it in https://github.com/scalar-labs/scalardb/pull/1607/commits/0007ba39a3d7ccc73ac7064e2fc81193ac0c4fc8

core/src/main/java/com/scalar/db/util/groupcommit/ThrowableRunnable.java
komamitsu: Fixed it in https://github.com/scalar-labs/scalardb/pull/1607/commits/e8faee28e7f1a7f7cdfb2b2d1d36cc8673e34e36

core/src/main/java/com/scalar/db/util/groupcommit/BackgroundWorker.java
```diff
@@ -0,0 +1,117 @@
...
+    T item = queue.peek();
+
+    if (item != null) {
+      boolean shouldRemove = processItem(item);
+      if (shouldRemove) {
+        // Remove the handled item.
+        T removed = queue.poll();
+        // Check if the removed group is expected just in case.
+        if (removed == null || !removed.equals(item)) {
+          logger.error(
```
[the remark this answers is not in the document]
komamitsu: Good point. It shouldn't happen in normal cases. Throwing AssertionError or IllegalStateException is an option. But throwing either of them from background worker doesn't make sense so much and I think just outputting error log is enough.

But, I noticed the error message probably looks unclear thanks to your comment. I updated it in https://github.com/scalar-labs/scalardb/pull/1607/commits/bdf344f560b33f02316c16a1c87e037782eaea5f

komamitsu:
@feeblefakie @brfrn169 Thanks for reviewing this PR!
This is a small FYI, but I renamed `asyncEmit()` to `delegateEmitTaskToWaiter()` in https://github.com/scalar-labs/scalardb/pull/1607/commits/545ad204a0c4ee05931c0a8f983f7ee0225d0217 since the old method name was misleading (actually, the emit task is executed by one of the slots in the group synchronously in `waitUntilEmit()`)

@brfrn169 The performance improvement task of DelayedSlotMoveWorker will be handled in another PR https://github.com/scalar-labs/scalardb/pull/1607/commits/545ad204a0c4ee05931c0a8f983f7ee0225d0217#diff-21491e996128b6ee7175154186a58be20ca714274559ba304eb39c59be8db4faR110-R113

core/src/main/java/com/scalar/db/util/groupcommit/Slot.java
```diff
@@ -0,0 +1,123 @@
...
+
+// A container of value which is stored in a group.
+@ThreadSafe
+class Slot<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> {
+  private final AtomicReference<Group<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V>> parentGroup =
+      new AtomicReference<>();
+  private final CHILD_KEY key;
+  // If a result value is null, the value is already emitted.
+  // Otherwise, the result lambda must be emitted by the receiver's thread.
+  private final CompletableFuture<ThrowableRunnable> completableFuture = new CompletableFuture<>();
```
brfrn169, line 21:
```suggestion
  private final CompletableFuture<ThrowableRunnable<Exception>> completableFuture = new CompletableFuture<>();
```

core/src/main/java/com/scalar/db/util/groupcommit/Slot.java
```diff
@@ -0,0 +1,123 @@
...
+
+  void setValue(V value) {
+    this.value.set(Objects.requireNonNull(value));
+  }
+
+  void waitUntilEmit() throws GroupCommitException {
+    try {
+      // If a result value is null, the value is already emitted.
+      // Otherwise, the result lambda must be emitted by the receiver's thread.
+      ThrowableRunnable taskToEmit = completableFuture.get();
```
brfrn169, line 54:
```suggestion
      ThrowableRunnable<Exception> taskToEmit = completableFuture.get();
```

core/src/main/java/com/scalar/db/util/groupcommit/Slot.java
```diff
@@ -0,0 +1,123 @@
...
+      throw new GroupCommitException(
+          String.format("Group commit was interrupted, Group: %s", parentGroup.get()), e);
+    } catch (ExecutionException e) {
+      Throwable cause = e.getCause();
+      if (cause instanceof GroupCommitException) {
+        throw (GroupCommitException) cause;
+      }
+      throw new GroupCommitException(
+          String.format("Group commit failed. Group: %s", parentGroup.get()), cause);
+    } catch (Throwable e) {
```
brfrn169, line 69:
```suggestion
    } catch (Exception e) {
```

core/src/main/java/com/scalar/db/util/groupcommit/Slot.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  // Marks this slot as a failure.
+  void markAsFailed(Exception e) {
+    completableFuture.completeExceptionally(e);
+  }
+
+  // Delegates the emit task to the client. The client receiving this task needs to handle the emit
+  // task.
+  void delegateTaskToWaiter(ThrowableRunnable task) {
```
brfrn169, line 112:
```suggestion
  void delegateTaskToWaiter(ThrowableRunnable<Exception> task) {
```

core/src/main/java/com/scalar/db/util/groupcommit/NormalGroup.java
```diff
@@ -0,0 +1,181 @@
...
+      if (isFirst) {
+        isFirst = false;
+        emitterSlot.set(slot);
+      }
+      values.add(slot.value());
+    }
+
+    // This task is passed to only the first slot, so the slot will be resumed.
+    // Other slots will be blocked until `markAsXxxx()` is called.
+    ThrowableRunnable<GroupCommitException> taskForEmitterSlot =
```
brfrn169, line 109:
```suggestion
    ThrowableRunnable<Exception> taskForEmitterSlot =
```

brfrn169 (commented):
Left very minor suggestions. It looks good to me, but let me review this once again later.

core/src/main/java/com/scalar/db/util/groupcommit/GroupManager.java
```diff
@@ -0,0 +1,214 @@
...
+  // ready. The new one is also
+  // registered to the group map and the cleanup queue.
+  //
+  // Returns true if any delayed slot is moved, false otherwise.
+  boolean moveDelayedSlotToDelayedGroup(
+      NormalGroup<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> normalGroup) {
+    long stamp = lock.writeLock();
+    try {
+      // TODO: NormalGroup.removeNotReadySlots() calls updateStatus() potentially resulting in
+      //       asyncEmit(). Maybe it should be called outside the lock.
```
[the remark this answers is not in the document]
komamitsu: Thanks! Fixed it in https://github.com/scalar-labs/scalardb/pull/1607/commits/865830013b893112149fa88ce705cd60aa465226

core/src/main/java/com/scalar/db/util/groupcommit/Slot.java
komamitsu: Thanks! Fixed it in https://github.com/scalar-labs/scalardb/pull/1607/commits/865830013b893112149fa88ce705cd60aa465226

core/src/main/java/com/scalar/db/util/groupcommit/Slot.java
komamitsu: Thanks! Fixed it in https://github.com/scalar-labs/scalardb/pull/1607/commits/865830013b893112149fa88ce705cd60aa465226

core/src/main/java/com/scalar/db/util/groupcommit/NormalGroup.java
komamitsu: Thanks! Fixed it in https://github.com/scalar-labs/scalardb/pull/1607/commits/865830013b893112149fa88ce705cd60aa465226

core/src/main/java/com/scalar/db/util/groupcommit/BackgroundWorker.java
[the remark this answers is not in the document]
feeblefakie: Looking good!
Yeah, it cannot stop processing even such a case, so having the updated comment makes sense.

feeblefakie (approved):
LGTM! Thank you!

PR #1641, opened 2024-04-10
Use the priority queue for DelayedSlotMoveWorker to easily reduce the latency to be handled in `util/groupcommit`

komamitsu:
## Description

This is a minor performance improvement for DelayedSlotMoveWorker by using priority queue. The idea was suggested by @brfrn169.

In the current implementation, the worker uses a normal FIFO queue. Due to the characteristic of the timeout logic, a NormalGroup which will be timed-out soon can be enqueued at the tail and it might cause latency until being handled by the worker. By using priority queue, the latency would be reduced.

I also tried to modify the wait duration decision logic based on the head item, but I didn't see a significant performance improvement and left the simple logic as is.

## Related issues and/or PRs

I mentioned this change in https://github.com/scalar-labs/scalardb/pull/1607 as a TODO.

## Changes made

- Use `PriorityBlockingQueue` instead of `LinkedBlockingQueue` in `DelayedSlotMoveWorker`

## Checklist

I observed a slight performance improvement with this change. The benchmark was YCSB-F
image

## Release notes

Performance improvement of the group commit by using priority queue in the background worker.


feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/util/groupcommit/DelayedSlotMoveWorker.java
```diff
@@ -20,11 +23,17 @@ class DelayedSlotMoveWorker<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V>
...
         // be updated.
-        RetryMode.MOVE_TO_TAIL);
+        RetryMode.RE_ENQUEUE);
     this.groupManager = groupManager;
     this.groupCleanupWorker = groupCleanupWorker;
   }
 
+  @Override
+  BlockingQueue<NormalGroup<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V>> createQueue() {
+    return new PriorityBlockingQueue<>(
```
feeblefakie, line 33: Does this mean an item with the smallest timeout value is taken first?
It might be helpful to add a comment on that.

feeblefakie (approved):
Just add an minor comment.

core/src/main/java/com/scalar/db/util/groupcommit/DelayedSlotMoveWorker.java
komamitsu: Thanks for the feedback! Addressed it in https://github.com/scalar-labs/scalardb/pull/1641/commits/7e66c31d4ffa168c47d2952d7b3344b579f38148

core/src/main/java/com/scalar/db/util/groupcommit/GroupSizeFixWorker.java
```diff
@@ -34,6 +36,11 @@ private void enqueueItemToNextQueue(
     }
   }
 
+  @Override
+  BlockingQueue<NormalGroup<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V>> createQueue() {
+    return new LinkedBlockingQueue<>();
```
brfrn169, line 41: I thought we could also use the priority queue (with `groupSizeFixTimeoutAtMillis()`) in `GroupSizeFixWorker` as follows. What do you think?

```java
  @Override
  BlockingQueue<NormalGroup<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V>> createQueue() {
    return new PriorityBlockingQueue<>(
        64, Comparator.comparingLong(NormalGroup::groupSizeFixTimeoutAtMillis));
  }
```

brfrn169 (commented):
Overall LGTM. Left a comment. Please take a look when you have time!

core/src/main/java/com/scalar/db/util/groupcommit/GroupSizeFixWorker.java
komamitsu: Good question! The reason why using the normal queue is as follows:
- Queued groups are removed once processed, without being re-enqueued
- No need for a priority queue since the order of queued groups is basically consistent with the timeout order

I also added comments in https://github.com/scalar-labs/scalardb/pull/1641/commits/17a604bb584300387cfac72e004a388cb630d222

brfrn169: Understood. Thank you!

brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #1650, opened 2024-04-11
Wait until all groups get done when GroupCommitter is closed (was "Abort all in-flight groups on close in the group commit")

komamitsu:
## Description

If an application using the group commit feature doesn't abort in-flight transactions (in-flight `slots` in terms of the group commit package) that failed before commit phase (before calling `ready()` method in terms of the group commit package), the client threads that have those failed in-flight transactions will wait until `old-group-timeout` is triggered even if `close()` is called. What if the application recreates the transaction manager including the GroupCommitter while there are failed in-flight transactions? The new GroupCommitter doesn't have any information about the in-flight transactions, resulting in that no one aborts them and the clients will wait forever. This case might sounds very weird, but https://github.com/scalar-labs/scalar-jepsen tests this type of situation, great. To prevent the client threads from waiting forever even in this case, ~~all the groups should be aborted~~ GroupCommitter should wait until all groups get done when `GroupCommitter.close()` is called.

This issue happens if the following steps are executed
1. `client#1` begins `txn#1`
  i. The NormalGroup contains `[txn#1(empty)]` now
2. `client#2` begins `txn#2`
  i. The NormalGroup contains `[txn#1(empty), txn#2(empty)]` now
3. `txn#1` failes in the CRUD (execution) phase, and `client#1` doesn't abort it for some reason
  i. The NormalGroup contains `[txn#1(empty), txn#2(empty)]` now
4. `txn#2` executes some CRUD operations and try committing
  i. The NormalGroup contains `[txn#1(empty), txn#2(snapshot-for-txn#2)]` now
5. The application closes the TransactionManager by calling `close()` which results in calling `GroupCommitter.close()`, then all the background workers stop
  i. The NormalGroup contains `[txn#1(empty), txn#2(snapshot-for-txn#2)]` now
6. `client#2` is still waiting on the CompletableFuture inside the slot for `txn#2`
  i. The NormalGroup contains `[txn#1(empty), txn#2(snapshot-for-txn#2)]` now
7. No background workers taking care of the NormalGroup is running. `client#2` gets stuck since the NormalGroup won't be READY due to the open slot for `txn#1`

[This comment](https://github.com/scalar-labs/scalardb/pull/1650#discussion_r1566716082) contains an example code to reproduce this issue. It might be helpful to understand.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/1607

## Changes made

This PR makes GroupCommitter ~~abort all the in-flight slots by calling a newly added `GroupManager#abortAllGroups()`~~ wait until all groups get done when `GroupCommitter.close()` is called. To test the behaviors, some initialization codes are refactored.

## Checklist

None

## Release notes

N/A (The group commit package isn't used yet).

### DIFF
--- a/core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitMonitor.java
+++ b/core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitMonitor.java
@@ -1,0 +1,47 @@
+package com.scalar.db.util.groupcommit;
+
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.google.common.util.concurrent.Uninterruptibles;
+import java.io.Closeable;
+import java.time.Instant;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class GroupCommitMonitor implements Closeable {
+  private static final Logger logger = LoggerFactory.getLogger(GroupCommitMonitor.class);
+  private final ExecutorService executorService;
+
+  GroupCommitMonitor(String label, Supplier<Metrics> metricsSupplier) {
+    executorService =
+        Executors.newSingleThreadExecutor(
+            new ThreadFactoryBuilder()
+                .setDaemon(true)
+                .setNameFormat(label + "-group-commit-monitor-%d")
+                .build());
+    startExecutorService(metricsSupplier);
+  }
+
+  private void startExecutorService(Supplier<Metrics> metricsSupplier) {
+    Runnable print =
+        () -> logger.debug("[MONITOR] Timestamp={}, Metrics={}", Instant.now(), metricsSupplier);
+
+    executorService.execute(
+        () -> {
+          while (!executorService.isShutdown()) {
+            print.run();
+            Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
+          }
+          print.run();
+        });
+  }
+
+  @Override
+  public void close() {
+    MoreExecutors.shutdownAndAwaitTermination(executorService, 10, TimeUnit.SECONDS);
+  }
+}
--- a/core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
+++ b/core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
@@ -30,7 +30,7 @@ public class GroupCommitter<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> implem
       delayedSlotMoveWorker;
   private final GroupCleanupWorker<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> groupCleanupWorker;
 
-  // Executors
+  // Monitor
   private final GroupCommitMonitor groupCommitMonitor;
 
   // This contains logics of how to treat keys.
--- a/core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
+++ b/core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
@@ -1,14 +1,8 @@
 package com.scalar.db.util.groupcommit;
 
-import com.google.common.util.concurrent.MoreExecutors;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import com.google.common.util.concurrent.Uninterruptibles;
+import com.google.common.annotations.VisibleForTesting;
 import com.scalar.db.util.groupcommit.KeyManipulator.Keys;
 import java.io.Closeable;
-import java.time.Instant;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 import javax.annotation.concurrent.ThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -37,7 +31,7 @@ public class GroupCommitter<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> implem
   private final GroupCleanupWorker<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> groupCleanupWorker;
 
   // Executors
-  private final ExecutorService monitorExecutorService;
+  private final GroupCommitMonitor groupCommitMonitor;
 
   // This contains logics of how to treat keys.
   private final KeyManipulator<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY> keyManipulator;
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitMonitor.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.Closeable;
+import java.time.Instant;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class GroupCommitMonitor implements Closeable {
```
komamitsu, line 15: This class was extracted for testing.

core/src/main/java/com/scalar/db/util/groupcommit/GroupCommitter.java
```diff
@@ -53,28 +47,18 @@ public GroupCommitter(
       String label,
       GroupCommitConfig config,
       KeyManipulator<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY> keyManipulator) {
-    logger.info("Staring GroupCommitter. Label: {}, Config: {}", label, config);
+    logger.info("Starting GroupCommitter. Label: {}, Config: {}", label, config);
     this.keyManipulator = keyManipulator;
-    this.groupManager = new GroupManager<>(config, keyManipulator);
-    this.groupCleanupWorker =
-        new GroupCleanupWorker<>(label, config.timeoutCheckIntervalMillis(), groupManager);
+    this.groupManager = createGroupManager(config, keyManipulator);
```
komamitsu, line 52: This refactoring is needed for testing

core/src/main/java/com/scalar/db/util/groupcommit/GroupManager.java
```diff
@@ -211,4 +216,15 @@ int sizeOfNormalGroupMap() {
   int sizeOfDelayedGroupMap() {
     return delayedGroupMap.size();
   }
+
+  void abortAllGroups() {
```
brfrn169, line 220: Currently, we don't explicitly abort ongoing transactions when closing transaction manager instances.

> What if the application recreates the transaction manager including the GroupCommitter while there are failed in-flight transactions? The new GroupCommitter doesn't have any information about the in-flight transactions, resulting in that no one aborts them and the clients will wait forever.

Sorry, I don't fully understand why the clients will wait forever. This means that the new GroupCommitter instance interferes the existing GroupCommitter instance? I thought the GroupCommitter instances independently ran.

komamitsu: Yeah, this issue is a very uncommon and complicated... Actually, new GroupCommitter doesn't matter in this case. Maybe the following sample code which reproduces the issue is helpful.

```java
 Properties properties = new Properties();
  ...
  properties.put("scalar.db.consensus_commit.coordinator.group_commit.enabled", "true");
  // This is too large, but emulating that a lot of groups are contained in DelayedSlotMoveWorker.
  properties.put("scalar.db.consensus_commit.coordinator.group_commit.delayed_slot_move_timeout_millis", "10000");
  // The default value is 60 seconds. Shoten this to easily to see if the issue happens.
  properties.put("scalar.db.consensus_commit.coordinator.group_commit.old_group_abort_timeout_seconds", "20");
  TransactionFactory factory = TransactionFactory.create(properties);
  DistributedTransactionManager transactionManager = factory.getTransactionManager();
  ExecutorService executorService = Executors.newCachedThreadPool();

  // Transaction#1 should fail in the execution phase.
  Future<?> future1 = executorService.submit(() -> {
    DistributedTransaction txn;
    try {
      txn = transactionManager.begin();
    } catch (TransactionException e) {
      throw new RuntimeException(e);
    }

    try {
      txn.get(
          Get.newBuilder()
              .namespace("ycsb")
              .table("usertable")
             // Wrong key name to fail in the execution phase.
              .partitionKey(Key.ofInt("xxxxxxx_ycsb_key", 42))
              .build());
      txn.commit();
    } catch (TransactionException e) {
      // Don't abort the transaction.
      throw new RuntimeException(e);
    }
  });

  // Transaction#2 should be suspended in the commit phase.
  Future<?> future2 = executorService.submit(() -> {
    DistributedTransaction txn;
    try {
      txn = transactionManager.begin();
    } catch (TransactionException e) {
      throw new RuntimeException(e);
    }

    try {
      Optional<Result> result = txn.get(
          Get.newBuilder()
              .namespace("ycsb")
              .table("usertable")
              .partitionKey(Key.ofInt("ycsb_key", 42))
              .build());
      txn.commit();
      return result;
    } catch (TransactionException e) {
      throw new RuntimeException(e);
    }
  });

  executorService.shutdown();

  TimeUnit.SECONDS.sleep(2);
  // At this moment, the group contains these slots
  // - Slot#1 (transaction#1)
  //   - No value (Snapshot) is set since the transaction failed before committing
  //   - If the client aborts the transaction, this slot is removed and Slot#2 can proceed.
  //   - Also, in normal case, this should be moved to a DelayedGroup (or aborted) by DelayedSlotMoveWorker,
  //     and Slot#2 can move forward. But the worker can be delayed and the user can close the transaction manager before that
  // - Slot#2 (transaction#2)
  //   - A value (Snapshot) is set. And the client thread is waiting until the group is committed

  transactionManager.close();
  // All the background workers are closed. No one is taking care of the information of in-flight groups.

  try {
    future1.get();
  } catch (ExecutionException e) {
    // Expected
    System.out.println("Caught an exception from the first transaction expectedly: " + e.getMessage());
  } catch (InterruptedException e) {
    throw new RuntimeException(e);
  }

  try {
    long start = System.currentTimeMillis();
    // This gets stuck unfortunately forever even waiting more than
    // `old_group_abort_timeout_seconds` since no background worker is running.
    future2.get();
    System.out.println("Got the result from the second transaction. Duration: " + (System.currentTimeMillis() - start));
  } catch (ExecutionException | InterruptedException e) {
    throw new RuntimeException(e);
  }

  // transactionManager.close();
```

Without this PR's change, the above example code lets the client thread for transaction#2 stuck forever.

brfrn169: Thank you for the explanation. I may not fully understand the issue yet, but I think we should wait for the queues for the worker threads to be empty when closing GroupCommitter. What do you think?

feeblefakie, line 220: I would like to understand it correctly.
Calling this is for earlier recovery and not required for safety and liveness, right?
I'm a little concerned about the `the clients will wait forever` part in the description since it sounds like no liveness is guaranteed.
There might be a case where a client crashes before aborting all groups (calling this method) so I think the correctness should be guaranteed without this.

komamitsu: > Calling this is for earlier recovery and not required for safety and liveness, right?

Thanks for the question. I think this is for liveness property.

In the following special case, the client of a transaction waits forever (maybe [this example code](https://github.com/scalar-labs/scalardb/pull/1650#discussion_r1566716082) is more helpful to understand):
1. `client#1` begins `txn#1`. The NormalGroup contains `[txn#1(empty)]` now
2. `client#2` begins `txn#2`. The NormalGroup contains `[txn#1(empty), txn#2(empty)]` now
3. `txn#1` failes in the CRUD (execution) phase, and `client#1` doesn't abort it for some reason.  The NormalGroup contains `[txn#1(empty), txn#2(empty)]` now
4. `txn#2` executes some CRUD operations and try committing. The NormalGroup contains `[txn#1(empty), txn#2(snapshot-for-txn#2)]` now
5. The application closes the TransactionManager by calling `close()` which results in calling `GroupCommitter.close()`, then all the background workers stop. The NormalGroup contains `[txn#1(empty), txn#2(snapshot-for-txn#2)]` now
6. `client#2` is still waiting on the CompletableFuture inside the slot for `txn#2`. The NormalGroup contains `[txn#1(empty), txn#2(snapshot-for-txn#2)]` now
7. No background workers taking care of the NormalGroup is running. `client#2` gets stuck since the NormalGroup won't be READY due to the open slot for `txn#1`

> where a client crashes before aborting all groups (calling this method) so I think the correctness should be guaranteed without this.

In this case (the client didn't close the TransactionManager, right?), DelayedSlotMoveWorker will either move `txn#1` to a DelayedGroup or remove timed-out `txn#1` from the NormalGroup so that `txn#2` will be committed. The point is if `GroupCommitter.close()` is called.

This PR for the case a client closes the TransactionManager while there are ongoing transactions and the client keeps waiting on the transactions.

brfrn169: IMO, if users close the transaction manager instance even though there are ongoing transactions, I believe that responsibility lies with the users. Therefore, I think we can simply go with option #2.

komamitsu: Yeah, that sounds reasonable. The option 3 I suggested was just based on the option 2 by adding the timeout. We can incrementally add the timeout mechanism to the option 2 later if needed.

But, on second thougtht, I just noticed what if clients keep starting new transactions which I faced in https://github.com/scalar-labs/scalar-jepsen. In this case, probably it won't finish and the test will finally fail. So, let me go with option 3 just in case.

komamitsu: Hmm. The both option 2 and 3 should make GroupCommitter reject new requests (i.e., `reserve()`) once it's closed. With this logic, option 2 also can handle the situation I mentioned in the previous comment. I'll take the simpler option 2 😅

komamitsu: Addressed it in https://github.com/scalar-labs/scalardb/pull/1650/commits/dd1a0bb94ddc002decce8104ddc103120d82ff5c

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #1679, opened 2024-04-16
Introduce insert mode for Consensus Commit Put operations

brfrn169:
## Description

This PR introduces the insert mode for Consensus Commit Put operations. When the insert mode is enabled for a put operation, the preparation write for the put operation is executed without the target record in the read set. Therefore, regardless of whether to read the target record or not, we try to write the target record for the put operation as an initial record. One of the use cases of this feature is to implement the SQL INSERT statement behavior.

Users can enable the insert mode with the `enableInsertMode()` method (or the `insertModeEnabled()` method) of the put builder as follows:
```java
Put put =
    Put.newBuilder()
        .namespace("ns")
        .table("tbl")
        .partitionKey(partitionKey)
        .clusteringKey(clusteringKey)
        .floatValue("c4", 1.23F)
        .doubleValue("c5", 4.56)
        .enableInsertMode()
        .build();
```

## Related issues and/or PRs

N/A

## Changes made

- Added the insert mode related methods (`enableInsertMode()`, `disableInsertMode()`, and `insertModeEnabled()`) to the put builder.
- Modified the logic of `PrepareMutationComposer` to write a target record of a put operation as an initial record when the insert mode is enabled for the put operation.
- Included unit tests and integration tests for the insert mode.

## Checklist

- The insert mode is primarily for internal use, so we don't document this feature. For the same reason, we don't include a release note for this feature.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/api/OperationBuilder.java
+++ b/core/src/main/java/com/scalar/db/api/OperationBuilder.java
@@ -305,6 +305,30 @@ class OperationBuilder {
     T implicitPreReadEnabled(boolean implicitPreReadEnabled);
   }
 
+  interface InsertModeEnabled<T> {
+    /**
+     * Disables the insert mode for this put operation.
+     *
+     * @return the operation builder
+     */
+    T disableInsertMode();
+
+    /**
+     * Enables the insert mode for this put operation.
+     *
+     * @return the operation builder
+     */
+    T enableInsertMode();
+
+    /**
+     * Sets whether the insert mode is enabled or not for this put operation.
+     *
+     * @param insertModeEnabled whether the insert mode is enabled or not
+     * @return the operation builder
+     */
+    T insertModeEnabled(boolean insertModeEnabled);
+  }
+
   interface Limit<T> {
     /**
      * Sets the specified number of results to be returned
--- a/core/src/main/java/com/scalar/db/api/PutBuilder.java
+++ b/core/src/main/java/com/scalar/db/api/PutBuilder.java
@@ -237,7 +237,7 @@ public class PutBuilder {
 
     @Override
     public Buildable insertModeEnabled(boolean insertModeEnabled) {
-      this.implicitPreReadEnabled = insertModeEnabled;
+      this.insertModeEnabled = insertModeEnabled;
       return this;
     }
 
--- a/core/src/main/java/com/scalar/db/api/PutBuilder.java
+++ b/core/src/main/java/com/scalar/db/api/PutBuilder.java
@@ -10,6 +10,7 @@ import com.scalar.db.api.OperationBuilder.ClusteringKey;
 import com.scalar.db.api.OperationBuilder.Condition;
 import com.scalar.db.api.OperationBuilder.Consistency;
 import com.scalar.db.api.OperationBuilder.ImplicitPreReadEnabled;
+import com.scalar.db.api.OperationBuilder.InsertModeEnabled;
 import com.scalar.db.api.OperationBuilder.PartitionKeyBuilder;
 import com.scalar.db.api.OperationBuilder.TableBuilder;
 import com.scalar.db.api.OperationBuilder.Values;
@@ -78,12 +79,14 @@ public class PutBuilder {
           Consistency<Buildable>,
           Condition<Buildable>,
           Values<Buildable>,
-          ImplicitPreReadEnabled<Buildable> {
+          ImplicitPreReadEnabled<Buildable>,
+          InsertModeEnabled<Buildable> {
     final Map<String, Column<?>> columns = new LinkedHashMap<>();
     @Nullable Key clusteringKey;
     @Nullable com.scalar.db.api.Consistency consistency;
     @Nullable MutationCondition condition;
     boolean implicitPreReadEnabled;
+    boolean insertModeEnabled;
 
     private Buildable(@Nullable String namespace, String table, Key partitionKey) {
       super(namespace, table, partitionKey);
@@ -221,6 +224,24 @@ public class PutBuilder {
     }
 
     @Override
+    public Buildable disableInsertMode() {
+      insertModeEnabled = false;
+      return this;
+    }
+
+    @Override
+    public Buildable enableInsertMode() {
+      insertModeEnabled = true;
+      return this;
+    }
+
+    @Override
+    public Buildable insertModeEnabled(boolean insertModeEnabled) {
+      this.implicitPreReadEnabled = insertModeEnabled;
+      return this;
+    }
+
+    @Override
     public Put build() {
       Put put = new Put(partitionKey, clusteringKey);
       put.forNamespace(namespaceName).forTable(tableName);
@@ -232,6 +253,7 @@ public class PutBuilder {
         put.withCondition(condition);
       }
       put.setImplicitPreReadEnabled(implicitPreReadEnabled);
+      put.setInsertModeEnabled(insertModeEnabled);
 
       return put;
     }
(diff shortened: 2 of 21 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java
```diff
@@ -67,7 +67,7 @@ private void add(Put base, @Nullable TransactionResult result) throws ExecutionE
     putBuilder.intValue(Attribute.STATE, TransactionState.PREPARED.get());
     putBuilder.bigIntValue(Attribute.PREPARED_AT, current);
 
-    if (result != null) { // overwrite existing record
+    if (!base.isInsertModeEnabled() && result != null) { // overwrite existing record
```
brfrn169, line 70: When the insert mode is enabled for a put operation, we always try to write the record of the put operation as an initial record. In other words, even if the target record of the put operation is in the read set, we do not use it for the preparation write when the insert mode is enabled.

komamitsu:
@brfrn169 This change itself looks good to me 👍 

But I think some features (e.g., Semi-Sync Replication) which read partial orders between transaction IDs from the write operation history of a record and restore the final state from them won't work with this change.

For example, let's say there are 3 transactions (T1, T2 and T3: T1 < T2 < T3) containing a write operation on record R with the insert mode enabled. The operation history on the record R would be like this:

Put#1
  - version: 1
  - tx_id: T1
  - state: COMMITTED
  - before columns: empty
  - columns: values written by T1
Put#2
  - version: 1
  - tx_id: T2
  - state: COMMITTED
  - before columns: empty
  - columns: values written by T2
Put#3
  - version: 1
  - tx_id: T3
  - state: COMMITTED
  - before columns: empty
  - columns: values written by T3

This write operation history doesn't contain partial order information since `before_tx_id` is empty. So, if we apply the 3 write operations in the order of Put#3, Put#1 and then Put#2, the final state of R will be along with Put#2 not Put#3.

Also, we can't distinguish initial Put (== INSERT) and Put with insert mode enabled (== blind UPDATE). So, I don't think such a feature which restores the final state from historical write operations depending on the partial order works with this insert mode Put feature. Probably we should describe the limitation somewhere later.

brfrn169:
@komamitsu Thank you for taking a look!

> For example, let's say there are 3 transactions (T1, T2 and T3: T1 < T2 < T3) containing a write operation on record R with the insert mode enabled. The operation history on the record R would be like this:

Actually, in that scenario, only T1 would be committed, and T2 and T3 would be aborted due to conflicts. When enabling the insert mode, we always attempt to write the target record for the put operation as an initial record, which means it will succeed only when the target record doesn't exist.

We need this feature when implementing the SQL INSERT behavior. The SQL INSERT should fail if the target record exists.

Consider the following scenario:

1. Begin a transaction
2. SELECT record A
3. INSERT record A
4. Commit the transaction

In this scenario, we use the ScalarDB get operation for step 2 and the ScalarDB put operation for step 3.

Without the insert mode, if record A exists and the get operation in step 2 returns record A, the put operation in step 3 overwrites record A, resulting in unexpected UPDATE behavior.

With the insert mode, even if record A exists and the get operation in step 2 returns record A, the put operation in step 3 fails due to a conflict, resulting in the expected INSERT behavior.

core/src/main/java/com/scalar/db/api/PutBuilder.java
```diff
@@ -220,6 +223,24 @@ public Buildable implicitPreReadEnabled(boolean implicitPreReadEnabled) {
...
+
+    @Override
+    public Buildable enableInsertMode() {
+      insertModeEnabled = true;
+      return this;
+    }
+
+    @Override
+    public Buildable insertModeEnabled(boolean insertModeEnabled) {
+      this.implicitPreReadEnabled = insertModeEnabled;
```
komamitsu, line 240:
```suggestion
      this.insertModeEnabled = insertModeEnabled;
```
Typo?

komamitsu:
> only T1 would be committed, and T2 and T3 would be aborted due to conflicts

@brfrn169 I see. Thanks for the explanation. So, to achieve this, all the Put operations should have `putIfNotExists()`? I thought the condition wouldn't be set when looking at the example in the description. `enableInsertMode()` itself doesn't enable the condition, right?

```java
Put put =
    Put.newBuilder()
        .namespace("ns")
        .table("tbl")
        .partitionKey(partitionKey)
        .clusteringKey(clusteringKey)
        .floatValue("c4", 1.23F)
        .doubleValue("c5", 4.56)
        .enableInsertMode()
        .build();
```

> Without the insert mode, if record A exists and the get operation in step 2 returns record A, the put operation in step 3 overwrites record A, resulting in unexpected UPDATE behavior.

I think I'm still missing something, but this Put for INSERT should have `putIfNotExists` and fail even without the insert mode ?


BTW, I feel "insert mode" sounds a little bit SQL-ish. "creation mode" may be another option? I don't have a strong opinion on it, though.

core/src/main/java/com/scalar/db/api/PutBuilder.java
brfrn169: Ah good catch! Thanks. Fixed in b745e97cd911b17c0802512a2024eead068c5cc6.

brfrn169:
@komamitsu 

> So, to achieve this, all the Put operations should have putIfNotExists()? I thought the condition wouldn't be set when looking at the example in the description. enableInsertMode() itself doesn't enable the condition, right?

An actual storage put operation for a put operation with the insert mode has `putIfNotExists`:
https://github.com/scalar-labs/scalardb/blob/b745e97cd911b17c0802512a2024eead068c5cc6/core/src/main/java/com/scalar/db/transaction/consensuscommit/PrepareMutationComposer.java#L92

> I think I'm still missing something, but this Put for INSERT should have putIfNotExists and fail even without the insert mode ?

Good point. Yes, we can still specify `putIfNotExists` explicitly for the put operation for INSERT. However, in Consensus Commit, when we specify `putIfNotExists` for the put operation, we internally need to read the record and check if the record exists or not, which causes one extra read. Instead, we can enable the insert mode to execute INSERT behavior without extra read.

> BTW, I feel "insert mode" sounds a little bit SQL-ish. "creation mode" may be another option? I don't have a strong opinion on it, though.

Actually, we plan to add `insert()`, `upsert()` and `update()` methods to the transaction API. So maybe we can use the `insert` word here. Thanks.

komamitsu (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/api/OperationBuilder.java
```diff
@@ -305,6 +305,30 @@ interface ImplicitPreReadEnabled<T> {
     T implicitPreReadEnabled(boolean implicitPreReadEnabled);
   }
 
+  interface InsertModeEnabled<T> {
```
feeblefakie, line 308: It might be better to have `PutMode` enum (or something) and have `INSERT`, `UPSERT`, and `UPDATE` to match the corresponding SQL commands?
Not fully sure, but it might be worth thinking now.

brfrn169: Let me explain the background. We plan to introduce `insert()`, `upsert()`, and `update()` methods to the transaction API (`TransactionCrudOperable`). Additionally, we intend to mark the `put()` method in the transaction API as deprecated. The insert mode will be used in the `insert()` method.

What do you think?

feeblefakie (approved):
LGTM! Thank you!

PR #1682, opened 2024-04-18
Support dynamic arbitrary filtering on NoSQL databases

jnmt:
## Description

This PR adds cross-partition scan filtering (i.e., `ScanAll` with a where clause`) for non-JDBC databases. It filters results based on the specified arbitrary conditions in the ScalarDB layer. It also limits results in the ScalarDB layer if it's a cross-partition scan with filtering to count the results correctly.

Sorry for the big PR, but the basic concept is the same for all three databases. See inline comments for details.

## Related issues and/or PRs

N/A

## Changes made

- Add dynamic filtering in Cassandra, Cosmos DB, and DynamoDB.
  - Add filterable (and limitable) `Scanner` implementations.
  - Move the function to evaluate conditional expressions from `Snapshot` to `ScalarDbUtils`.
  - Add integration tests with conditions and limits for cross-partition scan.
- Fix warning for incorrectly specified projections when getting keys in results.

## Checklist

Dynamic filtering will also be applied for `Get`, `GetWithIndex`, `Scan`, and `ScanWithIndex` in this PR, but the builders to actually run those queries (and the integration tests) will be added in other PRs since this PR is already very big.

## Release notes

Added dynamic arbitrary filtering for non-JDBC databases.


core/src/integration-test/java/com/scalar/db/storage/cassandra/CassandraCrossPartitionScanIntegrationTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.scalar.db.api.DistributedStorageCrossPartitionScanIntegrationTestBase;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Properties;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+public class CassandraCrossPartitionScanIntegrationTest
+    extends DistributedStorageCrossPartitionScanIntegrationTestBase {
```
jnmt, line 11: I only added storage-level tests for each database since transaction-level tests would be tested enough using JDBC databases and they would increase CI time. But please feel free to comment if more tests should be added.

core/src/main/java/com/scalar/db/storage/cassandra/FilterableScannerImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+  private final Set<Conjunction> conjunctions;
+  private int left;
+
+  @SuppressFBWarnings("EI_EXPOSE_REP2")
+  public FilterableScannerImpl(
+      Scan scan, ResultSet resultSet, ResultInterpreter resultInterpreter) {
+    super(resultSet, resultInterpreter);
+    this.conjunctions = scan.getConjunctions();
+    this.left = scan.getLimit() > 0 ? scan.getLimit() : -1;
```
jnmt, line 29: We need to manage left items that can be returned before reaching the limit.

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
```diff
@@ -121,11 +119,15 @@ public Scanner scan(Scan scan) throws ExecutionException {
...
-        new ResultInterpreter(scan.getProjections(), metadataManager.getTableMetadata(scan)));
+    ResultInterpreter resultInterpreter =
+        new ResultInterpreter(scan.getProjections(), metadataManager.getTableMetadata(scan));
+    if (scan.getConjunctions().isEmpty()) {
+      return new ScannerImpl(handlers.select().handle(scan), resultInterpreter);
+    } else {
+      // Ignore limit to control it in FilterableScannerImpl
+      Scan scanAll = Scan.newBuilder(scan).limit(0).build();
+      return new FilterableScannerImpl(scan, handlers.select().handle(scanAll), resultInterpreter);
+    }
```
jnmt, line 130: I introduce a new filterable scanner implementation so that the regular ScanAll (w/o filtering) does not have to check the existence of filters every time when iterating. It's also nice to keep the logic simple.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageCrossPartitionScanIntegrationTestBase.java
```diff
@@ -560,7 +560,6 @@ private Scan prepareScanWithLike(boolean isLike, String pattern) {
         .table(CONDITION_TEST_TABLE)
         .all()
         .where(condition)
-        .ordering(Ordering.asc(PARTITION_KEY_NAME))
```
jnmt, line 563: I removed orderings (including at other places) to commonly use these tests in non-JDBC databases except for order-specifc tests.

jnmt:
Sorry, I faced an issue when using this with consensus commit. Let me investigate and maybe fix it.

core/src/integration-test/java/com/scalar/db/storage/cassandra/ConsensusCommitCrossPartitionScanIntegrationTestWithCassandra.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+public class ConsensusCommitCrossPartitionScanIntegrationTestWithCassandra
+    extends ConsensusCommitCrossPartitionScanIntegrationTestBase {
+
+  @Override
+  protected Properties getProps(String testName) {
+    Properties properties = CassandraEnv.getProperties(testName);
+    properties.setProperty(ConsensusCommitConfig.ISOLATION_LEVEL, "SERIALIZABLE");
```
jnmt, line 17: Even though serializable isolation is not recommended in non-JDBC databases, as we warn when using it, I think it's better to use serializable for testing purposes so as not to miss bugs regarding extra-read.

It might be better to use serializable for other consensus commit integration tests if no concerns. Any thoughts?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -427,10 +386,10 @@ void toSerializableWithExtraRead(DistributedStorage storage)
...
             try {
               Scan scan = entry.getKey();
-              // only get tx_id and tx_version columns because we use only them to compare
-              scan.clearProjections();
-              scan.withProjection(Attribute.ID).withProjection(Attribute.VERSION);
-              ScalarDbUtils.addProjectionsForKeys(scan, getTableMetadata(scan));
+              // For JDBC databases or non-JDBC databases without filtering, we only get tx_id,
+              // tx_version, and primary key columns because we use only them to compare. For
+              // non-JDBC databases with filtering, we also need filtering columns to compare.
+              projectRequiredColumns(storage, scan);
```
jnmt, line 392: This is the fix for a bug with the serializable consensus commit. We need non-primary key columns to perform filtering correctly in the ScalarDB layer if they are used in the conditions, even for the extra-read purpose.

jnmt:
Now ready for review! PTAL, when you get a chance since the issue was fixed in 9729ce8.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
komamitsu: [minor] Does it mean that the values of `scanSet` doesn't contain filtering columns in the case of JDBC (thanks to the condition pushdown?), while the values of `scanSet` contain filtering columns in the case of non-JDBC?

If so, I guess it would be a bit more helpful to understand if `Scan` tells whether filtering columns are contained with a method like `scan.containsFilteringColumnsInKey()` so that we can revise the condition in `projectRequiredColumns` like this?

```java
if (scan.containsFilteringColumnsInKey()) {
  scan.getConjunctions()
      .forEach(
          conjunction ->
              conjunction
   ...
```

core/src/main/java/com/scalar/db/storage/cosmos/FilterableScannerImpl.java
```diff
@@ -0,0 +1,99 @@
...
+        if (left == 0) {
+          return Optional.empty();
+        }
+        Optional<Result> result = getResultWithFiltering(currentPageRecords.next());
+        if (result.isPresent()) {
+          return result;
+        }
+      } else {
+        // Otherwise, advance to the next page
+        currentPageRecords = recordsPages.next().getResults().iterator();
```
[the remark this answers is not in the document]
jnmt: Good point. Thank you. Yeah, we don't have to call `next()` if there are no results left that can be returned. Moreover, we don't have to check `left` several times in `one()` since decreasing `left`

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: I might not get your point perfectly, but it doesn't mean whether the user's original `Scan` has filtering columns in projections or not. Do you think this condition `if (!(storage instanceof JdbcDatabase))` is hard to understand the background? For JDBC databases, we don't have to care about the existence of filtering columns in projections, so I didn't want to check it excessively.

By the way, thanks to your question, I might find an issue regarding projections...

komamitsu: Thanks for the reply!

What I thought was as follows:
- Why filtering columns were needed for non-JDBC (w/ filtering) ?
- The results of this scan are put to `currentReadMap` as a part of the key
- Then, the map is looked-up with the values of `scanSet` (`TransactionResult latestResult = currentReadMap.get(key)`)
- So, including filtering columns in projection is for consistency of the key types related to the `equals()`'s logic?

But I think I had the wrong idea... Can you elaborate a bit more why filtering columns are needed for projection in the case of non-JDBC with filtering?

jnmt: Ahh, the reason is not related to the key consistency of `currentReadMap` since the validation is always handled with the primary key basis.

We originally project columns except for `tx_id`, `tx_version`, and primary key columns for efficiency since they are enough for the re-scan for the validation phase (extra-read). We now want to filter results in the ScalarDB layer for non-JDBC databases. So, we need to additionally have filtering columns here since they are used even for the re-scan, the same as the first scan.

komamitsu: Yeah, I misread the code around Key class... Sorry for bothering you.

> We originally project columns except for tx_id, tx_version, and primary key columns

`s/except for/only for/` ?

But, I think I don't still get why columns for filtering need to be projected. My understanding is as follows:
- `Scan` instances stored in `scanSet` contain the original conditions
- `FilterableScannerImpl#getResultWithFiltering` filters records based on the conditions. Whether the columns for filtering are projected isn't related to the filtering. (The columns for filtering included in the conditions are automatically scanned at the underlying storage layer, right?)
- The columns for filtering of the result in `currentReadMap` are not used
- So, the columns for filtering don't need to be projected?

I think I have some wrong ideas of them. Can you point out it?

jnmt: > Yeah, I misread the code around Key class... Sorry for bothering you.

No no, it's OK. I think this part is so complicated and worth while to share the background.

> > We originally project columns except for tx_id, tx_version, and primary key columns
> 
> `s/except for/only for/` ?

Sorry, "only for".

> But, I think I don't still get why columns for filtering need to be projected. My understanding is as follows:
> 
> * `Scan` instances stored in `scanSet` contain the original conditions

Right. Precisely, the list of projections is basically cleared in `CrudHandler` for the sake of transaction processing. In that sense, it's a bit different from the original one specified by users.

> * `FilterableScannerImpl#getResultWithFiltering` filters records based on the conditions. Whether the columns for filtering are projected isn't related to the filtering. (The columns for filtering included in the conditions are automatically scanned at the underlying storage layer, right?)

Right. We always need columns specified in the where clause to filter results. (*)

> * The columns for filtering of the result in `currentReadMap` are not used

Right. `currentReadMap` is for the read set verification. So, primary key columns are enough, but, of course, the map should be based on the result of the scan with the filtering conditions in the current database state.

> * So, the columns for filtering don't need to be projected?

No. We used them to get the current results above.

Let's chat a bit if you have further questions : )

komamitsu: Thanks for the detailed answers!

I found my misunderstanding...

I thought `Cassandra.scan()` implicitly projects the columns for filtering when there are any conditions in the scan. What I expected was like this:

```java
--- a/core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
@@ -14,6 +14,7 @@ import com.scalar.db.api.Mutation;
 import com.scalar.db.api.Put;
 import com.scalar.db.api.Result;
 import com.scalar.db.api.Scan;
+import com.scalar.db.api.ScanBuilder.BuildableScanOrScanAllFromExisting;
 import com.scalar.db.api.Scanner;
 import com.scalar.db.common.AbstractDistributedStorage;
 import com.scalar.db.common.TableMetadataManager;
@@ -125,7 +126,9 @@ public class Cassandra extends AbstractDistributedStorage {
       return new ScannerImpl(handlers.select().handle(scan), resultInterpreter);
     } else {
       // Ignore limit to control it in FilterableScannerImpl
-      Scan scanAll = Scan.newBuilder(scan).limit(0).build();
+      BuildableScanOrScanAllFromExisting builder = Scan.newBuilder(scan);
+      withProjectionsForFilteringIfMissing(builder);
+      Scan scanAll = builder.limit(0).build();
       return new FilterableScannerImpl(scan, handlers.select().handle(scanAll), resultInterpreter);
     }
   }
```

BTW, assuming the above example works, it is a kind of provisioning the columns for filtering at storage layer while the change of this PR is provisioning the columns for filtering at an upper layer. Which do you think better?

jnmt: Right. As you thought, we needed to handle projections properly in the storage layer, and it would be better to hide such a storage matter from the upper layer. So, your question got to the core of the matter from the beginning. Thanks again. Fixed in 6b1e92a.

core/src/main/java/com/scalar/db/storage/cosmos/FilterableScannerImpl.java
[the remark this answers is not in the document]
jnmt: Fixed in new `FilterableScanner` of 6b1e92a.

core/src/main/java/com/scalar/db/storage/dynamo/FilterableQueryScanner.java
```diff
@@ -0,0 +1,44 @@
...
+import com.scalar.db.api.Scan.Conjunction;
+import com.scalar.db.api.Scanner;
+import com.scalar.db.storage.dynamo.request.PaginatedRequest;
+import com.scalar.db.util.ScalarDbUtils;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.Optional;
+import java.util.Set;
+import javax.annotation.Nonnull;
+
+public class FilterableQueryScanner extends QueryScanner implements Scanner {
```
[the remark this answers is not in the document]
jnmt: Fixed in new `FilterableScanner` of 6b1e92a.

jnmt:
@komamitsu Based on the feedback, I updated the following in 6b1e92a. PTAL again when you get a chance.

- Fixed a bug where both projections and conditions are given and the projections do not include the columns specified in conditions.
- Refactored filtering and limiting in the storage layer using a wrapper scanner.

brfrn169 (commented):
Did my first review for this PR and left several comments. Please take a look when you have time! I'll review it again later.

Also, I have a question: it seems that for Cassandra, Cosmos DB, and DynamoDB, we always execute dynamic filtering for the conjunctions. Are there any cases where we can push down the conjunctions to the databases?

Additionally, I think we could support conjunctions for the normal Scan and ScanWithIndex. What are your thoughts on this?

jnmt:
@brfrn169 Thanks for the comments. Those are very good points.

We can push down the conditions with some limitations. For example, Cosmos DB and DynamoDB do not support a LIKE operator. Cassandra has it, but only for index columns. Since all those databases scan all partitions in any case, I guess the performance benefit could not be so significant in spite of the effort for per-storage implementation. But, if it’s highly prioritized, we can improve one by one (in the future, IMHO).

For Scan and ScanWithIndex, my first thought was that supporting them was a bit complicated for the benefits and use cases, especially when it comes to considering ScalarDB SQL, as I mentioned in a Demo session. But after reconsideration, it’s very nice to have for non-JDBC database users.

Since we additionally need the following, and they will get a bit big changes to some extent, I think it’s good to separate PRs. What do you think?

- API changes so that we can call `where` after specifying keys
- JDBC database changes for better query building when mixing partition-scan conditions
- Non-JDBC database changes with dynamic filtering (maybe not so big deal)
- (ScalarDB SQL changes for a kind of optimization when mixing partition-scan conditions)

core/src/main/java/com/scalar/db/common/FilterableScanner.java
```diff
@@ -0,0 +1,78 @@
...
+  @Nonnull
+  public Iterator<Result> iterator() {
+    if (scannerIterator == null) {
+      scannerIterator = new ScannerIterator(this);
+    }
+    return scannerIterator;
+  }
+
+  @Override
+  public void close() {}
```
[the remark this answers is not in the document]
jnmt: Good catch! Thank you!

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
```diff
@@ -46,11 +48,9 @@ public class Cassandra extends AbstractDistributedStorage {
   public Cassandra(DatabaseConfig config) {
     super(config);
 
-    if (config.isCrossPartitionScanFilteringEnabled()
-        || config.isCrossPartitionScanOrderingEnabled()) {
+    if (config.isCrossPartitionScanOrderingEnabled()) {
       throw new IllegalArgumentException(
-          CoreError.CASSANDRA_CROSS_PARTITION_SCAN_WITH_FILTERING_OR_ORDERING_NOT_SUPPORTED
```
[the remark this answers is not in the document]
jnmt: Ahh, we do not maintain old ones for reference, do we? OK, then I will remove them.

core/src/main/java/com/scalar/db/storage/cassandra/ScannerImpl.java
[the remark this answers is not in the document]
jnmt: Sorry, my bad. Will fix.

core/src/main/java/com/scalar/db/common/ProjectedResult.java
```diff
@@ -0,0 +1,146 @@
...
+    return getKey(original.getClusteringKey());
+  }
+
+  private Optional<Key> getKey(Optional<Key> key) {
+    if (!key.isPresent()) {
+      return Optional.empty();
+    }
+    for (Value<?> value : key.get()) {
+      if (!containedColumnNames.contains(value.getName())) {
+        logger.warn("Full key doesn't seem to be projected into the result");
```
komamitsu, line 59: In this case, it means `Cassandra.scan()` or equivalents didn't setup proper projections? If so, it's not an expected situation and throwing an AssertionError or IllegalStateException would be good, I think.

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -194,13 +195,19 @@ private Scanner executeFullScan(ScanAll scan, TableMetadata tableMetadata) {
...
     ScanRequest.Builder builder = ScanRequest.builder().tableName(dynamoOperation.getTableName());
 
-    if (scan.getLimit() > 0) {
+    // Ignore limit to control it in FilterableQueryScanner if scan has condition(s)
+    if (scan.getLimit() > 0 && scan.getConjunctions().isEmpty()) {
       builder.limit(scan.getLimit());
     }
 
-    if (!scan.getProjections().isEmpty()) {
+    List<String> projections = new ArrayList<>(scan.getProjections());
```
komamitsu, line 203: Is it a bit simpler if these lines are moved to right before instantiating FilterableScanner below?

core/src/main/java/com/scalar/db/common/ProjectedResult.java
jnmt: Honestly, I just followed the same manner as `FilteredResult` in the consensus commit, but `IllegalStateException` with a message like `CoreError.COLUMN_NOT_FOUND` would be better as you mentioned.

@brfrn169 Do you remember why you chose a warning rather than an exception?

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
jnmt: Thanks, but we need to do this before creating `ScanRequest requestWrapper`, so it's not possible IIUC your point.

komamitsu: Yeah, I overlooked `builder` was updated in this block...

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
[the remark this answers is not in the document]
brfrn169: I think we can remove old error messages.

@feeblefakie @josh-wong What do you think?

docs/configurations.md
```diff
@@ -157,7 +157,7 @@ For details about client configurations, see the ScalarDB Cluster [client config
 
 ## Cross-partition scan configurations
 
-By enabling the cross-partition scan option below, the `Scan` operation can retrieve all records across partitions. In addition, you can specify arbitrary conditions and orderings in the cross-partition `Scan` operation by enabling `cross_partition_scan.filtering` and `cross_partition_scan.ordering`, respectively. Currently, the cross-partition scan with filtering and ordering is available only for JDBC databases. To enable filtering and ordering, `scalar.db.cross_partition_scan.enabled` must be set to `true`.
+By enabling the cross-partition scan option below, the `Scan` operation can retrieve all records across partitions. In addition, you can specify arbitrary conditions and orderings in the cross-partition `Scan` operation by enabling `cross_partition_scan.filtering` and `cross_partition_scan.ordering`, respectively. Currently, the cross-partition scan with ordering is available only for JDBC databases. To enable filtering and ordering, `scalar.db.cross_partition_scan.enabled` must be set to `true`.
```
josh-wong, line 160: Just one minor suggestion for clarity.

```suggestion
By enabling the cross-partition scan option as described below, the `Scan` operation can retrieve all records across partitions. In addition, you can specify arbitrary conditions and orderings in the cross-partition `Scan` operation by enabling `cross_partition_scan.filtering` and `cross_partition_scan.ordering`, respectively. Currently, the cross-partition scan with ordering is available only for JDBC databases. To enable filtering and ordering, `scalar.db.cross_partition_scan.enabled` must be set to `true`.
```

josh-wong (approved):
I've left one minor suggestion. Other than that, LGTM. Thank you!🙇🏻‍♂️

jnmt:
I will close this PR because we discussed and decided to first refactor `Scan.Conjunction`, `ScanBuilder.ConditionSetBuilder`, and so on so that we can support the where clause and dynamic filtering in `Get`. I will re-open once the refactoring PR is merged.

core/src/integration-test/java/com/scalar/db/storage/cassandra/ConsensusCommitCrossPartitionScanIntegrationTestWithCassandra.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  @Override
+  protected Map<String, String> getCreationOptions() {
+    return Collections.singletonMap(CassandraAdmin.REPLICATION_FACTOR, "1");
+  }
+
+  @Test
+  @Override
+  @Disabled
```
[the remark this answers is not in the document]
jnmt: Thank you! Fixed in 580fae5 and edd250c.

core/src/main/java/com/scalar/db/common/FilterableScanner.java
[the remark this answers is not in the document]
jnmt: Fixed in 580fae5.

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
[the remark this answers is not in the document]
jnmt: Fixed in 580fae5. Note that, as discussed with Toshi, we need to maintain deleted messages to avoid reusing already deleted message IDs somehow, especially the most recent ones.

core/src/main/java/com/scalar/db/storage/cosmos/Cosmos.java
```diff
@@ -46,11 +46,9 @@ public class Cosmos extends AbstractDistributedStorage {
   public Cosmos(DatabaseConfig databaseConfig) {
     super(databaseConfig);
 
-    if (databaseConfig.isCrossPartitionScanFilteringEnabled()
-        || databaseConfig.isCrossPartitionScanOrderingEnabled()) {
+    if (databaseConfig.isCrossPartitionScanOrderingEnabled()) {
       throw new IllegalArgumentException(
-          CoreError.COSMOS_CROSS_PARTITION_SCAN_WITH_FILTERING_OR_ORDERING_NOT_SUPPORTED
```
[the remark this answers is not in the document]
jnmt: Fixed in 580fae5.

core/src/main/java/com/scalar/db/storage/dynamo/Dynamo.java
```diff
@@ -52,11 +52,9 @@ public class Dynamo extends AbstractDistributedStorage {
   public Dynamo(DatabaseConfig databaseConfig) {
     super(databaseConfig);
 
-    if (databaseConfig.isCrossPartitionScanFilteringEnabled()
-        || databaseConfig.isCrossPartitionScanOrderingEnabled()) {
+    if (databaseConfig.isCrossPartitionScanOrderingEnabled()) {
       throw new IllegalArgumentException(
-          CoreError.DYNAMO_CROSS_PARTITION_SCAN_WITH_FILTERING_OR_ORDERING_NOT_SUPPORTED
```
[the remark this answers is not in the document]
jnmt: Fixed in 580fae5.

core/src/main/java/com/scalar/db/storage/cassandra/ScannerImpl.java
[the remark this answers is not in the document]
jnmt: Fixed in 580fae5.

core/src/main/java/com/scalar/db/storage/cosmos/ScannerImpl.java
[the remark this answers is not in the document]
jnmt: Fixed in 580fae5.

core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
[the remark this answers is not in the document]
jnmt: Fixed in 580fae5.

core/src/main/java/com/scalar/db/common/ProjectedResult.java
jnmt: @komamitsu Thank you for the suggestion. I fixed it, including other similar classes in this PR based on the discussion with @brfrn169.

core/src/integration-test/java/com/scalar/db/storage/cassandra/ConsensusCommitCrossPartitionScanIntegrationTestWithCassandra.java
brfrn169: > Even though serializable isolation is not recommended in non-JDBC databases

Does this mean that we don't recommend using `SERIALIZABLE` with the cross-partition scan in non-JDBC databases?

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
```diff
@@ -100,19 +98,34 @@ public Optional<Result> get(Get get) throws ExecutionException {
...
-    Row next = resultSet.one();
-    if (next != null) {
-      throw new IllegalArgumentException(
-          CoreError.GET_OPERATION_USED_FOR_NON_EXACT_MATCH_SELECTION.buildMessage(get));
+    Scanner scanner = null;
+    try {
+      if (get.getConjunctions().isEmpty()) {
+        scanner = getInternal(get);
+      } else {
+        scanner = new FilterableScanner(get, getInternal(copyAndPrepareForDynamicFiltering(get)));
```
brfrn169, line 106: Actually since we already copied the get object here (https://github.com/scalar-labs/scalardb/pull/1682/files#diff-660ca76c9b64d94000e0854e2a18fdc5e6a69c028a74b8cacfa4ef733d4604c2L100), I don't think we need to copy the get object here with `copyAndPrepareForDynamicFiltering()` again.

brfrn169 (commented):

Torch3333 (approved):
LGTM, thank you!

core/src/integration-test/java/com/scalar/db/storage/cassandra/ConsensusCommitCrossPartitionScanIntegrationTestWithCassandra.java
jnmt: Yes. I mean this warning: https://scalardb.scalar-labs.com/docs/latest/configurations/#cross-partition-scan-configurations. Even so, I thought these tests should be done with extra-read just in case.

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
jnmt: Thank you for the comment. I wanted to pass the original `get` for `FilterableScanner` and the different modified `get` for `getInternal()` since we need the original limit value and projections in `FilterableScanner`. It might be possible to avoid copying the whole `get` operation somehow, but I think the current one is not so bad from the readability perspective. What do you think?

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/common/FilterableScanner.java
```diff
@@ -0,0 +1,87 @@
...
+
+@NotThreadSafe
+public class FilterableScanner implements Scanner {
+
+  private final Scanner scanner;
+  private final List<String> projections;
+  private final Set<Conjunction> conjunctions;
+  private int left;
+
+  private ScannerIterator scannerIterator;
```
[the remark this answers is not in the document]
jnmt: Fixed in 74ab4fb. Thank you! I didn't know the `@LazyInit` annotation :P

core/src/main/java/com/scalar/db/storage/cassandra/Cassandra.java
brfrn169: That makes sense. Thanks.

brfrn169 (approved):
LGTM! Thank you!

PR #1697, opened 2024-04-23
Add Insert, Upsert, and Update to TransactionCrudOperable

brfrn169:
## Description

This PR adds the Insert, Upsert, and Update operations to `TransactionCrudOperable`. For the specification of these operations, see the Javadoc of `TransactionCrudOperable`.

## Related issues and/or PRs

N/A

## Changes made

- Added the `Insert`, `Upsert`, and `Update` classes as subclasses of `Mutate`.
- Added the `insert()`, `upsert()`, and `update()` methods to `TransactionCrudOperable`.
- Deprecated the `put()` methods and some other methods in `TransactionCrudOperable`.
- Implemented the `insert()`, `upsert()`, and `update()` methods for Consensus Comment and JDBC transactions.

## Checklist

- Documentation will be added in a separate PR.

## Release notes

Added the Insert, Upsert, and Update operations to the transactional API.

### DIFF
--- a/core/src/main/java/com/scalar/db/api/ConditionBuilder.java
+++ b/core/src/main/java/com/scalar/db/api/ConditionBuilder.java
@@ -84,6 +84,26 @@ public class ConditionBuilder {
   }
 
   /**
+   * Returns a builder object for a UpdateIf condition.
+   *
+   * @param conditionalExpression a condition expression for a UpdateIf condition
+   * @return a builder object
+   */
+  public static UpdateIfBuilder updateIf(ConditionalExpression conditionalExpression) {
+    return new UpdateIfBuilder(conditionalExpression);
+  }
+
+  /**
+   * Creates a UpdateIf condition with the specified conditional expressions.
+   *
+   * @param conditionalExpressions condition expressions for a UpdateIf condition
+   * @return a UpdateIf condition
+   */
+  public static UpdateIf updateIf(List<ConditionalExpression> conditionalExpressions) {
+    return new UpdateIf(conditionalExpressions);
+  }
+
+  /**
    * Builds a conditional expression with the specified column and operator.
    *
    * <p>This method is primarily for internal use. Breaking changes can and will be introduced to
@@ -845,6 +865,47 @@ public class ConditionBuilder {
           || conditionalExpression.getOperator().equals(Operator.NOT_LIKE)) {
         throw new IllegalArgumentException(
             CoreError.CONDITION_BUILD_ERROR_CONDITION_NOT_ALLOWED_FOR_DELETE_IF.buildMessage(
+                conditionalExpression));
+      }
+    }
+  }
+
+  public static class UpdateIfBuilder {
+
+    private final List<ConditionalExpression> conditionalExpressions;
+
+    private UpdateIfBuilder(ConditionalExpression conditionalExpression) {
+      check(conditionalExpression);
+      conditionalExpressions = new ArrayList<>();
+      conditionalExpressions.add(conditionalExpression);
+    }
+
+    /**
+     * Adds a condition for a UpdateIf condition.
+     *
+     * @param conditionalExpression a condition for a UpdateIf condition
+     * @return a builder object
+     */
+    public UpdateIfBuilder and(ConditionalExpression conditionalExpression) {
+      check(conditionalExpression);
+      conditionalExpressions.add(conditionalExpression);
+      return this;
+    }
+
+    /**
+     * Builds a UpdateIf condition with the specified conditional expressions.
+     *
+     * @return a UpdateIf condition
+     */
+    public UpdateIf build() {
+      return new UpdateIf(conditionalExpressions);
+    }
+
+    private void check(ConditionalExpression conditionalExpression) {
+      if (conditionalExpression.getOperator().equals(Operator.LIKE)
+          || conditionalExpression.getOperator().equals(Operator.NOT_LIKE)) {
+        throw new IllegalArgumentException(
+            CoreError.CONDITION_BUILD_ERROR_CONDITION_NOT_ALLOWED_FOR_UPDATE_IF.buildMessage(
                 conditionalExpression));
       }
     }
--- a/core/src/main/java/com/scalar/db/api/Delete.java
+++ b/core/src/main/java/com/scalar/db/api/Delete.java
@@ -6,10 +6,12 @@ import com.google.common.base.MoreObjects;
 import com.scalar.db.api.DeleteBuilder.BuildableFromExisting;
 import com.scalar.db.api.DeleteBuilder.Namespace;
 import com.scalar.db.io.Key;
+import java.util.Optional;
+import javax.annotation.Nonnull;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
- * A command to delete an entry from {@link DistributedStorage}.
+ * A command to delete an entry from a storage.
  *
  * @author Hiroyuki Yamada
  */
@@ -120,6 +122,12 @@ public class Delete extends Mutation {
   @Deprecated
   public Delete withCondition(MutationCondition condition) {
     return (Delete) super.withCondition(condition);
+  }
+
+  @Nonnull
+  @Override
+  public Optional<MutationCondition> getCondition() {
+    return super.getCondition();
   }
 
   /**
--- a/core/src/main/java/com/scalar/db/api/Insert.java
+++ b/core/src/main/java/com/scalar/db/api/Insert.java
@@ -1,0 +1,131 @@
+package com.scalar.db.api;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableMap;
+import com.scalar.db.io.Column;
+import com.scalar.db.io.Key;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/** A command to insert an entry to a storage. */
+@NotThreadSafe
+public class Insert extends Mutation {
+
+  private final Map<String, Column<?>> columns;
+
+  Insert(
+      @Nullable String namespace,
+      String tableName,
+      Key partitionKey,
+      @Nullable Key clusteringKey,
+      Map<String, Column<?>> columns) {
+    super(namespace, tableName, partitionKey, clusteringKey, null);
+    this.columns = ImmutableMap.copyOf(columns);
+  }
+
+  public Map<String, Column<?>> getColumns() {
+    return columns;
+  }
+
+  /** @deprecated As of release 3.13.0. Will be removed in release 5.0.0. */
+  @Deprecated
+  @Nonnull
+  @Override
+  public Optional<MutationCondition> getCondition() {
+    throw new UnsupportedOperationException();
+  }
+
+  /** @deprecated As of release 3.6.0. Will be removed in release 5.0.0. */
+  @Deprecated
+  @Override
+  public Mutation withCondition(MutationCondition condition) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public Consistency getConsistency() {
+    throw new UnsupportedOperationException();
+  }
+
+  /** @deprecated As of release 3.6.0. Will be removed in release 5.0.0 */
+  @Deprecated
+  public Operation withConsistency(Consistency consistency) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public void accept(OperationVisitor v) {
+    v.visit(this);
+  }
+
+  /**
+   * Indicates whether some other object is "equal to" this object. The other object is considered
+   * equal if:
+   *
+   * <ul>
+   *   <li>both super class instances are equal and
+   *   <li>it is also an {@code Insert} and
+   *   <li>both instances have the same values
+   * </ul>
+   *
+   * @param o an object to be tested for equality
+   * @return {@code true} if the other object is "equal to" this object otherwise {@code false}
+   */
+  @Override
+  public boolean equals(Object o) {
+    if (!super.equals(o)) {
+      return false;
+    }
+    if (o == this) {
+      return true;
+    }
+    if (!(o instanceof Insert)) {
+      return false;
+    }
+    Insert other = (Insert) o;
+    return columns.equals(other.columns);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(super.hashCode(), columns);
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("namespace", forNamespace())
+        .add("table", forTable())
+        .add("partitionKey", getPartitionKey())
+        .add("clusteringKey", getClusteringKey())
+        .add("columns", getColumns())
+        .toString();
+  }
+
+  /**
+   * Build a {@code Insert} operation using a builder.
+   *
+   * @return a {@code Insert} operation builder
+   */
+  public static InsertBuilder.Namespace newBuilder() {
+    return new InsertBuilder.Namespace();
+  }
+
+  /**
+   * Build a {@code Insert} operation from an existing {@code Insert} object using a builder. The
+   * builder will be parametrized by default with all the existing {@code Insert} object attributes.
+   *
+   * @param insert an existing {@code Insert} operation
+   * @return a {@code Insert} operation builder
+   */
+  public static InsertBuilder.BuildableFromExisting newBuilder(Insert insert) {
+    checkNotNull(insert);
+    return new InsertBuilder.BuildableFromExisting(insert);
+  }
+}
--- a/core/src/main/java/com/scalar/db/api/InsertBuilder.java
+++ b/core/src/main/java/com/scalar/db/api/InsertBuilder.java
@@ -1,0 +1,345 @@
+package com.scalar.db.api;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.scalar.db.api.OperationBuilder.ClearClusteringKey;
+import com.scalar.db.api.OperationBuilder.ClearNamespace;
+import com.scalar.db.api.OperationBuilder.ClearValues;
+import com.scalar.db.api.OperationBuilder.ClusteringKey;
+import com.scalar.db.api.OperationBuilder.PartitionKeyBuilder;
+import com.scalar.db.api.OperationBuilder.TableBuilder;
+import com.scalar.db.api.OperationBuilder.Values;
+import com.scalar.db.io.BigIntColumn;
+import com.scalar.db.io.BlobColumn;
+import com.scalar.db.io.BooleanColumn;
+import com.scalar.db.io.Column;
+import com.scalar.db.io.DoubleColumn;
+import com.scalar.db.io.FloatColumn;
+import com.scalar.db.io.IntColumn;
+import com.scalar.db.io.Key;
+import com.scalar.db.io.TextColumn;
+import java.nio.ByteBuffer;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+public class InsertBuilder {
+
+  public static class Namespace
+      implements OperationBuilder.Namespace<Table>, OperationBuilder.Table<PartitionKey> {
+
+    Namespace() {}
+
+    @Override
+    public Table namespace(String namespaceName) {
+      checkNotNull(namespaceName);
+      return new Table(namespaceName);
+    }
+
+    @Override
+    public PartitionKey table(String tableName) {
+      checkNotNull(tableName);
+      return new PartitionKey(null, tableName);
+    }
+  }
+
+  public static class Table extends TableBuilder<PartitionKey> {
+
+    private Table(String namespaceName) {
+      super(namespaceName);
+    }
+
+    @Override
+    public PartitionKey table(String tableName) {
+      checkNotNull(tableName);
+      return new PartitionKey(namespace, tableName);
+    }
+  }
+
+  public static class PartitionKey extends PartitionKeyBuilder<Buildable> {
+
+    private PartitionKey(@Nullable String namespaceName, String tableName) {
+      super(namespaceName, tableName);
+    }
+
+    @Override
+    public Buildable partitionKey(Key partitionKey) {
+      checkNotNull(partitionKey);
+      return new Buildable(namespaceName, tableName, partitionKey);
+    }
+  }
+
+  public static class Buildable extends OperationBuilder.Buildable<Insert>
+      implements ClusteringKey<Buildable>, Values<Buildable> {
+    final Map<String, Column<?>> columns = new LinkedHashMap<>();
+    @Nullable Key clusteringKey;
+
+    private Buildable(@Nullable String namespace, String table, Key partitionKey) {
+      super(namespace, table, partitionKey);
+    }
+
+    @Override
+    public Buildable clusteringKey(Key clusteringKey) {
+      checkNotNull(clusteringKey);
+      this.clusteringKey = clusteringKey;
+      return this;
+    }
+
+    @Override
+    public Buildable booleanValue(String columnName, boolean value) {
+      columns.put(columnName, BooleanColumn.of(columnName, value));
+      return this;
+    }
+
+    @Override
+    public Buildable booleanValue(String columnName, @Nullable Boolean value) {
+      if (value != null) {
+        return booleanValue(columnName, value.booleanValue());
+      }
+      columns.put(columnName, BooleanColumn.ofNull(columnName));
+      return this;
+    }
+
+    @Override
+    public Buildable intValue(String columnName, int value) {
+      columns.put(columnName, IntColumn.of(columnName, value));
+      return this;
+    }
+
+    @Override
+    public Buildable intValue(String columnName, @Nullable Integer value) {
+      if (value != null) {
+        return intValue(columnName, value.intValue());
+      }
+      columns.put(columnName, IntColumn.ofNull(columnName));
+      return this;
+    }
+
+    @Override
+    public Buildable bigIntValue(String columnName, long value) {
+      columns.put(columnName, BigIntColumn.of(columnName, value));
+      return this;
+    }
+
+    @Override
+    public Buildable bigIntValue(String columnName, @Nullable Long value) {
+      if (value != null) {
+        return bigIntValue(columnName, value.longValue());
+      }
+      columns.put(columnName, BigIntColumn.ofNull(columnName));
+      return this;
+    }
+
+    @Override
+    public Buildable floatValue(String columnName, float value) {
+      columns.put(columnName, FloatColumn.of(columnName, value));
+      return this;
+    }
+
+    @Override
+    public Buildable floatValue(String columnName, @Nullable Float value) {
+      if (value != null) {
+        return floatValue(columnName, value.floatValue());
+      }
+      columns.put(columnName, FloatColumn.ofNull(columnName));
+      return this;
+    }
+
+    @Override
+    public Buildable doubleValue(String columnName, double value) {
+      columns.put(columnName, DoubleColumn.of(columnName, value));
+      return this;
+    }
+
+    @Override
+    public Buildable doubleValue(String columnName, @Nullable Double value) {
+      if (value != null) {
+        return doubleValue(columnName, value.doubleValue());
+      }
+      columns.put(columnName, DoubleColumn.ofNull(columnName));
+      return this;
+    }
+
+    @Override
+    public Buildable textValue(String columnName, @Nullable String value) {
+      columns.put(columnName, TextColumn.of(columnName, value));
+      return this;
+    }
+
+    @Override
+    public Buildable blobValue(String columnName, @Nullable byte[] value) {
+      columns.put(columnName, BlobColumn.of(columnName, value));
+      return this;
+    }
+
+    @Override
+    public Buildable blobValue(String columnName, @Nullable ByteBuffer value) {
+      columns.put(columnName, BlobColumn.of(columnName, value));
+      return this;
+    }
+
+    @Override
+    public Buildable value(Column<?> column) {
+      columns.put(column.getName(), column);
+      return this;
+    }
+
+    @Override
+    public Insert build() {
+      return new Insert(namespaceName, tableName, partitionKey, clusteringKey, columns);
+    }
+  }
+
+  public static class BuildableFromExisting extends Buildable
+      implements OperationBuilder.Namespace<BuildableFromExisting>,
+          OperationBuilder.Table<BuildableFromExisting>,
+          OperationBuilder.PartitionKey<BuildableFromExisting>,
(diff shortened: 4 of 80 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -232,4 +238,47 @@ public static boolean isAfterImageColumn(String columnName, TableMetadata tableM
...
+    if (update.getCondition().isPresent()) {
+      update
+          .getCondition()
+          .ifPresent(c -> buildable.condition(ConditionBuilder.putIf(c.getExpressions())));
+    } else {
+      buildable.condition(ConditionBuilder.putIfExists());
+    }
+    buildable.enableImplicitPreRead();
+    return buildable.build();
+  }
```
brfrn169, line 283: For Update in Consensus Commit, we enable the implicit pre-read, and if no condition is specified in the Update object, we set `putIfExists` for the put operation, or if a condition is specified, we set the same condition for the put operation.

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +47,132 @@ public interface TransactionCrudOperable {
...
    * @throws CrudException if the transaction CRUD operation fails due to transient or nontransient
    *     faults. You can try retrying the transaction from the beginning, but the transaction may
    *     still fail if the cause is nontranient
-   * @throws UnsatisfiedConditionException if the mutation condition is not satisfied
+   * @throws UnsatisfiedConditionException if a condition is specified, and if the condition is not
+   *     satisfied or the entry does not exist
+   * @deprecated As of release 3.13.0. Will be removed in release 5.0.0.
    */
+  @Deprecated
   void put(Put put) throws CrudConflictException, CrudException, UnsatisfiedConditionException;
```
brfrn169, line 64: Deprecated the put operation in the transactional API.

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +47,132 @@ public interface TransactionCrudOperable {
...
    *     faults. You can try retrying the transaction from the beginning, but the transaction may
    *     still fail if the cause is nontranient
-   * @throws UnsatisfiedConditionException if the mutation condition is not satisfied
+   * @throws UnsatisfiedConditionException if a condition is specified, and if the condition is not
+   *     satisfied or the entry does not exist
+   * @deprecated As of release 3.13.0. Will be removed in release 5.0.0.
    */
+  @Deprecated
   void put(List<Put> puts)
       throws CrudConflictException, CrudException, UnsatisfiedConditionException;
```
brfrn169, line 84: Ditto.

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +47,132 @@ public interface TransactionCrudOperable {
...
    *     faults. You can try retrying the transaction from the beginning, but the transaction may
    *     still fail if the cause is nontranient
-   * @throws UnsatisfiedConditionException if the mutation condition is not satisfied
+   * @throws UnsatisfiedConditionException if a condition is specified, and if the condition is not
+   *     satisfied or the entry does not exist
+   * @deprecated As of release 3.13.0. Will be removed in release 5.0.0.
    */
+  @Deprecated
   void delete(List<Delete> deletes)
       throws CrudConflictException, CrudException, UnsatisfiedConditionException;
```
brfrn169, line 121: Also, deprecated this method.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -124,11 +137,101 @@ public void delete(Delete delete) throws CrudException {
...
+
+    PutBuilder.Buildable buildable =
+        Put.newBuilder()
+            .namespace(insert.forNamespace().orElse(null))
+            .table(insert.forTable().orElse(null))
+            .partitionKey(insert.getPartitionKey());
+    insert.getClusteringKey().ifPresent(buildable::clusteringKey);
+    insert.getColumns().values().forEach(buildable::value);
+    buildable.condition(ConditionBuilder.putIfNotExists());
+    Put put = buildable.build();
```
brfrn169, line 160: For Insert in JDBC transactions, we convert the Insert object to the Put object with `putIfNotExists()`.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -124,11 +137,101 @@ public void delete(Delete delete) throws CrudException {
...
+    upsert = copyAndSetTargetToIfNot(upsert);
+
+    PutBuilder.Buildable buildable =
+        Put.newBuilder()
+            .namespace(upsert.forNamespace().orElse(null))
+            .table(upsert.forTable().orElse(null))
+            .partitionKey(upsert.getPartitionKey());
+    upsert.getClusteringKey().ifPresent(buildable::clusteringKey);
+    upsert.getColumns().values().forEach(buildable::value);
+    Put put = buildable.build();
```
brfrn169, line 187: For Upsert in JDBC transactions, we convert the Upsert object to the Put object.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -124,11 +137,101 @@ public void delete(Delete delete) throws CrudException {
...
+            .table(update.forTable().orElse(null))
+            .partitionKey(update.getPartitionKey());
+    update.getClusteringKey().ifPresent(buildable::clusteringKey);
+    update.getColumns().values().forEach(buildable::value);
+    if (update.getCondition().isPresent()) {
+      buildable.condition(ConditionBuilder.putIf(update.getCondition().get().getExpressions()));
+    } else {
+      buildable.condition(ConditionBuilder.putIfExists());
+    }
+    Put put = buildable.build();
```
brfrn169, line 216: For Update in JDBC transactions, we convert the Update object to the Put object. And if no condition is specified in the Update object, we set `putIfExists()` for the put operation, or if a condition is specified, we set the same condition for the put operation.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -127,14 +139,63 @@ public void delete(List<Delete> deletes) throws CrudException {
...
+  @Override
+  public void update(Update update) throws CrudException {
+    update = copyAndSetTargetToIfNot(update);
+    ScalarDbUtils.checkUpdate(update);
+    Put put = ConsensusCommitUtils.createPutForUpdate(update);
+    checkMutation(put);
+    try {
+      crud.put(put);
+    } catch (UnsatisfiedConditionException e) {
+      if (update.getCondition().isPresent()) {
```
komamitsu, line 172: If the target record specified with the keys doesn't exist even with any condition, I think `RecordNotFoundException` might be more consistent. What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
komamitsu: > if a condition is specified, we set the same condition for the Put object.

Even w/o `putIfExists`, an update operation [fails](https://github.com/scalar-labs/scalardb/blob/c49d06482361c8e98e0ad5b6472319e066968fd9/core/src/main/java/com/scalar/db/transaction/consensuscommit/MutationConditionsValidator.java#L50) if the target record doesn't exist, right?

core/src/main/java/com/scalar/db/api/InsertBuilder.java
```diff
@@ -0,0 +1,345 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+public class InsertBuilder {
+
+  public static class Namespace
+      implements OperationBuilder.Namespace<Table>, OperationBuilder.Table<PartitionKey> {
+
+    Namespace() {}
```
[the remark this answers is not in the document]
brfrn169: I think this prevents this class from being instantiated outside of the package. What do you think?

core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
```diff
@@ -349,6 +354,13 @@ public void check(List<? extends Mutation> mutations) throws ExecutionException
     }
   }
 
+  private void checkMutationType(Mutation mutation) {
+    if (!(mutation instanceof Put) && !(mutation instanceof Delete)) {
+      throw new IllegalArgumentException(
+          CoreError.OPERATION_CHECK_ERROR_UNSUPPORTED_MUTATION_TYPE.buildMessage(mutation));
```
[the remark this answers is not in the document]
brfrn169: Fixed in 4ecdc2bc0a101db3e3d541bef0be01df06751ddd. Thanks.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
brfrn169: Actually, in Consensus Commit, we can identify whether the condition is not satisfied or the record does not exist if some condition is specified. However, in JDBC transactions, we cannot. This is because for Update with a condition, JDBC transactions execute a `Put` operation with the same condition of the `Update` operation. In this case, `NoMutationExecutionException` is thrown in both cases if the condition is not satisfied or the record does not exist. For more details, please see `JdbcTransaction.java`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
brfrn169: Yes, that's correct. If a condition is specified, we set the same condition (`putIf`) for the Put object. In this case, if the target record doesn't exist, the update operation fails.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
komamitsu: Thanks for providing the context. Understood.

It seems there is a bit of inconsistency in that this method doesn't throw `RecordNotFoundException` even if the target record doesn't exist with a condition, while the method throws `RecordNotFoundException` without a condition. I guess it might be confusing and complicate error handling. I'm curious if there is a specific benefit to introduce `RecordNotFoundException` beyond the inconsistency. What do you think?

brfrn169: > I'm curious if there is a specific benefit to introduce RecordNotFoundException beyond the inconsistency.

Instead of focusing on a specific benefit, consider what exception should be thrown when users don't specify a condition for Update and the target record doesn't exist. Should we also throw `UnsatisfiedConditionException` in such cases?

I think there are three situations throwing an exception:

1. When a condition is specified for Update:
    1. The target record doesn't exist
    2. The condition isn't satisfied
2. When no condition is specified:
    1. The target record doesn't exist

However, we can't differentiate between whether the condition isn't satisfied (1-i) or the target record doesn't exist (1-ii) when a condition is specified for Update. Therefore, we throw `UnsatisfiedConditionException` in those cases.

The problem is what to throw in situation 2-i. Currently, the code throws `RecordNotFoundException`, but honestly I didn't want to introduce a new exception. Do you have any ideas on this?

komamitsu: I think we have some options as follows. Let me order them in my preference:

1. This method always throws `RecordNotFoundException` whenever the target record doesn't exist even with a condition
2. This method always throws `UnsatisfiedConditionException` for simplicity and consistency. Considering only the case that uses no condition, `RecordNotFoundException` would be useful. But with a condition, users need to check the existence of the target record by themselves after catching `UnsatisfiedConditionException`. So, consistent error handling by checking the target record existence by users might be better. `RecordNotFoundException` is removed
3. The current behavior in this PR

For 1, as you explained, it doesn't work considering JDBC Transaction. So, option#2? I don't think `RecordNotFoundException` is so helpful for users, although I'm missing something.

brfrn169: I'm concerned that users might be confused if they receive an `UnsatisfiedConditionException` even when they don't specify any condition with the Update operation. That's why I introduced `RecordNotFoundException`. Considering this, I'm not sure if option#2 is a good idea. What do you think?

brfrn169: By the way, when users specify a condition with the Put or Delete operation, `UnsatisfiedConditionException` is thrown when the condition is not satisfied or the target record doesn't exist. So this existing behavior is the same when a condition is specified for Update.

komamitsu: > I'm concerned that users might be confused if they receive an UnsatisfiedConditionException even when they don't specify any condition with the Update operation

Good point. Indeed, the name of exception sounds a bit confusing. Always throwing `RecordNotFoundException` whose description is like "The target record is not found. This exception is also thrown when not satisfying a given condition" in the option 2 might be better.

> So this existing behavior is the same when a condition is specified for Update.

Yeah, my understanding aligns with this.

I think it's important to prevent users from being surprised with the different exceptions (`RecordNotFoundException` and `UnsatisfiedConditionException`) when the target record doesn't exist, and option 2 would be okay. So throwing `RecordSatisfyingConditionNotFoundException` when a condition is specified would be another option? This is almost same as the current PR's behavior.

1. When a condition is specified for Update:
  i. The target record doesn't exist ===> `RecordSatisfyingConditionNotFoundException`
  ii. The condition isn't satisfied ===> `RecordSatisfyingConditionNotFoundException`
2. When no condition is specified:
  i. The target record doesn't exist ===> `RecordNotFoundException`

I hope users think `RecordSatisfyingConditionNotFoundException` is a kind of extension of `RecordNotFoundException`, at least they're in the same category.

brfrn169: Umm, but, throwing `RecordSatisfyingConditionNotFoundException` in Update is not consistent with Put and Delete? As I mentioned, in Put and Delete, `UnsatisfiedConditionException` is thrown when the condition is not satisfied or the target record doesn't exist.

The current situation is as follows:

For Put or Delete:

1. When a condition is specified for Put or Delete:
  i. The target record doesn't exist ===> `UnsatisfiedConditionException`
  ii. The condition isn't satisfied ===> `UnsatisfiedConditionException`
2. When no condition is specified ===> No exception is thrown in any cases

For Update:

1. When a condition is specified for Update:
  i. The target record doesn't exist ===> `UnsatisfiedConditionException`
  ii. The condition isn't satisfied ===> `UnsatisfiedConditionException`
2. When no condition is specified:
  i. The target record doesn't exist ===> `RecordNotFoundException`

What do you think?

komamitsu: > throwing RecordSatisfyingConditionNotFoundException in Update is not consistent with Put and Delete?

Ah, sorry. I didn't consider the consistency between Put (and Delete) and Update since Update is newly added. If we need to keep the consistency, yeah, `UnsatisfiedConditionException` still needs to be used?

Also, I thought the existing Put and Delete also throw `UnsatisfiedConditionException` when no condition is specified. But it was the wrong idea. The original option 2 which always throws `UnsatisfiedConditionException` won't be chosen as far as the consistency between Put and Update needs to be kept.

So, the room to decide is whether to throw `RecordNotFoundException` from Update when no condition is specified, I guess. It seems Update can follow the behavior of Put and Delete without throwing the new exception and I didn't get a strong reason to introduce the new one. But I don't have a strong opinion on it.

brfrn169: I think one possible design is as follows:

In the current design in this PR, we can only specify `UpdateIf` for the Update operation, but I think we can also introduce `UpdateIfExists`. In this case, the exceptions are thrown as follows:

1. When `UpdateIf` is specified for Update:
  i. If the target record doesn't exist ===> `UnsatisfiedConditionException`
  ii. If the condition isn't satisfied ===> `UnsatisfiedConditionException`
2. When `UpdateIfExists` is specified for Update:
  i. If the target record doesn't exist (and the condition isn't satisfied) ===> `UnsatisfiedConditionException`
3. When no condition is specified ===> No exceptions are thrown in any case

With this approach, we don't need to introduce any new exceptions. What do you think?

BTW, the accurate scenarios for Put and Delete are as follows:

For Put:

1. When `PutIf` is specified for Put:
  i. If the target record doesn't exist ===> `UnsatisfiedConditionException`
  ii. If the condition isn't satisfied ===> `UnsatisfiedConditionException`
2. When `PutIfExists` is specified for Put:
  i. If the target record doesn't exist (and the condition isn't satisfied) ===> `UnsatisfiedConditionException`
3. When `PutIfNotExists` is specified for Put:
  i. If the target record exist (and the condition isn't satisfied) ===> `UnsatisfiedConditionException`
4. When no condition is specified ===> No exceptions are thrown in any case

For Delete:

1. When `DeleteIf` is specified for Delete:
  i. If the target record doesn't exist ===> `UnsatisfiedConditionException`
  ii. If the condition isn't satisfied ===> `UnsatisfiedConditionException`
2. When `DeleteIfExists` is specified for Delete:
  i. If the target record doesn't exist (and the condition isn't satisfied) ===> `UnsatisfiedConditionException`
3. When no condition is specified ===> No exceptions are thrown in any case

komamitsu: > I think we can also introduce UpdateIfExists

It seems reasonable and consistent to me! With this explicit condition, throwing `UnsatisfiedConditionException` won't surprise users.

Also, thank you for letting me know the detailed conditions behavior of Put and Delete. I still had a few wrong ideas in the previous comment...
I think introducing `UpdateIf` condition makes Update more consistent with Put and Delete.

core/src/main/java/com/scalar/db/api/Delete.java
```diff
@@ -122,6 +124,12 @@ public Delete withCondition(MutationCondition condition) {
     return (Delete) super.withCondition(condition);
   }
 
+  @Nonnull
+  @Override
+  public Optional<MutationCondition> getCondition() {
```
komamitsu, line 129: [trivial] This can be removed?

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +46,129 @@ public interface TransactionCrudOperable {
...
    *
    * @param put a {@code Put} command
    * @throws CrudConflictException if the transaction CRUD operation fails due to transient faults
    *     (e.g., a conflict error). You can retry the transaction from the beginning
    * @throws CrudException if the transaction CRUD operation fails due to transient or nontransient
    *     faults. You can try retrying the transaction from the beginning, but the transaction may
    *     still fail if the cause is nontranient
-   * @throws UnsatisfiedConditionException if the mutation condition is not satisfied
+   * @throws UnsatisfiedConditionException if a condition is specified, and if the condition is not
+   *     satisfied or the entry does not exist
```
komamitsu, line 59: Same as above

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
```diff
@@ -127,11 +134,28 @@ public synchronized void delete(Delete delete) throws CrudException {
...
+    /** @deprecated As of release 3.13.0. Will be removed in release 5.0.0. */
+    @Deprecated
     @Override
     public synchronized void delete(List<Delete> deletes) throws CrudException {
       super.delete(deletes);
     }
 
+    @Override
+    public void insert(Insert insert) throws CrudException {
+      super.insert(insert);
```
komamitsu, line 146: [trivial] This can be removed?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -232,4 +242,61 @@ public static boolean isAfterImageColumn(String columnName, TableMetadata tableM
...
+      if (update.getCondition().get() instanceof UpdateIf) {
+        update
+            .getCondition()
+            .ifPresent(c -> buildable.condition(ConditionBuilder.putIf(c.getExpressions())));
+      } else {
+        assert update.getCondition().get() instanceof UpdateIfExists;
+        buildable.condition(ConditionBuilder.putIfExists());
+      }
+    } else {
+      buildable.condition(ConditionBuilder.putIfExists());
```
komamitsu, line 288: This condition is implicitly added, but `ConsensusCommit.update()` checks if the original `Update` has any condition not a converted `Put` when `UnsatisfiedConditionException` is thrown from `CrudHandler.put()`.

So, this 3 is achieved, right?

1. When `UpdateIf` is specified for Update:
  i. If the target record doesn't exist ===> `UnsatisfiedConditionException`
  ii. If the condition isn't satisfied ===> `UnsatisfiedConditionException`
2. When `UpdateIfExists` is specified for Update:
  i. If the target record doesn't exist (and the condition isn't satisfied) ===> `UnsatisfiedConditionException`
3. When no condition is specified ===> No exceptions are thrown in any case

brfrn169: Yes, this is for 3. When no condition is specified, the Update operation is executed only when the target record exists, and if the target record doesn't exist, we do nothing.

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
brfrn169: Sorry, what do you mean?

core/src/main/java/com/scalar/db/api/Delete.java
brfrn169: Actually, `Mutation.getCondition()` is deprecated and will be removed in the future. However, `Delete.getCondition()` is not deprecated and will remain. Therefore, we don't add `@Deprecated` to this method.

core/src/main/java/com/scalar/db/api/ConditionBuilder.java
```diff
@@ -83,6 +83,35 @@ public static DeleteIfExists deleteIfExists() {
...
+   *
+   * @param conditionalExpressions condition expressions for an UpdateIf condition
+   * @return an UpdateIf condition
+   */
+  public static UpdateIf updateIf(List<ConditionalExpression> conditionalExpressions) {
+    return new UpdateIf(conditionalExpressions);
+  }
+
+  /**
+   * Creates a UpdateIfExists condition.
```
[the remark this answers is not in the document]
brfrn169: Fixed in ad802e94063f32ce5cad5089b44920eed22ea890. Thanks.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedDistributedTransactionManager.java
brfrn169: I forgot to add `synchronized` to this method, and I've added it in ad802e94063f32ce5cad5089b44920eed22ea890. To add `synchronized`, I think we need this method.

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
komamitsu: Ah, sorry. I commented something above this, but I removed it since I misunderstood it. I forgot to remove this one. Never mind.

komamitsu (approved):
LGTM! 👍

feeblefakie (commented):
The code looks good! Thank you!
I left several English suggestions on the Javadoc comments.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/api/Delete.java
```diff
@@ -6,10 +6,12 @@
 import com.scalar.db.api.DeleteBuilder.BuildableFromExisting;
 import com.scalar.db.api.DeleteBuilder.Namespace;
 import com.scalar.db.io.Key;
+import java.util.Optional;
+import javax.annotation.Nonnull;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
- * A command to delete an entry from {@link DistributedStorage}.
+ * A command to delete an entry from a storage.
```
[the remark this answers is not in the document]
josh-wong: Sorry for the delayed reply. Your suggestion looks good to me!👍

core/src/main/java/com/scalar/db/api/ScanAll.java
```diff
@@ -6,9 +6,8 @@
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
- * A command to retrieve all the entries of the database. The scan range of clustering key and
- * {@link Ordering} cannot be specified for this command. The number of {@link Result} can be
- * limited.
+ * A command to retrieve all the entries of a storage. The scan range of clustering key and {@link
```
josh-wong, line 9:
```suggestion
 * A command to retrieve all the entries from the underlying storage. The scan range of clustering key and {@link
```

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +46,129 @@ public interface TransactionCrudOperable {
 
   /**
    * Inserts/Updates an entry to the storage through a transaction with the specified {@link Put}
```
josh-wong, line 48: Since I don't think `Insert` and `Update` are interchangeable, how about if we change this part as follows?:

```suggestion
   * Inserts an entry into or updates an entry in the underlying storage through a transaction with the specified {@link Put}
```

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +46,129 @@ public interface TransactionCrudOperable {
...
-   * @throws UnsatisfiedConditionException if the mutation condition is not satisfied
+   * @throws UnsatisfiedConditionException if a condition is specified, and if the condition is not
+   *     satisfied or the entry does not exist
+   * @deprecated As of release 3.13.0. Will be removed in release 5.0.0.
    */
+  @Deprecated
   void put(Put put) throws CrudConflictException, CrudException, UnsatisfiedConditionException;
 
   /**
    * Inserts/Updates multiple entries to the storage through a transaction with the specified list
```
josh-wong, line 66: Since I don't think `Insert` and `Update` are interchangeable, how about if we change this part as follows?:

```suggestion
   * Inserts multiple entries into or updates multiple entries in the underlying storage through a transaction with the specified list
```

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +46,129 @@ public interface TransactionCrudOperable {
...
+   *     satisfied or the entry does not exist
+   * @deprecated As of release 3.13.0. Will be removed in release 5.0.0. Use {@link #mutate(List)}
+   *     instead.
    */
+  @Deprecated
   void put(List<Put> puts)
       throws CrudConflictException, CrudException, UnsatisfiedConditionException;
 
   /**
    * Deletes an entry from the storage through a transaction with the specified {@link Delete}
```
josh-wong, line 87:
```suggestion
   * Deletes an entry from the underlying storage through a transaction with the specified {@link Delete}
```

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +46,129 @@ public interface TransactionCrudOperable {
...
    *     still fail if the cause is nontranient
-   * @throws UnsatisfiedConditionException if the mutation condition is not satisfied
+   * @throws UnsatisfiedConditionException if a condition is specified, and if the condition is not
+   *     satisfied or the entry does not exist
    */
   void delete(Delete delete)
       throws CrudConflictException, CrudException, UnsatisfiedConditionException;
 
   /**
    * Deletes entries from the storage through a transaction with the specified list of {@link
```
josh-wong, line 104:
```suggestion
   * Deletes entries from the underlying storage through a transaction with the specified list of {@link
```

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -44,63 +46,129 @@ public interface TransactionCrudOperable {
...
+   * @throws CrudConflictException if the transaction CRUD operation fails due to transient faults
+   *     (e.g., a conflict error). You can retry the transaction from the beginning
+   * @throws CrudException if the transaction CRUD operation fails due to transient or nontransient
+   *     faults. You can try retrying the transaction from the beginning, but the transaction may
+   *     still fail if the cause is nontranient
+   */
+  void insert(Insert insert) throws CrudConflictException, CrudException;
+
+  /**
+   * Inserts or updates an entry in the storage through a transaction with the specified {@link
```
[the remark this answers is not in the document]
josh-wong: In this case, since we need to use the same phrase consistently, the only way I can think of to get around this is to structure the sentence as follows:

```suggestion
   * Inserts an entry into or updates an entry in the underlying storage through a transaction with the specified {@link
```

core/src/main/java/com/scalar/db/api/Update.java
```diff
@@ -0,0 +1,124 @@
...
+import com.scalar.db.io.Column;
+import com.scalar.db.io.Key;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/** A command to update an entry to a storage. */
```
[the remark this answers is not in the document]
josh-wong: Assuming that we're referring to an entry that exists in of the underlying storage, I think we should use `in` instead of `of`.

```suggestion
/** A command to update an entry in the underlying storage. */
```

josh-wong (changes requested):

core/src/main/java/com/scalar/db/api/Insert.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Override
+  public Consistency getConsistency() {
+    throw new UnsupportedOperationException();
+  }
+
+  /** @deprecated As of release 3.6.0. Will be removed in release 5.0.0 */
+  @Deprecated
+  public Operation withConsistency(Consistency consistency) {
```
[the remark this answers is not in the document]
brfrn169: Fixed in cae04c10db86c7dcb272aece7a2a54a481b7f9da. Thanks.

brfrn169:
@feeblefakie @josh-wong I've applied your English suggestions. Please take a look again when you have time! 🙇

feeblefakie (approved):
LGTM! Thank you!

josh-wong (approved):
LGTM! Thank you!🙇‍♂️

PR #1710, opened 2024-05-01
Add YugabyteDB adapter

komamitsu:
## Description

YugabyteDB supports the replication over multiple regions and it would be very important database for ScalarDB users who are interested in disaster recovery. In this PR, we support YugabyteDB as a ScalarDB underlying JDBC storage by the following changes:
- Support `jdbc:yugabytedb:` JDBC connection string to use YugabyteDB's Smart JDBC driver which supports multiple endpoints and failover functionality.
- Updated unit test and integration test cases to take care of the characteristic of YugabyteDB

## Related issues and/or PRs

None

## Changes made

- Add `RdbEngineYugabyte`
  - It inherits `RdbEnginePostgreSQL` thanks to the compatibility
  - It uses the YugabyteDB Smart JDBC driver https://github.com/yugabyte/pgjdbc
  - It uses the existing Apache Commons DBCP connection pool library
    - I first tried HikariCP as the document introduced as the combination example of a connection pool library and the driver. But with the answers from Yugabyte support team and some investigations, I confirmed the JDBC driver works with the existing DBCP. I also observed the combination of HikariCP and YugabyteDB driver unexpectedly uses several times larger number of connections than the configuration (I didn't look into it, though). So, I decided to use DBCP as-is
- Support escaped commas in `scalar.db.contact_points`
  - To configure multiple YugabyteDB endpoints, comma separated endpoints need to be handled as a single contact point not as multiple contact points
- Update the unit test for YugabyteDB adapter
- Update the integration test for YugabyteDB adapter. Especially take care of the following limitation of YugabyteDB
  - Can't use floating point types for a key lookup https://github.com/yugabyte/yugabyte-db/issues/18461
  - DDL fails when other concurrent DDLs are executed

## Checklist

N/A

## Release notes

Added YugabyteDB adapter as one of JDBC storages


core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -645,7 +645,7 @@ public void close() {
     try {
       dataSource.close();
     } catch (SQLException e) {
-      logger.error("Failed to close the dataSource", e);
+      logger.warn("Failed to close the dataSource", e);
```
komamitsu, line 648: Strictly speaking, this isn't related to this PR, but just for consistency with other error handlings.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -222,7 +222,7 @@ public void close() {
     try {
       dataSource.close();
     } catch (SQLException e) {
-      logger.error("Failed to close the dataSource", e);
+      logger.warn("Failed to close the dataSource", e);
```
komamitsu, line 225: Strictly speaking, this isn't related to this PR, but just for consistency with other error handlings.

docs/scalardb-supported-databases.md
```diff
@@ -140,3 +140,14 @@ For recommendations when using JDBC databases, see [Recommendations](requirement
 | **ScalarDB 3.9**  | ✅        |
 | **ScalarDB 3.8**  | ❌        |
 | **ScalarDB 3.7**  | ❌        |
+
+### YugabyteDB
+
+| Version           | YugabyteDB 2 |
+|:------------------|:-------------|
+| **ScalarDB 3.12** | ❌           |
```
komamitsu, line 148: No released version supports this feature, so maybe we can remove this table?

integration-test/src/main/java/com/scalar/db/api/DistributedStorageMultiplePartitionKeyIntegrationTestBase.java
```diff
@@ -96,14 +106,23 @@ protected int getThreadNum() {
...
+  protected boolean isFloatTypeKeySupported() {
+    return true;
+  }
+
+  private void createTables()
+      throws java.util.concurrent.ExecutionException, InterruptedException, ExecutionException {
     List<Callable<Void>> testCallables = new ArrayList<>();
 
     Map<String, String> options = getCreationOptions();
+    admin.createNamespace(getNamespaceName(), true, options);
```
komamitsu, line 122: The latency of DDLs is basically very high with YugabyteDB, and many DDLs takes time. I noticed this integration test class didn't need multiple namespaces. So, I changed it to use a single namespace to reduce the duration for testing. Let me know if you have any concern on this change.

integration-test/src/main/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
```diff
@@ -338,6 +343,9 @@ private void createTables_ShouldCreateTables() throws Exception {
   private void deleteTables_ShouldDeleteTables() throws Exception {
     // Act
     int exitCode = executeWithArgs(getCommandArgsForDeletion(CONFIG_FILE_PATH, SCHEMA_FILE_PATH));
+    if (exitCode != 0 && couldFailToReadNamespaceAfterDeletingTable()) {
```
komamitsu, line 346: This case could fail and retry is needed with YugabyteDB.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageMultiplePartitionKeyIntegrationTestBase.java
komamitsu: Ah, got it. The integration test case failed with Cosmos DB due to the limitation `Collection create forbidden as collection count in database offer has exceeded 25` (only 25...?) https://github.com/scalar-labs/scalardb/actions/runs/8906213495/job/24458020986?pr=1710.

I'll revert the change.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineYugabyte.java
```diff
@@ -0,0 +1,10 @@
+package com.scalar.db.storage.jdbc;
+
+import java.sql.Driver;
+
+class RdbEngineYugabyte extends RdbEnginePostgresql {
+  @Override
+  public Driver getDriver() {
+    return new com.yugabyte.Driver();
```
brfrn169, line 8: It looks like we don't use HikariCP for YugabyteDB. So, HikariCP is not required for YugabyteDB?

brfrn169 (commented):

docs/scalardb-supported-databases.md
brfrn169: Let's keep this table. We can add a row for ScalarDB 3.13 when we release ScalarDB 3.13.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageMultiplePartitionKeyIntegrationTestBase.java
komamitsu: Done

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineYugabyte.java
komamitsu: Right.

The combination of Apache Commons DBCP and YugabyteDB driver isn't introduced as an example on the official document while HikariCP is introduced. But YugabyteDB support team said we can use DBCP as well. I also looked over the source code of the YugabyteDB driver and it seems to work well.

As for HikariCP, I think we can integrate it with the following steps as a separate project:
1. Benchmark ScalarDB with all the JDBC storages using HikariCP
2. If the performance of 1 isn't significantly better than the current performance, we don't need to move forward on it
3. Introduce a new property `scalar.db.jdbc.connection_pool.library` or something which can be either of `dbcp` or `hikaricp`
4. Use `dbcp` by default in version 3 and use `hikari` by default in version 4

integration-test/src/main/java/com/scalar/db/api/DistributedStorageMultipleClusteringKeyScanIntegrationTestBase.java
```diff
@@ -114,6 +115,15 @@ protected int getThreadNum() {
...
   }
 
+  protected boolean isParallelDdlSupported() {
+    return true;
+  }
+
+  // TODO: Remove this once https://github.com/yugabyte/yugabyte-db/issues/22140 is fixed.
+  protected boolean isYugabyteDb() {
+    return false;
+  }
```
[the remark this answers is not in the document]
komamitsu: Probably JUnit 5 would fail due to `org.junit.jupiter.engine.execution.ConditionEvaluationException: Failed to evaluate condition [org.junit.jupiter.api.condition.DisabledIfCondition]: Could not find method [isYugabyteDb] in class [$(non-JDBC-integration-test)]`?, I guess.

integration-test/src/main/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
komamitsu: Good point. Only this test case requires long wait (e.g., 15 seconds) on my laptop, so I chose the retry.

But, I just found `--ysql_enable_db_catalog_version_mode=false` option for `bin/yugabyted` and it prevented "Catalog Version Mismatch" error with the test cases. I'll use the option for the integration tests and remove unnecessary waits as much as possible.

komamitsu: Fixed it with https://github.com/scalar-labs/scalardb/pull/1710/commits/cf633158438e27644c4c5d6d73855d3a0e9ebc3d and https://github.com/scalar-labs/scalardb/pull/1710/commits/f0269b18d410c9de80e0233d0aade382febf7b74.

The option especially works for this case. Some other test cases still need waits, so I left them as is.

docs/backup-restore.md
```diff
@@ -205,4 +211,8 @@ If you used `pg_dump` to create the backup file, use the `psql` command to resto
 
 Use the `.restore` command as specified in [Special commands to sqlite3 (dot-commands)](https://www.sqlite.org/cli.html#special_commands_to_sqlite3_dot_commands_).
 </div>
+<div id="YugabyteDB_Managed1" class="tabcontent" markdown="1">
+
+You can restore from the scheduled backup within the backup retention period. See [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```
feeblefakie, line 216: YugabyteDB seems to provide a PITR feature.
https://docs.yugabyte.com/preview/manage/backup-restore/point-in-time-recovery/
So, we probably need to mention it here.
(Also, we need to enable the PITR feature by creating a snapshot schedule, so we might need to add a tab for YugabyteDB in the `Back up with explicit pausing` section.) 

But, I'm not sure about the relationships between the usual backup/restore and PITR in YugabyteDB.
It looks like [this doc](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/) doesn't describe PITR and snapshots at all, and the above doc doesn't really describe usual backup and restore.

Could you take a deeper look at the docs and check the features?

feeblefakie (commented):
I left some comments on the backup and restore doc. PTAL when you get a chance!

integration-test/src/main/java/com/scalar/db/api/DistributedStorageMultipleClusteringKeyScanIntegrationTestBase.java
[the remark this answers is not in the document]
brfrn169: How about adding the following methods to `JdbcDatabaseMultipleClusteringKeyScanIntegrationTest`?

```java
  // TODO: Remove this once https://github.com/yugabyte/yugabyte-db/issues/22140 is fixed.
  @DisabledIf("isYugabyteDb")
  @Test
  @Override
  public void scan_WithSecondClusteringKeyRange_ShouldReturnProperResult()
      throws java.util.concurrent.ExecutionException, InterruptedException {
    super.scan_WithSecondClusteringKeyRange_ShouldReturnProperResult();
  }

  // TODO: Remove this once https://github.com/yugabyte/yugabyte-db/issues/22140 is fixed.
  @DisabledIf("isYugabyteDb")
  @Test
  @Override
  public void scan_WithSecondClusteringKeyRangeWithSameValues_ShouldReturnProperResult()
      throws java.util.concurrent.ExecutionException, InterruptedException {
    super.scan_WithSecondClusteringKeyRangeWithSameValues_ShouldReturnProperResult();
  }
```

I think with this, we remove `isYugabyteDb()` things from `DistributedStorageMultipleClusteringKeyScanIntegrationTestBase`.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineYugabyte.java
brfrn169: Sounds good! Thanks.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageMultipleClusteringKeyScanIntegrationTestBase.java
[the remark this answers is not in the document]
komamitsu: Ah, I see. The override seems to work. I'm okay to take it. I'll update the PR.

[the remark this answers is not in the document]
komamitsu: Done in https://github.com/scalar-labs/scalardb/pull/1710/commits/a981f86b58c7faf6ae2eabc3c7fb02699b8677d3.

brfrn169 (approved):
LGTM! Thank you!

brfrn169:
@komamitsu Ah, I just noticed that the integration test for YugabyteDB is very slow, even slower than the one for Cosmos DB, which was previously the slowest. Could this be related to the wait time?

komamitsu:
@brfrn169 Good point. I left the issue since the YDB tests was a little bit faster than the Cosmos DB tests, but yeah the YDB test set is often slower than the Cosmos DB's one.

I think the main cause of the slowness is the sequential DDL executions https://github.com/scalar-labs/scalardb/pull/1710/files#diff-a4be19d0f2c918cb2cca4bef9281647ff4b7963df40d777053232a73177f2868R2061-R2075. One possible idea is to execute the many DDLs in parallel and retry if it fails. I'll try it.

docs/backup-restore.md
komamitsu: @feeblefakie Good catch!

I confirmed the PITR feature with local YugabyteDB via `yb-admin` cli command worked fine.

But, as for the managed YugabyteDB, the web console only provides the full and incremental backup/restore not PITR. Also, it seems impossible to conduct the PITR via `yb-admin` cli since the managed YugabyteDB doesn't expose TCP port:7100 which is used for `yb-admin`. I'll ask YugabyteDB support team about it.

komamitsu:
> One possible idea is to execute the many DDLs in parallel and retry if it fails. I'll try it.

@brfrn169 I tried this idea, but it seemed YugabyteDB didn't efficiently handle the many parallel DDLs and the total duration was even larger and some failures happened. I'll continue to look for other improvement points (maybe as another PR?)

komamitsu:
@brfrn169 We've already talked on another chat, but let me summarize it here.

Creating a ScalarDB table with DESC order which implicitly creates an index takes more than 2 seconds on YugabyteDB. It takes total 10 minutes and it's the main cause of the long duration of integration test.

In the current implementation, ScalarDB creates an index whenever any DESC order is specified.

https://github.com/scalar-labs/scalardb/blob/11374d4a7b427945da7aebbc0594a51eda981173/core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java#L677-L680

https://github.com/scalar-labs/scalardb/blob/11374d4a7b427945da7aebbc0594a51eda981173/core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java#L267-L269

But, actually no extra index is needed when all the orders are DESC as the default primary key based b-tree can be utilized.

```
postgres=# explain analyze select * from test_tbl where pk = 5 order by ck1 asc, ck2 asc limit 5;
                                                                  QUERY PLAN                                                                   
-----------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.57..1.06 rows=5 width=12) (actual time=0.081..0.091 rows=5 loops=1)
   ->  Index Only Scan using test_tbl_pkey on test_tbl  (cost=0.57..950593.32 rows=9705382 width=12) (actual time=0.077..0.086 rows=5 loops=1)
         Index Cond: (pk = 5)
         Heap Fetches: 0
 Planning Time: 0.269 ms
 Execution Time: 0.129 ms
(6 rows)

postgres=# explain analyze select * from test_tbl where pk = 5 order by ck1 asc, ck2 desc limit 5;
                                                                     QUERY PLAN                                                                      
-----------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.68..1.39 rows=5 width=12) (actual time=0.110..0.112 rows=5 loops=1)
   ->  Incremental Sort  (cost=0.68..1387335.51 rows=9705382 width=12) (actual time=0.107..0.108 rows=5 loops=1)
         Sort Key: ck1, ck2 DESC
         Presorted Key: ck1
         Full-sort Groups: 1  Sort Method: quicksort  Average Memory: 25kB  Peak Memory: 25kB
         ->  Index Only Scan using test_tbl_pkey on test_tbl  (cost=0.57..950593.32 rows=9705382 width=12) (actual time=0.067..0.075 rows=6 loops=1)
               Index Cond: (pk = 5)
               Heap Fetches: 0
 Planning Time: 0.357 ms
 Execution Time: 1.689 ms
(10 rows)

postgres=# explain analyze select * from test_tbl where pk = 5 order by ck1 desc, ck2 desc limit 5;
                                                                       QUERY PLAN                                                                       
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.57..1.06 rows=5 width=12) (actual time=0.270..0.282 rows=5 loops=1)
   ->  Index Only Scan Backward using test_tbl_pkey on test_tbl  (cost=0.57..950593.32 rows=9705382 width=12) (actual time=0.265..0.275 rows=5 loops=1)
         Index Cond: (pk = 5)
         Heap Fetches: 0
 Planning Time: 0.273 ms
 Execution Time: 0.317 ms
(6 rows)
```

So, I'll optimize the ScalarDB table creation by skipping index creation if possible as another PR. This optimization will reduce the long duration of the YugabyteDB integration test.

docs/backup-restore.md
komamitsu: @feeblefakie I received a reply from Yugabyte team. YugabyteDB Managed doesn't provide PITR feature at the moment. I'll talk with their product team later to get their plan of the feature. I'll update this document once the feature is released in the future.

BTW, I only described YugabyteDB Managed not OSS YugabyteDB since I think it's not easy to deploy the cluster by  users. Do you think we should describe OSS YugabyteDB as well?

feeblefakie: Thank you!
I understand the current situation.
I think, for now, we can assume users use YugabyteDB Managed.

feeblefakie: When we back up with explicit pausing, we should probably use on-demand backup during a paused period, 
so I think we should mention so in the backup-side doc.
Here, I think we should probably mention that we restore from the backups taken by the on-demand backups, not from the scheduled backups.
What do you think?

komamitsu: @feeblefakie Thanks for the feedbacks! I overlooked the need for explicit pausing when dealing with multiple databases.

I added a description about YugabyteDB with explicit pausing https://github.com/scalar-labs/scalardb/pull/1710/files#diff-0847747898d2c35d3723a35f59e28fbdc704e3d2146f2c5ca5e01850ac410badR136-R141.

Regarding the backup types (on-demand or scheduled backup), I think either is fine as long as the backup timing is synced with explicit pausing. I didn't elaborate it in the newly added description, but let me know if you think we should.

docs/schema-loader-import.md
```diff
@@ -131,20 +131,20 @@ Data types not listed in the above are not supported. The following are some com
...
-| double precision  | DOUBLE   |                       |
-| integer           | INT      |                       |
-| real              | FLOAT    |                       |
-| smallint          | INT      | [*2](#warn-data-size) |
-| text              | TEXT     |                       |
+<div id="PostgreSQL_YugabyteDB" class="tabcontent" markdown="1">
+
+| PostgreSQL/YugabyteDB | ScalarDB | Notes                 |
+|-----------------------|----------|-----------------------|
+| bigint                | BIGINT   | [*1](#warn-data-size) |
```
josh-wong, line 138: I don't think this is specific to this PR, but the links to `#warn-data-size` don't seem to work. Is there a section or footnote within this document that this link should point to?

docs/schema-loader.md
```diff
@@ -616,25 +616,33 @@ Auto-scaling for Cosmos DB for NoSQL is enabled only when this option is set to
...
-| BOOLEAN   | boolean   | boolean (JSON)      | BOOL     | boolean  | boolean          | number(1)      | bit             | boolean |
-| INT       | int       | number (JSON)       | N        | int      | int              | int            | int             | int     |
-| BIGINT    | bigint    | number (JSON)       | N        | bigint   | bigint           | number(19)     | bigint          | bigint  |
-| FLOAT     | float     | number (JSON)       | N        | double   | float            | binary_float   | float(24)       | float   |
-| DOUBLE    | double    | number (JSON)       | N        | double   | double precision | binary_double  | float           | double  |
-| TEXT      | text      | string (JSON)       | S        | longtext | text             | varchar2(4000) | varchar(8000)   | text    |
-| BLOB      | blob      | string (JSON)       | B        | longblob | bytea            | RAW(2000)      | varbinary(8000) | blob    |
+| ScalarDB  | Cassandra | Cosmos DB for NoSQL | DynamoDB | MySQL    | PostgreSQL/YugabyteDB | Oracle         | SQL Server      | SQLite  |
```
josh-wong, line 619: Since PostgreSQL is not the same as YugabyteDB, how about if we separate them into different columns in this table? (For example, Cassandra and SQLite have the same data types, but they are in separate columns.)

> [!NOTE]
>
> If we decide to separate PostgreSQL and YugabyteDB here, we should also separate them in other tables and tabs.

docs/backup-restore.md
```diff
@@ -67,6 +68,10 @@ Use the `.backup` command with the `.timeout` command as specified in [Special c
 
 For an example, see [BASH: SQLite3 .backup command](https://stackoverflow.com/questions/23164445/bash-sqlite3-backup-command).
 </div>
+<div id="YugabyteDB_Managed1" class="tabcontent" markdown="1">
+
+Clusters are backed up automatically based on the backup policy, and these backups are retained for a specific duration. You can also perform on-demand backups. See [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```
josh-wong, line 73:
```suggestion
Clusters are backed up automatically based on the backup policy, and these backups are retained for a specific duration. You can also perform on-demand backups, as described in [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```

docs/backup-restore.md
```diff
@@ -127,6 +133,12 @@ You must enable the PITR feature for DynamoDB tables. If you're using [ScalarDB
 
 To specify a transactionally consistent restore point, pause your application that is using ScalarDB with DynamoDB as described in [Back up with explicit pausing](#back-up-with-explicit-pausing).
 </div>
+<div id="YugabyteDB_Managed2" class="tabcontent" markdown="1">
+
+Clusters are backed up automatically based on the backup policy, and these backups are retained for a specific duration. You can also perform on-demand backups. See [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```
josh-wong, line 138:
```suggestion
Clusters are backed up automatically based on the backup policy, and these backups are retained for a specific duration. You can also perform on-demand backups, as described in [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```

docs/backup-restore.md
```diff
@@ -127,6 +133,12 @@ You must enable the PITR feature for DynamoDB tables. If you're using [ScalarDB
...
+
+If you're using YugabyteDB with other databases, to ensure transactional consistency across databases, you'll need to pause the application using ScalarDB or [ScalarDB Cluster (redirects to the Enterprise docs site)](https://scalardb.scalar-labs.com/docs/latest/scalardb-cluster/) during the backup process, as described in [Back up with explicit pausing](#back-up-with-explicit-pausing).
```
josh-wong, line 140: Just a minor revision to reduce eliminate. (I originally read this part as `pause the application by using ScalarDB`, which I believe is incorrect.)

```suggestion
If you're using YugabyteDB with other databases, to ensure transactional consistency across databases, you'll need to pause the application that is using ScalarDB or [ScalarDB Cluster (redirects to the Enterprise docs site)](https://scalardb.scalar-labs.com/docs/latest/scalardb-cluster/) during the backup process, as described in [Back up with explicit pausing](#back-up-with-explicit-pausing).
```

docs/backup-restore.md
```diff
@@ -205,4 +218,8 @@ If you used `pg_dump` to create the backup file, use the `psql` command to resto
 
 Use the `.restore` command as specified in [Special commands to sqlite3 (dot-commands)](https://www.sqlite.org/cli.html#special_commands_to_sqlite3_dot_commands_).
 </div>
+<div id="YugabyteDB_Managed3" class="tabcontent" markdown="1">
+
+You can restore from the scheduled or on-demand backup within the backup retention period. See [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```
josh-wong, line 223:
```suggestion
You can restore from the scheduled or on-demand backup within the backup retention period, as described in [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```

josh-wong (changes requested):

docs/schema-loader-import.md
komamitsu: Good catch. The link seems to try to point https://github.com/scalar-labs/scalardb/blob/master/docs/schema-loader-import.md?plain=1#L272, but GitHub markdown automatically adds a prefix `user-content-`. So, it should be `user-content-warn-data-size`?

docs/schema-loader.md
komamitsu: Good question!

I don't think the data types of YugabyteDB and PostgreSQL happen to be the same. YugabyteDB has purposefully achieved the compatibility with PostgreSQL:
- https://docs.yugabyte.com/preview/explore/ysql-language-features/postgresql-compatibility/
- https://www.yugabyte.com/postgresql/postgresql-compatibility/

> Supporting [SQL](https://www.yugabyte.com/distributed-sql/sql/) (Structured Query Language), the standard language for relational databases, is also essential. This includes support for the same data types, operators, and functions used in PostgreSQL, along with similar syntax for creating and manipulating tables, indexes, and views.

I don't think they are the 100% same. But I think we can treat YugabyteDB as compatible as PostgreSQL for now to reduce our maintenance effort. We can separate them if the difference gets significant in the future. 

But I'm okay to separate them in this PR.

docs/backup-restore.md
komamitsu: @josh-wong Thanks for the suggestion! I basically agree on it.
My slight concern is that the YugabyteDB doc is also helpful to understand scheduled backups not only on-demand backups. With the change you suggested, it seems the doc is only related to on-demand ones?

josh-wong: Oh, right😅 How about we use the following instead?

```suggestion
Clusters are backed up automatically based on the backup policy, and these backups are retained for a specific duration. You can also perform on-demand backups. For details on performing backups, see [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```

komamitsu: @josh-wong 💯

komamitsu: @josh-wong Can I apply this to other sentences as well? Or you've been already working on it?

docs/schema-loader-import.md
josh-wong: Oooh, I see. I tried adding the footnotes within an admonition like you've done here, but doing so doesn't seem to be supported in Docusaurus. I'm guessing it has to do with the admonition component in Docusaurus, which isn't in HTML like the one in Jekyll is. 

> [!NOTE]
>
> I'm guessing the reasoning behind that is footnotes are their own section, so they shouldn't need to be inside an admonition. Books, academic papers, and such list their footnotes at the bottom of the pages as well, so Docusaurus is probably just following that standard.

Instead, how about if we use the standard [footnote feature in Markdown](https://www.markdownguide.org/extended-syntax/#footnotes)? This make a "Footnote" section at the end of the doc that users can refer to.[^1]

[^1]: This is an example of a footnote in Markdown.

docs/schema-loader.md
josh-wong: Thanks for the explanation! I'm OK with keeping them together as is for now👍

docs/backup-restore.md
josh-wong: Sorry for the slow reply. I've updated my suggestions!

feeblefakie (commented):
I left one suggestion. PTAL!

docs/backup-restore.md
komamitsu: @josh-wong Thanks a lot!

docs/backup-restore.md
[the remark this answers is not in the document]
komamitsu: > to ensure transactional consistency across databases

Oops. I unexpectedly removed something about backup-ish taste from this... Maybe `when taking a backup` or something should've been added.

Your suggestion seems better in terms of simplicity. I'll merge it.

BTW, the reason why I put `If you're using YugabyteDB with other databases` was to clarify this description exists in `Back up with explicit pausing` section here only considering multi database scenario (`A[no]->D` below), not due to missing transaction support (`B[no]->D`). Maybe we can revisit this later as well as describing other RDBMS here considering multi database usage (`A[no]->D`).

```
flowchart TD
  A[Are you using a single database with ScalarDB?]
  A -->|Yes| B[Does the database have transaction support?]
  B -->|Yes| C[Perform <a href='#back-up-without-explicit-pausing'>back up without explicit pausing</a>]
  B ---->|No| D[Perform <a href='#back-up-with-explicit-pausing'>back up with explicit pausing</a>]
  A ---->|No| D
```

docs/schema-loader-import.md
komamitsu: @josh-wong Thanks for the investigation. The footnote feature looks good. But, the content in the warning block is important and it should be still in the warning block more than in footnotes to keep them eye-catching?

I noticed the warning block is located below all the links pointing the block. I guess users can find the warning block without any links. So, how about replacing the links with just a comment like `See the below warning block (#2)`?

komamitsu: BTW, I think we can fix the issue in a separate PR later.

docs/backup-restore.md
```diff
@@ -127,6 +133,12 @@ You must enable the PITR feature for DynamoDB tables. If you're using [ScalarDB
 
 To specify a transactionally consistent restore point, pause your application that is using ScalarDB with DynamoDB as described in [Back up with explicit pausing](#back-up-with-explicit-pausing).
 </div>
+<div id="YugabyteDB_Managed2" class="tabcontent" markdown="1">
+
+Clusters are backed up automatically based on the backup policy, and these backups are retained for a specific duration. You can also perform on-demand backups. For details on performing backups, see [YugabyteDB Managed: Back up and restore clusters](https://docs.yugabyte.com/preview/yugabyte-cloud/cloud-clusters/backup-clusters/).
```
feeblefakie, line 138: Sorry, I meant this paragraph should probably be removed because it sounds like automatic backups based on the policy are just fine.
What do you think?
(I came up with the next paragraph to emphasize taking backups should be done during a paused duration.)

feeblefakie (approved):
I left one minor suggestion, but other than that, LGTM, so I'm approving this.

docs/backup-restore.md
komamitsu: Oh, sorry, I misread your comment and just merged the change w/o removing the first paragraph.

I thought the first paragraph was okay since it described what the backups on YugabyteDB Managed are like, and then the second paragraph explained that pausing was necessary in that case. But, I agree that users who read only the first paragraph might think no pause is needed. I'll remove it.

docs/schema-loader-import.md
josh-wong: @komamitsu 
> I noticed the warning block is located below all the links pointing the block. I guess users can find the warning block without any links. So, how about replacing the links with just a comment like See the below warning block (#2)?

I'm OK with this method 👍

josh-wong (approved):
LGTM! Thank you!🙇‍♂️

docs/backup-restore.md
komamitsu: @feeblefakie Addressed it in https://github.com/scalar-labs/scalardb/pull/1710/commits/92098ba2f99a3ba4daee59bb057e40eb72416ad4. Thanks for the great feedbacks!

brfrn169:
@komamitsu It looks like the integration test for YugabyteDB is still slow even after applying the optimization?

komamitsu:
@brfrn169 Yes. Sill slow basically. I think it's been totally improved by about 1 ~ 2 minutes, though.

The drawback of applying the optimization is that [this issue](https://github.com/scalar-labs/scalardb/pull/1710/files/189e292b1dba4d68ff9aea8f01b4e3dea49b0cee#r1591940745) has been reproduced and I increased the wait from 2 seconds to 10 seconds, total about 1 minute wait is added.

Maybe we should revert [this retry](https://github.com/scalar-labs/scalardb/pull/1710/files/189e292b1dba4d68ff9aea8f01b4e3dea49b0cee#r1592330864) so that we can shorten the wait to 1 or 2 seconds (total down to about 10 seconds). What do you think?

komamitsu:
> Maybe we should revert [this retry](https://github.com/scalar-labs/scalardb/pull/1710/files/189e292b1dba4d68ff9aea8f01b4e3dea49b0cee#r1592330864) so that we can shorten the wait to 1 or 2 seconds (total down to about 10 seconds). What do you think?

@brfrn169 The change would be like this https://github.com/scalar-labs/scalardb/pull/1745/files

komamitsu:
Per the discussion with @brfrn169, we concluded to leave it as is for now since the retry won't drastically improve the situation (e.g., 10 minutes) and the duration of YugabyteDB integration tests is basically same as Cosmos DB duration.

PR #1715, opened 2024-05-08
Refactor scan with filtering

jnmt:
## Description

This PR refactors `ScanAll` with a where clause so that we can support `Get`, `GetWithIndex` `Scan`, and `ScanWithIndex` with a where clause in a consistent way in the upcoming updates. After careful discussion, we decided to introduce a few backward-incompatible changes since cross-partition scan with filtering is a relatively new feature and we expected their users to be none or less.

## Related issues and/or PRs

- #1682 (upcoming after merging this)

## Changes made

- Move `Conjunction` from `Scan` to `Selection`
- Move `ConditionSetBuilder` and `{And,Or}ConditionSet` from `ScanBuilder` to `api`
- Refactor `ScanBuilder`
  - Rename `BuildableScanAllFromExistingWithX` to `BuildableScanFromExistingWithX` so that it can be commonly used for `Scan`, `ScanWithIndex` and `ScanAll`
  - Consolidate member variables in `BuildableScanFromExistingWithWhere` using a composite buildable
  - Introduce `OngoingWhere` for the common internal condition management

## Checklist

The following PRs will be upcoming in this order.

- Dynamic filtering for `ScalAll` in #1682
- Dynamic filtering for `Scan` and `ScanWithIndex`
- Dynamic filtering for `Get` and `GetWithIndex`

## Release notes

Refactored scan with filtering.


core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -936,525 +905,299 @@ public Scan build() {
...
-      if (o == this) {
-        return true;
-      }
-      if (!(o instanceof AndConditionSet)) {
-        return false;
-      }
-      AndConditionSet other = (AndConditionSet) o;
-      return conditions.equals(other.conditions);
-    }
+  private static class OngoingWhere {
```
jnmt, line 1140: This is introduced for simply and commonly managing the internal condition state for both `BuildableScanAllWithWhere` and `BuildableScanFromExistingWithXX`. Since it can also be used for `GetBuilder,` this class may be moved to somewhere commonplace.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -621,9 +591,9 @@ public static class BuildableScanOrScanAllFromExisting extends BuildableScan
           OperationBuilder.Table<BuildableScanOrScanAllFromExisting>,
```
jnmt, line 591: Since `BuildableScanOrScanAllFromExisting` is also used for changing an existing `ScanWithIndex`, it might be better `BuildableScanAllFromExisting`. But `BuildableScanOrScanAllFromExisting` is older than the cross-partition scan with filtering, so I think breaking it would be a bit aggressive. Please let me know if you have any opinions.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
jnmt: Also, I avoided using just `Where` to distinguish the `Where` interface, but the name might be a bit weird since it will also be used after closing the where clause. `InternalWhere`, `WhereClause`, or `WhereBuilder`? What do you think?

jnmt:
Sorry, I found some null checks were accidentally deleted when refactoring. Fixed in [797b95e](https://github.com/scalar-labs/scalardb/pull/1715/commits/797b95e43746aba3768c0ac3d8867476dd62db73).

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -936,525 +910,299 @@ public Scan build() {
...
-      this.namespaceName = buildable.namespaceName;
-      this.tableName = buildable.tableName;
-      this.condition = null;
-      this.conjunctions.addAll(buildable.conjunctions);
-      this.disjunctions.addAll(buildable.disjunctions);
-      this.limit = buildable.limit;
-      this.orderings.addAll(buildable.orderings);
-      this.projections.addAll(buildable.projections);
-      this.consistency = buildable.consistency;
+    BuildableScanOrScanAllFromExisting buildableScanFromExisting;
```
komamitsu, line 924:
```suggestion
    private BuildableScanOrScanAllFromExisting buildableScanFromExisting;
```
This can be `private`?

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -936,525 +910,299 @@ public Scan build() {
...
+    protected final OngoingWhere where;
```
komamitsu, line 925: IntelliJ shows `Class 'OngoingWhere' is exposed outside its defined visibility scope`. So, this should be package-private?

```suggestion
    final OngoingWhere where;
```

core/src/main/java/com/scalar/db/api/ScanBuilder.java
komamitsu: `WhereBuilder` sounds to me that it has `build()` method. So, it might be worth adding `build()` that returns `Collection<Conjunction>` or something.

`WhereClause` itself sounds good to me, but I noticed the term `Clause` isn't used in this file...

I think `WhereCondition` might be an option as the class contains only conditions.

core/src/main/java/com/scalar/db/api/ConditionSetBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Adds a conditional expression for a {@code AndConditionSet}.
+     *
+     * @param condition a conditional expression for a {@code AndConditionSet}.
+     * @return a builder object
+     */
+    public BuildableAndConditionSet and(ConditionalExpression condition) {
+      conditions.add(condition);
+      return new BuildableAndConditionSet(conditions);
```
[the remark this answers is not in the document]
jnmt: Right, that's well-mannered. Is the following expected?

```java
  @Immutable
  public static class BuildableAndConditionSet {

    private final Set<ConditionalExpression> conditions;

    BuildableAndConditionSet(Set<ConditionalExpression> conditions) {
      this.conditions = ImmutableSet.copyOf(conditions);
    }

    public BuildableAndConditionSet and(ConditionalExpression condition) {
      return new BuildableAndConditionSet(
          new ImmutableSet.Builder<ConditionalExpression>()
              .addAll(conditions)
              .add(condition)
              .build());
    }
```

core/src/main/java/com/scalar/db/api/ScanBuilder.java
jnmt: Thanks for the suggestion. I changed my mind a bit. Since the interface `Where` is only used twice in this file, using it with the modifier (i.e., `OperationBuilder.Where`) is not so bad looking. I'm going with this if no further concerns.

core/src/main/java/com/scalar/db/api/ConditionSetBuilder.java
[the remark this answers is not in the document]
jnmt: Thanks! Unfortunately, we cannot just return `this` here since we need to prevent users from calling `or` once they start with `and`. So, recreating a set without affecting the current set seems the best solution here.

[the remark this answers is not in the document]
komamitsu: Ah, you're right. I thought it as a usual builder pattern...

Also, I think I was thinking of defensive copy and immutability when I writing the first comment, but I forgot it when writing the second comment. Sorry for the confusion.

As we talked on another chat, immutable approach is basically better as long as there is no memory memory-pressure concerns.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -514,54 +499,33 @@ public static class BuildableScanAllWithWhere
...
-    protected final String namespaceName;
-    protected final String tableName;
-    @Nullable protected final ConditionalExpression condition;
-    protected final Set<Set<ConditionalExpression>> conjunctions = new HashSet<>();
-    protected final Set<Set<ConditionalExpression>> disjunctions = new HashSet<>();
-    protected final List<String> projections = new ArrayList<>();
-    protected final List<Scan.Ordering> orderings = new ArrayList<>();
-    protected int limit;
-    @Nullable protected com.scalar.db.api.Consistency consistency;
+    protected BuildableScanAll buildableScanAll;
```
komamitsu, line 502: [trivial] `BuildableScanAll` is mutable, so you can also make this variable `final` and simply use it like this.

```java
    @Override
    public BuildableScanAllWithWhere projection(String projection) {
      buildableScanAll.projection(projection);
      return this;
    }
```

The current code can be used as-is if `BuildableScanAll` changes into immutable in the future, so I don't have a strong opinion on it , though.

jnmt: Thanks again. Definitely. Fixed them in 8320ae8 as you mentioned due to no plan for changing them to immutable so far.

komamitsu (approved):
LGTM! Thank you!

brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!
(Very sorry for the late review 🙇 )

PR #1728, opened 2024-05-14
Integrate `util/groupcommit` package with the coordinator table commit

komamitsu:
## Description

We've already [added `util/groupcommit` package](https://github.com/scalar-labs/scalardb/pull/1607), but not used yet. This PR integrate the group commit with the coordinator table writes.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/1607

## Changes made

image

- Add `CoordinatorGroupCommitter` that inherits `GroupCommitter` and `CommitHandlerWithGroupCommit` that inherits `CommitHandler`
  - They write a record to `coordinator.state` table as `Emitter.emit()`
  - There are 2 types of `coordinator.state` table record
    - In the case of normal group commit, the columns of a record are as follows:
      - `tx_id`: `$parent_id` (a key of the normal group contains multiple transactions)
      - `tx_child_ids`: `$child_id_1,$child_id_2,...,$child_id_n` (values of the normal group contains multiple transactions)
    - In the case of delayed group commit which contains a single transaction isolated from a normal group due to delayed transaction, the columns of a record are as follows:
      - `tx_id`: `$parent_id:$child_id` (a key and a value of the delayed group contains a single transaction)
      - `tx_child_ids`: Empty
- Modify `ConsensusCommitConfig`, `ConsensusCommit`, `ConsensusCommitManager`, `TwoPhaseConsensusCommit`, and `TwoPhaseConsensusCommitManager` to use `CommitHandlerWithGroupCommit`
- Modify `util/groupcommit` package for fixing some issues occurred in this integration
  - Sorry, this includes an improvement changing `oldGroupAbortTimeoutSeconds` to `oldGroupAbortTimeoutMillis` that can be handled in another PR...
- Add some unit tests and modify some existing tests 
- Modify some existing integration tests
  - `ConsensusJdbcEnv` is introduced to inject the group commit related properties.

## Checklist

### Transaction ID conventions

- Parent ID
  - A key of normal group that contains multiple transactions
  - `coordinator.state.tx_id` will be set to this ID when storing transactions in a normal group
  - This consists of 24 length 62 type characters `[0-9A-Za-z]`
- Child ID
  - A key to identify each transaction in a group
  - `coordinator.state.tx_child_ids` will contain this type of ID when storing transactions in a normal group
  - This will be UUID v4 when automatically generating on ScalarDB side or an arbitrary string passed by users via `begin(String txId)`
- Full ID
  - A concatenated string of Parent ID and Child ID for a transaction
  - `coordinator.state.tx_id` will be set to this ID when storing transactions in a delayed group that contains only a single transaction (`coordinator.state.tx_child_ids` will be empty in this case)
  - The format of this ID is `$ParentID:$ChildID` whose total length is 61 (24 + 36 + 1). So we don't need to extend the maximum length of `coordinator.state.tx_id`

### Limitations

- [Users need to use a transaction ID returned from `ConsensusCommitManager.begin(String txId)`](https://github.com/scalar-labs/scalardb/blob/ebbb0274bb01fd52dc553a333aff5183434c44b6/docs/api-guide.md#custom-transaction-id-passed-by-users)
- [In the two phase commit interface, users must commit a transaction on the coordinator service before on participant services](https://github.com/scalar-labs/scalardb/blob/ebbb0274bb01fd52dc553a333aff5183434c44b6/docs/api-guide.md#commit-in-two-phase-commit-interface)

image

## Release notes

Added Group Commit feature for Coordinator Table


core/src/main/java/com/scalar/db/util/groupcommit/DelayedGroup.java
```diff
@@ -17,7 +17,7 @@ class DelayedGroup<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V>
       FULL_KEY fullKey,
       Emittable<EMIT_KEY, V> emitter,
       KeyManipulator<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY> keyManipulator) {
-    super(emitter, keyManipulator, 1, config.oldGroupAbortTimeoutSeconds());
+    super(emitter, keyManipulator, 1, config.oldGroupAbortTimeoutMillis());
```
komamitsu, line 20: Strictly speaking, this refactoring isn't directly related to this PR.

core/src/test/java/com/scalar/db/transaction/consensuscommit/CommitHandlerTest.java
```diff
@@ -23,41 +23,65 @@
...
   private static final String ANY_TEXT_4 = "text4";
+  private static final String ANY_ID = "id";
   private static final int ANY_INT_1 = 100;
   private static final int ANY_INT_2 = 200;
 
-  @Mock private DistributedStorage storage;
-  @Mock private Coordinator coordinator;
-  @Mock private TransactionTableMetadataManager tableMetadataManager;
-  @Mock private ConsensusCommitConfig config;
+  @Mock protected DistributedStorage storage;
```
komamitsu, line 48: This class is inherited by `CommitHandlerWithGroupCommitTest` and these variables are used by it.

docs/api-guide.md
```diff
@@ -1197,6 +1197,90 @@ In the sample code, the transaction is retried three times maximum and sleeps fo
...
+   ...
+
+   // When the group commit feature is enabled, a custom transaction ID passed by users can't be used as is.
+   // logger.info("The transaction state: {}", manager.getState(myTxId));
+   logger.info("The transaction state: {}", manager.getState(transaction.getId()));
+```
+
+##### Commit in Two-Phase Commit Interface
+
+The group commit feature manages all in-flight transactions in memory. When the feature is enabled with the two-phase commit interface, the information must be solely maintained by the coordinator service to prevent conflicts caused by participant services' inconsistent writes to the coordinator table, which may contain different transaction distributions over groups. Consequently, `com.scalar.db.transaction.consensuscommit.TwoPhaseConsensusCommit.commit()` must be called first by the coordinator service before being called by participant services.
```
josh-wong, line 1243: Is there a more accurate word we can use instead of `in-flight`, like `pending` or `in-progress` be accurate?

```suggestion
The group commit feature manages all in-flight transactions in memory. When the feature is enabled with the two-phase commit interface, the information must be solely maintained by the coordinator service to prevent conflicts caused by participant services' inconsistent writes to the Coordinator table, which may contain different transaction distributions over groups. Consequently, `com.scalar.db.transaction.consensuscommit.TwoPhaseConsensusCommit.commit()` must be called first by the coordinator service before being called by participant services.
```

josh-wong (changes requested):
Looking good! I've added some comments and suggestions. PTAL!

docs/api-guide.md
```diff
@@ -1197,6 +1197,90 @@ In the sample code, the transaction is retried three times maximum and sleeps fo
 
 <div class="notice--info">{{ notice--info | markdownify }}</div>
 
+### Group commit for the Coordinator table
+
+The Coordinator table used for Consensus Commit transactions is a vital data store, and it is recommended to use robust storage for it. However, utilizing more robust storage options, such as internally leveraging multi-AZ or multi-region replication, may lead to increased latency when writing records to the storage, resulting in poor throughput performance. ScalarDB provides a group commit feature for the Coordinator table that groups multiple record writes into a single write operation, improving write throughput. However, it may increase latency.
```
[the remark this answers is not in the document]
komamitsu: Good point. Basically group commit tends to increase latency as it adds an additional step to wait multiple requests. On the other hand, it may reduces workload on the database by reducing request, resulting in reducing the latency. So, we can say it may depends on the situation.

docs/api-guide.md
komamitsu: `ongoing transactions` might be better?

josh-wong: `ongoing transactions` sounds good to me! Thanks!

docs/api-guide.md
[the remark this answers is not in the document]
josh-wong: I see. How about the following then, which would make it a little more generic?

```suggestion
The Coordinator table that is used for Consensus Commit transactions is a vital data store, and using robust storage for it is recommended. However, utilizing more robust storage options, such as internally leveraging multi-AZ or multi-region replication, may lead to increased latency when writing records to the storage, resulting in poor throughput performance. 

ScalarDB provides a group commit feature for the Coordinator table that groups multiple record writes into a single write operation, improving write throughput. In this case, latency may increase or decrease, depending on your environment.
```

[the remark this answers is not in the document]
komamitsu: > depending on your environment

This sounded a bit too broad scope. How about "depending on the underlying database (or storage) and the workload" or something?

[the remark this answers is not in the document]
josh-wong: Thank you for clarifying! I wasn't sure what the scope was, so your suggestion sounds good to me!

```suggestion
The Coordinator table that is used for Consensus Commit transactions is a vital data store, and using robust storage for it is recommended. However, utilizing more robust storage options, such as internally leveraging multi-AZ or multi-region replication, may lead to increased latency when writing records to the storage, resulting in poor throughput performance.

ScalarDB provides a group commit feature for the Coordinator table that groups multiple record writes into a single write operation, improving write throughput. In this case, latency may increase or decrease, depending on the underlying database and the workload.
```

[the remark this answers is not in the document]
komamitsu: @josh-wong Thank you!

docs/api-guide.md
```diff
@@ -1197,6 +1197,92 @@ In the sample code, the transaction is retried three times maximum and sleeps fo
...
+### Group commit for the Coordinator table
+
+The Coordinator table that is used for Consensus Commit transactions is a vital data store, and using robust storage for it is recommended. However, utilizing more robust storage options, such as internally leveraging multi-AZ or multi-region replication, may lead to increased latency when writing records to the storage, resulting in poor throughput performance.
+
+ScalarDB provides a group commit feature for the Coordinator table that groups multiple record writes into a single write operation, improving write throughput. In this case, latency may increase or decrease, depending on the underlying database and the workload.
+
+To enable the group commit feature, add the following configuration:
+
+#### Configurations
+
```
josh-wong, line 1209: I don't think we need a heading here since we introduce the code block in the previous sentence (`To enable the group commit feature, add the following configuration:`).

```suggestion

```

docs/api-guide.md
```diff
@@ -1197,6 +1197,92 @@ In the sample code, the transaction is retried three times maximum and sleeps fo
...
+   DistributedTransaction transaction = manager.begin(myTxId);
+
+   ...
+
+   // When the group commit feature is enabled, a custom transaction ID passed by users can't be used as is.
+   // logger.info("The transaction state: {}", manager.getState(myTxId));
+   logger.info("The transaction state: {}", manager.getState(transaction.getId()));
+```
+
+##### Commit in Two-Phase Commit Interface
```
josh-wong, line 1243:
```suggestion
##### Commit in two-phase commit interface
```

docs/api-guide.md
```diff
@@ -1197,6 +1197,92 @@ In the sample code, the transaction is retried three times maximum and sleeps fo
...
+# scalar.db.consensus_commit.coordinator.group_commit.group_size_fix_timeout_millis=40
+# scalar.db.consensus_commit.coordinator.group_commit.delayed_slot_move_timeout_millis=800
+# scalar.db.consensus_commit.coordinator.group_commit.old_group_abort_timeout_millis=30000
+# scalar.db.consensus_commit.coordinator.group_commit.timeout_check_interval_millis=10
+# scalar.db.consensus_commit.coordinator.group_commit.metrics_monitor_log_enabled=true
+```
+
+#### Limitations
+
+##### Custom transaction ID passed by users
```
josh-wong, line 1224: So that we're not stacking headings without text between them, I've added the following sentence. Please feel free to revise it as necessary.

```suggestion
#### Limitations

This section describes the limitations of the group commit feature.

##### Custom transaction ID passed by users
```

josh-wong (approved):

brfrn169 (commented):
Overall LGTM, but I'll review this again later. Left several comments. Please take a look when you have time!

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
```diff
@@ -137,19 +157,34 @@ TwoPhaseCommitTransaction join(String txId, Isolation isolation, SerializableStr
       return resume(txId);
     }
 
-    return createNewTransaction(txId, isolation, strategy);
+    // Participant services don't use the group commit feature even if it's enabled. They simply use
```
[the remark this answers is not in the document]
komamitsu: Sure thing!

image

How GroupCommitter places transactions over multiple groups depends on various timings, including some types of timeouts (e.g., moving delayed transactions to a DelayedGroup) and events (e.g., `GroupCommitter.ready(SnapShot)` is called after the `prepare` and `validate` phases are done). So, it's unlikely that the Coordinator service and all the Participant services will have the same the groups with the same transaction distributions in-memory. So, if the participant services use the group commit feature, they may insert transaction groups different from other nodes, including the Coordinator service.

See
https://github.com/scalar-labs/scalardb/blob/ebbb0274bb01fd52dc553a333aff5183434c44b6/docs/api-guide.md#commit-in-two-phase-commit-interface also.

core/src/integration-test/java/com/scalar/db/storage/jdbc/ConsensusCommitCrossPartitionScanIntegrationTestWithJdbcDatabase.java
```diff
@@ -8,6 +8,6 @@ public class ConsensusCommitCrossPartitionScanIntegrationTestWithJdbcDatabase
 
   @Override
   protected Properties getProps(String testName) {
-    return JdbcEnv.getProperties(testName);
+    return ConsensusCommitJdbcEnv.getProperties(testName);
```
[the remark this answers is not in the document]
komamitsu: IIRC, this is a kind of trade-off topic between the cost and completeness, based on the memory of a discussion we had before. The reason why I picked JDBC is the current main target of the group commit feature is YugabyteDB. 

But, on second thought, it might be better to add the group commit integration tests with other storages. Maybe scheduled weekly integration test with all the storages is a good trade-off? If it's okay, I'll take care of it as another PR later. What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorGroupCommitter.java
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      CHARS_FOR_PRIMARY_KEY_SIZE = CHARS_FOR_PRIMARY_KEY.length;
+    }
+
+    // Use Random instead of ThreadLocalRandom in favor of global randomness.
+    private final Random random = new Random();
+
+    @Override
+    public String generateParentKey() {
```
feeblefakie, line 68: Random string generation might affect the performance, so it would be great if you could do a simple benchmark with multiple threads.

I previously encountered crazily slow random string generation when using an Apache Commons library, so I used a method taken from the benchbase one instead.
https://github.com/scalar-labs/scalardb-benchmarks/blob/master/src/main/java/com/scalar/db/benchmarks/ycsb/YcsbCommon.java#L109

So, it would be great if you could compare this with the one above.

feeblefakie (commented):
Overall, looking good!
Left some comments and suggestions. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -99,8 +171,11 @@ private Optional<Coordinator.State> get(Get get) throws CoordinatorException {
 
   @VisibleForTesting
   Put createPutWith(Coordinator.State state) {
-    return new Put(new Key(Attribute.toIdValue(state.getId())))
-        .withValue(Attribute.toStateValue(state.getState()))
+    Put put = new Put(new Key(Attribute.toIdValue(state.getId())));
+    if (!state.getChildIds().isEmpty()) {
+      put.withValue(Attribute.toChildIdsValue(Joiner.on(',').join(state.getChildIds())));
```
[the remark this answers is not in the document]
komamitsu: Good point! Yeah, I overlooked it. `State` class already knows how to handle the column including what's the delimiter. So [I took care of it in `State` class.](https://github.com/scalar-labs/scalardb/pull/1728/commits/e12a3d84fc5e523e3be4d05198b9b90d1bdd4b01)

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandlerWithGroupCommit.java
```diff
@@ -84,7 +84,10 @@ public void commitState(Snapshot snapshot)
     commitStateViaGroupCommit(snapshot);
   }
 
-  private void groupCommitState(String parentId, List<Snapshot> snapshots)
+  // TODO: Emitter interface should have `emitNormalGroup()` and `emitDelayedGroup()` separately and
```
komamitsu, line 87: Found there was room to improve this method and related ones to make which type of ID is passed clearer, so added this TODO.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -52,24 +62,86 @@ public class Coordinator {
...
   @SuppressFBWarnings("EI_EXPOSE_REP2")
   public Coordinator(DistributedStorage storage, ConsensusCommitConfig config) {
     this.storage = storage;
     coordinatorNamespace = config.getCoordinatorNamespace().orElse(NAMESPACE);
+    keyManipulator = new CoordinatorGroupCommitKeyManipulator();
   }
 
   public Optional<Coordinator.State> getState(String id) throws CoordinatorException {
+    if (keyManipulator.isFullKey(id)) {
+      // The following read operation order is important. Reading a coordinator state is likely to
```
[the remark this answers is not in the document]
komamitsu: Thanks for the feedback! I agree with you. I moved these logics for the group commit to [getStateForGroupCommit()](https://github.com/scalar-labs/scalardb/pull/1728/commits/dc5395653b325c71a4e2742eb8a8f15a0a6e3418#diff-8049e66d11b81cdd5a1dc82c4977b5336741bcee24dd7ddd54c6a6940f09497bR86-R105) and revised the comment. What you think?

Of course, chatting on the comment is welcome!

docs/configurations.md
```diff
@@ -38,16 +38,23 @@ The following basic configurations are available for the Consensus Commit transa
...
+| `scalar.db.consensus_commit.coordinator.group_commit.old_group_abort_timeout_millis`   | Timeout to abort an old ongoing group. A small value reduces resource consumption through aggressive aborts, but may also increase the likelihood of unnecessary aborts for long-running transactions.                                                                                | `60000`                                                           |
+| `scalar.db.consensus_commit.coordinator.group_commit.timeout_check_interval_millis`    | Interval for checking the group commit–related timeouts.                                                                                                                                                                                                                              | `20`                                                              |
```
Torch3333, line 56: Just a thought: I suppose finding the right configuration to achieve great performance depends on many factors and requires a lot of testing and monitoring. How do you estimate the degree of difficulty for customers to acquire this know-how?

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -52,24 +62,86 @@ public class Coordinator {
...
+      // occur for lazy recovery. The likelihood of lazy recovery attempts would increase if a
+      // transaction scanned by another transaction executing lazy recovery is delayed. In the group
+      // commit, a delayed transaction is specified with a full transaction ID, not with a parent
+      // transaction ID. Therefore, looking up with the full transaction ID should be tried first
+      // to minimize read operations as much as possible.
+
+      // Scan with the full ID for a delayed group that contains only a single transaction.
```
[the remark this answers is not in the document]
komamitsu: I [moved this logic](https://github.com/scalar-labs/scalardb/pull/1728/commits/dc5395653b325c71a4e2742eb8a8f15a0a6e3418#diff-8049e66d11b81cdd5a1dc82c4977b5336741bcee24dd7ddd54c6a6940f09497bR86-R105) for the group commit to `getStateForGroupCommit()` to make things clearer. 

> regarding the `Scan with the full ID`, should it be something like `Regard the ID as a full ID and scan with the ID` since we don't know if the given ID is a full ID or not.

Ah, whether the ID is a full ID is already checked. But I changed the argument name to a more clear one in the commit.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -52,24 +62,86 @@ public class Coordinator {
...
+      // The normal lookup logic can be used here.
+      Optional<State> state = get(createGetWith(id));
+      if (state.isPresent()) {
+        return state;
+      }
+
+      // Scan with the parent ID for a normal group that contains multiple transactions with
```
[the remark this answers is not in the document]
komamitsu: A parent ID is explicitly extracted from a full ID here. So, we can go with the current comment as is?

docs/configurations.md
komamitsu: Really good question... As you thought, many factors including underlying database spec and access pattern by the applications are related. I think it would be worth trying 2 or 3 values per configuration except for `old_group_abort_timeout_millis` and `timeout_check_interval_millis`. In the case of `slot_capacity`, trying `20`, `40`, `60` may be good to find the best configuration. But the combination of these configurations is also important. So, the number of attempts would be 3 (for `slot_capacity`) times 2 (for `group_size_fix_timeout_millis`) times 2 (for `delayed_slot_move_timeout_millis)`, then 12. It might be a bit tough for users, though.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorGroupCommitter.java
komamitsu: Thanks for sharing the concern. That makes sense.

I benchmarked the following combinations with [this commit](https://github.com/scalar-labs/scalardb/commit/3b2be71c72b6911655226143675fd92317c5bd2a).

a. Simple impl (the original `generateParentKey()` impl) + `Random` (as same as the current impl)
b. Faster impl (ported from `scalardb-benchmark`) + `Random`
c. Simple impl (the original `generateParentKey()` impl) + `ThreadLocalRandom`
d. Faster impl (ported from `scalardb-benchmark`) + `ThreadLocalRandom`

image

What I noticed from the result are:
- The implementation ported from `scalardb-benchmark` is about 10-15% faster than the original simple `generateParentKey()` impl when using `ThreadLocalRandom`
- Which to use `Random` or `ThreadLocalRandom` significantly affects the performance along with the number of threads. The average duration was 0.015 ms in case `a` with 16 threads which is the same number of the Linux machine's CPUs. The duration would increase when using a larger number of CPU machine

I think we should use `ThreadLocalRandom` just in case. When I benchmarked the group commit feature with `ThreadLocalRandom` before, I observed a lot of parent key conflicts at the beginning of benchmarking. I'll come up with a way to mitigate the issue and update this PR.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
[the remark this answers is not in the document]
brfrn169: Understood. Thank you for your explanation!

So the way to use the two-phase interface varies depending on whether group commit is enabled or not, which might be confusing for users. One option is to always disable group commit when using the two-phase interface. What do you think?

docs/configurations.md
Torch3333: I see. Thanks for laying out a practical approach to get closer to a good configuration. Adding that to the documentation would already be quite helpful for users.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorGroupCommitter.java
komamitsu: @feeblefakie BTW, I checked the distribution of the random character values when using the implementation ported from `scalardb-benchmark`. It seems skewed compared to the original simple impl. I also observed a similar distribution when directly testing https://github.com/scalar-labs/scalardb-benchmarks/blob/master/src/main/java/com/scalar/db/benchmarks/ycsb/YcsbCommon.java#L109.

I think it's okay to generate just a random value, but the simple one would be better when generating unique identifiers.

- The implementation ported from `scalardb-benchmark`.
```
[000] 12170
[001] 21654
[002] 13412
[003] 23733
[004] 15577
[005] 26786
[006] 15973
[007] 56956
[008] 14013
[009] 26212
[010] 14705
[011] 107740
[012] 13730
[013] 28530
[014] 19633
[015] 106851
[016] 10670
[017] 29521
[018] 19581
[019] 61491
[020] 17591
[021] 46623
[022] 22265
[023] 48223
[024] 16174
[025] 58574
[026] 17782
[027] 44353
[028] 16499
[029] 69603
[030] 18314
[031] 40725
[032] 8898
[033] 73426
[034] 17933
[035] 47698
[036] 12817
[037] 81226
[038] 18393
[039] 30052
[040] 12386
[041] 64607
[042] 14837
[043] 53717
[044] 15901
[045] 73947
[046] 17029
[047] 30590
[048] 11894
[049] 34586
[050] 22686
[051] 52105
[052] 16454
[053] 43558
[054] 22372
[055] 33789
[056] 15510
[057] 24684
[058] 17335
[059] 28752
[060] 18036
[061] 29118
```

- The original simple implementation
```
[000] 39054
[001] 38329
[002] 38803
[003] 39094
[004] 38654
[005] 38724
[006] 38692
[007] 38661
[008] 38667
[009] 38810
[010] 38764
[011] 38945
[012] 38641
[013] 38578
[014] 38603
[015] 38749
[016] 38753
[017] 38709
[018] 38565
[019] 38706
[020] 38555
[021] 38993
[022] 38613
[023] 38477
[024] 38939
[025] 38521
[026] 38813
[027] 38752
[028] 38829
[029] 38728
[030] 38606
[031] 38984
[032] 38803
[033] 38611
[034] 39012
[035] 38577
[036] 38516
[037] 38485
[038] 38580
[039] 38982
[040] 38378
[041] 38650
[042] 38782
[043] 38494
[044] 38518
[045] 38574
[046] 38892
[047] 38815
[048] 38335
[049] 38681
[050] 38510
[051] 38572
[052] 38581
[053] 38473
[054] 38734
[055] 38822
[056] 39052
[057] 39031
[058] 38996
[059] 38902
[060] 38711
[061] 38620
```

komamitsu: > When I benchmarked the group commit feature with ThreadLocalRandom before, I observed a lot of parent key conflicts at the beginning of benchmarking

This was my silly mistake. I stored the return value from `ThreadLocalRandom.current()`, but I should've not done that...

> Usages of this class should typically be of the form: ThreadLocalRandom.current().nextX(...) (where X is Int, Long, etc). When all usages are of this form, it is never possible to accidently share a ThreadLocalRandom across multiple threads.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html

Fixed it in https://github.com/scalar-labs/scalardb/pull/1728/commits/755b91144b7ab8f165f53799a1963f723e0b2dac

core/src/integration-test/java/com/scalar/db/storage/jdbc/ConsensusCommitCrossPartitionScanIntegrationTestWithJdbcDatabase.java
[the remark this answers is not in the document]
komamitsu: Yes. In the case of https://github.com/scalar-labs/scalardb/actions/runs/9202687096, the group commit integration test using PostgreSQL which executed only `**ConsensusCommit**` cases took 12 minutes, while the full integration test using PostgreSQL without the group commit feature took only 2 minutes. (It was basically because of frequent open/close `CoordinatorGroupCommitter` and garbage ongoing transactions that were not explicitly rollbacked, though.)

So, this is a very rough estimate, but it may take 6 times larger computation resources of the GitHub Actions in total. I think the combination of 1) the group commit integration test only using PostgreSQL in the regular workflow, and 2) testing using all the storages with the group commit feature enabled in a weekly or daily workflow. 1 is already implemented in this PR, and 2 will be added in another PR later. What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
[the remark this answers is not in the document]
komamitsu: > So the way to use the two-phase interface varies depending on whether group commit is enabled or not, which might be confusing for users.

Yes. Users need to change the commit order in their applications as a part of the limitation, and it may be confusing.

> One option is to always disable group commit when using the two-phase interface

You mean to throw an exception or something if `TwoPhaseCommitTransactionManager` is used with the group commit feature enabled, right? Indeed, the idea sounds simpler and less confusing. On the other hand, I'm wondering what if a user wants to use the two-phase commit interface on a high-latency underlying storage where the coordinator table is placed 🤔

docs/configurations.md
komamitsu: Thanks for the suggestion! I added [a footnote](https://github.com/scalar-labs/scalardb/compare/bdad881...e466a16403e72fb8688bd0990db4dc2b001c5666) for that since the table would be complicated if I put the note in the table. 

@josh-wong Can you take a look at the diff when you get a chance? Thanks

core/src/integration-test/java/com/scalar/db/storage/jdbc/ConsensusCommitCrossPartitionScanIntegrationTestWithJdbcDatabase.java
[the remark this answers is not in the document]
brfrn169: Okay, that sounds good. I thought we should run CI for the group commit feature with other storages as much as possible before pushing changes, because we should detect bugs early. However, it seems like this significantly impacts the regular CI time.

Let's add item 2 in another PR. Also, maybe we should add the `workflow_dispatch` trigger to the workflow so that we can manually run the CI for the group commit feature with other storages when we make changes related to the group commit feature.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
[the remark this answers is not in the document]
brfrn169: > You mean to throw an exception or something if TwoPhaseCommitTransactionManager is used with the group commit feature enabled, right?

Yes, I meant that.

> On the other hand, I'm wondering what if a user wants to use the two-phase commit interface on a high-latency underlying storage where the coordinator table is placed 🤔

Correct. But, actually, there are also some other optimizations and features we have to give up for the two-phase commit interface.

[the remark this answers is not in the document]
komamitsu: I see. That sounds a reasonable concern.

I guess it's related to the product specification. @feeblefakie Do you think we can/should prohibit the group commit feature in the two-phase commit interface?

feeblefakie (approved):
LGTM! Thank you!

docs/api-guide.md
```diff
@@ -1197,6 +1197,91 @@ In the sample code, the transaction is retried three times maximum and sleeps fo
...
 <div class="notice--info">{{ notice--info | markdownify }}</div>
 
+### Group commit for the Coordinator table
+
+The Coordinator table that is used for Consensus Commit transactions is a vital data store, and using robust storage for it is recommended. However, utilizing more robust storage options, such as internally leveraging multi-AZ or multi-region replication, may lead to increased latency when writing records to the storage, resulting in poor throughput performance.
+
+ScalarDB provides a group commit feature for the Coordinator table that groups multiple record writes into a single write operation, improving write throughput. In this case, latency may increase or decrease, depending on the underlying database and the workload.
+
+To enable the group commit feature, add the following configuration:
+```properties
```
josh-wong, line 1207:
````suggestion
To enable the group commit feature, add the following configuration:

```properties
````

File: docs/configurations.md [hunk 18]
```diff
@@ -38,16 +38,23 @@ The following basic configurations are available for the Consensus Commit transa
...
+| `scalar.db.consensus_commit.coordinator.group_commit.enabled`                          | Whether or not committing the transaction state is executed in batch mode.                                                                                                                                                                                                                | `false`                                                           |
+| `scalar.db.consensus_commit.coordinator.group_commit.slot_capacity`                    | Maximum number of slots in a group for the group commit feature. A large value improves the efficiency of group commit, but may also increase latency and the likelihood of transaction conflicts[^1].                                                                                    | `20`                                                              |
```
josh-wong, line 52:
```suggestion
| `scalar.db.consensus_commit.coordinator.group_commit.slot_capacity`                    | Maximum number of slots in a group for the group commit feature. A large value improves the efficiency of group commit, but may also increase latency and the likelihood of transaction conflicts.[^1]                                                                                    | `20`                                                              |
```

docs/configurations.md
```diff
@@ -38,16 +38,23 @@ The following basic configurations are available for the Consensus Commit transa
...
+| `scalar.db.consensus_commit.coordinator.group_commit.group_size_fix_timeout_millis`    | Timeout to fix the size of slots in a group. A large value improves the efficiency of group commit, but may also increase latency and the likelihood of transaction conflicts[^1].                                                                                                        | `40`                                                              |
```
josh-wong, line 53:
```suggestion
| `scalar.db.consensus_commit.coordinator.group_commit.group_size_fix_timeout_millis`    | Timeout to fix the size of slots in a group. A large value improves the efficiency of group commit, but may also increase latency and the likelihood of transaction conflicts.[^1]                                                                                                        | `40`                                                              |
```

docs/configurations.md
```diff
@@ -38,16 +38,23 @@ The following basic configurations are available for the Consensus Commit transa
...
+| `scalar.db.consensus_commit.coordinator.group_commit.delayed_slot_move_timeout_millis` | Timeout to move delayed slots from a group to another isolated group to prevent the original group from being affected by delayed transactions. A large value improves the efficiency of group commit, but may also increase the latency and the likelihood of transaction conflicts[^1]. | `1200`                                                            |
```
josh-wong, line 54:
```suggestion
| `scalar.db.consensus_commit.coordinator.group_commit.delayed_slot_move_timeout_millis` | Timeout to move delayed slots from a group to another isolated group to prevent the original group from being affected by delayed transactions. A large value improves the efficiency of group commit, but may also increase the latency and the likelihood of transaction conflicts.[^1] | `1200`                                                            |
```

docs/configurations.md
```diff
@@ -271,3 +278,5 @@ scalar.db.contact_points=indirect:<SCALARDB_CLUSTER_CONTACT_POINT>
 ```
 
 For details about client configurations, see the ScalarDB Cluster [client configurations (redirects to the Enterprise docs site)](https://scalardb.scalar-labs.com/docs/latest/scalardb-cluster/developer-guide-for-scalardb-cluster-with-java-api/#client-configurations).
+
+[^1]: It's worth benchmarking the performance with a few variations (e.g., 75% and 125% of the default value) on the same underlying storage that your application uses, considering your application's access pattern, to determine the optimal configuration as it really depends on those factors. Also, it's important to benchmark the combinations of these parameters (e.g., First, `slot_capacity:20` and `group_size_fix_timeout_millis:40`. Second, `slot_capacity:30` and `group_size_fix_timeout_millis:40`. Third, `slot_capacity:20` and `group_size_fix_timeout_millis:80`) in order to determine the optimal combination.
```
josh-wong, line 282:
```suggestion
[^1]: It's worth benchmarking the performance with a few variations (for example, 75% and 125% of the default value) on the same underlying storage that your application uses, considering your application's access pattern, to determine the optimal configuration as it really depends on those factors. Also, it's important to benchmark combinations of these parameters (for example, first, `slot_capacity:20` and `group_size_fix_timeout_millis:40`; second, `slot_capacity:30` and `group_size_fix_timeout_millis:40`; and third, `slot_capacity:20` and `group_size_fix_timeout_millis:80`) to determine the optimal combination.
```

josh-wong (approved):

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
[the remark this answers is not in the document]
komamitsu: @brfrn169 I had a chance to talk with @feeblefakie about that and he agreed with you. I'll revise this PR.

> Users need to change the commit order in their applications as a part of the limitation

BTW, although this is a kind of memo for me about future development when we'll enable the group commit with two-phase commit interface, we should add a check whether the transaction is committed in the coordinator table,  before committing records on participant services.

brfrn169 (approved):
LGTM! Thank you!

PR #1748, opened 2024-05-20
Remove broken links on schema-loader-import

komamitsu:
## Description

We noticed some internal links on `docs/schema-loader-import.md` were [broken](https://github.com/scalar-labs/scalardb/pull/1710#discussion_r1599275046). It seems because the DOM id was changed by implicitly adding a prefix. I guess specifying id with the prefix isn't so robust. So, in this PR, we just replacing the broken links with "See warning (x) below" comment.

Another option is to use footnotes. But the information in the warning block is important, so this PR leaves the warning block as is.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/1710#discussion_r1599275046

## Changes made

- Replace the broken links with "See warning (x) below" comment

## Checklist

None

## Release notes

N/A

### DIFF
--- a/docs/schema-loader-import.md
+++ b/docs/schema-loader-import.md
@@ -89,29 +89,29 @@ The following table shows the supported data types in each JDBC database and the
 
 | MySQL        | ScalarDB | Notes                 |
 |--------------|----------|-----------------------|
-| bigint       | BIGINT   | [*1](#warn-data-size) |
+| bigint       | BIGINT   | See warning (1) below |
 | binary       | BLOB     |                       |
 | bit          | BOOLEAN  |                       |
-| blob         | BLOB     | [*2](#warn-data-size) |
-| char         | TEXT     | [*2](#warn-data-size) |
+| blob         | BLOB     | See warning (2) below |
+| char         | TEXT     | See warning (2) below |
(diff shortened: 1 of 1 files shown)


komamitsu:
> Instead, how about if we use the standard [footnote feature in Markdown](https://www.markdownguide.org/extended-syntax/#footnotes)? This makes a "Footnote" section at the end of the doc that users can refer to.[^1]

@josh-wong Just an idea. But do you think it's possible to wrap (or decorate?) a footnote section to make it like a warning block? If it's possible, I'll update this PR.

josh-wong:
@komamitsu 
> @josh-wong Just an idea. But do you think it's possible to wrap (or decorate?) a footnote section to make it like a warning block? If it's possible, I'll update this PR.

Thanks for chatting with me about this via video chat!

As I showed in the local example during our discussion, let's use HTML to create an ordered list that includes anchors within the "Warning" admonition. Since the syntax in this doc will be updated to the syntax we use in Docusaurus soon, I will make that change after we've copied the docs in this repository to the ScalarDB docs repository. 

Also, I'll apply the same fix to other versions of this doc since those links are also broken on the docs site.

PR #1755, opened 2024-05-21
Allow directly executing CRUD operations with transaction managers

brfrn169:
## Description

This PR allows directly executing CRUD operations with transaction managers.

## Related issues and/or PRs

N/A

## Changes made

- Added CRUD operation methods (`get()`, `scan()`, `put()`, `insert()`, `upsert()`, `update()`, `delete()`, and `mutate()`) to `DistributedTransactionManager` and `TwoPhaseCommitTransactionManager`
  - Introduced `CrudOperable`, and made `TransactionCrudOperable` a subclass of `CrudOperable`
  - Also introduced `TransactionManagerCrudOperable` as a subclass of `CrudOperable`
  - Made `DistributedTransactionManager` and `TwoPhaseCommitTransactionManager` extend `TransactionManagerCrudOperable`

## Checklist

N/A

## Release notes

Allowed directly executing CRUD operations with transaction managers.


core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
```diff
@@ -1,56 +1,41 @@
...
 import java.util.Optional;
 
-/**
- * An interface for transactional CRUD operations. Note that the LINEARIZABLE consistency level is
- * always used in transactional CRUD operations, so {@link Consistency} specified for CRUD
- * operations is ignored.
- */
-public interface TransactionCrudOperable {
+/** An interface for transactional CRUD operations for transactions. */
+public interface TransactionCrudOperable extends CrudOperable<CrudException> {
```
komamitsu, line 10: This might be just a naming issue. IIUC, `CrudException` inherits from `TransactionException`, and `TransactionCrudOperable` inherits from `CrudOperable`. Their inheritance directions seem reversed from the perspective of class names. It's a bit confusing?

core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionManager.java
```diff
@@ -76,151 +71,159 @@ public Optional<String> getTable() {
...
 
-    @Override
-    public void upsert(Upsert upsert) throws CrudException {
-      checkIfActive();
-      super.upsert(upsert);
+  private void rollbackTransaction(DistributedTransaction transaction) throws CrudException {
+    try {
+      transaction.rollback();
+    } catch (RollbackException e) {
+      throw new CrudException(e.getMessage(), e, e.getTransactionId().orElse(null));
```
komamitsu, line 194: Uncommitted transaction will be rolled back, so just warning message or something is enough here?

core/src/main/java/com/scalar/db/common/TransactionDecorationTwoPhaseCommitTransactionManager.java
```diff
@@ -0,0 +1,224 @@
...
+  /**
+   * This class is to unify the call sequence of the transaction object. It doesn't care about the
+   * potential inconsistency between the status field on JVM memory and the underlying persistent
+   * layer.
+   */
+  @VisibleForTesting
+  static class StateManagedTransaction extends DecoratedTwoPhaseCommitTransaction {
+
+    private enum Status {
+      ACTIVE,
```
komamitsu, line 65: [minor] This is a bit off topic. But `PREPARED` and `VALIDATED` may seem active as well. So, `EXECUTING` or `CRUD` or something might be less confusing?

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
brfrn169: Good point. Thanks.

Actually, this naming strongly depends on backward compatibility. `TransactionCrudOperable` is an existing name, so we can't rename it to preserve backward compatibility.

The requirements here are as follows:

- Add CRUD operation methods to the transaction manager interfaces (`DistributedTransactionManager` and `TwoPhaseCommitTransactionManager`).
- Access the CRUD operation methods through the same abstraction for the transaction manager interfaces and the transaction interfaces (`DistributedTransaction` and `TwoPhaseCommitTransaction`).
- The exceptions thrown by the CRUD operation methods differ slightly between the transaction manager interfaces and the transaction interfaces.

Considering these requirements, we finally decided on the current naming. But do you have any better ideas for the naming?

core/src/main/java/com/scalar/db/common/TransactionDecorationTwoPhaseCommitTransactionManager.java
brfrn169: Yes, you are right. Let's discuss this in a separate PR. Thanks.

core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionManager.java
brfrn169: Correct. Let me add a warning message here. Thanks.

brfrn169: Fixed in 0b1c67ce83a356f118a5e8ab635a716fd563413b. Thanks.

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
komamitsu: Thanks for sharing the background.

What we can change is only the name of `CrudOperable` newly added in this PR. It seems to be only inherited by `TransactionManagerCrudOperable` and `TransactionCrudOperable`. How about renaming it to `TransactionCrudOperableBase` or something so that the inheritance relationships will be as follows?
- `TransactionCrudOperableBase`
  - `TransactionCrudOperable`
  - `TransactionManagerCrudOperable`
- `TransactionException`
  - `CrudException`

Current
- `CrudOperable`
  - `TransactionCrudOperable`
  - `TransactionManagerCrudOperable`
- `TransactionException`
  - `CrudException`

brfrn169: Actually, `TransactionManagerCrudOperable` is also newly added in this PR, so we can rename it as well.

The naming `XXXBase` sounds more appropriate for abstract classes to me, so I'm not sure it's suitable for interfaces. What do you think?

komamitsu: I see. I think `TransactionManagerCrudOperable` is alined with `TransactionCrudOperable`, so it can be left as is. Regarding `XxxxBase`, indeed it's less common. I'm also not sure if we should avoid it. Another option is `TransactionalCrudOperable` based on the comment `An interface for transactional CRUD operations`? It might be a bit confusing, though.

core/src/main/java/com/scalar/db/common/AbstractTwoPhaseCommitTransactionManager.java
```diff
@@ -12,34 +11,35 @@
...
-import java.util.concurrent.CopyOnWriteArrayList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public abstract class AbstractTwoPhaseCommitTransactionManager
-    implements TwoPhaseCommitTransactionManager, TwoPhaseCommitTransactionDecoratorAddable {
+    implements TwoPhaseCommitTransactionManager {
+
+  private static final Logger logger =
+      LoggerFactory.getLogger(AbstractDistributedTransactionManager.class);
```
Torch3333, line 35:
```suggestion
      LoggerFactory.getLogger(AbstractTwoPhaseCommitTransactionManager.class);
```

Torch3333 (approved):
Besides this comment, LGTM, thank you.

core/src/main/java/com/scalar/db/api/TransactionCrudOperable.java
brfrn169: Yeah, `TransactionalCrudOperable` is a possible idea. But, as you mentioned, it might still be confusing between `TransactionalCrudOperable` and `TransactionCrudOperable`.

If we consider `CrudOperable` vs. `TransactionalCrudOperable`, I prefer `CrudOperable`. But that's just my preference. If you have any other better ideas, strong opinions, or strong reasons, please let me know. Thanks.

core/src/main/java/com/scalar/db/common/AbstractTwoPhaseCommitTransactionManager.java
brfrn169: Good catch! Fixed in 194e18dae02328c1901b4d9b5254f80044b0adaa. Thanks.

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

PR #1763, opened 2024-05-23
Add arbitrary filtering for partition scan and index scan

jnmt:
## Description

This PR adds arbitrary filtering for partition scan and index scan.

## Related issues and/or PRs

- #1682

## Changes made

- Add interfaces in `ScanBuilder` so that users can write the where clause in `Scan` and `ScanWithIndex`
- Update the query builder for JDBC databases to push down conditions
- Update the check for preventing scan-after-write to care for arbitrary conditions

## Checklist

- [x] I have updated the documentation to reflect the changes. **Will be updated in another PR, including get operations with filitering**

- With this change, `Scan` and `ScanWithIndex` with filtering are supported not only on JDBC databases but also on NoSQL databases due to #1682.
- Arbitrary filtering for get operations will be added in an upcoming PR.

## Release notes

Added arbitrary filtering for partition scan and index scan.


core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -107,13 +107,12 @@ private Optional<Result> createGetResult(Snapshot.Key key, List<String> projecti
...
-    // cross-partition scan, and thus, the write command may not have columns used in the condition,
-    // which are necessary to determine overlaps. For a scan with clustering keys, we can determine
-    // overlaps without the actual scan, but we also check it here for consistent logic and
-    // readability.
+    // We verify if this scan does not overlap previous writes using the actual scan result. Because
+    // we support arbitrary conditions in the where clause of a scan (not only ScanAll, but also
+    // Scan and ScanWithIndex), we cannot determine whether the scan results will include a record
+    // whose key is the same as the key specified in the previous writes, without knowing the
+    // obtained keys in the actual scan. With this check, users can avoid seeing unexpected scan
+    // results that have not included previous writes yet.
```
jnmt, line 115: Now, we are about to support arbitrary filtering in all scan operations, and we need to use actual scan results in all types of scans to prevent scan-after-write properly. So, I changed the comment a little.

core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
```diff
@@ -258,8 +260,8 @@ private void checkOrderings(ScanAll scanAll, TableMetadata metadata) {
     }
   }
 
-  private void checkConjunctions(ScanAll scan, TableMetadata metadata) {
-    for (Conjunction conjunction : scan.getConjunctions()) {
+  private void checkConjunctions(Selection selection, TableMetadata metadata) {
+    for (Conjunction conjunction : selection.getConjunctions()) {
```
jnmt, line 264: Changed to `Selection` since get operations will use this check as well in an upcoming PR.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageIntegrationTestBase.java
```diff
@@ -526,6 +526,37 @@ public void scan_ScanWithLimitGiven_ShouldReturnGivenNumberOfResults()
         .isEqualTo(Optional.of(new IntValue(COL_NAME4, 2)));
   }
 
+  @Test
+  public void
+      scan_ScanWithClusteringKeyRangeAndConjunctionsGiven_ShouldRetrieveResultsOfBothConditions()
```
jnmt, line 531: I only added very basic tests for `Scan`/` ScanWithIndex` with the where clause since various patterns of conditions are being tested with cross-partition scan filtering.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -230,14 +230,21 @@ private boolean isWriteSetOverlappedWith(Scan scan) {
...
+      }
 
+      Put put = entry.getValue();
       if (!put.forNamespace().equals(scan.forNamespace())
           || !put.forTable().equals(scan.forTable())
           || !put.getPartitionKey().equals(scan.getPartitionKey())) {
         continue;
       }
 
+      if (!areConjunctionsOverlapped(put, scan)) {
```
komamitsu, line 244: I tried to remove this condition and ran the integration test. But any test case didn't fail. So, maybe there is room to add a test case for this?

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -205,9 +218,253 @@ public Scan build() {
...
+
+    @Override
+    public BuildableScanWithOngoingWhereOr or(AndConditionSet andConditionSet) {
+      checkNotNull(andConditionSet);
+      where.or(andConditionSet);
+      return this;
+    }
+  }
+
+  public static class BuildableScanWithWhere extends BuildableScan {
```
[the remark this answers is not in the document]
jnmt: No, it cannot since I think we should be able to do something like this.

```java
    BuildableScanWithWhere buildable =
        Scan.newBuilder()
            .namespace("cassandrans")
            .table("lineitem")
            .partitionKey(Key.ofInt("l_orderkey", 807))
            .where(ConditionBuilder.column("l_discount").isGreaterThanDouble(0.04));

    Scan scan1 = buildable.build();
    Scan scan2 = buildable.start(Key.ofInt("l_linenumber", 2)).build();
```

JFYI: I noticed `BuildableScanWithWhere` should not inherit `BuildableScan` above since `BuildableScan` does not build a scan with the where clause as long as users intentionally use (or cast) the `WithWhere` object, and users hardly know they should use it. I will fix it. Thank you for the chance to notice it.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: Definitely. We can test the case where clustering keys are overlapped, but conjunctions are not overlapped with put (it should not throw exceptions). Good catch, thank you!

core/src/main/java/com/scalar/db/api/ScanBuilder.java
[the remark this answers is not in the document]
jnmt: Sorry, I misunderstood it. We can make it `abstract`, but still I think it can be as is because we do not intend that users always willingly use the base one, but rather intend to properly use both the base one and inherited ongoing ones based on the situation.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
jnmt: Added in 383f30d.

Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM! 👍

brfrn169 (commented):
Overall LGTM. Left a super minor comment. Please take a look when you have time!

Let me review this later again before approving it just in case. Thanks.

core/src/main/java/com/scalar/db/api/ScanBuilder.java
```diff
@@ -835,15 +1279,6 @@ private void checkNotScanWithIndex() {
       }
     }
 
-    private void checkScanAll() {
-      if (!isScanAll) {
-        throw new UnsupportedOperationException(
-            CoreError
-                .SCAN_BUILD_ERROR_OPERATION_SUPPORTED_ONLY_WHEN_SCANNING_ALL_RECORDS_OF_DATABASE
```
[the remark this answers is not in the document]
jnmt: Right. Removed in 3583def. Good catch, thank you!

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #1793, opened 2024-05-28
Add single CRUD operation transaction manager

brfrn169:
## Description

This PR adds a single CRUD operation transaction manager. This transaction manager implementation does not allow beginning a transaction by calling `begin()`/`start()`. It only allows directly executing CRUD operations from the transaction manager.

Users can use this transaction manager implementation by specifying `single-crud-operation` in the `scalar.db.transaction_manager` property:

```properties
scalar.db.transaction_manager=single-crud-operation
```

## Related issues and/or PRs

N/A

## Changes made

- Added single CRUD operation transaction manager implementation.

## Checklist

N/A

## Release notes

Added a single CRUD operation transaction manager. This transaction manager implementation does not allow beginning a transaction by calling `begin()`/`start()`. It only allows directly executing CRUD operations from the transaction manager.


core/src/main/java/com/scalar/db/api/DistributedTransactionProvider.java
```diff
@@ -39,7 +40,10 @@ public interface DistributedTransactionProvider {
    * Creates an instance of {@link TwoPhaseCommitTransactionManager} for the transaction manager.
    *
    * @param config a database config
-   * @return an instance of {@link TwoPhaseCommitTransactionManager} for the transaction manager
+   * @return an instance of {@link TwoPhaseCommitTransactionManager} for the transaction manager. If
+   *     the transaction manager does not support the two-phase commit interface, returns {@code
+   *     null}.
    */
+  @Nullable
   TwoPhaseCommitTransactionManager createTwoPhaseCommitTransactionManager(DatabaseConfig config);
```
brfrn169, line 48: Made this method return null if the transaction manager implementation doesn't support the two-phase commit interface.

core/src/main/java/com/scalar/db/transaction/singlecrudoperation/SingleCrudOperationTransactionManager.java
```diff
@@ -0,0 +1,364 @@
...
+    super(databaseConfig);
+    this.storage = storage;
+  }
+
+  @Override
+  public DistributedTransaction begin() throws TransactionException {
+    throw new UnsupportedOperationException(
+        CoreError.SINGLE_CRUD_OPERATION_TRANSACTION_BEGINNING_TRANSACTION_NOT_ALLOWED
+            .buildMessage());
+  }
```
brfrn169, line 70: `SingleCrudOperationTransactionManager` doesn't allow beginning a transaction.

core/src/main/java/com/scalar/db/transaction/singlecrudoperation/SingleCrudOperationTransactionManager.java
```diff
@@ -0,0 +1,364 @@
...
+
+  @Override
+  public DistributedTransaction resume(String txId) throws TransactionNotFoundException {
+    throw new UnsupportedOperationException(
+        CoreError.SINGLE_CRUD_OPERATION_TRANSACTION_RESUMING_TRANSACTION_NOT_ALLOWED
+            .buildMessage());
+  }
+
+  @Override
+  public Optional<Result> get(Get get) throws CrudException {
```
brfrn169, line 145: `SingleCrudOperationTransactionManager` only allows executing CRUD operations.

integration-test/src/main/java/com/scalar/db/transaction/singlecrudoperation/SingleCrudOperationTransactionAdminIntegrationTestBase.java
```diff
@@ -0,0 +1,94 @@
...
+  @Disabled("Single CRUD operation transactions don't have Coordinator tables")
+  @Test
+  @Override
+  public void dropCoordinatorTables_IfExist_CoordinatorTablesDoNotExist_ShouldNotThrowAnyException()
+      throws ExecutionException {
+    super.dropCoordinatorTables_IfExist_CoordinatorTablesDoNotExist_ShouldNotThrowAnyException();
+  }
+
+  @Test
+  @Disabled("This case is not tested for Single CRUD operation transactions")
```
josh-wong, line 85:
```suggestion
  @Disabled("This case is not tested for single CRUD operation transactions")
```

josh-wong (approved):
I left one minor suggestion. Other than that, LGTM. Thank you!🙇🏻‍♂️

core/src/main/java/com/scalar/db/transaction/singlecrudoperation/SingleCrudOperationTransactionAdmin.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  @Override
+  public void truncateCoordinatorTables() {
+    // Do nothing since the single CRUD operation transactions don't have coordinator tables
+  }
+
+  @Override
+  public boolean coordinatorTablesExist() {
+    // Always return true since the single CRUD operation transactions don't have coordinator tables
```
komamitsu, line 143: > since the single CRUD operation transactions don't have coordinator tables

In this case, I think returning true is a confusing behavior since it's not checked actually. How about throwing `UnsupportedOperationException`?

Regarding other empty methods (e.g., `createCoordinatorTables`), users might think the operation finished successfully if the API call finishes quietly. It might be better to throw `UnsupportedOperationException` as well when they are called?

brfrn169: Good point. I didn't do that because if we do so, the application developer would need to switch the code depending on whether they are using Consensus Commit or not. To avoid this code switch, we always return `true` in `coordinatorTablesExist()` and do nothing in the other coordinator table-related methods. What do you think?

komamitsu: That sounds reasonable. I also saw the same implementation in `JdbcTransactionAdmin`.

I'm okay with the current implementation as-is, but how about adding a warning message just in case?

```java
  @Override
  public boolean coordinatorTablesExist() {
    logger.warn("Always true is returned since the single CRUD operation transactions don't have coordinator tables");  
```

brfrn169: Yeah, we can add warning messages for the coordinator table-related methods.

One thing I'm concerned about is that in ScalarDB Cluster, the cluster nodes try to create the coordinator tables if they don't exist. So, if we add warning messages, the messages are shown in the log of the cluster nodes, which might confuse users if they are using a non Consensus Commit transaction manager. What do you think?

komamitsu: I see. My idea on the concern is as follows:

- With the single CRUD operation transaction mode, those Admin APIs for the coordinator table do nothing. A return value from the APIs doesn't make sense. So, the warning message should more emphasize that like this?
```java
    logger.warn("Always true is returned since the single CRUD operation transactions don't have coordinator tables. Please use this Admin API with Consensus Commit transaction mode if you want to get the actual state");
```
- ScalarDB Cluster nodes automatically create the coordinator table if not exists
- Users see the warning message after using the Admin APIs with the single CRUD operation transaction mode, but they understand the Admin APIs did nothing and the return value made no sense. So, they won't be surprised even if seeing the coordinator table exists

So, I don't think warning log confuses users. But I don't have a strong opinion on it. So, we can go as-is without waring log or throwing exception.

komamitsu (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/singlecrudoperation/SingleCrudOperationTransactionAdmin.java
brfrn169: Added Javadoc in aecdf951b623ea4d77be27435ca397eddb5414ce. Thanks.

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/singlecrudoperation/SingleCrudOperationTransactionManager.java
```diff
@@ -0,0 +1,364 @@
...
+    Put put = buildable.consistency(Consistency.LINEARIZABLE).build();
+
+    try {
+      storage.put(put);
+    } catch (NoMutationException e) {
+      if (update.getCondition().isPresent()) {
+        throwUnsatisfiedConditionException(update);
+      }
+
+      // If the condition is not specified, it means that the record does not exist. In this case,
```
Torch3333, line 266: Sorry, I am confused by this comment. 
If the `update` condition is not specified, a `PutIfExists` condition is set. If the record already exists, the `NoMutationException` will be thrown, and we can ignore it, isn't it?

brfrn169: Sorry for the confusion.

If the `update` condition is not specified, we specify `ConditionBuilder.putIfExists()` for the put operation here:
https://github.com/scalar-labs/scalardb/pull/1793/files/aecdf951b623ea4d77be27435ca397eddb5414ce#diff-ca66c3466e760469ea1c0982cf3eb91f275cf80e0c5a7e808085afb8296374e5R255

Therefore, if a `NoMutationException` is thrown and the `update` condition is not specified, it means that the record doesn't exist. Does this explanation make sense?

Torch3333 (approved):
LGTM, thank you!

PR #1805, opened 2024-05-29
Split Group Commit Emitter method for normal group emit and delayed group emit

komamitsu:
## Description

In the current implementation, `Emittable` interface has only a single method `execute(String parentKeyOrFullKey, List<Snapshot> snapshots)`. As a result, the following methods need to handle both normal group commits and delayed group commits (containing a single transaction):

- https://github.com/scalar-labs/scalardb/blob/0315cbccaa75642e4a46105c6f11488c3f93b834/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandlerWithGroupCommit.java#L87-L90
- https://github.com/scalar-labs/scalardb/blob/0315cbccaa75642e4a46105c6f11488c3f93b834/core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java#L128-L136

These implementations should be split into separate methods for normal group commits and for delayed group commits in terms of readability and maintainability.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/1728

## Changes made

- Split `Emittable.execute(EMIT_KEY key, List<V> values)` into
  - `emitNormalGroup(PARENT_KEY parentKey, List<V> values)`
  - `emitDelayedGroup(FULL_KEY fullKey, V value)`
- Split `CommitHandlerWithGroupCommit.groupCommitState(String parentIdOrFullId, List<Snapshot> snapshots)` into 
  - `CommitHandlerWithGroupCommit.Emitter.emitNormalGroup(String parentId, List<Snapshot> snapshots)`
  - `CommitHandlerWithGroupCommit.Emitter.emitDelayedGroup(String fullId, Snapshot snapshot)`
- Update `Coordinator.putStateForGroupCommit(String parentIdOrFullId, ...)
  - to `putStateForGroupCommit(String parentId, ...)` so that the method only handles normal group commits
- Split `EMIT_KEY` generic type into `EMIT_PARENT_KEY` and `EMIT_FULL_KEY` generic types
  - Sorry, most changes in this PR come from this
- Update related unit tests including some refactoring and improvements

## Checklist

None

## Release notes

N/A


core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandlerWithGroupCommit.java
```diff
@@ -84,33 +86,45 @@ public void commitState(Snapshot snapshot)
...
+
+    public Emitter(Coordinator coordinator) {
+      this.coordinator = coordinator;
     }
 
-    List<String> transactionIds =
-        snapshots.stream().map(Snapshot::getId).collect(Collectors.toList());
+    @Override
+    public void emitNormalGroup(String parentId, List<Snapshot> snapshots)
+        throws CoordinatorException {
```
komamitsu, line 104: The original implementation of `groupCommitState()` is moved into this method.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -125,30 +125,23 @@ public void putState(Coordinator.State state) throws CoordinatorException {
...
 
-  // TODO: This method should be separated into `putStateForGroupCommitWithParentId()` and
-  //       `putStateForGroupCommitWithFullId()` so whether the id is a parent ID or a full ID is
-  //       clear.
   void putStateForGroupCommit(
-      String parentIdOrFullId,
-      List<String> fullIds,
-      TransactionState transactionState,
-      long createdAt)
+      String parentId, List<String> fullIds, TransactionState transactionState, long createdAt)
```
komamitsu, line 129: This method is now called only with a parent ID, not with a full ID.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorGroupCommitter.java
```diff
@@ -7,7 +7,7 @@
 import java.util.concurrent.ThreadLocalRandom;
 
 public class CoordinatorGroupCommitter
-    extends GroupCommitter<String, String, String, String, Snapshot> {
+    extends GroupCommitter<String, String, String, String, String, Snapshot> {
```
komamitsu, line 10: From here there will be many changes, but most of them are just adding an additional generic type (`EMIT_KEY` -> `EMIT_PARENT_ID` and `EMIT_FULL_ID`)

core/src/main/java/com/scalar/db/util/groupcommit/DelayedGroup.java
```diff
@@ -35,13 +35,13 @@ FULL_KEY fullKey(CHILD_KEY childKey) {
...
-    for (Slot<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_KEY, V> slot : slots.values()) {
+    for (Slot<PARENT_KEY, CHILD_KEY, FULL_KEY, EMIT_PARENT_KEY, EMIT_FULL_KEY, V> slot :
+        slots.values()) {
       // Pass `emitter` to ask the receiver's thread to emit the value
       slot.delegateTaskToWaiter(
           () ->
-              emitter.execute(
-                  keyManipulator.emitKeyFromFullKey(fullKey),
-                  Collections.singletonList(slot.value())));
+              emitter.emitDelayedGroup(
```
komamitsu, line 43: Call `Emittable.emitDelayedGroup()` for a transaction in a DelayedGroup

core/src/main/java/com/scalar/db/util/groupcommit/NormalGroup.java
```diff
@@ -113,12 +118,13 @@ public synchronized void delegateEmitTaskToWaiter() {
               logger.info("This group is already done, but trying to emit. Group: {}", this);
               return;
             }
-            emitter.execute(keyManipulator.emitKeyFromParentKey(parentKey), values);
+            emitter.emitNormalGroup(keyManipulator.emitParentKeyFromParentKey(parentKey), values);
```
komamitsu, line 121: Call `Emittable.emitNormalGroup()` for transactions in a NormalGroup

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandlerWithGroupCommit.java
```diff
@@ -84,33 +86,50 @@ public void commitState(Snapshot snapshot)
...
+      if (snapshots.isEmpty()) {
+        // This means all buffered transactions were manually rolled back. Nothing to do.
+        return;
+      }
 
-    // The id is either of a parent ID in the case of normal group commit or a full ID in the case
-    // of delayed commit.
-    coordinator.putStateForGroupCommit(
-        parentIdOrFullId, transactionIds, TransactionState.COMMITTED, System.currentTimeMillis());
+      // These transactions are contained in a normal group that has multiple transactions.
```
komamitsu, line 110: @josh-wong Can you take a look at this comment?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandlerWithGroupCommit.java
```diff
@@ -84,33 +86,50 @@ public void commitState(Snapshot snapshot)
...
-    return super.abortState(id);
+      logger.debug(
+          "Transaction {} (parent ID) is committed successfully at {}",
+          parentId,
+          System.currentTimeMillis());
+    }
+
+    @Override
+    public void emitDelayedGroup(String fullId, Snapshot snapshot) throws CoordinatorException {
+      // This transaction is contained in a delayed group that has only a single transaction.
```
komamitsu, line 126: @josh-wong Can you take a look at this comment?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -125,30 +125,23 @@ public void putState(Coordinator.State state) throws CoordinatorException {
...
-      List<String> childIds = new ArrayList<>(fullIds.size());
-      for (String fullId : fullIds) {
-        Keys<String, String, String> keys = keyManipulator.keysFromFullKey(fullId);
-        childIds.add(keys.childKey);
-      }
-      state = new State(parentIdOrFullId, childIds, transactionState, createdAt);
+
+    if (keyManipulator.isFullKey(parentId)) {
+      throw new AssertionError(
+          "This method is only for normal group commits that use a parent ID as the key");
```
komamitsu, line 134: @josh-wong Can you take a look at this error message (although this shouldn't occur) ?

josh-wong (approved):

Torch3333 (approved):
LGTM, thank you.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -125,30 +125,23 @@ public void putState(Coordinator.State state) throws CoordinatorException {
     put(put);
   }
 
-  // TODO: This method should be separated into `putStateForGroupCommitWithParentId()` and
-  //       `putStateForGroupCommitWithFullId()` so whether the id is a parent ID or a full ID is
-  //       clear.
   void putStateForGroupCommit(
```
brfrn169, line 128: Sorry, this is not directly related to this PR, but I'm wondering if we could separate this group commit related logic to another class. For example, can we create a subclass of this `Coordinator` class, something like `CoordinatorWithGroupCommit`, and move the group commit related logic to this class?

brfrn169 (commented):
Overall, LGTM. I left a comment. Please take a look when you have time!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
feeblefakie: Agree. It is worth thinking about it.

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
komamitsu: @brfrn169 @feeblefakie Thanks for the feedback!

I first thought the same thing like the relationship of `CommitHandler` and `CommitHandlerWithGroupCommit`, either of them is used based on the configuration. But, in case of `Coordinator`, it could need to handle full IDs generated for the group commit even if the group commit feature is disabled as follows:

1. An application starts with the group commit feature is enabled
2. A transaction creates a PREPARE state record that has a full transaction ID and gets crashed w/o commit or abort
3. The application restarts with the group commit feature is disabled
4. A transaction reads the PREPARE state record and tries to check if it's committed or aborted via `Coordinator` 
5. But the `Coordinator` can handle only normal transaction IDs. So, `Coordinator` needs to be able to handle the full ID despite the group commit feature is disabled

Of course, we can create `CoordinatorWithGroupCommit` that inherits `Coordinator` and is able to handle both normal transaction ID and group commit related transaction ID, and always use it:
- `CoordinatorWithGroupCommit.getState(String id)` will be as follows
  - If `id` is for normal transaction ID, it's delegated to `super.getState(id)`
  - Otherwise, the `id` is handled with `CoordinatorWithGroupCommit.getStateForGroupCommit(id)`
- Transaction Managers always use `CoordinatorWithGroupCommit` instead of `Coordinator`. In other word, `Coordinator` will be like abstract class.

I wasn't sure this was better. But if you think it's better than the current fat `Coordinator`, I'll handle the refactoring as another PR. What do you think?

brfrn169: Okay, understood. I didn't assume the situation for disabling after enabling and the group commit feature. Let's leave it as it is.

BTW, do we really need to assume the situation? I feel that makes the things complicated (the code, tests, etc.).

One more question, if we have mixed multiple transaction manager instances that enable and disable the group commit feature, does that work?

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
komamitsu: > BTW, do we really need to assume the situation? I feel that makes things complicated (the code, tests, etc.).

Good point. I think it's a very corner case, but it might potentially occur. If the transaction state is committed before crashing in the above example case, the committed transaction might be rolled back by `Coordinator` which can't find the transaction with full ID after the restart. It seems a critical case and should be prevented as much as possible.

Also, I agree with your concern. In this case, we need to only consider the scenario where `Coordinator.getState(fullID)` is called with the group committer disabled, so probably don't need so much complicated things. (I'll add the test case later...)

> if we have mixed multiple transaction manager instances that enable and disable the group commit feature, does that work?

Yes. I think it works as far as a transaction started with a transaction manager that enables the group commit feature isn't handled by another transaction manager that disables the feature, or vise versa.
- Transaction Manager (TM1) with the group commit feature enabled starts Transaction 1
- There is another Transaction Manager (TM2) with the group commit feature disabled 
- Transaction 1 needs to be handled by TM1 always, not by TM2

PR #2177, opened 2024-08-13
Change commit conflict behavior

jnmt:
## Description

This PR changes the behavior of the commit conflict when a transaction finds the committed-state record with the same transaction ID in the coordinator state table.

**Before:**
- ConsensusCommit: ignored
- TwoPhaseConsensusCommit: ignored

**After:**
- ConsensusCommit: throw an exception
- TwoPhaseConsensusCommit: ignored

The reason for this change is to prevent users from accidentally (or maliciously in ScalarDL) committing transactions with the same transaction ID. We wanted to do it in both the regular consensus commit transaction and the two-phase one for consistency. But we couldn’t do it in the two-phase one because each service tries to put the state ‘committed’ in the two-phase commit transaction. Although we also discussed changing this kind of style in the two-phase consensus commit, we decided not to do it so far.

## Related issues and/or PRs

N/A

## Changes made

- Throw an exception when a transaction finds the committed-state record with the same transaction ID in the coordinator state table.
- Fix a message in the related part.

## Checklist

N/A

## Release notes

Changed the behavior of the commit operation for the Consensus Commit transaction.


core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -709,7 +709,7 @@ public enum CoreError implements ScalarDbError {
   CONSENSUS_COMMIT_CONFLICT_OCCURRED_WHEN_COMMITTING_STATE(
       Category.CONCURRENCY_ERROR,
       "0015",
-      "The committing state in the coordinator failed. The transaction has been aborted",
+      "A transaction conflict occurred when committing the state to the coordinator failed. The transaction has been aborted",
```
jnmt, line 712: This is not directly related to this change. But I included it since I think the message is something wrong and it's commit conflict thing.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -125,11 +125,13 @@ public ConsensusCommitManager(
 
   // `groupCommitter` must be set before calling this method.
   private CommitHandler createCommitHandler() {
+    // Create CommitHandler(WithGroupCommit) with throwExceptionIfCommittedTransactionExists true
+    // because we want to detect committing a transaction with the duplicated transaction ID.
     if (isGroupCommitEnabled()) {
       return new CommitHandlerWithGroupCommit(
-          storage, coordinator, tableMetadataManager, parallelExecutor, groupCommitter);
+          storage, coordinator, tableMetadataManager, parallelExecutor, groupCommitter, true);
```
jnmt, line 132: In the case of the group commit, different transaction IDs are internally created for different groups, even if the same transaction IDs are given from applications. I think it’s still worth enabling this to check the uniqueness of groups, just in case.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
```diff
@@ -90,7 +90,10 @@ public TwoPhaseConsensusCommitManager(DatabaseConfig databaseConfig) {
     coordinator = new Coordinator(storage, config);
     parallelExecutor = new ParallelExecutor(config);
     recovery = new RecoveryHandler(storage, coordinator, tableMetadataManager);
-    commit = new CommitHandler(storage, coordinator, tableMetadataManager, parallelExecutor);
+    // Create CommitHandler with throwExceptionIfCommittedTransactionExists false different from the
+    // case for ConsensusCommitManager because each service tries to put the state ‘committed’ in
+    // the two-phase commit transaction, and we want to ignore the existing committed state record.
```
josh-wong, line 95: I believe we decided that `transactions with a two-phase commit interface` was the accurate phrase. Should this wording match that?

```suggestion
    // the transaction with a two-phase commit interface, and we want to ignore the existing committed state record.
```

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -2114,6 +2115,44 @@ private void commit_WriteSkewOnExistingRecordsWithSnapshot_ShouldProduceNonSeria
     assertThat(thrown2).isInstanceOf(CommitConflictException.class);
   }
 
+  @Test
+  @DisabledIf(
+      value = "isGroupCommitEnabled",
+      disabledReason =
+          "When the group commit is enabled, ScalarDB internally creates different transaction IDs for this case.")
```
josh-wong, line 2122:
```suggestion
          "When group commit is enabled, ScalarDB internally creates different transaction IDs for this case.")
```

josh-wong (changes requested):
I've added a couple comments and suggestions. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
jnmt: Yes, exactly speaking. I applied it with formatting in 344aade. Thank you!

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
jnmt: Fixed in 344aade.

josh-wong (approved):
LGTM! Thank you!🙇‍♂️

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -125,11 +125,13 @@ public ConsensusCommitManager(
 
   // `groupCommitter` must be set before calling this method.
   private CommitHandler createCommitHandler() {
+    // Create CommitHandler(WithGroupCommit) with throwExceptionIfCommittedTransactionExists true
+    // because we want to detect committing a transaction with the duplicated transaction ID.
```
feeblefakie, line 129:
```suggestion
    // because we want to detect committing a transaction with a duplicated transaction ID.
```
Maybe?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -125,11 +125,13 @@ public ConsensusCommitManager(
 
   // `groupCommitter` must be set before calling this method.
   private CommitHandler createCommitHandler() {
+    // Create CommitHandler(WithGroupCommit) with throwExceptionIfCommittedTransactionExists true
```
feeblefakie, line 128:
```suggestion
    // Create a CommitHandler(WithGroupCommit) with throwExceptionIfCommittedTransactionExists true
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
```diff
@@ -90,7 +90,11 @@ public TwoPhaseConsensusCommitManager(DatabaseConfig databaseConfig) {
     coordinator = new Coordinator(storage, config);
     parallelExecutor = new ParallelExecutor(config);
     recovery = new RecoveryHandler(storage, coordinator, tableMetadataManager);
-    commit = new CommitHandler(storage, coordinator, tableMetadataManager, parallelExecutor);
+    // Create CommitHandler with throwExceptionIfCommittedTransactionExists false different from the
```
feeblefakie, line 93:
```suggestion
    // Create CommitHandler with throwExceptionIfCommittedTransactionExists false, which is different from the
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
```diff
@@ -90,7 +90,11 @@ public TwoPhaseConsensusCommitManager(DatabaseConfig databaseConfig) {
...
+    // case for ConsensusCommitManager because each service tries to put the state 'committed' in
```
feeblefakie, line 94:
```suggestion
    // case for ConsensusCommitManager, because each service could put the state 'committed' in
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
```diff
@@ -90,7 +90,11 @@ public TwoPhaseConsensusCommitManager(DatabaseConfig databaseConfig) {
...
+    // the transaction with a two-phase commit interface, and we want to ignore the existing
+    // committed state record.
```
feeblefakie, line 96: The last sentence sounds a bit weird to me.
How about something like this?

We want to avoid cases where some services get exceptions due to conflicting commits with the same transaction ID.

feeblefakie (commented):
The code looks good. Thank you!
I left minor feedback on the comments. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
jnmt: Thanks! According to Grammarly, the comma is not necessary, and I chose `can put` since we can still do it regardless of this PR. PTAL at b433e33.

core/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseConsensusCommitManager.java
jnmt: Great, thank you. Revised, including the above comment, in b433e335d.

brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -95,6 +98,14 @@ protected void handleCommitConflict(Snapshot snapshot, Exception cause)
               CoreError.CONSENSUS_COMMIT_CONFLICT_OCCURRED_WHEN_COMMITTING_STATE.buildMessage(),
               cause,
               snapshot.getId());
+        } else if (state.equals(TransactionState.COMMITTED)) {
+          if (throwExceptionIfCommittedTransactionExists) {
+            throw new CommitConflictException(
```
komamitsu, line 103: Let me ask about a corner case. If a proper transaction that started with the transaction ID and detected here the coordinator state conflict crashes for some reasons before rolling back PREPARE records, future lazy recoveries will roll-forward based on the unexpected (malicious?) commit?

jnmt: @komamitsu Umm, that’s a really good point… Regardless of crashing, the lazy recovery will roll forward the records based on the committed state of the unrelated transaction. That's the same thing as not throwing the exception after all... Let me think about it more.

jnmt:
After the discussion again, we decided to leave the behavior as is since it’s hard to guarantee correctness if we accept the same transaction IDs for different transactions. For ScalarDL, we will continue discussing how to prevent it. Sorry guys, for the lack of consideration. Thank you for pointing it out, Mitsu.

PR #2245, opened 2024-09-20
Change default key column size and make it configurable for MySQL and Oracle

jnmt:
## Description

This PR changes the default key column size and makes it configurable for MySQL and Oracle.

Background:
Currently, the default key column size for MySQL and Oracle is 64 bytes due to the underlying database limitation. It is sometimes short for certain applications. For example, ScalarDL uses the binary name of the class to identify a contract, and it often exceeds 64 bytes.

Solution:
We change the default value to 128 bytes. In addition, we make it configurable because the underlying databases limit the total size of primary key columns; i.e., the larger default key column size affects the total number of primary key columns. So, users may want to create a primary key with a large number of small key columns size.

## Related issues and/or PRs

N/A

## Changes made

- Add options for changing the key column size for MySQL and Oracle and use 128 bytes as the default

## Checklist

This change is backward-compatible since it doesn't affect existing tables.

## Release notes

Added options for changing the key column size for MySQL and Oracle and used 128 bytes as the default.


core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
```diff
@@ -47,6 +47,11 @@ public class JdbcConfig {
   public static final String ADMIN_CONNECTION_POOL_MAX_TOTAL =
       PREFIX + "admin.connection_pool.max_total";
 
+  public static final String MYSQL_VARIABLE_KEY_COLUMN_SIZE =
+      PREFIX + "mysql.variable_key_column_size";
+  public static final String ORACLE_VARIABLE_KEY_COLUMN_SIZE =
+      PREFIX + "oracle.variable_key_column_size";
```
jnmt, line 53: I introduced two separate settings (only) for MySQL and Oracle for the following reasons.

- PostgreSQL also has a limit of 10485760 for each key column (but not for total). So, users might be confused if the common setting `jdbc.variable_key_column_size` does not affect it.
- We can also make the limit of PostgreSQL configurable, but the common settings would decrease the current limit since 10485760 is too large for MySQL and Oracle.
- I guess there is no motivation to decrease 10485760 for PostgreSQL.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -60,8 +59,7 @@ public JdbcDatabase(DatabaseConfig databaseConfig) {
             databaseConfig.getMetadataCacheExpirationTimeSecs());
 
     OperationChecker operationChecker = new OperationChecker(databaseConfig, tableMetadataManager);
-    QueryBuilder queryBuilder = new QueryBuilder(rdbEngine);
-    jdbcService = new JdbcService(tableMetadataManager, operationChecker, queryBuilder);
+    jdbcService = new JdbcService(tableMetadataManager, operationChecker, rdbEngine);
```
jnmt, line 62: Push down `RdbEngineStrategy` since we don't have to create it again at the deeper level (i.e., `ConditionMutator`).

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -201,9 +212,9 @@ public String getDataTypeForEngine(DataType scalarDbDataType) {
   public String getDataTypeForKey(DataType dataType) {
     switch (dataType) {
       case TEXT:
-        return "VARCHAR(64)";
+        return "VARCHAR(" + keyColumnSize + ")";
```
komamitsu, line 215: I guess this column definition is also used for `coordinator.state.tx_id` column, which usually contains a UUID value or [a UUID value plus 25 byte prefix for the group commit feature](https://github.com/scalar-labs/scalardb/blob/master/core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorGroupCommitter.java#L40-L44). So, a validation that confirms the key column size is equal to or larger than 64 would be needed somewhere?

jnmt: @komamitsu Thank you! That's a really good point. We should check it maybe in `JdbcConfig`. I will add it.

jnmt: @komamitsu Revised in b04a069. PTAL!

komamitsu (approved):
LGTM! 👍

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
Torch3333: What is the reasoning for adding `variable` in the configuration name? 
In essence, it is a configuration, implying that the value is configurable (== variable). Following the same logic, we could add `variable` in all the configuration names.

jnmt: @Torch3333 Thank you for the question. It's a good point. It doesn't mean it's configurable, but the size of the variable (data-type) key column. I would like to distinguish it from other fixed-size data types like `int` and `double` since this option only affects data types (e.g., `varchar` and `varbinary`) with variable length. Still weird?

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
```diff
@@ -167,6 +191,22 @@ public JdbcConfig(DatabaseConfig databaseConfig) {
...
+            MYSQL_VARIABLE_KEY_COLUMN_SIZE,
+            DEFAULT_VARIABLE_KEY_COLUMN_SIZE);
+
+    oracleVariableKeyColumnSize =
+        getInt(
+            databaseConfig.getProperties(),
+            ORACLE_VARIABLE_KEY_COLUMN_SIZE,
+            DEFAULT_VARIABLE_KEY_COLUMN_SIZE);
+
+    if (mysqlVariableKeyColumnSize < 64 || oracleVariableKeyColumnSize < 64) {
```
feeblefakie, line 206: I think it's probably better to make `64` a static variable like `MINIMUM_VARIABLE_KEY_COLUMN_SIZE.
We should probably note why 64.

feeblefakie (approved):
Left one suggestion.
Other than that, LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
Torch3333: Oh, ok. I completely missed this interpretation. I agree with this reasoning. 👍 
Thank for the explanation.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
jnmt: @feeblefakie Definitely, thank you! I updated it with the reason based on Mitsu's comment in 1072e98.

jnmt:
@josh-wong Can you take a look at the message part? I forgot to add you as a reviewer when revising this PR to add the error message.

feeblefakie (approved):
LGTM! Thank you!

josh-wong (approved):
LGTM! Thank you!🙇🏻‍♂️

PR #2265, opened 2024-09-26
Refactor ConcatenationVisitor to use ColumnVisitor

Torch3333:
## Description

Refactor `ConcatenationVisitor` that uses the deprecated `ValueVisitor` to use the `ColumnVisitor` instead.

The background is I take care of some refactoring to facilitate the implementation of time-related types.

## Related issues and/or PRs

N/A

## Changes made

Refactor `ConcatenationVisitor` that uses the deprecated `ValueVisitor` to use the `ColumnVisitor` instead.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
+++ b/core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
@@ -38,6 +38,7 @@ public class ConcatenationVisitor implements ColumnVisitor {
    */
   @Override
   public void visit(BooleanColumn column) {
+    assert !column.hasNullValue();
     columns.add(String.valueOf(column.getBooleanValue()));
   }
 
@@ -48,6 +49,7 @@ public class ConcatenationVisitor implements ColumnVisitor {
    */
   @Override
   public void visit(IntColumn column) {
+    assert !column.hasNullValue();
     columns.add(String.valueOf(column.getIntValue()));
   }
 
@@ -58,6 +60,7 @@ public class ConcatenationVisitor implements ColumnVisitor {
    */
   @Override
   public void visit(BigIntColumn column) {
+    assert !column.hasNullValue();
     columns.add(String.valueOf(column.getBigIntValue()));
   }
 
@@ -68,6 +71,7 @@ public class ConcatenationVisitor implements ColumnVisitor {
    */
   @Override
   public void visit(FloatColumn column) {
+    assert !column.hasNullValue();
     columns.add(String.valueOf(column.getFloatValue()));
   }
 
@@ -78,6 +82,7 @@ public class ConcatenationVisitor implements ColumnVisitor {
    */
   @Override
   public void visit(DoubleColumn column) {
+    assert !column.hasNullValue();
     columns.add(String.valueOf(column.getDoubleValue()));
   }
 
@@ -88,6 +93,7 @@ public class ConcatenationVisitor implements ColumnVisitor {
    */
   @Override
   public void visit(TextColumn column) {
+    assert !column.hasNullValue();
     column.getValue().ifPresent(columns::add);
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
```diff
@@ -19,88 +19,89 @@
  * @author Yuji Ito
  */
 @NotThreadSafe
-public class ConcatenationVisitor implements ValueVisitor {
-  private final List<String> values;
+public class ConcatenationVisitor implements ColumnVisitor {
```
Torch3333, line 22: All changes originates from this change of interface.

core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
```diff
@@ -19,88 +19,89 @@
...
 
   public ConcatenationVisitor() {
-    values = new ArrayList<>();
+    columns = new ArrayList<>();
   }
 
   public String build() {
-    // TODO What if the string or blob value includes `:`?
-    return String.join(":", values);
+    // TODO What if the string or blob colum includes `:`?
```
komamitsu, line 30:
```suggestion
    // TODO What if the string or blob column includes `:`?
```

core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
```diff
@@ -19,88 +19,89 @@
...
+   * @param column a {@code BlobColumn} to be set
    */
   @Override
-  public void visit(BlobValue value) {
-    // Use Base64 encoding
-    value
-        .get()
-        .ifPresent(b -> values.add(Base64.getUrlEncoder().withoutPadding().encodeToString(b)));
+  public void visit(BlobColumn column) {
+    if (!column.hasNullValue()) {
```
komamitsu, line 101: Just out of curiosity: why not use `column.getValue().ifPresent()` as well as `visit(TextColumn column)`?

Torch3333: Thanks for asking.
`column.getValue()` returns an `Optional<ByteBuffer>` but I need to obtain a `byte[]` which can only be obtained with `column.getBlobValueAsBytes()`

komamitsu: I guess you can call `java.nio.ByteBuffer.array()` to get the byte array since the ByteBuffer instance seems byte-array backed one. But it instantiates a temporary ByteBuffer instance. So, either is fine with me 👍

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
Torch3333: Yes, you are right, but I prioritize not instantiating the `ByteBuffer` over a syntax simplification.

core/src/main/java/com/scalar/db/storage/cosmos/ConcatenationVisitor.java
```diff
@@ -19,88 +19,89 @@
...
    *
-   * @param value a {@code BooleanValue} to be set
+   * @param column a {@code BooleanColumn} to be set
    */
   @Override
-  public void visit(BooleanValue value) {
-    values.add(String.valueOf(value.get()));
+  public void visit(BooleanColumn column) {
+    columns.add(String.valueOf(column.getBooleanValue()));
   }
```
brfrn169, line 42: Since this class is supposed to be used to create concatenated keys (partition keys and clustering keys), the columns should not be null. So I think we can add an assertion as follows:

```suggestion
  public void visit(BooleanColumn column) {
    assert !column.hasNullValue();
    columns.add(String.valueOf(column.getBooleanValue()));
  }
```

The same applies to the other methods.

brfrn169 (approved):
LGTM! Thank you!

PR #2271, opened 2024-10-03
Add an extension point to collect uncommitted read-set and write-set

komamitsu:
## Description

In the current implementation, it's not possible to develop an extension module injected via class loader that accesses the uncommitted read-set and write-set of a consensus commit transaction. This PR adds an extension point to collect uncommitted read-set and write-set to make it easy to develop extension modules to access those information.

## Related issues and/or PRs

None

## Changes made

- Add `SnapshotHandler` that receives uncommitted `Snapshot`
- Add `CommitHandler.setSnapshotHandler()`
- Relax the access scopes of some information about `ConsensusCommitManager` and `Snapshot`

## Checklist

None

## Release notes

N/A


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -75,7 +75,7 @@ public ConsensusCommitManager(
     mutationOperationChecker = new ConsensusCommitMutationOperationChecker(tableMetadataManager);
   }
 
-  ConsensusCommitManager(DatabaseConfig databaseConfig) {
+  public ConsensusCommitManager(DatabaseConfig databaseConfig) {
```
komamitsu, line 78: To make it easy to inherit `ConsensusCommitManager`. This change is helpful to utilize the new extension point.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -299,4 +300,18 @@ static String convertUnsatisfiedConditionExceptionMessageForUpdate(
...
     return message;
   }
+
+  /**
+   * Returns the next `tx_version` based on the current value.
+   *
+   * @param currentTxVersion The current `tx_version`, if it exists, or null otherwise.
+   * @return The next `tx_version`.
+   */
+  public static int getNextTxVersion(@Nullable Integer currentTxVersion) {
```
komamitsu, line 310: This logic how to determine the next Tx version may be reused by an extension module injected via class loader.

feeblefakie (commented):
Thank you!
Left some questions. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -79,6 +85,8 @@ public void commit(Snapshot snapshot) throws CommitException, UnknownTransaction
       throw e;
     }
 
+    snapshotHandleFuture.ifPresent(SnapshotHandleFuture::get);
```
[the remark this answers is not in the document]
komamitsu: @feeblefakie 

> Can we just set Function or BiFunction like it does with `setSnapshotHandler` and invoke it here?

Yeah, good point. I think using `Function` should also work.

I named the interface `SnapshotHandle**Future**` since I wanted to encourage `setSnapshotHandler` users to implement asynchronous handlers that don't block `prepare()` and `validate()`. Also, the reason why I wrapped `Future` with `SnapshotHandleFuture` is for maintainability considering future modifications just in case. But, I'm okay to go with simple `Function`.

> Also, I feel SnapshotHandler sounds too general.

Ah, I agree. `BeforeCommitSnapshotHook` seems good. But it might sound like "validated" snapshots. How about `PreparedSnapshotHook`? Either is fine with me, thought.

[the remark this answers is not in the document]
feeblefakie: Thanks for the explanation.
Using `Future` for a non-blocking hook makes sense in this case since Snapshot won't be changed after the prepare phase, so I'm fine with it.
For the name, how about `AfterPrepareSnapshotHook`?
I'm also fine with `BeforeCommitSnapshotHook` and `PreparedSnapshotHook`.
I think you can decide.

feeblefakie (approved):
LGTM! Thank you!
I would like the class name to be changed, but any of the current options is fine with me.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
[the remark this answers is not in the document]
komamitsu: @feeblefakie Thanks for the comment!

> For the name, how about AfterPrepareSnapshotHook?
> I'm also fine with BeforeCommitSnapshotHook and PreparedSnapshotHook.

Strictly speaking, snapshots are sent asynchronously before `prepare()` completes. So, I hesitate a bit to use the terms `before` or `after` that emphasizes the order. `PreparedSnapshotHook` might imply that snapshots are **already prepared**, but I think it can be also interpreted that snapshots are in the process of being prepared, and relatively better. I'll go with `PreparedSnapshotHook`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -253,6 +254,39 @@ public void to(MutationComposer composer)
...
 
+  /**
+   * Returns the read set. This is only for internal use. It doesn't copy the values for performance
+   * reasons.
+   *
+   * @return the read set. Don't modify the value.
+   */
+  @SuppressFBWarnings(value = {"EI_EXPOSE_REP"})
+  public Map<Key, Optional<TransactionResult>> getReadSet() {
+    return readSet;
```
brfrn169, line 265: How about wrapping it with `new HashMap()` or `ImmutableMap.copyOf()`?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -253,6 +254,39 @@ public void to(MutationComposer composer)
...
+
+  /**
+   * Returns the write set. This is only for internal use. It doesn't copy the values for
+   * performance reasons.
+   *
+   * @return the write set. Don't modify the value.
+   */
+  @SuppressFBWarnings(value = {"EI_EXPOSE_REP"})
+  public Map<Key, Put> getWriteSet() {
+    return writeSet;
```
brfrn169, line 276: Ditto.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -253,6 +254,39 @@ public void to(MutationComposer composer)
...
+
+  /**
+   * Returns the delete set. This is only for internal use. It doesn't copy the values for
+   * performance reasons.
+   *
+   * @return the delete set. Don't modify the value.
+   */
+  @SuppressFBWarnings(value = {"EI_EXPOSE_REP"})
+  public Map<Key, Delete> getDeleteSet() {
+    return deleteSet;
```
brfrn169, line 287: Ditto.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
brfrn169: So, `protected` is more appropriate?
```suggestion
  protected ConsensusCommitManager(DatabaseConfig databaseConfig) {
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -51,8 +56,9 @@ protected void onPrepareFailure(Snapshot snapshot) {}
   protected void onValidateFailure(Snapshot snapshot) {}
 
   public void commit(Snapshot snapshot) throws CommitException, UnknownTransactionStatusException {
+    Optional<PreparedSnapshotHookFuture> snapshotHookFuture;
     try {
-      prepare(snapshot);
+      snapshotHookFuture = prepare(snapshot);
```
brfrn169, line 61: Maybe we can move the snapshotHandler logic outside the `prepare()` method as follows?
```suggestion
      snapshotHandleFuture = handleSnapshotBeforePreparation(snapshot);
      prepare(snapshot);
```

What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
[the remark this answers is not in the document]
komamitsu: Addressed it in https://github.com/scalar-labs/scalardb/pull/2271/commits/037913934c5e7dca44262c280b48136300fdeb92

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
```diff
@@ -0,0 +1,10 @@
+package com.scalar.db.transaction.consensuscommit;
+
+public interface PreparedSnapshotHook {
```
brfrn169, line 3: `PreparedSnapshotHook` sounds like a hook called after preparation to me. Actually, the state of the snapshot changes after preparation.
https://github.com/scalar-labs/scalardb/blob/037913934c5e7dca44262c280b48136300fdeb92/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java#L459

Maybe `BeforePreparationSnapshotHook` would be a better name?

In the future, we might add something like `BeforeValidationSnapshotHook` or `BeforeCommitSnapshotHook`.

What do you think?

brfrn169 (commented):

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
komamitsu: > PreparedSnapshotHook sounds like a hook called after preparation to me.

Indeed, I have the same concern slightly. https://github.com/scalar-labs/scalardb/pull/2271#discussion_r1792738799

It's possible that the prepared records are asynchronously written before a snapshot is sent. So, I was also wondering if `BeforePreparationSnapshotHook`-ish name was reasonable or not. But, I think it can mean the snapshot is sent before the prepare phase operation. So, I think it's okay. I'll change it to `BeforePreparationSnapshotHook`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
komamitsu: The snapshot hook mechanism introduced in this PR could impact the performance and memory consumption of applications that use ScalarDB. So, I think all the operations in a snapshot hook should be fast and low-memory-footprint as much as possible. As described in the method comment, it would be great if a snapshot hook can read those write set and read set without any coping entries.

I also thought the following iteration method, but it's too usage-specific.
```java
  public void iterateWriteSet(BiConsumer<Put, Optional<TransactionResult>> consumer) {
    for (Map.Entry<Key, Put> entry : writeSet.entrySet()) {
      Optional<TransactionResult> result = Optional.empty();
      if (readSet.containsKey(entry.getKey())) {
        result = readSet.get(entry.getKey());
      }
      consumer.accept(entry.getValue(), result);
    }
  }
```

I think using `java.util.Collections#unmodifiableMap()` is a reasonable option since it doesn't have its own table but delegates read operations to the original map. The performance impact should be negligible. What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
komamitsu: Ah, good point! I missed the improvement after we decided to use Snapshot itself... I'll update the PR.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: I see. Sorry, I missed the method comment. That sounds reasonable.

> I think using `java.util.Collections#unmodifiableMap()` is a reasonable option since it doesn't have its own table but delegates read operations to the original map. The performance impact should be negligible. What do you think?

Actually, I’m not sure about the differences in performance between `Collections#unmodifiableMap()`, `new HashMap()`, and `ImmutableMap.copyOf()`, so let's keep it as is for now.

By the way, we already have getter methods for `writeSet` and `deleteSet`, and they copy `writeSet` and `deleteSet`:
https://github.com/scalar-labs/scalardb/blob/037913934c5e7dca44262c280b48136300fdeb92/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java#L157-L163

For the same reason, we probably don’t want to copy them. So could you please modify these methods as well and add similar method comments, even though it’s not directly related to this PR?

komamitsu: > so let's keep it as is for now.

👍 

As for `getPutsInWriteSet()` and `getDeletesInDeleteSet()`, they wrap the maps' values with ArrayList constructor, which also internally copies received values. So, I'll change the return type to `Collection<Put>` and `Collection<Delete>`, respectively.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
komamitsu: Thanks! I confirmed it works. Addressed it in https://github.com/scalar-labs/scalardb/pull/2271/commits/d2d100361ad864defd4d624a1b285fc800424a81

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
komamitsu: Addressed them in https://github.com/scalar-labs/scalardb/pull/2271/commits/71b40a4168c24dbfdbd8eb4ef7852b071cbc2395

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
komamitsu: Addressed it in https://github.com/scalar-labs/scalardb/pull/2271/commits/0fe41c9412a03a215c00f1f0bf30ae210493fd8f

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
komamitsu: Addressed them in https://github.com/scalar-labs/scalardb/pull/2271/commits/067d69df83b55412f8c8147f71882d892da2a258.

@brfrn169  I also noticed the original PR didn't properly handle errors when invoking the snapshot hook and waiting the snapshot future, resulting in throwing non-CommitException and resource leaks. I enhanced the error handling in the commit. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
feeblefakie: I think it might be better to do a simple benchmarking if we care about the performance.
`ImmutableMap.copyOf` actually tries to avoid copying if it is safe as described in the doc, so it might not cause much overhead.
https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableMap.html#copyOf-java.util.Map-

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
feeblefakie: As Toshi said, snapshot states might be changed in the case of extra-write. So, we shouldn't trigger the hook asynchronously?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
komamitsu: Yeah, thanks for the heads-up. I micro-benchmarked yesterday with the following code. Sorry, I should've mentioned this benchmark.

```java
package org.example;

import com.google.common.collect.ImmutableMap;
import java.util.Collections;
import java.util.HashMap;
import java.util.Map;

public class IteratorTest {
  private static final Map<String, Integer> MAP = new HashMap<>();
  static {
    MAP.put("one", 1);
    MAP.put("two", 2);
    MAP.put("three", 3);
    MAP.put("four", 4);
    MAP.put("five", 5);
    MAP.put("six", 6);
    MAP.put("seven", 7);
    MAP.put("eight", 8);
    MAP.put("nine", 9);
    MAP.put("ten", 10);
  }

  private static void testDirect() {
    Map<String, Integer> map = MAP;
    for (Map.Entry<String, Integer> entry : map.entrySet()) {
      entry.getKey();
      entry.getValue();
    }
  }

  private static void testUnmod() {
    Map<String, Integer> map = Collections.unmodifiableMap(MAP);
    for (Map.Entry<String, Integer> entry : map.entrySet()) {
      entry.getKey();
      entry.getValue();
    }
  }

  private static void testNewMap() {
    Map<String, Integer> map = new HashMap<>(MAP);
    for (Map.Entry<String, Integer> entry : map.entrySet()) {
      entry.getKey();
      entry.getValue();
    }
  }

  private static void testImmutable() {
    Map<String, Integer> map = ImmutableMap.copyOf(MAP);
    for (Map.Entry<String, Integer> entry : map.entrySet()) {
      entry.getKey();
      entry.getValue();
    }
  }

  public static void main(String[] args) {
    int n = 1000000;

    for (int i = 0; i < n; i++) { testDirect(); }

    {
      long start = System.currentTimeMillis();
      for (int i = 0; i < n; i++) {
        testDirect();
      }
      System.out.println("Direct: duration=" + (System.currentTimeMillis() - start));
    }

    for (int i = 0; i < n; i++) { testUnmod(); }
    {
      long start = System.currentTimeMillis();
      for (int i = 0; i < n; i++) {
        testUnmod();
      }
      System.out.println("UnmodiMap: duration=" + (System.currentTimeMillis() - start));
    }

    for (int i = 0; i < n; i++) { testNewMap(); }
    {
      long start = System.currentTimeMillis();
      for (int i = 0; i < n; i++) {
        testNewMap();
      }
      System.out.println("NewMap: duration=" + (System.currentTimeMillis() - start));
    }

    for (int i = 0; i < n; i++) { testImmutable(); }
    {
      long start = System.currentTimeMillis();
      for (int i = 0; i < n; i++) {
        testImmutable();
      }
      System.out.println("Immutable: duration=" + (System.currentTimeMillis() - start));
    }
  }
}
```

The result was:
```
> Task :IteratorTest.main()
Direct: duration=20
UnmodiMap: duration=21
NewMap: duration=89
Immutable: duration=130
```

As for the Javadoc, I also looked into the implementation. It doesn't copy only when the original map is `ImmutableMap` instance. In the case of `Snapshot`. the original maps are `HashMap` and I think they'll be copied.

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
komamitsu: @feeblefakie 

Ah, you're right... I just remembered that @brfrn169 and I discussed this case and planed to disable the snapshot hook feature when extra-write is enabled, since extra-write would be sunset-ed (so, this PR should've checked the configuration...).

Triggering the hook after preparation might increase the total latency. So, I think the invocation of the hook should be asynchronous without waiting for the completion of `prepare` phase operation, although it might increase the possibility of transactions that abort in `prepare` phase being handled by the snapshot handler to some extent. What do you think?

brfrn169: One possible solution is for the Snapshot getters to return copies of the read-set and write-set to avoid the impact of state changes?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -34,6 +37,8 @@ public class CommitHandler {
   private final TransactionTableMetadataManager tableMetadataManager;
   private final ParallelExecutor parallelExecutor;
 
+  @LazyInit @Nullable private BeforePreparationSnapshotHook beforePreparationSnapshotHook;
```
brfrn169, line 40: It's a minor thing, but I thought `@LazyInit` also implied that the field was nullable. Should we add `@Nullable` alongside `@LazyInit` as well?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -50,7 +55,45 @@ protected void onPrepareFailure(Snapshot snapshot) {}
...
+    }
+    try {
+      snapshotHookFuture.get();
+    } catch (Exception e) {
+      abortState(snapshot.getId());
+      rollbackRecords(snapshot);
+      // TODO: This method is actually a part of validation phase. But the callback method name
+      //       `onValidateFailure()` should be renamed to more reasonable one.
+      onValidateFailure(snapshot);
+      throw new CommitException(e.getMessage(), e, snapshot.getId());
```
brfrn169, line 90: Ditto.
```suggestion
      throw new CommitException(
          CoreError.WAITING_BEFORE_PREPARATION_SNAPSHOT_HOOK_FAILED.buildMessage(e.getMessage()),
          e,
          snapshot.getId());
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
komamitsu: > I thought `@LazyInit` also implied that the field was nullable.

Does it point that the field can be null until it's lazily initialized? If so, I think only `@LazyInit` may be enough. `@Nullable` is added since the field can be permanently null if no snapshot hook is set. Does it answer your question?

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
komamitsu: @brfrn169 Oh, do you think the read-set or the write-set could be changed during the `prepare` or `validate` phase even if extra-write is out of the scope?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
komamitsu: Good catch! I'll define and use the new error code.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: That makes sense. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
brfrn169: For now, if extra-write is out of scope, the read-set and write-set should not change. So, if we don’t allow users to set `BeforePreparationSnapshotHook` when extra-write is enabled, that should be fine. However, I’m concerned that this might not be guaranteed in the future. Therefore, it might be safer for the getters to return copies of the read-set and write-set, since `BeforePreparationSnapshotHook` is executed asynchronously. Otherwise, thread-safety issues could arise. What do you think?

komamitsu: @brfrn169 

> I’m concerned that this might not be guaranteed in the future.

I understand your concern. It seems reasonable.

Okay. Indeed, the micro-benchmark showed copying the map had some performance impact, but it didn't necessarily indicate ScalarDB application incurs significant performance impact. So, I'll copy the map in this PR just in case. And, I'll compare with a zero-copy version later in a ScalarDB application from the perspective of performance and resource consumption.

BTW, I assume we don't need to revert https://github.com/scalar-labs/scalardb/pull/2271/commits/71b40a4168c24dbfdbd8eb4ef7852b071cbc2395#diff-4e6eb5a222519a33681ef588e24c3a8eabc8bdec7e89e79bb5c96c3ce8898759R158-R176 since it's called synchronously. Is my understanding correct?

komamitsu: @brfrn169 

> I’m concerned that this might not be guaranteed in the future.

Ah, in this case, we may need to pass changed read-set and write-set to the snapshot hook, although it depends. It would require CommitHandler to pass changed read-set and write-set to the snapshot hook after PREPARED records are written. So, I don't think just copying the maps is sufficient.

How about going with the current no-copy approach in this PR for now? If we need to change read-set and write-set after PREPARED records are written in the future, we can modify the implementation introduced by this PR to copy the maps and delay the timing to pass changed read-set and write
-set to the snapshot hook handler. What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -50,7 +55,45 @@ protected void onPrepareFailure(Snapshot snapshot) {}
...
 
+  private Optional<BeforePreparationSnapshotHookFuture> invokeBeforePreparationSnapshotHook(
+      Snapshot snapshot) throws UnknownTransactionStatusException, CommitException {
+    try {
+      if (beforePreparationSnapshotHook == null) {
+        return Optional.empty();
+      }
+
+      return Optional.of(beforePreparationSnapshotHook.handle(tableMetadataManager, snapshot));
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
komamitsu: [Done](https://github.com/scalar-labs/scalardb/pull/2271/commits/964f5fa6e3d7d8438a4786293cca8f130e5a3e7c)

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
komamitsu: [Done!](https://github.com/scalar-labs/scalardb/pull/2271/commits/2e5db19cf6858c8dfae1843757832f109a198744)

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
brfrn169: > I don’t think just copying the maps is sufficient.

Ah yes, if we assume the preparation, validation, and commit phases might change the state of the snapshot (read-set, write-set, etc.), I agree that simply copying the maps is not sufficient.

Ideally, to make the hook more general, I think we should pass an immutable state of the snapshot at the time suggested by the hook’s name.

For example:
- For `BeforePreparationSnapshotHook`, we should pass the immutable state before preparation.
- For `BeforeValidationSnapshotHook`, we should pass the immutable state before validation.
- For `BeforeCommitSnapshotHook`, we should pass the immutable state before commit.

(We won’t add `BeforeValidationSnapshotHook` and `BeforeCommitSnapshotHook` in this PR though.)

However, in the current implementation, the snapshot passed to the hook is mutable, and since the hook could be executed asynchronously, if the preparation, validation, and commit phases change the state of the snapshot, the hook could access the state at an unpredictable time.

> How about going with the current no-copy approach in this PR for now? If we need to change read-set and write-set after PREPARED records are written in the future, we can modify the implementation introduced by this PR to copy the maps and delay the timing to pass changed read-set and write
-set to the snapshot hook handler. What do you think?

I’m fine with that, but should we add some validation to check if extra-write is used when `BeforePreparationSnapshotHook` is set?

komamitsu: > However, in the current implementation, the snapshot passed to the hook is mutable, and since the hook could be executed asynchronously, if the preparation, validation, and commit phases change the state of the snapshot, the hook could access the state at an unpredictable time.

I think you have a point. It's hard to create an immutable snapshot. So, do you think we should give up passing a snapshot and just passes the immutable read/write-sets like this?

```java
public class ReadWriteSets {
   // These are copied from Snapshot
   final Map<Key, TransactionResult> readSet;
   final Map<Key, Put> writeSet;
   final Map<Key, Delete> deleteSet;
}

public interface BeforePreparationReadWriteSetsHook {
  BeforePreparationReadWriteSetsHookFuture handle(ReadWriteSets readWriteSets);
}
```

I'm okay with it.

As for extra-write, yeah I'll add a validation for that in this PR later.

brfrn169: > So, do you think we should give up passing a snapshot and just passes the immutable read/write-sets like this?

Yes, that’s what I had in mind. Thanks.

> As for extra-write, yeah I'll add a validation for that in this PR later.

Maybe, if we pass the immutable read/write-sets to the hook, we may be able to support extra-write as well.

The extra-write modifies the write-set in the preparation phase, but actually I don’t think we need to replicate it to the secondary site since it’s only for concurrency control.

core/src/main/java/com/scalar/db/transaction/consensuscommit/BeforePreparationSnapshotHook.java
```diff
@@ -0,0 +1,10 @@
+package com.scalar.db.transaction.consensuscommit;
+
+public interface BeforePreparationSnapshotHook {
+  BeforePreparationSnapshotHookFuture handle(
```
brfrn169, line 4: Sorry, one more thing. Why not use the Java `Future` as follows?
```suggestion
  Future<Void> handle(
```

komamitsu: I first thought using the custom future would be helpful if we add a new API like `cancel()` to the future. But it also requires some changes in `CommitHandler` to call new APIs. Since we can postpone implementing the custom future until we actually need such new APIs, I'll go ahead and merge your suggestion to simplify things for now.

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
komamitsu: > The extra-write modifies the write-set in the preparation phase, but actually I don’t think we need to replicate it to the secondary site since it’s only for concurrency control.

It seems like this causes inconsistent `tx_id` between the primary site and the backup site. Records read by a transaction will be updated with the transaction ID in the primary site, while the same records won't be updated in the backup site.

> As for extra-write, yeah I'll add a validation for that in this PR later.

[Done](https://github.com/scalar-labs/scalardb/pull/2271/commits/5b771342cf4171716f95b2612ed3aef835634ffe)

core/src/main/java/com/scalar/db/transaction/consensuscommit/BeforePreparationSnapshotHook.java
komamitsu: [Done](https://github.com/scalar-labs/scalardb/pull/2271/commits/657558cb56c213c81ce84702dcfbdc5717b0c1aa)

core/src/main/java/com/scalar/db/transaction/consensuscommit/PreparedSnapshotHook.java
komamitsu: > > So, do you think we should give up passing a snapshot and just passes the immutable read/write-sets like this?
>
> Yes, that’s what I had in mind. Thanks.

[Done](https://github.com/scalar-labs/scalardb/pull/2271/commits/8f0817d88b87168095a743f5fa3f10d273aea655)

@brfrn169 I noticed `new ArrayList(map.entrySet())` is twice faster than returning `new HashMap(map)`. So, I added the method `List<Entry<Key, Put>> getKeysAndPutsInWriteSet()` instead of `Map<Key, Put> getWriteSetMap()`. I think the new method is very similar to the existing method `List<Put> getPutsInWriteSet()`. Probably we can replace the usage of `List<Put> getPutsInWriteSet()` with the new method. What do you think?

komamitsu: >> As for extra-write, yeah I'll add a validation for that in this PR later.
>
> Done

On second thought, whether to prohibit EXTRA WRITE depends on the usage of write sets. On the above discussions, I was only thinking of the use case of replication. But, write set captured before `prepare` might work with EXTRA WRITE in other use cases. So,  I think a validation to prohibit using EXTRA WRITE should be handled by a plug-in side not in `scalardb:core`. Probably I should revert https://github.com/scalar-labs/scalardb/pull/2271/commits/8f0817d88b87168095a743f5fa3f10d273aea655 ...

komamitsu: > > > So, do you think we should give up passing a snapshot and just passes the immutable read/write-sets like this?
> >
> > Yes, that’s what I had in mind. Thanks.

> [Done](https://github.com/scalar-labs/scalardb/pull/2271/commits/8f0817d88b87168095a743f5fa3f10d273aea655)

Oops. What I did was not along with this idea. I'll address this...

komamitsu: > On second thought, whether to prohibit EXTRA WRITE depends on the usage of write sets. On the above discussions, I was only thinking of the use case of replication. But, write set captured before prepare might work with EXTRA WRITE in other use cases. So, I think a validation to prohibit using EXTRA WRITE should be handled by a plug-in side not in scalardb:core. Probably I should revert https://github.com/scalar-labs/scalardb/commit/8f0817d88b87168095a743f5fa3f10d273aea655 ...

[Reverted](https://github.com/scalar-labs/scalardb/pull/2271/commits/882e476250eb0917822e0f021104bdca11be87b3)

komamitsu:
@feeblefakie JFYI: I added the following changes after you approves this PR based on some second thoughts and discussion with @brfrn169. Feel free to share your concerns or thoughts.

- Renamed `PreparedSnapshotHook` to `BeforePreparationSnapshotHook`
  - Reason: `Prepared` still sounds "after-prepare-phase"
- Made `BeforePreparationSnapshotHook.handle()` simply return `Future<Void>` for now
  - Reason: I initially used `PreparedSnapshotHookFuture` since I had considered the potential extendability of the return value. But, we would need to `CommitHandler` as well if we add a new API to the returned Future-ish value.  Using `PreparedSnapshotHookFuture` isn't so useful at the moment.
- Passed immutable `Snapshot.ReadWriteSet` instead of mutable `Snapshot` itself to the snapshot hooks.
  - Reason: For safety considering future changes.

@brfrn169 I updated this PR. PTAL when you have a chance.

brfrn169 (approved):
Left a minor comment. Other than that, LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -50,7 +55,52 @@ protected void onPrepareFailure(Snapshot snapshot) {}
...
+      snapshotHookFuture.get();
+    } catch (Exception e) {
+      abortState(snapshot.getId());
+      rollbackRecords(snapshot);
+      // TODO: This method is actually a part of validation phase. But the callback method name
+      //       `onValidateFailure()` should be renamed to more reasonable one.
+      onValidateFailure(snapshot);
+      throw new CommitException(
+          CoreError.HANDLING_BEFORE_PREPARATION_SNAPSHOT_HOOK_FAILED.buildMessage(e.getMessage()),
+          e,
```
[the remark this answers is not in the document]
komamitsu: `Future` (== `FutureTask`) wraps an exception occurs in `get()` with `new ExecutionException((Throwable)cause)`. [`new Throwable(Throwable cause)`](https://github.com/openjdk/jdk/blob/b9b0bd0871886eb65f87864f262424b119f2c748/src/java.base/share/classes/java/lang/Throwable.java#L336-L344) sets `detailedMessage` to `cause.toString()`. So, I think it's okay in this case.

```java
    ExecutorService executorService = Executors.newCachedThreadPool();
    Future<Object> future = executorService.submit(() -> {
      throw new RuntimeException("This is the original exception");
    });
    executorService.shutdown();

    try {
      future.get();
    }
    catch (Exception e) {
      System.out.println("Message from the exception: " + e.getMessage());
      System.out.println("Message from the cause: " + e.getCause());
    }
```

```
Message from the exception: java.lang.RuntimeException: This is the original exception
Message from the cause: java.lang.RuntimeException: This is the original exception
```

Torch3333 (approved):
LGTM, thank you!

PR #2291, opened 2024-10-24
Fix scan with limit behavior in DynamoDB adapter

brfrn169:
## Description

This PR fixes the scan with limit behavior in the DynamoDB adapter. In this bug, the scan with a limit doesn’t work correctly when the scan results exceed 1 MB and require multiple requests to DynamoDB.

## Related issues and/or PRs

N/A

## Changes made

Added inline comments. Please take a look at them.

## Checklist

N/A

## Release notes

Fixed a bug where the scan with a limit doesn’t work correctly when the scan results exceed 1 MB and require multiple requests to DynamoDB.


core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
```diff
@@ -49,18 +58,23 @@ private boolean hasNext() {
       return true;
     }
     if (lastEvaluatedKey != null) {
-      handleResponse(request.execute(lastEvaluatedKey));
+      if (remainingLimit != null) {
+        handleResponse(request.execute(lastEvaluatedKey, remainingLimit));
+      } else {
+        handleResponse(request.execute(lastEvaluatedKey));
+      }
```
brfrn169, line 65: If we have the renaming limit, specify the limit and send a request to DynamoDB.

core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
```diff
@@ -49,18 +58,23 @@ private boolean hasNext() {
...
       return itemsIterator.hasNext();
     }
     return false;
   }
 
   private void handleResponse(PaginatedRequestResponse response) {
     List<Map<String, AttributeValue>> items = response.items();
-    totalResultCount += items.size();
+    if (remainingLimit != null) {
+      remainingLimit -= items.size();
```
brfrn169, line 74: We subtract the remaining limit by the result count.

core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
```diff
@@ -49,18 +58,23 @@ private boolean hasNext() {
...
+    }
     itemsIterator = items.iterator();
-    if ((request.limit() == null || totalResultCount < request.limit())
-        && response.hasLastEvaluatedKey()) {
+    if ((remainingLimit == null || remainingLimit > 0) && response.hasLastEvaluatedKey()) {
```
brfrn169, line 77: If we have the remaining limit, we should send an additional request with the remaining limit to DynamoDB.

feeblefakie, line 77: I'm confused so I'm probably wrong, but should it be something like below?
```suggestion
    if ((remainingLimit != null && remainingLimit > 0) && response.hasLastEvaluatedKey()) {
```
I'm probably a bit confused with the meaning of null `remainingLimit`.

feeblefakie (commented):
Overall, looking good. 
Left a question just in case.

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
```diff
@@ -142,16 +142,16 @@ private Scanner executeScanWithIndex(Selection selection, TableMetadata tableMet
 
     builder.expressionAttributeNames(expressionAttributeNames);
 
+    int limit = 0;
     if (selection instanceof Scan) {
       Scan scan = (Scan) selection;
-      if (scan.getLimit() > 0) {
-        builder.limit(scan.getLimit());
```
komamitsu, line 148: I'm not sure what was the root cause, but I guess this doesn't expectedly work. Is my guess correct?

brfrn169: Yes, this doesn’t work as expected. The limit parameter actually functions more like a page size (the number of records in a single request) rather than the total number of records to retrieve.

core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
brfrn169: The logic is as follows:

- If `remainingLimit` is `null` (meaning the user hasn’t specified a limit in Scan), and the response includes the last evaluated key, we need to send an additional request to DynamoDB.
- If `remainingLimit` is not `null` (meaning the user has specified a limit in Scan) and `remainingLimit` is greater than `0`, and if the response includes the last evaluated key, we also need to send an additional request to DynamoDB.
- Otherwise, no further requests to DynamoDB are necessary.

So I think the current code is correct.

komamitsu (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java
Torch3333: I don't understand why you removed the line `builder.limit(scan.getLimit());`
Could you explain why you don't set the limit on the initial dynamo query only in the case of a ScanWithIndex (`SelectStatementHandler.executeScan()` and `SelectStatementHandler.executeFullScan` still set the limit on the request builder) ?

brfrn169: @Torch3333 Instead of using `builder.limit(scan.getLimit());` here, we call the method in the following part (`handleResponse(request.execute(limit));`):
https://github.com/scalar-labs/scalardb/pull/2291/files#diff-08e57801daa3c3c5ed64e1afbf4b043a9b2a581af89d92e64712af5cf051af81R35-R41

So we are actually still calling `builder.limit()` elsewhere.

Torch3333: I see, thanks for the explanation. 

Since the limit is set all cases in
 - https://github.com/scalar-labs/scalardb/blob/fe602d4d32e34aaa2806c2d40dbed0d3e9090794/core/src/main/java/com/scalar/db/storage/dynamo/request/QueryRequest.java#L33
 - https://github.com/scalar-labs/scalardb/blob/fe602d4d32e34aaa2806c2d40dbed0d3e9090794/core/src/main/java/com/scalar/db/storage/dynamo/request/ScanRequest.java#L33
 
 Do you think we could safely remove these two lines?
 - https://github.com/scalar-labs/scalardb/blob/fe602d4d32e34aaa2806c2d40dbed0d3e9090794/core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java#L175
 - https://github.com/scalar-labs/scalardb/blob/fe602d4d32e34aaa2806c2d40dbed0d3e9090794/core/src/main/java/com/scalar/db/storage/dynamo/SelectStatementHandler.java#L199

brfrn169: Ah yes, good catch! I forgot to remove those. I’ll fix them. thanks!

brfrn169: Fixed in 244babbb226362abd0517b981d10931bea0321ac. Thanks.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/storage/dynamo/QueryScanner.java
feeblefakie: I got a better understanding. Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #2292, opened 2024-10-24
Add validation for primary key columns in Cosmos DB adapter

brfrn169:
## Description

This PR adds validation for primary key columns in the Cosmos DB adapter. The validation ensures that the text values of the primary key columns do not contain illegal characters (`:`, `/`, `\`, `#`, and `?`).

The reasons for this validation are as follows:

- The colon (`:`) is used by `ConcatenationVisitor` as a key separator.
- The characters `/`, `\`, `#`, and `?` are restricted because we concatenate primary key column values for the `Id` property of the record, and these characters cannot be used in the property. For more details, see [this documentation](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.resource.id?view=azure-dotnet#remarks).

## Related issues and/or PRs

N/A

## Changes made

- Added validation for primary key columns in the Cosmos DB adapter.
- Also changed the transaction ID delimiter for Group Commit from `:` to `$` since `:` is disallowed in the Cosmos DB adapter.

## Checklist

N/A

## Release notes

Added validation for primary key columns in the Cosmos DB adapter. The validation ensures that the text values of the primary key columns do not contain illegal characters (`:`, `/`, `\`, `#`, and `?`).


core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
```diff
@@ -2,36 +2,118 @@
...
+import com.scalar.db.io.BlobColumn;
+import com.scalar.db.io.BooleanColumn;
+import com.scalar.db.io.ColumnVisitor;
 import com.scalar.db.io.DataType;
+import com.scalar.db.io.DoubleColumn;
+import com.scalar.db.io.FloatColumn;
+import com.scalar.db.io.IntColumn;
+import com.scalar.db.io.TextColumn;
 
 public class CosmosOperationChecker extends OperationChecker {
```
brfrn169, line 26: Added the validation for primary key columns here.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CoordinatorGroupCommitter.java
```diff
@@ -38,7 +38,7 @@ public static boolean isEnabled(ConsensusCommitConfig config) {
   public static class CoordinatorGroupCommitKeyManipulator
       implements KeyManipulator<String, String, String, String, String> {
     private static final int PRIMARY_KEY_SIZE = 24;
-    private static final char DELIMITER = ':';
+    private static final char DELIMITER = '$';
```
brfrn169, line 41: Also changed the delimiter for Group Commit from `:` to `$` since `:` is disallowed in the Cosmos DB adapter.

komamitsu (approved):
LGTM! 👍 

BTW my understanding is that, after this validation is introduced, ScalarDB tables that are stored in other underlying databases and contain those characters (`[:/\\#?]`) in their keys can't be migrated to Cosmos DB without converting the keys. Is that correct?

Torch3333 (approved):
LGTM, thank you!

jnmt (approved):
LGTM, thank you! We might also need to mention this limitation somewhere on the ScalarDL side.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
```diff
@@ -2,36 +2,118 @@
...
+
+        @Override
+        public void visit(DoubleColumn column) {}
+
+        @Override
+        public void visit(TextColumn column) {
+          String value = column.getTextValue();
+          assert value != null;
+
+          if (value.indexOf(':') != -1
```
feeblefakie, line 50: Should the invalid characters be managed as a list and defined as a static variable?
(here, we can do a simple loop to check if a given column includes the characters.)

The list can be joined and also used by the message for `COSMOS_PRIMARY_KEY_CONTAINS_ILLEGAL_CHARACTER` so that we don't have to maintain the characters in different places separately?

feeblefakie (commented):
Overall, looking good!
Left one comment/suggestion. PTAL!

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
brfrn169: @feeblefakie Thank you. Fixed in 4a226599cb071027749a1376a6469a5a3e92d4d9.

Regarding the message for `COSMOS_PRIMARY_KEY_CONTAINS_ILLEGAL_CHARACTER`, I didn’t make any modifications in this commit. If we make the illegal characters a variable, the message in the error code document appears as follows:
```
The value of the column %s in the primary key contains an illegal character. Primary key columns must not contain any of the following characters: %s. Value: %s
```

As you can see, it doesn’t show which characters are illegal; it only displays `%s`. That’s why I chose not to replace it with a variable. What do you think?

brfrn169:
@komamitsu

> BTW my understanding is that, after this validation is introduced, ScalarDB tables that are stored in other underlying databases and contain those characters ([:/\\#?]) in their keys can't be migrated to Cosmos DB without converting the keys. Is that correct?

That's correct. To address this, we’re considering creating guideline documentation advising that, if you might migrate your data in the future, you should avoid using the characters, even if you’re not currently using Cosmos DB.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -659,6 +659,13 @@ public enum CoreError implements ScalarDbError {
       "The variable key column size must be greater than or equal to 64",
       "",
       ""),
+  COSMOS_PRIMARY_KEY_CONTAINS_ILLEGAL_CHARACTER(
+      Category.USER_ERROR,
+      "0145",
+      "The value of the column %s in the primary key contains an illegal character. "
+          + "Primary key columns must not contain any of the following characters: ':', '/', '\\', '#', '?'. Value: %s",
```
josh-wong, line 666:
```suggestion
          + "Primary-key columns must not contain any of the following characters: ':', '/', '\\', '#', '?'. Value: %s",
```

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
```diff
@@ -2,36 +2,130 @@
...
 import com.scalar.db.io.DataType;
+import com.scalar.db.io.DoubleColumn;
+import com.scalar.db.io.FloatColumn;
+import com.scalar.db.io.IntColumn;
+import com.scalar.db.io.TextColumn;
 
 public class CosmosOperationChecker extends OperationChecker {
+
+  private static final char[] ILLEGAL_CHARACTERS_IN_PRIMARY_KEY = {
+    // Colons are not allowed in primary key columns due to the `ConcatenationVisitor` limitation
```
josh-wong, line 29:
```suggestion
    // Colons are not allowed in primary-key columns due to the `ConcatenationVisitor` limitation.
```

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
```diff
@@ -2,36 +2,130 @@
...
+    ':',
+
+    // The following characters are not allowed in primary key columns because they are restricted
```
josh-wong, line 32:
```suggestion
    // The following characters are not allowed in primary-key columns because they are restricted
```

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
```diff
@@ -2,36 +2,130 @@
...
+    // and cannot be used in the `Id` property of a Cosmos DB document. See the following link for
+    // more information:
+    // https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.databaseproperties.id?view=azure-dotnet#remarks
```
josh-wong, line 35:
```suggestion
    // and cannot be used in the `Id` property of a Cosmos DB document. For more information, see:
    // https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.databaseproperties.id?view=azure-dotnet#remarks
```

josh-wong (approved):

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
```diff
@@ -2,36 +2,130 @@
...
+    '/',
+    '\\',
```
kota2and3kan, line 37:
```suggestion
    '\',
```

The [document](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.resource.id?view=azure-dotnet#remarks) states the character `\\` (double backslashes) is restricted.

However, it seems that we cannot use the character `\` (single backslash) too.

By using ScalarDB 3.13.1, I created the test table and inserted test records that include `\` and `\\`.

```sql
0: scalardb> create table foo (a text primary key, b int);
No rows affected (0.508 seconds)
0: scalardb>
0: scalardb> insert into foo values ('aaa\bbb', 0);
1 row affected (0.27 seconds)
0: scalardb>
0: scalardb> insert into foo values ('aaa\\bbb', 0);
1 row affected (0.102 seconds)
```

After that, I retrieved both records respectively from the Azure Portal.

As a result, both `aaa\bbb` and `aaa\\bbb` return the following errors.

- `\` (single backslash)
  ```console
  Failed to read item aaa\bbb:
   Illegal characters ['/', '\', '#'] cannot be used in Resource ID
  ```
- `\\` (double backslashes)
  ```console
  Failed to read item aaa\\bbb:
   Illegal characters ['/', '\', '#'] cannot be used in Resource ID
  ```

So, I think the document description `\\` is a typo...
(I guess, they may want to represent the single `\`, by using `\` as an escape character, but it seems that it does not work properly. As a result, maybe, the document states `\\`...)

Therefore, I think it would be better to validate `\` (single backslash) instead of `\\` (double backslashes).

core/src/test/java/com/scalar/db/storage/cosmos/CosmosOperationCheckerTest.java
```diff
@@ -397,6 +418,284 @@ public void check_ForMutationsWithDeleteWithCondition_ShouldBehaveProperly() {
...
+            .namespace(NAMESPACE_NAME)
+            .table(TABLE_NAME)
+            .partitionKey(Key.ofText(PKEY1, "ab"))
+            .clusteringKey(Key.ofText(CKEY1, "a/b"))
+            .build();
+    Get get6 =
+        Get.newBuilder()
+            .namespace(NAMESPACE_NAME)
+            .table(TABLE_NAME)
+            .partitionKey(Key.ofText(PKEY1, "a\\b"))
```
kota2and3kan, line 467:
```suggestion
            .partitionKey(Key.ofText(PKEY1, "a\b"))
```
Ditto. I think we should validate `\` (single backslash) instead of `\\` (double backslashes).

core/src/test/java/com/scalar/db/storage/cosmos/CosmosOperationCheckerTest.java
```diff
@@ -397,6 +418,284 @@ public void check_ForMutationsWithDeleteWithCondition_ShouldBehaveProperly() {
...
+            .clusteringKey(Key.ofText(CKEY1, "ab"))
+            .build();
+    Get get7 =
+        Get.newBuilder()
+            .namespace(NAMESPACE_NAME)
+            .table(TABLE_NAME)
+            .partitionKey(Key.ofText(PKEY1, "ab"))
+            .clusteringKey(Key.ofText(CKEY1, "a\\b"))
```
kota2and3kan, line 475:
```suggestion
            .clusteringKey(Key.ofText(CKEY1, "a\b"))
```
Ditto. I think we should validate `\` (single backslash) instead of `\\` (double backslashes).

kota2and3kan (commented):
Thank you for the updates!
I left suggestions about `\`.
Please take a look when you have time! 🙇‍♂️

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
brfrn169: `\\` represents a single backslash (the first backslash is for escaping). So, the current code should be correct. Or, when you ran this fix, were backslashes not validated correctly?

kota2and3kan: Thank you for your reply!

> `\\` represents a single backslash (the first backslash is for escaping) in Java (and probably C#). So, the current code should be correct. 

Ah, I see... Sorry, I thought that `\\` does not mean `\` in the validation process when I read the code...

> Or, when you ran this fix, were backslashes not validated correctly?

I just noticed it in the code review before I tried the actual behavior, so I have not tested it yet.

I will test the validation behavior from now on. So, please give me some time.

kota2and3kan: I was able to confirm that all illegal characters (including `\`) are restricted on the ScalarDB side in my local environment as follows!

<details>
<summary> Test result</summary>

- Create a test table
  ```sql
  0: scalardb> create table foo (a text, b text, c int, primary key (a,b));
  No rows affected (0.553 seconds)
  0: scalardb>
  0: scalardb> describe foo;
  +------------+------+--------------+----------------+-----------------+-----------------+-----------+-------------+
  | columnName | type | isPrimaryKey | isPartitionKey | isClusteringKey | clusteringOrder | isIndexed | isEncrypted |
  +------------+------+--------------+----------------+-----------------+-----------------+-----------+-------------+
  | a          | TEXT | true         | true           | false           | null            | false     | false       |
  | b          | TEXT | true         | false          | true            | ASC             | false     | false       |
  | c          | INT  | false        | false          | false           | null            | false     | false       |
  +------------+------+--------------+----------------+-----------------+-----------------+-----------+-------------+
  3 rows selected (0.043 seconds)
  ```
- INSERT
  - Partition key
    ```sql
    0: scalardb> insert into foo values ('aaa/bbb', 'b', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa/bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> insert into foo values ('aaa\bbb', 'b', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa\bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> insert into foo values ('aaa#bbb', 'b', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa#bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> insert into foo values ('aaa?bbb', 'b', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa?bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> insert into foo values ('aaa:bbb', 'b', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa:bbb) (state=42000,code=501)
    ```
  - Clustering key
    ```sql
    0: scalardb> insert into foo values ('a', 'xxx/yyy', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx/yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> insert into foo values ('a', 'xxx\yyy', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx\yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> insert into foo values ('a', 'xxx#yyy', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx#yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> insert into foo values ('a', 'xxx?yyy', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx?yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> insert into foo values ('a', 'xxx:yyy', 0);
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx:yyy) (state=42000,code=501)
    ```
- SELECT
  - Partition key
    ```sql
    0: scalardb> select * from foo where a = 'aaa/bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa/bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> select * from foo where a = 'aaa\bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa\bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> select * from foo where a = 'aaa#bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa#bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> select * from foo where a = 'aaa?bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa?bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> select * from foo where a = 'aaa:bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa:bbb) (state=42000,code=501)
    ```
  - Clustering key
    ```sql
    0: scalardb> select * from foo where a = 'a' and b = 'xxx/yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx/yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> select * from foo where a = 'a' and b = 'xxx\yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx\yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> select * from foo where a = 'a' and b = 'xxx#yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx#yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> select * from foo where a = 'a' and b = 'xxx?yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx?yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> select * from foo where a = 'a' and b = 'xxx:yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx:yyy) (state=42000,code=501)
    ```
- UPDATE
  - Partition key
    ```sql
    0: scalardb> update foo set c = 100 where a = 'aaa/bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa/bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> update foo set c = 100 where a = 'aaa\bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa\bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> update foo set c = 100 where a = 'aaa#bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa#bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> update foo set c = 100 where a = 'aaa?bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa?bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> update foo set c = 100 where a = 'aaa:bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa:bbb) (state=42000,code=501)
    ```
  - Clustering key
    ```sql
    0: scalardb> update foo set c = 100 where a = 'a' and b = 'xxx/yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx/yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> update foo set c = 100 where a = 'a' and b = 'xxx\yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx\yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> update foo set c = 100 where a = 'a' and b = 'xxx#yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx#yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> update foo set c = 100 where a = 'a' and b = 'xxx?yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx?yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> update foo set c = 100 where a = 'a' and b = 'xxx:yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx:yyy) (state=42000,code=501)
    ```
- DELETE
  - Partition key
    ```sql
    0: scalardb> delete from foo where a = 'aaa/bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa/bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> delete from foo where a = 'aaa\bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa\bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> delete from foo where a = 'aaa#bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa#bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> delete from foo where a = 'aaa?bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa?bbb) (state=42000,code=501)
    0: scalardb>
    0: scalardb> delete from foo where a = 'aaa:bbb';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column a in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: aaa:bbb) (state=42000,code=501)
    ```
  - Clustering key
    ```sql
    0: scalardb> delete from foo where a = 'a' and b = 'xxx/yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx/yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> delete from foo where a = 'a' and b = 'xxx\yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx\yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> delete from foo where a = 'a' and b = 'xxx#yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx#yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> delete from foo where a = 'a' and b = 'xxx?yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx?yyy) (state=42000,code=501)
    0: scalardb>
    0: scalardb> delete from foo where a = 'a' and b = 'xxx:yyy';
    Error: Invalid query (INVALID_ARGUMENT: CORE-10145: The value of the column b in the primary key contains an illegal character. Primary-key columns must not contain any of the following characters: ':', '/', '\', '#', '?'. Value: xxx:yyy) (state=42000,code=501)
    ```
</details>

kota2and3kan (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
brfrn169: @kota2and3kan Thank you for the confirmation! 🙇

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -659,6 +659,13 @@ public enum CoreError implements ScalarDbError {
       "The variable key column size must be greater than or equal to 64",
       "",
       ""),
+  COSMOS_PRIMARY_KEY_CONTAINS_ILLEGAL_CHARACTER(
+      Category.USER_ERROR,
+      "0145",
+      "The value of the column %s in the primary key contains an illegal character. "
+          + "Primary-key columns must not contain any of the following characters in Cosmos DB: ':', '/', '\\', '#', '?'. Value: %s",
```
feeblefakie, line 666: I might miss your point, but can we do something like below?
It might be too much, but we can avoid managing the invalid characters in two places.

```suggestion
      "The value of the column %s in the primary key contains an illegal character. "
          + "Primary-key columns must not contain any of the following characters in Cosmos DB: "
          + Chars.asList(CosmosOperationChecker.ILLEGAL_CHARACTERS_IN_PRIMARY_KEY).stream()
              .map(c -> "'" + c + "'")
              .collect(Collectors.joining(", "))
          + ". Value: %s";
```

core/src/main/java/com/scalar/db/storage/cosmos/CosmosOperationChecker.java
feeblefakie: I might miss your point, but I left a suggestion. PTAL.
https://github.com/scalar-labs/scalardb/pull/2292#discussion_r1820525473

core/src/main/java/com/scalar/db/common/error/CoreError.java
brfrn169: @feeblefakie As discussed, we will leave it as is. This is because we should avoid creating a dependency from the Cosmos adapter to CoreError, as each adapter will be separated into its own sub-module in the future.

feeblefakie (approved):
LGTM! Thank you!

PR #2331, opened 2024-11-08
Optimize `Coordinator#getStateForGroupCommit()` by first looking up the coordinator table with parent transaction ID

komamitsu:
## Description

The current implementation of `Coordinator.getStateForGroupCommit()` checks transaction states that were stored with the Group Commit feature enabled in the following order:
1. First, it tries to find a record whose `tx_id` is the full transaction ID, returning the state if it's found. This record type is created when a transaction is delayed and committed individually.
2. Next, it tris to find a record whose `tx_id` is the parent transaction ID, returning the state if it's found and the `tx_sub_ids` contains the child transaction ID. This record type is created when a transaction gets commit-ready in time and group-committed.

This order is optimized for lazy-recovery, which is more likely to happen if the commit timing is delayed after PREPARE record is created. However, basically lazy-recovery doesn't happen so often.

We may create a new system that frequently looks up the coordinator table in the future. In that case, it might need to prioritize first finding group-committed transactions, not just delayed individually committed transaction. The current order doesn't fit this scenario.

This PR changes the lookup order to prioritize group-committed transactions first.

## Related issues and/or PRs

None

## Changes made

- Change the order of GET operations with full transaction ID and parent ID in `Coordinator#getStateForGroupCommit()`.
- Update the unit test along with the change.

## Checklist

None

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
@@ -83,27 +83,6 @@ public class Coordinator {
 
   @VisibleForTesting
   Optional<Coordinator.State> getStateForGroupCommit(String fullId) throws CoordinatorException {
-    // Reading a coordinator state is likely to occur during lazy recovery, as follows:
-    // 1. Transaction T1 starts and creates PREPARED state records but hasn't committed or aborted
-    //    yet.
-    // 2. Transaction T2 starts and reads the PREPARED state records created by T1.
-    // 3. T2 reads the coordinator table record for T1 to decide whether to roll back or roll
-    //    forward.
-    //
-    // The likelihood of step 2 would increase if T1 is delayed.
-    //
-    // With the group commit feature enabled, delayed transactions are isolated from a normal group
-    // that is looked up by a parent ID into a delayed group that is looked up by a full ID.
-    // Therefore, looking up with the full transaction ID should be tried first to minimize read
-    // operations as much as possible.
-
-    // Scan with the full ID for a delayed group that contains only a single transaction.
-    // The normal lookup logic can be used as is.
-    Optional<State> stateOfDelayedTxn = get(createGetWith(fullId));
-    if (stateOfDelayedTxn.isPresent()) {
-      return stateOfDelayedTxn;
-    }
-
     // Scan with the parent ID for a normal group that contains multiple transactions.
     Keys<String, String, String> idForGroupCommit = keyManipulator.keysFromFullKey(fullId);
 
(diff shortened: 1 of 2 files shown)


brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -83,41 +83,29 @@ public Optional<Coordinator.State> getState(String id) throws CoordinatorExcepti
...
     Get get = createGetWith(parentId);
     Optional<State> state = get(get);
-    return state.flatMap(
-        s -> {
-          if (s.getChildIds().contains(childId)) {
-            return state;
-          }
-          return Optional.empty();
-        });
+    // The current implementation is optimized for cases where the target transaction is properly
```
feeblefakie, line 93: The comment is a bit hard to follow to me, so can it be something like this?
Please also verify if it is correct.

```
The current implementation is optimized for a case where most transactions are group-committed.
Specifically, Coordinator first checks a transaction state by using a parent ID with one read operation.
If it cannot find it, which means the transaction is delayed and not included in a group, it will issue an additional read operation by using the full ID.
```

feeblefakie (commented):
The code looks good!
I left a suggestion on the comment.
PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
komamitsu: @feeblefakie Thanks! It seems better.

How about the following version revised based on your suggestion?

> The current implementation is optimized for cases where most transactions are group-committed. It first looks up a transaction state by the parent ID with a single read operation. If no matching transaction state is found (i.e., the transaction was delayed and committed individually), it issues an additional read operation using the full ID.

komamitsu: Updated the comment in https://github.com/scalar-labs/scalardb/pull/2331/commits/1b62d3c06a5be507f0db67a31659dd961c3b0a72

feeblefakie: It looks good!
Just one thing. Let's use consistent wording (using xxx ID).

> The current implementation is optimized for cases where most transactions are group-committed. It first looks up a transaction state using the parent ID with a single read operation. If no matching transaction state is found (i.e., the transaction was delayed and committed individually), it issues an additional read operation using the full ID.

feeblefakie (approved):
Left a comment on the Javadoc comment, other than that LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
komamitsu: Thanks! Fixed it in https://github.com/scalar-labs/scalardb/pull/2331/commits/6e3ece25d21df9bad6ca2e64432ff2ccdc3204d7.

PR #2340, opened 2024-11-14
Revisit behavior of multiple mutations for same record in transaction in Consensus Commit

brfrn169:
## Description

This PR revisits the behavior of multiple mutations for the same record in a transaction in Consensus Commit.

Expected behaviors for combinations of mutations on the same record are as follows:

1. Insert then Insert: the second Insert fails.
2. Insert then Upsert: merge Insert and Upsert (with implicit pre-read disabled), then execute.
3. Insert then Update: merge Insert and Update (with implicit pre-read disabled), then execute.
4. Insert then Delete: ~only execute Delete.~ not allowed in Consensus Commit.
5. Upsert then Insert: Insert fails.
6. Upsert then Upsert: merge the two Upserts, then execute.
7. Upsert then Update: merge Upsert and Update, then execute.
8. Upsert then Delete: only execute Delete.
9. Update then Insert: if the target record exists, Insert fails; if not, only execute Insert.
10. Update then Upsert: if the target record exists, only execute Upsert; if not, merge Update and Upsert, then execute.
11. Update then Update: if the target record exists, merge the two Updates; if not, do nothing.
12. Update then Delete: only execute Delete.
13. Delete then Insert: not allowed in Consensus Commit.
14. Delete then Upsert: not allowed in Consensus Commit.
15. Delete then Update: do nothing.
16. Delete then Delete: only execute the second Delete.

I've added some test cases to [ConsensusCommitIntegrationTestBase.java](https://github.com/scalar-labs/scalardb/pull/2340/files#diff-078305d8965c9dfa5f28cfd4c0fa81fb59065f5722cd5b1ebc4b584ae51c94ab) to check the behavior. 

To achieve this behavior, we need to make the following changes:

- Consider the write set for conditional updates.
- Add validation to ensure that Insert is not executed on an already written record in the same transaction.
- Updated the merge logic for the write set. Please see the inline comment for details.

[ADDED]

For the expected behaviors 1, 2, 3, and 4 (where the first operation is Insert), as @komamitsu pointed out in his comment, if the target record already exists, the transaction should fail.

For cases 1, 2, and 3, the transaction correctly fails in such situations. However, for case 4, it does not fail, as @komamitsu mentioned in the following comment:
https://github.com/scalar-labs/scalardb/pull/2340#issuecomment-2477995173

I think this behavior is confusing for users. To address this, we can make the following additional change:

- Add validation to ensure that Delete is not executed on a record that has already been inserted in the same transaction.

## Related issues and/or PRs

N/A

## Changes made

- Use the result that's merged with mutations in the write set for conditional updates.
- Added validation to ensure that Insert is not executed for already written record.
- Updated the merge logic for the write set.
- Refactored `Snapshot`.

## Checklist

The Insert, Upsert, and Update are converted to Put internally. See the following for details:
https://github.com/scalar-labs/scalardb/blob/78f3883c676de6a788b702ddd2e5f4e2667d24d4/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java#L247-L293 

## Release notes

Fixed the behavior of multiple mutations for the same record in a transaction in Consensus Commit.

### DIFF
--- a/core/src/main/java/com/scalar/db/common/error/CoreError.java
+++ b/core/src/main/java/com/scalar/db/common/error/CoreError.java
@@ -666,6 +666,8 @@ public enum CoreError implements ScalarDbError {
           + "Primary-key columns must not contain any of the following characters in Cosmos DB: ':', '/', '\\', '#', '?'. Value: %s",
       "",
       ""),
+  CONSENSUS_COMMIT_INSERTING_ALREADY_WRITTEN_DATA_NOT_ALLOWED(
+      Category.USER_ERROR, "0146", "Inserting already-written data is not allowed", "", ""),
 
   //
   // Errors for the concurrency error category
--- a/core/src/main/java/com/scalar/db/common/error/CoreError.java
+++ b/core/src/main/java/com/scalar/db/common/error/CoreError.java
@@ -668,6 +668,8 @@ public enum CoreError implements ScalarDbError {
       ""),
   CONSENSUS_COMMIT_INSERTING_ALREADY_WRITTEN_DATA_NOT_ALLOWED(
       Category.USER_ERROR, "0146", "Inserting already-written data is not allowed", "", ""),
+  CONSENSUS_COMMIT_DELETING_ALREADY_INSERTED_DATA_NOT_ALLOWED(
+      Category.USER_ERROR, "0147", "Deleting already-inserted data is not allowed", "", ""),
 
   //
   // Errors for the concurrency error category
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
@@ -26,7 +26,6 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.annotation.concurrent.NotThreadSafe;
@@ -101,9 +100,9 @@ public class CrudHandler {
         // transaction read it first. However, we update it only if a get operation has no
         // conjunction or the result exists. This is because we don’t know whether the record
         // actually exists or not due to the conjunction.
-        snapshot.put(key, result);
+        snapshot.putIntoReadSet(key, result);
       }
-      snapshot.put(get, result); // for re-read and validation
+      snapshot.putIntoGetSet(get, result); // for re-read and validation
       return;
     }
     throw new UncommittedRecordException(
@@ -117,7 +116,7 @@ public class CrudHandler {
       throws CrudException {
     TableMetadata metadata = getTableMetadata(key.getNamespace(), key.getTable());
     return snapshot
-        .mergeResult(key, snapshot.get(get), get.getConjunctions())
+        .getResult(key, get)
         .map(r -> new FilteredResult(r, projections, metadata, isIncludeMetadataEnabled));
   }
 
@@ -139,17 +138,12 @@ public class CrudHandler {
     List<String> originalProjections = new ArrayList<>(originalScan.getProjections());
     Scan scan = (Scan) prepareStorageSelection(originalScan);
 
-    Map<Snapshot.Key, TransactionResult> results = new LinkedHashMap<>();
-
-    Optional<Map<Snapshot.Key, TransactionResult>> resultsInSnapshot = snapshot.get(scan);
+    Optional<Map<Snapshot.Key, TransactionResult>> resultsInSnapshot = snapshot.getResults(scan);
     if (resultsInSnapshot.isPresent()) {
-      for (Entry<Snapshot.Key, TransactionResult> entry : resultsInSnapshot.get().entrySet()) {
-        snapshot
-            .mergeResult(entry.getKey(), Optional.of(entry.getValue()))
-            .ifPresent(result -> results.put(entry.getKey(), result));
-      }
-      return createScanResults(scan, originalProjections, results);
+      return createScanResults(scan, originalProjections, resultsInSnapshot.get());
     }
+
+    Map<Snapshot.Key, TransactionResult> results = new LinkedHashMap<>();
 
     Scanner scanner = null;
     try {
@@ -169,9 +163,9 @@ public class CrudHandler {
         // We always update the read set to create before image by using the latest record (result)
         // because another conflicting transaction might have updated the record after this
         // transaction read it first.
-        snapshot.put(key, Optional.of(result));
+        snapshot.putIntoReadSet(key, Optional.of(result));
 
-        snapshot.mergeResult(key, Optional.of(result)).ifPresent(value -> results.put(key, value));
+        snapshot.getResult(key).ifPresent(value -> results.put(key, value));
       }
     } finally {
       if (scanner != null) {
@@ -182,7 +176,7 @@ public class CrudHandler {
         }
       }
     }
-    snapshot.put(scan, results);
+    snapshot.putIntoScanSet(scan, results);
 
     return createScanResults(scan, originalProjections, results);
   }
@@ -213,10 +207,10 @@ public class CrudHandler {
         read(key, createGet(key));
       }
       mutationConditionsValidator.checkIfConditionIsSatisfied(
-          put, snapshot.getFromReadSet(key).orElse(null));
+          put, snapshot.getResult(key).orElse(null));
     }
 
-    snapshot.put(key, put);
+    snapshot.putIntoWriteSet(key, put);
   }
 
   public void delete(Delete delete) throws CrudException {
@@ -227,10 +221,10 @@ public class CrudHandler {
         read(key, createGet(key));
       }
       mutationConditionsValidator.checkIfConditionIsSatisfied(
-          delete, snapshot.getFromReadSet(key).orElse(null));
+          delete, snapshot.getResult(key).orElse(null));
     }
 
-    snapshot.put(key, delete);
+    snapshot.putIntoDeleteSet(key, delete);
   }
 
   public void readIfImplicitPreReadEnabled() throws CrudException {
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -9,6 +9,7 @@ import com.scalar.db.api.DistributedStorage;
 import com.scalar.db.api.Get;
 import com.scalar.db.api.Operation;
 import com.scalar.db.api.Put;
+import com.scalar.db.api.PutBuilder;
 import com.scalar.db.api.Result;
 import com.scalar.db.api.Scan;
 import com.scalar.db.api.ScanAll;
@@ -30,6 +31,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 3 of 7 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -166,7 +176,39 @@ public ReadWriteSets getReadWriteSets() {
...
+    }
+
+    Map<Key, TransactionResult> results = new LinkedHashMap<>();
+    for (Entry<Snapshot.Key, TransactionResult> entry : scanSet.get(scan).entrySet()) {
+      mergeResult(entry.getKey(), Optional.of(entry.getValue()))
+          .ifPresent(result -> results.put(entry.getKey(), result));
+    }
+
+    return Optional.of(results);
+  }
```
brfrn169, line 209: Added `getResult()` and `getResults()` methods to `Snapshot`. These methods return results that are merged with mutations in the write set, if necessary. Following this change, we can make the `mergeResult()` methods `private`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -213,10 +207,10 @@ public void put(Put put) throws CrudException {
         read(key, createGet(key));
       }
       mutationConditionsValidator.checkIfConditionIsSatisfied(
-          put, snapshot.getFromReadSet(key).orElse(null));
+          put, snapshot.getResult(key).orElse(null));
```
brfrn169, line 210: Use the added `getResult()` method to use the result that's merged with a mutation for the same record in the write set for conditional updates.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -227,10 +221,10 @@ public void delete(Delete delete) throws CrudException {
         read(key, createGet(key));
       }
       mutationConditionsValidator.checkIfConditionIsSatisfied(
-          delete, snapshot.getFromReadSet(key).orElse(null));
+          delete, snapshot.getResult(key).orElse(null));
```
brfrn169, line 224: Ditto. Use the added `getResult()` method to use the result that's merged with a mutation for the same record in the write set for conditional deletes.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -113,47 +115,55 @@ Isolation getIsolation() {
...
+  public void putIntoWriteSet(Key key, Put put) {
     if (deleteSet.containsKey(key)) {
       throw new IllegalArgumentException(
           CoreError.CONSENSUS_COMMIT_WRITING_ALREADY_DELETED_DATA_NOT_ALLOWED.buildMessage());
     }
     if (writeSet.containsKey(key)) {
+      if (put.isInsertModeEnabled()) {
+        throw new IllegalArgumentException(
+            CoreError.CONSENSUS_COMMIT_INSERTING_ALREADY_WRITTEN_DATA_NOT_ALLOWED.buildMessage());
+      }
```
brfrn169, line 141: Add validation to ensure that Insert is not executed on an already written record.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -113,47 +115,55 @@ Isolation getIsolation() {
...
+
+      // If the implicit pre-read is enabled for the new put, it should also be enabled for the
+      // merged put. However, if the previous put is in insert mode, this doesn’t apply. This is
+      // because, in insert mode, the read set is not used during the preparation phase. Therefore,
+      // we only need to enable the implicit pre-read if the previous put is not in insert mode
+      if (put.isImplicitPreReadEnabled() && !originalPut.isInsertModeEnabled()) {
+        putBuilder.enableImplicitPreRead();
+      }
+
+      writeSet.put(key, putBuilder.build());
```
brfrn169, line 156: Updated the merge logic for the write set. We should also merge the implicit pre-read, with one exception. Please see the comments for details.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -666,6 +666,8 @@ public enum CoreError implements ScalarDbError {
           + "Primary-key columns must not contain any of the following characters in Cosmos DB: ':', '/', '\\', '#', '?'. Value: %s",
       "",
       ""),
+  CONSENSUS_COMMIT_INSERTING_ALREADY_WRITTEN_DATA_NOT_ALLOWED(
+      Category.USER_ERROR, "0146", "Inserting already-written data is not allowed", "", ""),
```
brfrn169, line 670: @josh-wong Could you please review this message when you have time? 🙇

komamitsu:
> Insert then Upsert: merge Insert and Upsert (with implicit pre-read disabled), then execute.
> Insert then Update: merge Insert and Update (with implicit pre-read disabled), then execute.
> Insert then Delete: only execute Delete.

@brfrn169 I assume these 3 cases should fail if a record with the same key already exists, since the first Inserts should fail from semantic perspective. Is my understanding correct?

brfrn169:
> > Insert then Upsert: merge Insert and Upsert (with implicit pre-read disabled), then execute.
> > Insert then Update: merge Insert and Update (with implicit pre-read disabled), then execute.
> > Insert then Delete: only execute Delete.
> 
> @brfrn169 I assume these 3 cases should fail if a record with the same key already exists, since the first Inserts should fail from semantic perspective. Is my understanding correct?

@komamitsu Yes, that's correct. 

Strictly speaking, in Consensus Commit, when inserting a record, we don’t check if the target record exists or not. As a result, users encounter a conflict error only when committing the transaction, not when inserting the record.

komamitsu:
@brfrn169 Thanks for the details!

I executed this code with the implementation of this PR, but it didn't fail. I expected this failed since the INSERT conflicted with the existing record... What do you think?

```java
    TransactionFactory factory = TransactionFactory.create("/tmp/local-pg.properties");
    DistributedTransactionManager transactionManager = factory.getTransactionManager();
    try {
      DistributedTransaction tx = transactionManager.begin();
      System.out.println(">>>> " + tx.get(
          Get.newBuilder()
              .namespace("ycsb").table("usertable")
              .partitionKey(Key.ofInt("ycsb_key", 1))
              .build()));          // >>>> Optional[FilteredResult{payload=null, ycsb_key=1}]
      tx.insert(
          Insert.newBuilder()
              .namespace("ycsb").table("usertable")
              .partitionKey(Key.ofInt("ycsb_key", 1))
              .build());
      tx.delete(
          Delete.newBuilder()
              .namespace("ycsb").table("usertable")
              .partitionKey(Key.ofInt("ycsb_key", 1))
              .build());
      tx.commit();
    }
    finally{
      transactionManager.close();
    }
```

brfrn169:
@komamitsu One possible idea is to disallow deleting a record that has been inserted in the same transaction by adding validation. What do you think?

komamitsu:
@brfrn169 Thanks for the explanation. It matches my understanding from the source code.

> One possible idea is to disallow deleting a record that has been inserted in the same transaction by adding validation

This seems better to me in terms of consistency. But it might be a bit too strict in terms of convenience? I want to hear opinions from @feeblefakie as well. 


BTW, I also tried Upsert and Update after Insert as follows.

```java
    TransactionFactory factory = TransactionFactory.create("/Users/mitsunorikomatsu/tmp/local-pg.properties");
    DistributedTransactionManager transactionManager = factory.getTransactionManager();
    try {
      DistributedTransaction tx = transactionManager.begin();
      System.out.println(">>>> " + tx.get(
          Get.newBuilder()
              .namespace("ycsb").table("usertable")
              .partitionKey(Key.ofInt("ycsb_key", 1))
              .build()));
      tx.insert(
          Insert.newBuilder()
              .namespace("ycsb").table("usertable")
              .partitionKey(Key.ofInt("ycsb_key", 1))
              .build());
      tx.upsert(
          Upsert.newBuilder()
              .namespace("ycsb").table("usertable")
              .partitionKey(Key.ofInt("ycsb_key", 1))
              .value(TextColumn.of("payload", "hello"))
              .build());
      tx.commit();
    }
    finally{
      transactionManager.close();
    }
```

But it failed and the error message "The record being prepared already exists" seemed a bit odd to me. It also happens with the current `master` branch, though.

```
Exception in thread "main" com.scalar.db.exception.transaction.CommitConflictException: CORE-20013: The record being prepared already exists. Transaction ID: GMiUdvxZCcPyxLLaJt5yKnfR$70452dbf-7653-46d7-af8c-85b4f3337d65
	at com.scalar.db.transaction.consensuscommit.CommitHandler.commit(CommitHandler.java:124)
	at com.scalar.db.transaction.consensuscommit.ConsensusCommit.commit(ConsensusCommit.java:227)
	at com.scalar.db.common.DecoratedDistributedTransaction.commit(DecoratedDistributedTransaction.java:129)
	at com.scalar.db.common.TransactionDecorationDistributedTransactionManager$StateManagedTransaction.commit(TransactionDecorationDistributedTransactionManager.java:147)
	at com.scalar.db.common.DecoratedDistributedTransaction.commit(DecoratedDistributedTransaction.java:129)
	at com.scalar.db.common.ActiveTransactionManagedDistributedTransactionManager$ActiveTransaction.commit(ActiveTransactionManagedDistributedTransactionManager.java:168)
	at com.scalar.db.transaction.consensuscommit.CommitHandler.main(CommitHandler.java:361)
Caused by: com.scalar.db.exception.transaction.PreparationConflictException: CORE-20013: The record being prepared already exists. Transaction ID: GMiUdvxZCcPyxLLaJt5yKnfR$70452dbf-7653-46d7-af8c-85b4f3337d65
	at com.scalar.db.transaction.consensuscommit.CommitHandler.prepare(CommitHandler.java:184)
	at com.scalar.db.transaction.consensuscommit.CommitHandler.commit(CommitHandler.java:119)
	... 6 more
Caused by: com.scalar.db.exception.storage.NoMutationException: CORE-20000: No mutation was applied
	at com.scalar.db.storage.jdbc.JdbcDatabase.put(JdbcDatabase.java:115)
	at com.scalar.db.storage.multistorage.MultiStorage.put(MultiStorage.java:105)
	at com.scalar.db.storage.multistorage.MultiStorage.mutate(MultiStorage.java:130)
	at com.scalar.db.transaction.consensuscommit.CommitHandler.lambda$prepareRecords$0(CommitHandler.java:206)
	at com.scalar.db.transaction.consensuscommit.ParallelExecutor.lambda$executeTasksInParallel$0(ParallelExecutor.java:181)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
Caused by: com.scalar.db.exception.transaction.PreparationConflictException: CORE-20013: The record being prepared already exists. Transaction ID: GMiUdvxZCcPyxLLaJt5yKnfR$70452dbf-7653-46d7-af8c-85b4f3337d65

	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
Caused by: com.scalar.db.exception.storage.NoMutationException: CORE-20000: No mutation was applied

	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
```

brfrn169:
> But it failed and the error message "The record being prepared already exists" seemed a bit odd to me. It also happens with the current master branch, though.

@komamitsu I think the error indicates that you tried to insert a record that already exists. Maybe even if you remove the Upsert from the code, you still encounter the same error, right?

komamitsu:
@brfrn169 Oh, sorry, I forgot to call `rollback()` as error handling. Probably that's the cause 🙇 Please ignore the comment about Upsert and Update.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/common/error/CoreError.java
josh-wong: LGTM!👍

josh-wong (approved):
LGTM! Thank you!🙇🏻‍♂️

feeblefakie (approved):
LGTM! Thank you!

brfrn169:
@feeblefakie @komamitsu As discussed, we will add additional validation to ensure that Delete is not executed on a record that has already been inserted in the same transaction. Thanks.

komamitsu:
@brfrn169 Thanks for letting me that!

Similar validation will be introduced for the following cases?
> Insert then Upsert: merge Insert and Upsert (with implicit pre-read disabled), then execute.
> Insert then Update: merge Insert and Update (with implicit pre-read disabled), then execute.

brfrn169:
> Similar validation will be introduced for the following cases?
> 
> > Insert then Upsert: merge Insert and Upsert (with implicit pre-read disabled), then execute.
> > Insert then Update: merge Insert and Update (with implicit pre-read disabled), then execute.

@komamitsu In those cases, the transaction will correctly fail with the current logic, so I don’t think we need to add validation for them. Thanks.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -666,6 +666,10 @@ public enum CoreError implements ScalarDbError {
...
+  CONSENSUS_COMMIT_DELETING_ALREADY_INSERTED_DATA_NOT_ALLOWED(
+      Category.USER_ERROR, "0147", "Deleting already-inserted data is not allowed", "", ""),
```
brfrn169, line 672: @josh-wong Sorry, could you please review this message as well? 🙇

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -160,7 +160,16 @@ public void putIntoWriteSet(Key key, Put put) {
...
-    writeSet.remove(key);
+    Put put = writeSet.get(key);
+    if (put != null) {
+      if (put.isInsertModeEnabled()) {
+        throw new IllegalArgumentException(
+            CoreError.CONSENSUS_COMMIT_DELETING_ALREADY_INSERTED_DATA_NOT_ALLOWED.buildMessage());
+      }
+
+      writeSet.remove(key);
+    }
```
brfrn169, line 171: Added validation to ensure that Delete is not executed on a record that has already been inserted in the same transaction.

core/src/main/java/com/scalar/db/common/error/CoreError.java
komamitsu: @brfrn169 I ran a small code to reproduce this error. I noticed the namespace, table and key of the target record weren't shown in any error message. How about adding those information into this error message? I think this is a nice-to-have improvement and won't block this PR from being approved, though.

komamitsu (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/common/error/CoreError.java
josh-wong: @brfrn169 

> @josh-wong Sorry, could you please review this message as well? 🙇

LGTM!👍

feeblefakie (approved):
LGTM! Thank you!

PR #2390, opened 2024-12-05
Refactoring the callbacks for the group commit error handling in `CommitHandler`

komamitsu:
## Description

In the current implementation, `CommitHandler` has `onPrepareFailure()` and `onValidateFailure()`, but there is no need to separate them. This PR merges them to make it simpler.

## Related issues and/or PRs

None.

## Changes made

- Merge `onPrepareFailure()` and `onValidateFailure()` into `onFailureBeforeCommit()`
- Make the error handling more consistent

## Checklist

None

## Release notes

N/A (No behavior changed)

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -63,9 +63,9 @@ public class CommitHandler {
       return Optional.of(
           beforePreparationSnapshotHook.handle(tableMetadataManager, snapshot.getReadWriteSets()));
     } catch (Exception e) {
+      onFailureBeforeCommit(snapshot);
       abortState(snapshot.getId());
       rollbackRecords(snapshot);
-      onFailureBeforeCommit(snapshot);
       throw new CommitException(
           CoreError.HANDLING_BEFORE_PREPARATION_SNAPSHOT_HOOK_FAILED.buildMessage(e.getMessage()),
           e,
@@ -83,9 +83,9 @@ public class CommitHandler {
     try {
       snapshotHookFuture.get();
     } catch (Exception e) {
+      onFailureBeforeCommit(snapshot);
       abortState(snapshot.getId());
       rollbackRecords(snapshot);
-      onFailureBeforeCommit(snapshot);
       throw new CommitException(
           CoreError.HANDLING_BEFORE_PREPARATION_SNAPSHOT_HOOK_FAILED.buildMessage(e.getMessage()),
           e,
@@ -98,6 +98,7 @@ public class CommitHandler {
     try {
       prepare(snapshot);
     } catch (PreparationException e) {
+      onFailureBeforeCommit(snapshot);
       abortState(snapshot.getId());
       rollbackRecords(snapshot);
       if (e instanceof PreparationConflictException) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -65,11 +69,9 @@ private Optional<Future<Void>> invokeBeforePreparationSnapshotHook(Snapshot snap
       return Optional.of(
           beforePreparationSnapshotHook.handle(tableMetadataManager, snapshot.getReadWriteSets()));
     } catch (Exception e) {
+      onFailureBeforeCommit(snapshot);
```
komamitsu, line 72: Calling this callback first is safe since it doesn't throw any exception.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -104,28 +104,30 @@ public void commit(Snapshot snapshot) throws CommitException, UnknownTransaction
     try {
       prepare(snapshot);
     } catch (PreparationException e) {
+      onFailureBeforeCommit(snapshot);
```
komamitsu, line 107: Actually, the following `abortState()` internally calls a method equivalent to `onFailureBeforeCommit()` in case of `CommitHandlerWithGroupCommit`. So, this change doesn't change the behavior at all. But, I think the callback should be called here for consistent error handling.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169, line 72: For safety, I think we can catch the exception as follow:
```suggestion
      try {
        onFailureBeforeCommit(snapshot);
      } catch (Exception ex) {
        ...
      }
```

What do you think?

komamitsu: Yeah, you have a point. On the other hand, I think all the codes calling the callback need the same just-in-case error handling in terms of consistency. So, I wrote this comment https://github.com/scalar-labs/scalardb/pull/2390/files#diff-80b71c10ff00c8a739d8aec837a1730b5470df99295bebb354033c9aefc33bbaR55-R56 so that we don't need such a just-in-case error handling. But if you think the comment isn't enough, I'm okay to add the error handling to all the codes calling the callback. Either is fine with me 👍

komamitsu: @brfrn169 On second thought, I'll follow your suggestion since trusting a callback injected from outside is a bit too optimistic and the comment doesn't guarantee a callback won't throw an exception. Let me update the PR.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandlerWithGroupCommit.java
```diff
@@ -77,7 +72,12 @@ private void commitStateViaGroupCommit(Snapshot snapshot)
   }
 
   private void cancelGroupCommitIfNeeded(String id) {
-    groupCommitter.remove(id);
+    try {
+      groupCommitter.remove(id);
+    } catch (Exception e) {
+      logger.warn(
+          "Unexpectedly failed to remove the snapshot ID from the group committer. ID: {}", id);
```
brfrn169, line 79: Small nit:
```suggestion
      logger.warn(
          "Unexpectedly failed to remove the snapshot ID from the group committer. ID: {}", id, e);
```

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #2391, opened 2024-12-06
Add MariaDB Connector/J JDBC driver

Torch3333:
## Description

So far, ScalarDB has been using the Mysql JDBC driver for both MySQL and MariaDB, but the driver does not support fractional seconds for time-related types when used with MariaDB 10.11. 
The reason is that the MariaDB server returns a version number that does not follow the MySQL version standard and is incorrectly parsed by the MySQL driver which causes it to disable fractional seconds support. 
This issue does not occur with MariaDB 11.4 though.

To address the issue, the dedicated [MariaDB Connector/J](https://mariadb.com/kb/en/about-mariadb-connector-j/) JDBC driver will be used for any connection URL starting with `jdbc:mariadb` 

## Related issues and/or PRs

N/A

## Changes made

Add the MariaDB JDBC driver.

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes.
- [ ] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [ ] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [ ] Any dependent changes in other PRs have been merged and published.

## Release notes

Use the MariaDB Connector/J JDBC driver for any connection URL starting with `jdbc:mariadb`

PR #2420, opened 2024-12-19
Fix Slf4j providers not found error

Torch3333:
## Description

~~This updates slf4j to the latest version to resolve the following logger binding issue when running ScalarDB unit or integration test.~~
~~This forces Gradle to use the slf4j version defined by ScalarDB over any version pulled transitively.~~
This excludes the `slf4j-api` dependency from the maria-db driver to resolve the following logger binding issue when running ScalarDB unit or integration test.

```
SLF4J: No SLF4J providers were found.
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
SLF4J: Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
SLF4J: Ignoring binding found at [jar:file:/Users/vguilpain/.gradle/caches/modules-2/files-2.1/org.slf4j/slf4j-simple/1.7.36/a41f9cfe6faafb2eb83a1c7dd2d0dfd844e2a936/slf4j-simple-1.7.36.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See https://www.slf4j.org/codes.html#ignoredBindings for an explanation.
```

This problem appeared after adding the MariaDB driver dependency in https://github.com/scalar-labs/scalardb/pull/2391. This dependency pulls `slf4j-api` version 2.X, but `sl4j-simple` stays with version 1.7.X, causing a mismatch.

## Related issues and/or PRs

N/A

## Changes made

Update slf4j to the latest version.

## Checklist

N/A

## Release note

N/A

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ subprojects {
     ext {
         guiceVersion = '5.1.0'
         guavaVersion = '32.1.3-jre'
-        slf4jVersion = '1.7.36'
+        slf4jVersion = '2.0.16'
         cassandraDriverVersion = '3.11.5'
         azureCosmosVersion = '4.65.0'
         jooqVersion = '3.14.16'
--- a/build.gradle
+++ b/build.gradle
@@ -89,9 +89,5 @@ subprojects {
     installDist {
         duplicatesStrategy DuplicatesStrategy.EXCLUDE
     }
-
-    configurations.all {
-       resolutionStrategy.force "org.slf4j:slf4j-api:${slf4jVersion}", "org.slf4j:slf4j-simple:${slf4jVersion}"
-    }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ subprojects {
     ext {
         guiceVersion = '5.1.0'
         guavaVersion = '32.1.3-jre'
-        slf4jVersion = '2.0.16'
+        slf4jVersion = '1.7.36'
         cassandraDriverVersion = '3.11.5'
         azureCosmosVersion = '4.65.0'
         jooqVersion = '3.14.16'
@@ -88,6 +88,10 @@ subprojects {
 
     installDist {
         duplicatesStrategy DuplicatesStrategy.EXCLUDE
+    }
+
+    configurations.all {
+       resolutionStrategy.force "org.slf4j:slf4j-api:${slf4jVersion}", "org.slf4j:slf4j-simple:${slf4jVersion}"
     }
 }
 
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -107,7 +107,9 @@ dependencies {
     implementation "com.microsoft.sqlserver:mssql-jdbc:${sqlserverDriverVersion}"
     implementation "org.xerial:sqlite-jdbc:${sqliteDriverVersion}"
     implementation "com.yugabyte:jdbc-yugabytedb:${yugabyteDriverVersion}"
-    implementation "org.mariadb.jdbc:mariadb-java-client:${mariadDbDriverVersion}"
+    implementation ("org.mariadb.jdbc:mariadb-java-client:${mariadDbDriverVersion}") {
+        exclude group: 'org.slf4j', module: 'slf4j-api'
+    }
     implementation "org.apache.commons:commons-text:${commonsTextVersion}"
     testImplementation "org.junit.jupiter:junit-jupiter-api:${junitVersion}"
     testImplementation "org.junit.jupiter:junit-jupiter-params:${junitVersion}"


komamitsu:
@Torch3333 Can you tell me how to reproduce the warning message?

Torch3333:
@komamitsu 
Running any integration or unit test produces this error message on my IDE. 
image

komamitsu (approved):
LGTM! 👍

brfrn169 (approved):
LGTM! Thank you!

brfrn169:
@Torch3333 Ah, this change might break backward compatibility.

For example, if a user is using log4j with slf4j `1.x`, they need to include `log4j-slf4j-impl` as follows:
```gradle
    implementation "org.slf4j:slf4j-api:1.7.36"
    implementation "org.apache.logging.log4j:log4j-core:2.24.3"
    implementation "org.apache.logging.log4j:log4j-slf4j-impl:2.24.3"
```

However, if a user is using log4j with slf4j `2.x`, they need to include `log4j-slf4j2-impl` as follows:
```gradle
    implementation "org.slf4j:slf4j-api:2.0.16"
    implementation "org.apache.logging.log4j:log4j-core:2.24.3"
    implementation "org.apache.logging.log4j:log4j-slf4j2-impl:2.24.3"
```

Therefore, if we upgrade slf4j to `2.x`, users will need to change their log4j module to `log4j-slf4j2-impl`. What do you think?

komamitsu:
IIUC, the new way makes the library (`org.mariadb.jdbc:mariadb-java-client` -> `com.github.waffle:waffle-jna`) introduced `org.slf4j:slf4j-api:2.0.7` use `1.7.36`? It might cause another problem although it depends use cases.

```
|    +--- org.mariadb.jdbc:mariadb-java-client:3.5.1
|    |    \--- com.github.waffle:waffle-jna:3.3.0
|    |         +--- net.java.dev.jna:jna:5.13.0
|    |         +--- net.java.dev.jna:jna-platform:5.13.0
|    |         |    \--- net.java.dev.jna:jna:5.13.0
|    |         +--- org.slf4j:jcl-over-slf4j:2.0.7
|    |         |    \--- org.slf4j:slf4j-api:2.0.7 -> 1.7.36
|    |         +--- org.slf4j:slf4j-api:2.0.7 -> 1.7.36
```

I'm not sure if we need to continue using 1.x forever. I think our customers and we need to upgrade the dependency someday...

brfrn169:
> I'm not sure if we need to continue using 1.x forever. I think our customers and we need to upgrade the dependency someday...

@komamitsu I think we can upgrade in the next major version (`4.0.0`). Or do you think it should be done in the next minor version instead?

build.gradle
```diff
@@ -89,5 +89,9 @@ subprojects {
     installDist {
         duplicatesStrategy DuplicatesStrategy.EXCLUDE
     }
+
+    configurations.all {
+       resolutionStrategy.force "org.slf4j:slf4j-api:${slf4jVersion}", "org.slf4j:slf4j-simple:${slf4jVersion}"
```
brfrn169, line 94: Instead of this, how about excluding the `slf4j` dependency from the `mariadb` dependency?

komamitsu:
@brfrn169 

>  I think we can upgrade in the next major version (4.0.0)

Sounds okay!

```
|    +--- org.mariadb.jdbc:mariadb-java-client:3.5.1
|    |    \--- com.github.waffle:waffle-jna:3.3.0
|    |         +--- net.java.dev.jna:jna:5.13.0
|    |         +--- net.java.dev.jna:jna-platform:5.13.0
|    |         |    \--- net.java.dev.jna:jna:5.13.0
|    |         +--- org.slf4j:jcl-over-slf4j:2.0.7
|    |         |    \--- org.slf4j:slf4j-api:2.0.7 -> 1.7.36
|    |         +--- org.slf4j:slf4j-api:2.0.7 -> 1.7.36
```

(cc: @Torch3333)

BTW, I checked the source code of `com.github.waffle:waffle-jna` and confirmed it doesn't use SLF4J's Fluent API introduced in v2 (See https://www.slf4j.org/faq.html#compatibility). And the library is for Windows auth. So, it should be completely safe to keep using `org.slf4j:slf4j-api:1.7.36`.

build.gradle
komamitsu: Agreed. I think excluding it is a bit more polite way.

Torch3333: Yes, we can I.
I thought that using the `resolutionStrategy.force` was future proof in case another library we update pulls `sl4j` v2.

Torch3333:
> @brfrn169
> 
> > I think we can upgrade in the next major version (4.0.0)
> 
> Sounds okay!
> 
> ```
> |    +--- org.mariadb.jdbc:mariadb-java-client:3.5.1
> |    |    \--- com.github.waffle:waffle-jna:3.3.0
> |    |         +--- net.java.dev.jna:jna:5.13.0
> |    |         +--- net.java.dev.jna:jna-platform:5.13.0
> |    |         |    \--- net.java.dev.jna:jna:5.13.0
> |    |         +--- org.slf4j:jcl-over-slf4j:2.0.7
> |    |         |    \--- org.slf4j:slf4j-api:2.0.7 -> 1.7.36
> |    |         +--- org.slf4j:slf4j-api:2.0.7 -> 1.7.36
> ```
> 
> (cc: @Torch3333)
> 
> BTW, I checked the source code of `com.github.waffle:waffle-jna` and confirmed it doesn't use SLF4J's Fluent API introduced in v2 (See https://www.slf4j.org/faq.html#compatibility). And the library is for Windows auth. So, it should be completely safe to keep using `org.slf4j:slf4j-api:1.7.36`.

Ok, thank you for taking the time to check!

build.gradle
Torch3333: The Gradle doc explains that using the `resolutionStrategy.force` is a project-global excluding rule.

```
Excluding a particular transitive dependency does not guarantee that it does not show up in the dependencies of a given configuration. For example, some other dependency, which does not have any exclude rules, might pull in exactly the same transitive dependency. To guarantee that the transitive dependency is excluded from the entire configuration please use per-configuration exclude rules: [getExcludeRules](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.artifacts/-configuration/get-exclude-rules.html). In fact, in a majority of cases the actual intention of configuring per-dependency exclusions is really excluding a dependency from the entire configuration (or classpath).

If your intention is to exclude a particular transitive dependency because you don't like the version it pulls in to the configuration then consider using forced versions' feature: [force](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.artifacts/-resolution-strategy/force.html).
```
cf. https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.artifacts/-module-dependency/exclude.html

Could you share your concerns concerning the `resolutionStrategy.force` ?
I am ok but I just want to better understand your point of view.

komamitsu: The 2 ways have similar effects, and actually either is fine with me. IIUC, the slight difference is basically only global-scope or not. So, this might be also a matter of taste thing. I relatively prefer starting with small-scope effect, but I don't insist on it as I understand the pros of `resolutionStrategy.force`.

komamitsu (approved):
LGTM! 👍

build.gradle
brfrn169: I was concerned about the side effects of `resolutionStrategy.force` since it’s a project-global configuration. Also, I believe we’ve usually used the `exclude` configuration for similar issues in the past. Do you think `resolutionStrategy.force` is a better approach than the `exclude` configuration?

Torch3333: My understanding was that `resolutionStrategy.force` could be used as project-wide `exclude` which means:
1.  The forced dependency version was used project-wide when building ScalarDB
2. Other projects using ScalarDB as a dependency would also be presented with the forced dependency version, i.e. excluding the non-wanted version

But after testing locally, statement 2. is wrong. `slf4j-api:2.0.7` was still listed as a transitive dependency of  `com.github.waffle:waffle-jna`.

In comparison, when using the `exclude` syntax. `slf4j-api:2.0.7` was not listed as a dependency or pulled by `com.github.waffle:waffle-jna`, which is what we want to avoid the problem described by Toshi in https://github.com/scalar-labs/scalardb/pull/2420#issuecomment-2552885802

So, I will revise the PR to use the exclude syntax.

feeblefakie (approved):
LGTM! Thank you!

PR #789, opened 2023-01-10
Update slides links

brfrn169:
This PR updates the slides links to the Speaker Deck ones. Please take a look!

PR #456, opened 2022-01-09
Rename JdbcDatabaseAdmin to JdbcAdmin

brfrn169:
This PR renames `JdbcDatabaseAdmin` to `JdbcAdmin` because I think `JdbcAdmin` is a more proper name. Please take a look!

PR #2437, opened 2024-12-23
Add time related types crud

Torch3333:
## Description

> Provide a brief description about **why** this PR is necessary. Be sure to provide context.

## Related issues and/or PRs

> If this PR addresses or references any issues and/or other PRs, list them here. For example, "Fixes #123", "Related to #456", or "Addresses #789".

## Changes made

> Outline the specific changes made in this pull request. Include relevant details, such as added features, bug fixes, code refactoring, or improvements.

## Checklist

> Provide any additional information or notes that may be relevant to the reviewers or stakeholders.

## Release notes

> Provide release note text for this PR based on the guidelines at [Guidelines for Writing Release Notes](https://developers.scalar-labs.com/docs/style-guide/release-notes/).
> 
> For example:
> 
> Added validation for the Consensus Commit mutation operation.


integration-test/src/main/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
```diff
@@ -93,10 +111,32 @@ protected String getNamespaceBaseName() {
   }
 
   private void createTables() throws ExecutionException {
+    TableMetadata.Builder tableMetadata =
```
Torch3333, line 114: Up until now `DistributedTransactionIntegrationTestBase` was performing tests on a table composed of columns of INT type only.
I updated only the test that targeted the most generic use case for each operation to target all data types.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -131,4 +143,56 @@ default String getPattern(LikeExpression likeExpression) {
   default @Nullable String getSchemaName(String namespace) {
     return namespace;
   }
+
+  default Object encodeDate(DateColumn column) {
+    assert column.getDateValue() != null;
+    return column.getDateValue();
+  }
+
+  default Object encodeTime(TimeColumn column) {
```
komamitsu, line 152: Using Object type is hard for me to confirm if the implementation returns correct types. How about using generics like this?

```java
public interface RdbEngineStrategy<T_DATE, T_TIME, T_TIMESTAMP, T_TIMESTAMP_TZ> {

  ...

  T_DATE encodeDate(DateColumn column);
  T_TIME encodeTime(TimeColumn column);
  T_TIMESTAMP encodeTimestamp(TimestampColumn column);
  T_TIMESTAMP_TZ encodeTimestampTZ(TimestampTZColumn column);
}

class RdbEngineSqlServer implements RdbEngineStrategy<String, LocalTime, String, DateTimeOffset> {
  ...

  @Override
  public LocalTime encodeTime(TimeColumn column) {
    return column.getTimeValue();
  }

  @Override
  public String encodeDate(DateColumn column) {
    return column.getDateValue().format(DateTimeFormatter.BASIC_ISO_DATE);
  }

  ...
}
```

With this we can detect wrong type usage at compile time.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -206,10 +211,10 @@ public String getDataTypeForEngine(DataType scalarDbDataType) {
       case DATE:
         return "DATE";
       case TIME:
-        return "TIME";
+        return "TIME(6)";
```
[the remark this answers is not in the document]
Torch3333: In this notation `TIME(N)`, N is the number of digits for the fractional seconds. Since the ScalarDB TIME type has a precision of 1 microsecond, there is a 6 digits precision. 
Meanwhile, the TIMESTAMP and TIMESTAMPTZ have a precision of 1 millisecond, so they have 3 digits for the fractional seconds.
The precision for these 3 types is the best common precision for all storage.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSqlServer.java
```diff
@@ -336,4 +352,35 @@ public String getEscape(LikeExpression likeExpression) {
...
+    assert column.getTimestampValue() != null;
+    // Pass the timestamp value as text otherwise the dates before the Julian to Gregorian Calendar
+    // transition (October 15, 1582) will be offset by 10 days.
+    return column.getTimestampValue().format(DateTimeFormatter.ISO_DATE_TIME);
+  }
+
+  @Override
+  public DateTimeOffset encodeTimestampTZ(TimestampTZColumn column) {
+    assert column.getTimestampTZValue() != null;
+    // When using SQLServer DATETIMEOFFSET data type, we should use the SQLServer JDBC driver's
```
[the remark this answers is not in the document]
Torch3333: Yes, for encoding data, only using the SQLServer driver `microsoft.sql.DateTimeOffset` type worked for the range of data I tested. 
Decoding through a Java `java.time.OffsetDateTime` object worked,  if my memory is correct, `microsoft.sql.DateTimeOffset` was ok too.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
Torch3333: I like the idea of enforcing type but then the default implementation for the method in the interface cannot added.
So, this means we would need to have these methods in each `RdbEngine` even when the default behavior does not need to be overridden.  
```
  @Override
  public LocalDate encodeDate(DateColumn column) {
    assert column.getDateValue() != null;
    return column.getDateValue();
  }

  @Override
  public LocalTime encodeTime(TimeColumn column) {
    assert column.getTimeValue() != null;
    return column.getTimeValue();
  }

  @Override
  public LocalDateTime encodeTimestamp(TimestampColumn column) {
    assert column.getTimestampValue() != null;
    return column.getTimestampValue();
  }

  @Override
  public OffsetDateTime encodeTimestampTZ(TimestampTZColumn column) {
    assert column.getTimestampTZValue() != null;
    return column.getTimestampTZValue().atOffset(ZoneOffset.UTC);
  }
```
We gain type safety but we lose on the readability and increase code duplication. Do you think that the pros outweigh the cons ?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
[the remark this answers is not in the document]
komamitsu: I see. I had the wrong idea the ScalarDB TIME type has a precision of 1 millisecond. Thanks for the explanation!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
komamitsu: As you said, it's a trade-off. So, this might be a matter of taste, but I would take the type safety. Also, I think the code readability with the generics version is better for me. The code duplication is actually a drawback, but the logic and implementation are simple and I don't think it's a significant issue.

But, yeah it's a trade-off choice. I don't insist on it, and it's up to you.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
```diff
@@ -61,12 +68,20 @@ public static BasicDataSource initDataSource(
     dataSource.setMaxTotal(config.getConnectionPoolMaxTotal());
     dataSource.setPoolPreparedStatements(config.isPreparedStatementsPoolEnabled());
     dataSource.setMaxOpenPreparedStatements(config.getPreparedStatementsPoolMaxOpen());
+    // Add connection properties if it doesn't already exist in the connection string to not
+    // override the user settings
+    for (Entry<String, String> entry : rdbEngine.getConnectionProperties().entrySet()) {
+      if (!dataSource.getUrl().contains(entry.getKey())) {
```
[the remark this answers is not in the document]
Torch3333: I wanted to care for the rare case of the user wanting to override a driver property set by ScalarDB. This should not be necessary for the SQLServer `sendTimeAsDatetime`, but might be in the future.
But we can revise this method implementation later on if needed so I will go with your suggestion.
Thank you.

core/src/main/java/com/scalar/db/io/DateColumn.java
```diff
@@ -0,0 +1,138 @@
...
+          CoreError.OUT_OF_RANGE_COLUMN_VALUE_FOR_DATE.buildMessage(value));
+    }
+
+    this.name = Objects.requireNonNull(name);
+    this.value = value;
+  }
+
+  @Override
+  protected final void finalize() throws Throwable {
+    // TODO delete this method once https://github.com/scalar-labs/scalardb/pull/2421 is merge
```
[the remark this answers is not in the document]
Torch3333: Thanks, I will do that in a separate PR because rebasing the feature branch `[add_time_related_types](https://github.com/scalar-labs/scalardb/tree/add_time_related_types)` now would mess up this PR I think 🤔

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcDatabaseSinglePartitionKeyIntegrationTest.java
```diff
@@ -57,10 +60,21 @@ protected Column<?> getColumnWithMaxValue(String columnName, DataType dataType)
...
-  protected boolean isFloatTypeKeySupported() {
-    if (JdbcTestUtils.isYugabyte(rdbEngine)) {
-      return false;
-    }
-    return super.isFloatTypeKeySupported();
+  protected Set<DataType> getPartitionKeyTypes() {
+    return super.getPartitionKeyTypes().stream()
+        .filter(
+            type -> {
+              if (JdbcTestUtils.isOracle(rdbEngine) && type == DataType.TIMESTAMPTZ) {
```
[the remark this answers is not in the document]
Torch3333: You are right, I created a util method. Thank you.
https://github.com/scalar-labs/scalardb/pull/2437/commits/adcbfde4230ff71100c3fd6b83af07664149938f

core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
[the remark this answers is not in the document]
Torch3333: Updated it https://github.com/scalar-labs/scalardb/pull/2437/commits/bc66fee3b2da7db54ef00830eede7b4dd19f98e7

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
Torch3333: Alright, I updated it in https://github.com/scalar-labs/scalardb/pull/2437/commits/be71af53c8fea29cee89a6a177b67cb763be6377

komamitsu (approved):
LGTM! 👍

Great work!

brfrn169 (commented):
Overall LGTM. Left several comments. Please take a look when you have time!

integration-test/src/main/java/com/scalar/db/api/DistributedStorageSingleClusteringKeyScanIntegrationTestBase.java
```diff
@@ -79,8 +79,8 @@ protected String getNamespace() {
     return NAMESPACE;
   }
 
-  protected Set<DataType> getClusteringKeyTypes() {
-    return new HashSet<>(Arrays.asList(DataType.valuesWithoutTimesRelatedTypes()));
+  protected List<DataType> getClusteringKeyTypes() {
```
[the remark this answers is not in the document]
Torch3333: For two reasons: 
1. To uniformize the similar types of methods returning the data types in the storage integration test. Some class were using Set, others were List. 
2. For practical reasons since I added this method https://github.com/scalar-labs/scalardb/blob/dce86f75aa93d94a0134b68394ee6af671aba36c/core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcTestUtils.java#L76-L102 
that is used like this below to filter out non-testable data types on JDBC integration test and not requiring the method to handle a set or a list was simpler.
https://github.com/scalar-labs/scalardb/blob/dce86f75aa93d94a0134b68394ee6af671aba36c/core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcDatabaseMultiplePartitionKeyIntegrationTest.java#L80-L91

core/src/main/java/com/scalar/db/storage/TimeRelatedColumnEncodingUtils.java
```diff
@@ -0,0 +1,100 @@
...
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneOffset;
+
+/**
+ * This class provides utility methods for encoding and decoding time related column value for
+ * DynamoDB, CosmosDB and SQLite
+ */
+public final class TimeRelatedColumnEncodingUtils {
```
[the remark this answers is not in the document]
Torch3333: Yes, I revised it in https://github.com/scalar-labs/scalardb/pull/2437/commits/edb1814c4cdada34a6047d14691023f3470c8c24
Thank you.

core/src/main/java/com/scalar/db/io/Key.java
```diff
@@ -168,6 +172,58 @@ public Key(String columnName, ByteBuffer blobValue) {
...
+   *
+   * @param columnName a column name
+   * @param timestampTZValue a TIMESTAMPTZ value of the column as Instant type
+   * @deprecated As of release 3.6.0. Will be removed in release 5.0.0. Use {@link
+   *     #ofTimestampTZ(String, Instant)} instead
+   */
+  @Deprecated
+  public Key(String columnName, Instant timestampTZValue) {
+    columns = Collections.singletonList(TimestampTZColumn.of(columnName, timestampTZValue));
+  }
```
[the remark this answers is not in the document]
Torch3333: We don't. I added them because they were used in a non deprecated Key.of(..) static constructor but did not realize I could use do without creating them.
Thank you. 
It's revised in https://github.com/scalar-labs/scalardb/pull/2437/commits/0b774d21af6fa787e4fd841d42ed03d484d40b99

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
       ""),
   DATA_LOADER_ERROR_METHOD_NULL_ARGUMENT(
       Category.USER_ERROR, "0151", "Method null argument not allowed", "", ""),
+  OUT_OF_RANGE_COLUMN_VALUE_FOR_DATE(
+      Category.USER_ERROR,
+      "0152",
+      "This Date column value is out of the valid range. It must be from 1000-01-01 through 9999-12-12. Value: %s",
```
josh-wong, line 696: I tend to see column names spelled either all capitalized ([example](https://docs.aws.amazon.com/redshift/latest/dg/r_Datetime_types.html)) or all lowercase ([example](https://www.postgresql.org/docs/current/datatype-datetime.html)).

Should this capitalization stay as is? Or should we use one of the following instead?

## All capitalized

```suggestion
      "This DATE column value is out of the valid range. It must be between 1000-01-01 and 9999-12-12. Value: %s",
```

## All lowercase

```suggestion
      "This date column value is out of the valid range. It must be between 1000-01-01 and 9999-12-12. Value: %s",
```

> [!NOTE]
>
> The style we choose here should be applied to the other release notes added in this PR.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
...
+      "",
+      ""),
+  SUBMICROSECOND_PRECISION_NOT_SUPPORTED_FOR_TIME(
+      Category.USER_ERROR,
+      "0153",
+      "This Time column value precision cannot be smaller than a microsecond. Value: %s",
```
josh-wong, line 702: Just to be specific, I think we should change `a` to `one`.

```suggestion
      "This Time column value precision cannot be smaller than one microsecond. Value: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
...
+      "",
+      ""),
+  OUT_OF_RANGE_COLUMN_VALUE_FOR_TIMESTAMP(
+      Category.USER_ERROR,
+      "0154",
+      "This Timestamp column value is out of the valid range. It must be from 1000-01-01T00:00:00.000 through 9999-12-31T23:59:59.999. Value: %s",
```
josh-wong, line 708:
```suggestion
      "This Timestamp column value is out of the valid range. It must be between 1000-01-01T00:00:00.000 and 9999-12-31T23:59:59.999. Value: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
...
+      "",
+      ""),
+  SUBMILLISECOND_PRECISION_NOT_SUPPORTED_FOR_TIMESTAMP(
+      Category.USER_ERROR,
+      "0155",
+      "This Timestamp column value precision cannot be smaller than a millisecond. Value: %s",
```
josh-wong, line 714:
```suggestion
      "This Timestamp column value precision cannot be smaller than one millisecond. Value: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
...
+      "",
+      ""),
+  OUT_OF_RANGE_COLUMN_VALUE_FOR_TIMESTAMPTZ(
+      Category.USER_ERROR,
+      "0156",
+      "This TimestampTZ column value is out of the valid range. It must be from 1000-01-01T00:00:00.000Z through 9999-12-31T23:59:59.999Z. Value: %s",
```
josh-wong, line 720:
```suggestion
      "This TimestampTZ column value is out of the valid range. It must be between 1000-01-01T00:00:00.000Z to 9999-12-31T23:59:59.999Z. Value: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
...
+      "",
+      ""),
+  SUBMILLISECOND_PRECISION_NOT_SUPPORTED_FOR_TIMESTAMPTZ(
+      Category.USER_ERROR,
+      "0157",
+      "This TimestampTZ column value precision cannot be smaller than a millisecond. Value: %s",
```
josh-wong, line 726:
```suggestion
      "This TimestampTZ column value precision cannot be smaller than one millisecond. Value: %s",
```

josh-wong (changes requested):

core/src/main/java/com/scalar/db/common/error/CoreError.java
Torch3333: So far in the Javadoc comment, we are using the all-capitalized naming. 
If there are no objections, I will update the error code to use all-capitalized type names.

josh-wong: All-capitalized naming sounds good to me! I'll add a note about that to our style guide.

Torch3333: I applied all your suggestions, thank you.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
...
+  OUT_OF_RANGE_COLUMN_VALUE_FOR_DATE(
+      Category.USER_ERROR,
+      "0152",
+      "This DATE column value is out of the valid range. It must be between 1000-01-01 and 9999-12-12. Value: %s",
+      "",
+      ""),
+  SUBMICROSECOND_PRECISION_NOT_SUPPORTED_FOR_TIME(
+      Category.USER_ERROR,
+      "0153",
+      "This TIME column value precision cannot be smaller than one microsecond. Value: %s",
```
josh-wong, line 702: I just realized that, since we're talking about time, this should be `shorter` instead of `smaller`.

```suggestion
      "This TIME column value precision cannot be shorter than one microsecond. Value: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
...
+  OUT_OF_RANGE_COLUMN_VALUE_FOR_TIMESTAMP(
+      Category.USER_ERROR,
+      "0154",
+      "This TIMESTAMP column value is out of the valid range. It must be between 1000-01-01T00:00:00.000 and 9999-12-31T23:59:59.999. Value: %s",
+      "",
+      ""),
+  SUBMILLISECOND_PRECISION_NOT_SUPPORTED_FOR_TIMESTAMP(
+      Category.USER_ERROR,
+      "0155",
+      "This TIMESTAMP column value precision cannot be smaller than one millisecond. Value: %s",
```
josh-wong, line 714:
```suggestion
      "This TIMESTAMP column value precision cannot be shorter than one millisecond. Value: %s",
```

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -690,6 +690,42 @@ public enum CoreError implements ScalarDbError {
...
+  OUT_OF_RANGE_COLUMN_VALUE_FOR_TIMESTAMPTZ(
+      Category.USER_ERROR,
+      "0156",
+      "This TIMESTAMPTZ column value is out of the valid range. It must be between 1000-01-01T00:00:00.000Z to 9999-12-31T23:59:59.999Z. Value: %s",
+      "",
+      ""),
+  SUBMILLISECOND_PRECISION_NOT_SUPPORTED_FOR_TIMESTAMPTZ(
+      Category.USER_ERROR,
+      "0157",
+      "This TIMESTAMPTZ column value precision cannot be smaller than one millisecond. Value: %s",
```
josh-wong, line 726:
```suggestion
      "This TIMESTAMPTZ column value precision cannot be shorter than one millisecond. Value: %s",
```

josh-wong (changes requested):
Sorry, there was one minor thing I noticed about the wording, so I've added some suggestions. PTAL!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -4,18 +4,30 @@
...
+import java.util.Collections;
+import java.util.Map;
 import javax.annotation.Nullable;
 
 /**
  * An interface to hide the difference between underlying JDBC SQL engines in SQL dialects, error
  * codes, and so on. It's NOT responsible for actually connecting to underlying engines.
  */
-public interface RdbEngineStrategy {
+public interface RdbEngineStrategy<T_DATE, T_TIME, T_TIMESTAMP, T_TIMESTAMPTZ> {
```
feeblefakie, line 30: I don't have a strong opinion, but having four time-related generics in this interface might be too much since this handles many other data types.
How about pushing the time-related type things to other interface and classes?
For example, how about creating a time-specific strategy that converts the default type to a DB-specific type?
For example, we can create `RdbEngineTimeTypeStrategy`, and it converts time-related types as necessary as follows.

```
preparedStatement.setObject(index++, rdbEngineTimeType.convert(rdbEngine.encodeTimestampTZ(column)));
```

feeblefakie (commented):
Overall, looking good to me. Thank you!
I left a few comments. PTAL!

core/src/main/java/com/scalar/db/io/TimestampTZColumn.java
```diff
@@ -0,0 +1,150 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A {@code Column} for a TIMESTAMPTZ type. It represents a date-time in the UTC time zone in the
+ * ISO-8601 calendar system, such as 2017-06-19T16:15:30Z, and can be expressed with millisecond
+ * precision.
+ */
+@Immutable
+public class TimestampTZColumn implements Column<Instant> {
```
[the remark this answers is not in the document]
Torch3333: I also hesitated between `TimestampTZ` and `TimestampTz` but chose the first one for the following reasons: 
 - My understanding of the [Google Java style guide](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) differs from yours. Since `time zone` is written in two words, it recommends capitalizing the `T` and the `Z`.  
-  Compared to `TimestampColumn`, `TimestampTZColumn` stands out more than `TimestampTzColumn`. So, it is easier to spot the difference between the two types when reading code and writing code using auto-completion.
 - I could only find two Java classes with a similar naming and both used the capital letter version. For example in [Apache Hive](https://github.com/apache/hive/blob/master/common/src/java/org/apache/hadoop/hive/common/type/TimestampTZ.java) and [Spark](https://spark.apache.org/docs/latest/api/java/index.html?org%2Fapache%2Fspark%2Fsql%2Ftypes%2FTimestampNTZType.html)
 
 What do you think?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
Torch3333: It sounds good, thank you.
I reverted an earlier commit based on @komamitsu suggestion in https://github.com/scalar-labs/scalardb/pull/2437/commits/0eee567e6d4385413cb6d1f27f2e37698fc3864b than added your suggestion in 
https://github.com/scalar-labs/scalardb/pull/2437/commits/fe3be8d38000b0f8c7ffc62c27202c62a5bdf4c4

josh-wong (approved):
LGTM! Thank you!🙇🏻‍♂️

Does this PR need to have labels and projects selected? Or is it not necessary since it'll be merged into an existing branch where other work is being done?

Torch3333:
@josh-wong 

> LGTM! Thank you!🙇🏻‍♂️
> 
> Does this PR need to have labels and projects selected? Or is it not necessary since it'll be merged into an existing branch where other work is being done?

Yes, the current PR is merged into a temporary feature branch, so we don't add labels or projects. I will add tags and labels to the PR to merge the feature branch into master.

feeblefakie (approved):
LGTM! Thank you!

PR #2301, opened 2024-10-30
Add time-related types to the Admin and Schema loader

Torch3333:
## Description

This updates the Admin and Schema Loader components to be able to manage schema with the new time-related types:
- DATE
- TIME
- TIMESTAMP
- TIMESTAMPTZ

> **_NOTE:_** Cassandra does not support the TIMESTAMP type but we plan to support it in the future.

At the moment, I plan to have a second PR which includes the CRUD interface changes, and a third one which includes the changes to the import table feature.

## Related issues and/or PRs

N/A

## Changes made

Update the implementations of the `DistributedStorageAdmin` and Schema Loader to support managing schema with time-related types.

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

This is the first PR of several PR to support time-related on ScalarDB core, the changes of this feature will be merged on a feature branch [add_time_related_types](https://github.com/scalar-labs/scalardb/tree/add_time_related_types) before being merged into the master branch.

## Release notes

N/A

PR #2461, opened 2025-01-10
Revise import table feature for time-related types

Torch3333:
## Description

This PR brings three main changes to the import table feature:
1. Make the time-related types that could not be imported to be importable.
image

2. When importing a table, the default ScalarDB type mapping for a given column can now be overridden. This is only useful for some time-related types as of now:
    - Oracle DATE type which can be used a ScalarDB DATE (default mapping), TIME or TIMESTAMP
    - Oracle TIMESTAMP type which can be used as ScalarDB DATE or TIMESTAMP (default mapping)
    - Mysql DATETIME type which can be used as ScalarDB TIMESTAMP (default mapping) or TIMESTAMPTZ
    
    To do so, the user needs to define the columns requiring type override [through the Schema Loader import table schema file](https://github.com/scalar-labs/scalardb/pull/2461#discussion_r1915789689) or the [Admin API. ](https://github.com/scalar-labs/scalardb/pull/2461#discussion_r1915792640)
    
3. To verify that there is no encoding-decoding issue with the JDBC driver with all data types that ScalarDB supports through the import table feature, I added an integration test that performs a put then get on a generic value for each column.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/2301
- https://github.com/scalar-labs/scalardb/pull/2437

## Changes made

See the description above and the comments left below.

## Checklist

This PR will be merged into the feature branch https://github.com/scalar-labs/scalardb/compare/master...add_time_related_types

## Release notes

N/A

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
@@ -1,14 +1,39 @@
 package com.scalar.db.storage.jdbc;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.scalar.db.api.DistributedStorageAdminImportTableIntegrationTestBase.TestData;
+import com.scalar.db.api.Get;
+import com.scalar.db.api.Insert;
+import com.scalar.db.api.InsertBuilder;
+import com.scalar.db.api.Put;
+import com.scalar.db.api.PutBuilder;
 import com.scalar.db.api.TableMetadata;
 import com.scalar.db.config.DatabaseConfig;
+import com.scalar.db.io.BigIntColumn;
+import com.scalar.db.io.BlobColumn;
+import com.scalar.db.io.BooleanColumn;
+import com.scalar.db.io.Column;
 import com.scalar.db.io.DataType;
+import com.scalar.db.io.DateColumn;
+import com.scalar.db.io.DoubleColumn;
+import com.scalar.db.io.FloatColumn;
+import com.scalar.db.io.IntColumn;
+import com.scalar.db.io.Key;
+import com.scalar.db.io.TextColumn;
+import com.scalar.db.io.TimeColumn;
+import com.scalar.db.io.TimestampColumn;
+import com.scalar.db.io.TimestampTZColumn;
+import java.nio.charset.StandardCharsets;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -16,6 +41,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.annotation.Nullable;
 import org.apache.commons.dbcp2.BasicDataSource;
 
 public class JdbcAdminImportTestUtils {
@@ -24,16 +50,12 @@ public class JdbcAdminImportTestUtils {
       Arrays.asList(
           "BIGINT UNSIGNED",
           "BIT(8)",
-          "DATE",
-          "DATETIME",
           "DECIMAL(8,2)",
           "ENUM('a','b')",
           "SET('a','b')",
           "GEOMETRY",
           "JSON", // we remove this for MariaDB because it is an alias of a supported type, LONGTEXT
           "NUMERIC",
-          "TIME",
-          "TIMESTAMP",
           "YEAR");
   static final List<String> UNSUPPORTED_DATA_TYPES_PGSQL =
       Arrays.asList(
@@ -43,7 +65,6 @@ public class JdbcAdminImportTestUtils {
           "box",
           "cidr",
           "circle",
-          "date",
           "inet",
           "interval",
           "json",
@@ -60,10 +81,7 @@ public class JdbcAdminImportTestUtils {
           "polygon",
           "serial",
           "smallserial",
-          "time",
           "time with time zone",
-          "timestamp",
-          "timestamp with time zone",
           "tsquery",
           "tsvector",
           "txid_snapshot",
@@ -74,34 +92,24 @@ public class JdbcAdminImportTestUtils {
   static final List<String> UNSUPPORTED_DATA_TYPES_ORACLE =
       Arrays.asList(
           "BFILE",
-          "DATE",
           "FLOAT(54)",
           "INT",
           "INTERVAL YEAR(3) TO MONTH",
           "INTERVAL DAY(2) TO SECOND",
           "NUMBER(16,0)",
           "ROWID",
-          "TIMESTAMP",
-          "TIMESTAMP WITH TIME ZONE",
-          "TIMESTAMP WITH LOCAL TIME ZONE",
           "UROWID");
   static final List<String> UNSUPPORTED_DATA_TYPES_ORACLE_V20_OR_LATER =
       Collections.singletonList("JSON");
   static final List<String> UNSUPPORTED_DATA_TYPES_MSSQL =
       Arrays.asList(
-          "date",
-          "datetime",
-          "datetime2",
-          "datetimeoffset",
           "decimal(8,2)",
           "hierarchyid",
           "money",
           "numeric(8,2)",
           "rowversion",
-          "smalldatetime",
           "smallmoney",
           "sql_variant",
-          "time",
           "uniqueidentifier",
           "xml",
           "geometry",
@@ -122,7 +130,7 @@ public class JdbcAdminImportTestUtils {
   @Override
   protected final void finalize() {}
 
-  public Map<String, TableMetadata> createExistingDatabaseWithAllDataTypes(String namespace)
+  public List<TestData> createExistingDatabaseWithAllDataTypes(String namespace)
       throws SQLException {
     execute(rdbEngine.createSchemaSqls(namespace));
     if (JdbcTestUtils.isMysql(rdbEngine)) {
@@ -179,9 +187,14 @@ public class JdbcAdminImportTestUtils {
     columns.put("col18", "TINYBLOB");
     columns.put("col19", "MEDIUMBLOB");
     columns.put("col20", "LONGBLOB");
-    columns.put("col21", "BINARY(255)");
+    columns.put("col21", "BINARY(5)");
+    columns.put("col22", "DATE");
+    columns.put("col23", "TIME(6)");
+    columns.put("col24", "DATETIME(6)");
+    columns.put("col25", "DATETIME(6)"); // override to TIMESTAMPTZ
+    columns.put("col26", "TIMESTAMP(6)");
     if (isMariaDB()) {
-      columns.put("col22", "JSON");
+      columns.put("col27", "JSON");
     }
     return columns;
   }
@@ -212,12 +225,32 @@ public class JdbcAdminImportTestUtils {
             .addColumn("col19", DataType.BLOB)
             .addColumn("col20", DataType.BLOB)
             .addColumn("col21", DataType.BLOB)
+            .addColumn("col22", DataType.DATE)
+            .addColumn("col23", DataType.TIME)
+            .addColumn("col24", DataType.TIMESTAMP)
+            .addColumn("col25", DataType.TIMESTAMPTZ)
+            .addColumn("col26", DataType.TIMESTAMPTZ)
             .addPartitionKey("pk1")
             .addPartitionKey("pk2");
     if (isMariaDB()) {
-      builder.addColumn("col22", DataType.TEXT);
+      builder.addColumn("col27", DataType.TEXT);
     }
     return builder.build();
+  }
+
+  private Map<String, Column<?>> prepareInsertColumnsForMysql(TableMetadata metadata) {
+    ImmutableList.Builder<Column<?>> customColumns = new ImmutableList.Builder<>();
+    customColumns.add(
+        TimestampTZColumn.of(
+            "col26", LocalDateTime.of(2005, 10, 11, 8, 35).toInstant(ZoneOffset.UTC)));
+    if (isMariaDB()) {
+      customColumns.add(TextColumn.of("col27", "{\"foo\": \"bar\"}"));
+    }
+    return prepareInsertColumnsWithGenericAndCustomValues(metadata, customColumns.build());
+  }
+
+  private Map<String, DataType> prepareOverrideColumnsTypeForMysql() {
+    return ImmutableMap.of("col25", DataType.TIMESTAMPTZ);
   }
 
   private LinkedHashMap<String, String> prepareColumnsForPostgresql() {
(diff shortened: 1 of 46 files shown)


core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -179,9 +187,14 @@ private LinkedHashMap<String, String> prepareColumnsForMysql() {
     columns.put("col18", "TINYBLOB");
     columns.put("col19", "MEDIUMBLOB");
     columns.put("col20", "LONGBLOB");
-    columns.put("col21", "BINARY(255)");
+    columns.put("col21", "BINARY(5)");
```
Torch3333, line 190: I changed the column size to facilitate inserting a BLOB value for the simple write-read integration test. For this data type, the value is right-padded if the value length is inferior to the column size. 
I did a similar change for some other data types that are right padded.

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -559,7 +658,195 @@ private int getMajorVersion() {
...
+        throw new AssertionError();
+    }
+  }
+
   public void close() throws SQLException {
     dataSource.close();
   }
+
+  @SuppressWarnings("UseCorrectAssertInTests")
+  public static class JdbcTestData implements TestData {
```
Torch3333, line 719: I refactored the `JdbcAdminImportTestUtils` classes to store the test data for importable and non-importable tables in this object to improve readability.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageAdminImportTableIntegrationTestBase.java
```diff
@@ -147,4 +156,84 @@ private void importTable_ForNonExistingTable_ShouldThrowIllegalArgumentException
             () -> admin.importTable(getNamespace(), "non-existing-table", Collections.emptyMap()))
         .isInstanceOf(IllegalArgumentException.class);
   }
+
+  public void importTable_ForImportedTable_ShouldPutThenGetCorrectly() throws ExecutionException {
```
Torch3333, line 160: To verify that is no encoding-decoding issue with the JDBC driver with all data types that ScalarDB supports through the import table feature, I added an integration test that performs a put then get on a generic value for each column.

schema-loader/sample/import_schema_sample.json
```diff
@@ -1,6 +1,10 @@
 {
   "sample_namespace1.sample_table1": {
-    "transaction": true
+    "transaction": true,
+    "override-columns-type": {
```
Torch3333, line 4: By setting a columns and the desired data type inside the `override-columns-type` JSON object of the import table schema file , user can override the default type mapping.
This is only useful for some type-related types on Mysql and Oracle now.

core/src/main/java/com/scalar/db/api/Admin.java
```diff
@@ -451,7 +451,28 @@ default void addNewColumnToTable(
...
+   * @throws IllegalArgumentException if the table is already managed by ScalarDB, if the target
+   *     table does not exist, or if the table does not meet the requirement of ScalarDB table
+   * @throws ExecutionException if the operation fails
+   */
+  void importTable(
+      String namespace,
+      String table,
+      Map<String, String> options,
+      Map<String, DataType> overrideColumnsType)
       throws ExecutionException;
```
Torch3333, line 476: Add the `Map<String, DataType> overrideColumnsType` to override the type for columns in the `Admin.importTable(...)` method.

core/src/main/java/com/scalar/db/api/DistributedStorageAdmin.java
```diff
@@ -52,7 +54,26 @@ public interface DistributedStorageAdmin extends Admin, AutoCloseable {
...
+   * @param overrideColumnsType a map of column data type by column name. Only set the column for
+   *     which you want to override the default data type mapping.
+   * @throws IllegalArgumentException if the table does not exist
+   * @throws IllegalStateException if the table does not meet the requirement of ScalarDB table
+   * @throws ExecutionException if the operation fails
+   * @return import table metadata in the ScalarDB format
+   */
+  TableMetadata getImportTableMetadata(
+      String namespace, String table, Map<String, DataType> overrideColumnsType)
+      throws ExecutionException;
```
Torch3333, line 76: Add the `Map<String, DataType> overrideColumnsType` to override the type for columns in the `Admin.getImportTableMetadata(...)` method.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -726,6 +726,12 @@ public enum CoreError implements ScalarDbError {
       "This TIMESTAMPTZ column value precision cannot be shorter than one millisecond. Value: %s",
       "",
       ""),
+  JDBC_IMPORT_DATA_TYPE_OVERRIDE_NOT_SUPPORTED(
+      Category.USER_ERROR,
+      "0158",
+      "The underlying storage data type %s is not supported as ScalarDB %s data type: %s",
+      "",
+      ""),
```
Torch3333, line 734: @josh-wong 
Do you mind checking the English of this error message?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -310,6 +315,21 @@ public DataType getDataTypeForScalarDb(
...
+              CoreError.JDBC_IMPORT_DATA_TYPE_NOT_SUPPORTED.buildMessage(
+                  typeName, columnDescription));
+        }
+        return DataType.DATE;
+      case TIME:
+        return DataType.TIME;
+        // Both MySQL TIMESTAMP and DATETIME data types are mapped to the TIMESTAMP JDBC type
+      case TIMESTAMP:
+        if (overrideDataType == DataType.TIMESTAMPTZ || typeName.equalsIgnoreCase("TIMESTAMP")) {
+          return DataType.TIMESTAMPTZ;
```
komamitsu, line 330: With this PR, the following requirement can be addressed?
- The underlying database has some timestamps in DATETIME columns
- The timestamps in the columns are JST-based
- The user wants to import the timestamps to ScalarDB as `ScalarDB.TIMESTAMPZ` not `ScalarDB.TIMESTAMP`

I guess timezone issues would occur since the DATETIME columns don't have timezone information and there is no way to specify the timezone offset between JST and UTC for table import. Is my understanding correct?

Torch3333: > I guess timezone issues would occur since the DATETIME columns don't have timezone information 

It depends on how the user inserted the timestamp:
- if it was inserted without offset such as `2021-01-01 15:00:00`, then the read `ScalarDB.TIMESTAMPZ` value is `2021-01-01 15:00:00Z`
- if it was inserted with the offset of JST towards the UTC timezone such as ``2021-01-01 15:00:00+09:00`, then the read `ScalarDB.TIMESTAMPZ` value is `2021-01-01 06:00:00Z`. This syntax is only allowed from MySQL 8.0.19.

>  there is no way to specify the timezone offset between JST and UTC for table import
Yes, this is correct.

ScalarDB assumes `ScalarDB.TIMESTAMPZ` values are stored in UTC so indeed, if the datetime is inserted without offset towards UTC, then there is a timezone mismatch.

We can consider this a limitation and mention in the documentation that importing `MySQL.DATETIME` as `ScalarDB.TIMESTAMPZ` is possible as long as the data time value is stored in UTC.

What do you think?

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -179,9 +187,18 @@ private LinkedHashMap<String, String> prepareColumnsForMysql() {
     columns.put("col18", "TINYBLOB");
     columns.put("col19", "MEDIUMBLOB");
     columns.put("col20", "LONGBLOB");
-    columns.put("col21", "BINARY(255)");
+    columns.put("col21", "BINARY(5)");
+    columns.put("col22", "DATE");
+    columns.put("col23", "TIME(6)");
+    columns.put("col24", "DATETIME(6)");
+    columns.put("col25", "DATETIME(6)"); // override to TIMESTAMPTZ
+    // With Mysql 5.7, if a TIMESTAMP column is not explicitly declared with the NULL attribute, it
```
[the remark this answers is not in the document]
Torch3333: Thanks, addressed in https://github.com/scalar-labs/scalardb/pull/2461/commits/be1dd02ea10f0610f3a016afe33c369d239373c1

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -559,7 +662,195 @@ private int getMajorVersion() {
...
+    public static JdbcTestData createImportableTable(
+        String tableName,
+        String createTableSql,
+        TableMetadata tableMetadata,
+        Map<String, DataType> overrideColumnsType,
+        Map<String, Column<?>> columns) {
+      return new JdbcTestData(
+          tableName, createTableSql, overrideColumnsType, tableMetadata, columns);
+    }
+    /**
```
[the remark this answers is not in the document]
Torch3333: Thanks, addressed in https://github.com/scalar-labs/scalardb/pull/2461/commits/be1dd02ea10f0610f3a016afe33c369d239373c1

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
komamitsu: > if it was inserted without offset such as 2021-01-01 15:00:00, then the read ScalarDB.TIMESTAMPZ value is 2021-01-01 15:00:00Z

This is the case I was considering.

I understand there are pros and cons to using MySQL.DATETIME and MySQL.TIMESTAMP from the perspective of timezone info and time range. So, probably this issue should be considered from product requirement perspective.

This is just an idea, but if there is an option like `original-timezone` or something with which users can specify the original DB's timezone (JST in the above example), ScalarDB table import can store the timezone in `metadata` table so that ScalarDB handles the column as JST? With the feature, MySQL.DATETIME `2021-01-01 15:00:00` in the underlying database table would be handled as ScalarDB.TIMESTAMPZ `2021-01-01 06:00:00 UTC`. This might be a bit too much, though.

komamitsu: > mention in the documentation that importing MySQL.DATETIME as ScalarDB.TIMESTAMPZ is possible as long as the data time value is stored in UTC

This is also an option for the corner case with minimal effort! 👍

brfrn169 (approved):
Left a minor comment; otherwise, LGTM! Thank you!

It seems we have some limitations with the time-related types, and I think that’s fine for now. Let’s document these limitations, and we can consider extending type support if customers request it in the future.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
Torch3333: @komamitsu 
> This is just an idea, but if there is an option like original-timezone or something with which users can specify the original DB's timezone (JST in the above example), ScalarDB table import can store the timezone in metadata table so that ScalarDB handles the column as JST?

It's doable but as you guessed, this requires a certain amount of changes in the codebase for a corner case. So I am more in favor of treating it as a limitation at the moment.

Torch3333: @brfrn169 @feeblefakie 
What do you think about this corner case? Should we address it or just mention on the documentation that this is a limitation?

Torch3333: Ah sorry, I just saw @brfrn169 [message](https://github.com/scalar-labs/scalardb/pull/2461#pullrequestreview-2554665835) now

```
It seems we have some limitations with the time-related types, and I think that’s fine for now. Let’s document these limitations, and we can consider extending type support if customers request it in the future.
```

core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
```diff
@@ -7,6 +7,7 @@
 import org.apache.commons.dbcp2.BasicDataSource;
 
 public final class JdbcUtils {
+
```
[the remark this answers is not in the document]
Torch3333: Indeed, thank you. 
https://github.com/scalar-labs/scalardb/pull/2461/commits/ca0f42185c7ade0140021cec6b351b8170f0cb4c

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
feeblefakie: I agree with Toshi. Let's mention the limitations for now.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
```diff
@@ -307,6 +313,29 @@ public DataType getDataTypeForScalarDb(
...
+          if (overrideDataType == DataType.TIME) {
+            return DataType.TIME;
+          }
+          if (overrideDataType == DataType.TIMESTAMP) {
+            return DataType.TIMESTAMP;
+          }
+          return DataType.DATE;
+        }
+        // handles "timestamp" type
+        if (overrideDataType == DataType.TIME) {
```
feeblefakie, line 328: I don't fully check the source code, so I'm not fully sure, but I'm wondering if we do some validation if the specified override data type is compatible with the original data type?
So, for example, if INT is specified as an override data type for TIMESTAMP, does it just use DataType.TIMESTAMP without warning messages?
If so, should it check the compatibility and warn?

feeblefakie (commented):
Overall, looking good to me.
Left a question for understanding it properly.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineOracle.java
Torch3333: > I don't fully check the source code, so I'm not fully sure, but I'm wondering if we do some validation if the specified override data type is compatible with the original data type?

Yes, let me explain it.

The [AbstractRdbEngine.getDataTypeForScalarDbInternal()](https://github.com/scalar-labs/scalardb/blob/ca0f42185c7ade0140021cec6b351b8170f0cb4c/core/src/main/java/com/scalar/db/storage/jdbc/AbstractRdbEngine.java#L31C21-L31C51) method will only return a data type that equals the overrideDataType if it is allowed.

If `AbstractRdbEngine.getDataTypeForScalarDbInternal()` returns a data type that does not equal the desired overrideDataType, then an exception is thrown in this encapsulating method defined in `AbstractRdbEngine`
 https://github.com/scalar-labs/scalardb/blob/ca0f42185c7ade0140021cec6b351b8170f0cb4c/core/src/main/java/com/scalar/db/storage/jdbc/AbstractRdbEngine.java#L18-L28 
 
 I adopted this design to avoid having to add repetitive exception throws in `AbstractRdbEngine.getDataTypeForScalarDbInternal()` but I agree it might be a bit difficult to understand the logic.

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/common/error/CoreError.java
josh-wong: The following suggestion is just a minor change.

```suggestion
  JDBC_IMPORT_DATA_TYPE_OVERRIDE_NOT_SUPPORTED(
      Category.USER_ERROR,
      "0158",
      "The underlying-storage data type %s is not supported as the ScalarDB %s data type: %s",
      "",
      ""),
```

josh-wong (changes requested):
I've added one minor suggestion. PTAL!

josh-wong (approved):
LGTM! Thank you!🙇🏻‍♂️

PR #2598, opened 2025-04-16
Support IBM Db2

Torch3333:
## Description

This adds support for IBM Db2 LUV for versions 11 and 12.

## Changes made

Add a new JDBC adapter for Db2. Please look at my review comments below for the highlights of the changes.
 
## Checklist

N/A

## Release notes

Add support for IBM Db2 LUV 11 and 12


core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
```diff
@@ -53,8 +53,11 @@ public class JdbcConfig {
       PREFIX + "mysql.variable_key_column_size";
   public static final String ORACLE_VARIABLE_KEY_COLUMN_SIZE =
       PREFIX + "oracle.variable_key_column_size";
+  public static final String DB2_VARIABLE_KEY_COLUMN_SIZE = PREFIX + "db2.variable_key_column_size";
   public static final String ORACLE_TIME_COLUMN_DEFAULT_DATE_COMPONENT =
       PREFIX + "oracle.time_column.default_date_component";
+  public static final String DB2_TIME_COLUMN_DEFAULT_DATE_COMPONENT =
+      PREFIX + "db2.time_column.default_date_component";
```
Torch3333, line 60: The ScalarDB TIME type supports the microsecond fraction of a second, but the Db2 TIME type does not support fractional seconds. For this reason, I chose to use the Db2 TIMESTAMP type by default to map the ScalarDB TIME type. The drawback is that a fixed date component is stored with the TIME value, which increases the data volume.

Similarly to Oracle, the fixed date component value used to store value in the TIMESTAMP type is configurable with `scalar.db.jdbc.db2.time_column.default_date_component`

core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
```diff
@@ -92,6 +92,10 @@ public static BasicDataSource initDataSourceForTableMetadata(
     dataSource.setMinIdle(config.getTableMetadataConnectionPoolMinIdle());
     dataSource.setMaxIdle(config.getTableMetadataConnectionPoolMaxIdle());
     dataSource.setMaxTotal(config.getTableMetadataConnectionPoolMaxTotal());
+    for (Entry<String, String> entry : rdbEngine.getConnectionProperties().entrySet()) {
+      dataSource.addConnectionProperty(entry.getKey(), entry.getValue());
+    }
```
Torch3333, line 97: Now we add custom JDBC connection properties if set in the RdbEngine for all JDBC data source used by: 
- `DistributedStorage` => added previously
- `DistributedStorageAdmin` => newly added
- `TableMetadataManager` => newly added

This is motivated by the fact that the Db2 driver requires a configuration property `retrieveMessagesFromServerOnGetMessage` to print a verbose error message that does not contain only the error code.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineDb2.java
```diff
@@ -0,0 +1,502 @@
...
+        return "INT";
+      case TEXT:
+        return "VARCHAR(32672)";
+      case DATE:
+        return "DATE";
+      case TIME:
+        return "TIMESTAMP(6)";
+      case TIMESTAMP:
+      case TIMESTAMPTZ:
+        return "TIMESTAMP(3)";
```
Torch3333, line 85: The list default type mapping for ScalarDB data types.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineDb2.java
```diff
@@ -0,0 +1,502 @@
...
+  }
+
+  @Override
+  public RdbEngineTimeTypeStrategy<String, LocalDateTime, String, String> getTimeTypeStrategy() {
+    return timeTypeEngine;
+  }
+
+  @Override
+  public Map<String, String> getConnectionProperties() {
+    ImmutableMap.Builder<String, String> props = new ImmutableMap.Builder<>();
```
Torch3333, line 440: Add several JDBC driver properties, see the comments for the reason.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -268,16 +269,16 @@ private void createTableInternalSqlsAfterCreateTable(
         rdbEngine.createTableInternalSqlsAfterCreateTable(
             hasDifferentClusteringOrders(metadata), schema, table, metadata, ifNotExists);
     try {
-      execute(connection, stmts);
+      execute(connection, stmts, ifNotExists ? null : rdbEngine::throwIfDuplicatedIndexWarning);
```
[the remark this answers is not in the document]
Torch3333: Yes, only Db2 works like that. 
To make sure we are on the same page, I'd like to explain the current implementation in detail. 

Only an exception will be thrown with Db2 if a duplicate index code is matched. 
https://github.com/scalar-labs/scalardb/blob/adcf4d57b9d1a300f088e867b4b507dff6bd05e5/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineDb2.java#L281-L292

But for other storage, the warning won't even be parsed.
https://github.com/scalar-labs/scalardb/blob/adcf4d57b9d1a300f088e867b4b507dff6bd05e5/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java#L220-L228

Regarding your concern,

> My slight concern is this change might affect existing production services that use ScalarDB's other JDBC adapters, where negligible warnings have been emitted silently

This implementation should not impact other storage, even if warnings are emitted. 
What do you think ?

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -930,17 +932,41 @@ private void updateTableMetadata(
...
     if (Strings.isNullOrEmpty(sql)) {
       return;
     }
     try (Statement stmt = connection.createStatement()) {
       stmt.execute(sql);
+      throwSQLWarningIfNeeded(handler, stmt);
+    }
+  }
+
+  private static void throwSQLWarningIfNeeded(SQLWarningHandler handler, Statement stmt)
```
feeblefakie, line 949:
```suggestion
  private static void throwSqlWarningIfNeeded(SqlWarningHandler handler, Statement stmt)
```

feeblefakie (commented):
Overall, looking good.
I left one minor naming suggestion.
I'm also concerned about the blob type size.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -41,6 +44,10 @@ public interface RdbEngineStrategy {
 
   String getDataTypeForKey(DataType dataType);
 
+  default String getDataTypeForSecondaryIndex(DataType dataType) {
```
brfrn169, line 47: It seems like we should add the `@Nullable` annotation:
```suggestion
  @Nullable
  default String getDataTypeForSecondaryIndex(DataType dataType) {
```

Could you please add it to `String getDataTypeForKey(DataType dataType);` as well. Also, please add it to the implementation methods?

integration-test/src/main/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
```diff
@@ -216,7 +216,7 @@ protected Map<String, Object> getAlteredSchemaJsonMap() {
                         isTimestampTypeSupported()
                             ? ImmutableMap.of("col10", "TIMESTAMPTZ", "col11", "TIMESTAMP")
                             : ImmutableMap.of("col10", "TIMESTAMPTZ"))
-                    .put("col12", "TEXT")
+                    .put("col12", "INT")
```
brfrn169, line 219: Why do we need to change this type?

brfrn169 (commented):
Left a few comments. Please take a look when you have time!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineDb2.java
```diff
@@ -0,0 +1,497 @@
...
+    return new com.ibm.db2.jcc.DB2Driver();
+  }
+
+  @Override
+  public String getDataTypeForEngine(DataType scalarDbDataType) {
+    switch (scalarDbDataType) {
+      case BIGINT:
+        return "BIGINT";
+      case BLOB:
+        return "VARBINARY(32672)";
```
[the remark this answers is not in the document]
Torch3333: It is possible to use BLOB, but this data type cannot be an index. 
And with the purpose of creating an index, altering a BLOB column to a VARBINARY is not allowed. 

I assume that's why it was decided to map ScalarDB BLOB to Oracle RAW(2000), while Oracle also provides BLOB type that allows much bigger sizes than RAW.

So, it all boils down to whether there is a use case for creating an index on a blob column; if not, then we can choose to use BLOB types for the default mapping. And to stay consistent, we should do the same for Oracle.

integration-test/src/main/java/com/scalar/db/schemaloader/SchemaLoaderIntegrationTestBase.java
Torch3333: The `createTableThenAlterTables_ShouldExecuteProperly` will create an index on this column. However, since index creation for a TEXT column fails for Db2 on the CI, I changed the type to INT to avoid disabling this specific test for Db2.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
Torch3333: Thank you.
f0ce19bac0ac0e1b1df8d21809a4e35fbf90d239

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
Torch3333: Thank you.
f0ce19bac0ac0e1b1df8d21809a4e35fbf90d239

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineDb2.java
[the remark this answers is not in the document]
komamitsu: I see. That's an interesting limitation.

From the user perspective of ScalarDB, especially as the developer of SSR (which serializes a group of many items and stores it in a BLOB column), the 32KB limitation could be a blocker for that use case.

How about using DB2 `VARBINARY` for partition/clustering keys and secondary indexes, and DB2 `BLOB` for regular columns? As for cases where adding a new secondary index on DB2 `BLOB` type column, I think it should fail by validations, though.

If it works well, migrating the Oracle `RAW` type to `BLOB` type keeping compatibility might be an option in the future?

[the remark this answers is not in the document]
brfrn169: Let me clarify.

First of all, as noted in the following documentation, some adopters already use different data types for partition key, clustering key, and secondary index key columns compared to regular columns:

https://scalardb.scalar-labs.com/docs/latest/schema-loader/#data-type-mapping-between-scalardb-and-other-databases

That said, the reason we don’t use Oracle `BLOB` or DB2 `BLOB` for regular columns is due to an issue that arises when creating indexes on those columns.

When creating an index on a column that uses a different data type for secondary index keys, we need to alter the column’s data type before creating the index. However, Oracle and DB2 do not allow altering the data type from `BLOB` to `REAL` or `VARBINARY`, and vice versa.

Due to this limitation, we decided to use `REAL` or `VARBINARY` for regular columns as well, to avoid type-alteration issues during index creation.

I think there’s one more point to consider.

We currently have two ways to create an index:
1. Using the `createTable()` API by specifying secondary indexes
2. Using the `createIndex()` API

For case 2, we need to alter the data type as mentioned above. But for case 1, we don’t need to alter the data type because the column is created as an index key column from the beginning.

Given that, I think we have two possible options:
1. As @Torch3333 mentioned, we forbid creating an index on a `BLOB` column of an existing table. In this case, users can create an index through the `createTable()` API by specifying secondary indexes, but not via the `createIndex()` API.
2. As we discussed in the meeting, we disallow creating indexes on `BLOB` columns altogether so that we can consistently use Oracle `BLOB` or DB2 `BLOB`.

Personally, I feel option 2 is more consistent and would cause less confusion for users. What do you think?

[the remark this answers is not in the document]
feeblefakie: I discussed it with Toshi, and we want to go with option 2 if you are OK.
I think non-indexable Blob data is acceptable for most users.
For this PR, I think we can merge it without the change.
(And, let's work on expanding the Blob data size and adding some validation in another PR if OK.)

feeblefakie (approved):
LGTM! Thank you!

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineDb2.java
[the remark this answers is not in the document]
Torch3333: Alright, thanks. 
Let's work on changes in a separate PR.

PR #2601, opened 2025-04-21
Fix behavior of scan operation with limit in Consensus Commit

brfrn169:
## Description

Currently, there are two issues related to the scan operation with a limit in Consensus Commit:
1. A scan operation with a limit returns incorrect results after a record within the scan range is deleted.
    - A test reproducing this issue can be found here:
https://github.com/scalar-labs/scalardb/pull/2601/files#diff-7d8b7163e906546d45d9a7b662f97eb671e39030217ab77ef22325b0f74c2b5eR3268-R3316

2. If a record is inserted into the scan range after executing a scan with a limit, the transaction aborts.
    - A test reproducing this issue can be found here:
https://github.com/scalar-labs/scalardb/pull/2601/files#diff-7d8b7163e906546d45d9a7b662f97eb671e39030217ab77ef22325b0f74c2b5eR3318-R3364

This PR addresses both issues.

## Related issues and/or PRs

N/A

## Changes made

Added some inline comments. Please take a look for the details when you have time!

## Checklist

N/A

## Release notes

Fixed behavior of scan operation with limit in Consensus Commit.

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -79,7 +80,11 @@ public class CrudHandler {
     Get get = (Get) prepareStorageSelection(originalGet);
     Snapshot.Key key = new Snapshot.Key(get);
     readUnread(key, get);
-    return createGetResult(key, get, originalProjections);
+
+    TableMetadata metadata = getTableMetadata(get);
+    return snapshot
+        .getResult(key, get)
+        .map(r -> new FilteredResult(r, originalProjections, metadata, isIncludeMetadataEnabled));
   }
 
   @VisibleForTesting
@@ -113,14 +118,6 @@ public class CrudHandler {
         snapshot.getId());
   }
 
-  private Optional<Result> createGetResult(Snapshot.Key key, Get get, List<String> projections)
-      throws CrudException {
-    TableMetadata metadata = getTableMetadata(key.getNamespace(), key.getTable());
-    return snapshot
-        .getResult(key, get)
-        .map(r -> new FilteredResult(r, projections, metadata, isIncludeMetadataEnabled));
-  }
-
   public List<Result> scan(Scan scan) throws CrudException {
     List<Result> results = scanInternal(scan);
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -513,11 +469,116 @@ void toSerializable(DistributedStorage storage)
...
+    } finally {
+      if (scanner != null) {
+        try {
+          scanner.close();
+        } catch (IOException e) {
+          logger.warn("Failed to close the scanner", e);
+        }
+      }
+    }
+  }
```
brfrn169, line 575: To address issue 2, I’ve rewritten the validation logic for the scan set.

The purpose of this validation is to ensure that the scan results have not been modified by other transactions.

Key considerations for the validation:
- Since the validation is performed after the prepare-record phase, the scan might include prepared records if the transaction has performed operations that affect the scan result.
  - In such cases, those prepared records can be safely ignored.
- For scan operations with a limit, we only need to validate the records up to the number of records that were returned during the execution phase.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -141,33 +138,49 @@ private List<Result> scanInternal(Scan originalScan) throws CrudException {
...
+      Exception exception;
+      if (e.getCause() instanceof ExecutionException) {
+        exception = (ExecutionException) e.getCause();
+      } else {
+        exception = e;
       }
+      throw new CrudException(
+          CoreError.CONSENSUS_COMMIT_SCANNING_RECORDS_FROM_STORAGE_FAILED.buildMessage(),
+          exception,
+          snapshot.getId());
```
brfrn169, line 183: Added exception handling to account for cases where the scanner iterator may throw a `RuntimeException`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -141,33 +138,49 @@ private List<Result> scanInternal(Scan originalScan) throws CrudException {
...
-              snapshot.getId());
-        }
+      @Nullable Integer limit = null;
+      if (scan.getLimit() == 0) {
+        scanner = scanFromStorage(scan);
+      } else {
+        limit = scan.getLimit();
+
+        // Get a scanner without the limit if the scan has a limit
+        scanner = scanFromStorage(Scan.newBuilder(scan).limit(0).build());
```
komamitsu, line 159: This looks good. But, it seems to fetch all records. So, how about this optimization?
- At this line, set the limit to the original limit plus the size of `snapshot.deleteSet` considering cases where the scan range overlaps with `snapshot.deleteSet`
- Skip a result if it's contained in `snapshot.deleteSet`

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -513,11 +469,116 @@ void toSerializable(DistributedStorage storage)
...
+      while (latestResult.isPresent() && resultsEntry != null) {
+        TransactionResult latestTransactionResult = new TransactionResult(latestResult.get());
+        Key key = new Key(scan, latestTransactionResult);
+
+        // Ignore records that this transaction has prepared (and that are in the write set or
+        // delete set)
+        if (latestTransactionResult.getId() != null && latestTransactionResult.getId().equals(id)) {
+          latestResult = scanner.one();
+
+          if (resultsEntry.getKey().equals(key)) {
```
komamitsu, line 510: I might be wrong, but this implementation assumes `resultsIterator` and `scanner` return values in the same order? I'm not sure why `results` is not used as a Map as is...

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -513,11 +469,116 @@ void toSerializable(DistributedStorage storage)
...
+      Iterator<Entry<Key, TransactionResult>> resultsIterator = results.entrySet().iterator();
+
+      Optional<Result> latestResult = scanner.one();
+
+      Entry<Key, TransactionResult> resultsEntry = null;
+      if (resultsIterator.hasNext()) {
+        resultsEntry = resultsIterator.next();
+      }
+
+      while (latestResult.isPresent() && resultsEntry != null) {
```
komamitsu, line 501: Is the original limit not taken into account in this method?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: > this implementation assumes resultsIterator and scanner return values in the same order?

Yes. The implementation of the map `results` is `LinkedHashMap`, so the order is preserved.

This logic assumes that the scanner for the same scan returns records in the same order. What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
komamitsu: Yeah, I'm not sure all the storages will be affected or not. But, I guess either of taking shared locks or equivalents, loading records in memory as a read-set, updating read-timestamps, or etc would internally occur in the underlying database.

I also think users want to use LIMIT when the number of records associated with a partition key will be large. So, pushing down the limit condition (or a bit modified one) as much as possible should be useful.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: Right. We don't need to consider the original limit. We only need to consider the number of the scan result here.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
komamitsu: Glad to hear that `LinkedHashMap` here. It would be great if the parameter type is explicitly `LinkedHashMap`.

As for the underlying database/storage, I don't think they always return scanned records in the same order. They seem to often return records in the same order, but it depends on temporary physical data structure that can change.

brfrn169: For partition scans, the storage abstraction guarantees that records are returned in clustering order, so I think it’s fine in that case. For cross-partition scans and index-based scans, there’s no such ordering guarantee. However, in most cases, the records seem to be returned in a consistent order, so I assumed we could rely on that.

That said, if we can’t assume consistent ordering, it would complicate things quite a bit. Let me discuss this. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Understood. Since the current code doesn’t fetch all records at once and retrieves them one by one through the `Scanner` instance, I don’t think it causes any significant performance issues at this point. I’d prefer to keep the code simple for now and address this as a separate optimization task later if any performance issues arise. Thanks!

komamitsu: > I’d prefer to keep the code simple for now

I guess the reason why `validateScanResults()` uses the double iterators is related to this change to use no-limit. If so, honestly speaking, the implementation using the double iterators doesn't seem simple to me, compared to the original implementation.

brfrn169:
I’ll close this PR since I discussed with @feeblefakie, and we decided to go with another approach.

@komamitsu Thank you for your review!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Let's discuss it in another PR. Thanks!

PR #2610, opened 2025-04-23
Disallow scan operations after delete in Consensus Commit

brfrn169:
## Description

Currently, scan operations after a delete operation on the same record are allowed in Consensus Commit. However, this can cause issues with scan operations with a limit, as explained in https://github.com/scalar-labs/scalardb/pull/2601. 

Additionally, scan operations after a write operation on the same record are already disallowed. Therefore, disallowing scan operations after a delete operation should be a natural extension of the existing restriction.

This PR updates the code to also disallow scan operations after a delete operation in Consensus Commit.

## Related issues and/or PRs

N/A

## Changes made

Added some inline comments. Please take a look for the details when you have time!

## Checklist

N/A

## Release notes

Disallows scan operations after a delete on the same record in Consensus Commit to prevent inconsistent behavior and align with the existing restriction on scans after writes.


core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -278,15 +263,43 @@ public void to(MutationComposer composer)
...
+  }
+
+  private boolean isWriteSetOrDeleteSetOverlappedWith(
+      Scan scan, Map<Snapshot.Key, TransactionResult> results) {
+    // Check if the scan is overlapped with the delete set
+    for (Map.Entry<Key, Delete> entry : deleteSet.entrySet()) {
+      if (results.containsKey(entry.getKey())) {
+        return true;
+      }
+    }
```
brfrn169, line 292: This is a main change for this PR. Added scan overlap check for deletes.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -151,23 +142,21 @@ private List<Result> scanInternal(Scan originalScan) throws CrudException {
...
+      Exception exception;
+      if (e.getCause() instanceof ExecutionException) {
+        exception = (ExecutionException) e.getCause();
+      } else {
+        exception = e;
       }
+      throw new CrudException(
+          CoreError.CONSENSUS_COMMIT_SCANNING_RECORDS_FROM_STORAGE_FAILED.buildMessage(),
+          exception,
+          snapshot.getId());
```
brfrn169, line 159: Added exception handling to account for cases where the scanner iterator may throw a `RuntimeException`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -151,23 +142,21 @@ private List<Result> scanInternal(Scan originalScan) throws CrudException {
...
         Snapshot.Key key = new Snapshot.Key(scan, r);
-
-        // We always update the read set to create before image by using the latest record (result)
-        // because another conflicting transaction might have updated the record after this
-        // transaction read it first.
-        snapshot.putIntoReadSet(key, Optional.of(result));
-
-        snapshot.getResult(key).ifPresent(value -> results.put(key, value));
+        processScanResult(key, scan, result);
+        results.put(key, result);
```
brfrn169, line 147: We no longer need to merge the scan result with the delete set since we now disallow scan after deletes.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -204,14 +203,7 @@ public Optional<Map<Snapshot.Key, TransactionResult>> getResults(Scan scan) thro
...
       return Optional.empty();
     }
-
-    Map<Key, TransactionResult> results = new LinkedHashMap<>();
-    for (Entry<Snapshot.Key, TransactionResult> entry : scanSet.get(scan).entrySet()) {
-      mergeResult(entry.getKey(), Optional.of(entry.getValue()))
-          .ifPresent(result -> results.put(entry.getKey(), result));
-    }
-
-    return Optional.of(results);
```
brfrn169, line 214: Ditto. We no longer need to merge the scan result with the delete set since we now disallow scan after deletes.

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -456,8 +456,12 @@ public enum CoreError implements ScalarDbError {
...
       "",
       ""),
-  CONSENSUS_COMMIT_READING_ALREADY_WRITTEN_DATA_NOT_ALLOWED(
-      Category.USER_ERROR, "0106", "Reading already-written data is not allowed", "", ""),
+  CONSENSUS_COMMIT_READING_ALREADY_WRITTEN_OR_DELETED_DATA_NOT_ALLOWED(
+      Category.USER_ERROR,
+      "0106",
+      "Reading already-written or already-deleted data is not allowed",
+      "",
+      ""),
```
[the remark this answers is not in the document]
josh-wong: LGTM!

josh-wong (approved):
LGTM! Thank you!🙇🏻‍♂️

core/src/main/java/com/scalar/db/common/error/CoreError.java
```diff
@@ -456,8 +456,12 @@ public enum CoreError implements ScalarDbError {
       "Getting data neither in the read set nor the delete set is not allowed",
       "",
       ""),
-  CONSENSUS_COMMIT_READING_ALREADY_WRITTEN_DATA_NOT_ALLOWED(
-      Category.USER_ERROR, "0106", "Reading already-written data is not allowed", "", ""),
+  CONSENSUS_COMMIT_READING_ALREADY_WRITTEN_OR_DELETED_DATA_NOT_ALLOWED(
+      Category.USER_ERROR,
+      "0106",
+      "Reading already-written or already-deleted data is not allowed",
```
komamitsu, line 462:
```suggestion
      "Reading data already-written or already-deleted by the same transaction is not allowed",
```
The original message might sound like it includes cases where another transaction updates or deletes the record. What do you think?

brfrn169: You are right. And I just realized we should say `Scanning` instead of `Reading` since we disallow only scanning already-written or already-deleted data.

```suggestion
      "Scanning data already-written or already-deleted by the same transaction is not allowed",
```

@josh-wong Could you please take a look at this from an English perspective?

brfrn169: @josh-wong I’ve updated similar error messages in the same manner in 9ab7b475b39fd4001f494483ed6a14789413aa36. Could you please double-check them when you have a chance? 🙇 Apologies for the repeated request.

josh-wong: No problem at all. The new changes LGTM!👍

komamitsu (approved):
LGTM! 👍

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
feeblefakie: Is it safe not to check the overlap between the scan range and deleteSet?
(so, deleting non-returned results is safe?)

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -278,15 +263,43 @@ public void to(MutationComposer composer)
...
+   * Because we support arbitrary conditions in the where clause of a scan (not only ScanAll, but
+   * also Scan and ScanWithIndex), we cannot determine whether the scan results will include a
+   * record whose key is the same as the key specified in the previous writes or deletes, without
+   * knowing the obtained keys in the actual scan. With this check, users can avoid seeing
+   * unexpected scan results that have not included previous writes or deletes yet.
+   *
+   * @param scan the scan to be verified
+   * @param results the results of the scan
+   */
+  public void verify(Scan scan, Map<Snapshot.Key, TransactionResult> results) {
```
feeblefakie, line 277:
```suggestion
  public void verifyNoOverlap(Scan scan, Map<Snapshot.Key, TransactionResult> results) {
```
I feel `verify` is probably too general.
Additionally, since the method checks for overlap and throws an exception if one is found, we should specify what is verified in the method name.

feeblefakie (commented):
Left a suggestion and a question. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: Thank you for asking the question! That’s a really good point — I actually had the same question myself.

My conclusion is that we don’t need to check for overlap between the scan range and the `deleteSet`, for the following reasons:

- We check for overlap between scans and updates because we don’t have a full snapshot of the transaction — only the `writeSet` and `deleteSet`.
- The results returned to the user should reflect a merge of the underlying database data and the updates made in the current transaction (i.e., the `writeSet` and `deleteSet`).
- However, in some cases — especially with scan operations — handling that correctly is quite complex, since we don't have a complete snapshot.
- That’s why we disallow certain combinations of operations to avoid introducing complicated logic.
- So we only need to check for overlap when the operation could actually change the scan result.
- With that in mind, regarding the question “Do we need to check for overlap between the scan range and the `deleteSet`?”:
  - If the actual scan result doesn’t include any records affected by the `deleteSet`, we can conclude that the delete operations don’t impact the scan result — and no further check is needed.

That said, I might be missing something — what do you think?

BTW, we check for overlap between the scan range and the `writeSet` because writing to records within that range could change the scan result.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: Fixed in dc79a3fde5a51bf733ad34ca16d120d70f56b720. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -278,15 +263,116 @@ public void to(MutationComposer composer)
...
   }
 
-  private boolean isWriteSetOverlappedWith(Scan scan) {
+  /**
+   * Verifies if this scan does not overlap previous writes or deletes using the actual scan result.
+   * Because we support arbitrary conditions in the where clause of a scan (not only ScanAll, but
+   * also Scan and ScanWithIndex), we cannot determine whether the scan results will include a
+   * record whose key is the same as the key specified in the previous writes or deletes, without
+   * knowing the obtained keys in the actual scan. With this check, users can avoid seeing
+   * unexpected scan results that have not included previous writes or deletes yet.
```
feeblefakie, line 272: How about the following comment? It's a bit long but will probably be helpful in the future.

---


Verifies that the scan does not overlap with previous write or delete operations to prevent incorrect results. For instance, consider the following records in the database, where X is the partition key and Y is the clustering key: R1(X=1, Y=1), R2(X=1, Y=2), R3(X=1, Y=3), and R4(X=1, Y=4). 

If a transaction performs a write (insert) operation for a new record R5(X=1, Y=5) followed by a scan operation with partition key X=1, the overlap check is crucial. Without this check, the scan might return {R1, R2, R3, R4}, which is incorrect. The correct result should include the newly inserted record, yielding {R1, R2, R3, R4, R5}.

Similarly, if a transaction deletes R1(X=1, Y=1) and then issues a scan operation with partition key X=1 and a limit of 3, the result without the overlap check would incorrectly return {R2, R3}. The correct result should be {R2, R3, R4}.

These examples illustrate basic cases, but the situation becomes more complex when considering the ordering of results and limit constraints. Therefore, ScalarDB currently avoids such overlaps instead of attempting to handle more intricate scenarios.

The use of scanned results, in addition to the specified scan range, is important because we support arbitrary conditions in the WHERE clause of scan operations using Scan, ScanAll, and ScanWithIndex. Specifically, without knowing the keys obtained in the actual scan, it is impossible to determine whether the scanned results include any records that match the keys from previous writes or deletes.

feeblefakie (approved):
Although I left a comment for the Javadoc comment, the code looks good to me.
Please check the comment and apply the necessary changes before merging the PR.

PR #2612, opened 2025-04-25
Relax visibility of Coordinator's methods for use by other ScalarDB components

komamitsu:
## Description

`com.scalar.db.transaction.consensuscommit.Coordinator` has some implementation that are not public or reusable. For example, the current implementation doesn't allow external ScalarDB components to directly look up a state by the parent ID for the coordinator group commit. It's sometimes useful for optimization (e.g., for caching coordinator state records using parent IDs as keys) in external components.

It's possible to inherit `Coordinator` class in those external components and implements the same logic by themselves, but it would result in duplicated logics and implementations over multiple components.

This PR makes those logics and implementations accessible in `Coordinator` class so that they can be reused when necessary.

## Related issues and/or PRs

None

## Changes made

- Extracted the implementation for getting a state by ID as `public Optional<Coordinator.State> getStateByIdOnly(String id)` to make it reusable 
- Made `Coordinator.State#getChildIds()` public

## Checklist

None

## Release notes

N/A


core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
```diff
@@ -106,7 +122,20 @@ Optional<Coordinator.State> getStateForGroupCommit(String fullId) throws Coordin
...
+   * this method is public, but it's only for internal use.
+   *
+   * @param id the ID of the coordinator state
+   * @return the coordinator state
+   * @throws CoordinatorException if the coordinator state cannot be retrieved
+   */
+  public Optional<Coordinator.State> getStateByIdOnly(String id) throws CoordinatorException {
+    Get get = createGetWith(id);
+    return get(get);
   }
```
brfrn169, line 139: I feel that the already exposed getState() method behaves the same way when the ID is not a full ID for the coordinator group commit.
So, we need this method purely for performance reasons, without checking the ID format. Is that correct?

brfrn169 (commented):
The change looks good to me. Left a question. Please take a look when you have time!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Coordinator.java
komamitsu: @brfrn169 Sorry for lack of explanation. The current implementation doesn't allow external components to directly look up a state by the parent ID for the coordinator group commit. It's sometimes useful for optimization (e.g., for caching coordinator state records with parent IDs as keys) in external components.

I realized the method names introduced in this PR were a bit unclear from the use case perspective. I updated them in https://github.com/scalar-labs/scalardb/pull/2612/commits/496c8f8ab487804bdfdf6c68fd38cf0435dce6f6. I hope it clarifies the code.

brfrn169 (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #2621, opened 2025-05-07
Improve validation logic in Consensus Commit to correctly handle scan with limit

brfrn169:
## Description

This PR improves the serializable validation logic for scan operations to fix an issue where a transaction consistently aborts if a record is inserted into the scan range after a scan with a limit has already been executed.

A test that reproduces this issue can be found here:  
https://github.com/scalar-labs/scalardb/pull/2621/files#diff-7d8b7163e906546d45d9a7b662f97eb671e39030217ab77ef22325b0f74c2b5eR3746-R3796

In this case, although there are no actual conflicts, the transaction still fails.

The current validation logic re-executes the same scan to verify that the scan result has not changed. Since this validation occurs after the prepare-record phase, the re-executed scan may include prepared records if the same transaction has performed operations that affect the scan result. Therefore, those records should be ignored to ensure correct validation.

Currently, there is an issue when a record is inserted into the scan range after a scan with a limit has already been executed.

Suppose we have the following data in this order: `record2`, `record3`, and `record4`.

1. Begin a transaction.  
2. Execute a scan operation with a limit of `2`, and retrieve `record2` and `record3`.  
3. Insert `record1` before `record2`.  
4. Commit the transaction.

In this case, the current validation logic re-executes the scan with a limit of `2`, and retrieves `record1` (a prepared record by the same transaction) and `record2`. Since the validation logic ignores prepared records, the scan result is considered to be `record2`. It then compares the original scan result (`record2` and `record3`) with the re-executed scan result (`record2`), detects a mismatch, and incorrectly aborts the transaction.

To address this issue, the scan should be re-executed without the limit while ignoring any prepared records. This is the core idea behind this change.

## Related issues and/or PRs

N/A

## Changes made

- Improved the serializable validation logic.

## Checklist

N/A

## Release notes

Fixed a bug in Consensus Commit with `SERIALIZABLE` isolation level where a transaction consistently aborts if a record was inserted into the scan range after executing a scan with a limit.


core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -556,11 +509,149 @@ void toSerializable(DistributedStorage storage)
...
+        // Get a scanner without the limit if the scan has a limit
+        scanner = storage.scan(Scan.newBuilder(scan).limit(0).build());
+      }
+
+      //
+      // Initialize the iterators for the scan results and the latest scan results
+      //
+      Optional<Result> latestResult = scanner.one();
+
+      Iterator<Entry<Key, TransactionResult>> resultsIterator = results.entrySet().iterator();
```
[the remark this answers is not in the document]
komamitsu: As we talked on https://github.com/scalar-labs/scalardb/pull/2601#discussion_r2053683357, I don't think the order of records returned from the underlying database is always guaranteed.

So, I'm wondering if the following case could occur, although they might be based on my misunderstanding:

- Potential false positive case (1)
  - At the timestamp (t0), there are 3 records that have different clustering keys (Cn) at the same partition key: C0, C1 and C2
  - Transaction T1 scans the partition key without `LIMIT` or `ORDER`. It fetches all the records in the physical order of C2, C1 and C0
  - T1 re-scans the partition key at the validation phase. It fetches all the records in the physical order of C0, C1 and C2 (the physical order can be changed)
  - The first records (C2 and C0) don't match and `throwExceptionDueToAntiDependency()` is called

- Potential false positive case (2)
  - At the timestamp (t0), there are 3 records that have different clustering keys (Cn) at the same partition key: C0, C1 and C2
  - Transaction T1 scans the partition key with `LIMIT 2` without `ORDER`. It fetches C1 and C2
  - T1 re-scans the partition key at the validation phase. It fetches C0, C1 and C2
  - C0 exists only in the re-scanning `scanner`. `throwExceptionDueToAntiDependency()` is called

- Potential false negative case
  - At the timestamp (t0), there are 3 records that have different clustering keys (Cn) at the same partition key: C0, C1 and C2
  - Transaction T1 scans the partition key with `LIMIT 2` without `ORDER`. It fetches C1 and C2
  - Another transaction T2 deletes C0 and commits
  - T1 re-scans the partition key at the validation phase. It fetches C1 and C2
  - C0 existed in the scan range and was removed during T1, but the validation phase is passed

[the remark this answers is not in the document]
brfrn169: In the ScalarDB data model, records within a partition are stored in clustering order. Therefore, for partition scans, the underlying storage should guarantee that records are returned in clustering order:

https://scalardb.scalar-labs.com/docs/latest/data-modeling/#clustering-key

So you’re talking about cross-partition scans or scans using secondary indexes, right?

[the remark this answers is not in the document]
brfrn169: > If we depends on the order of LinkedHashMap here, I think we should explicitly use LinkedHashMap for the argument type and the instant variable types instead of Map.

Fixed in 898c2f398dbb57c93ca35a7d2e0aa9d38c21eefa. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -556,11 +509,149 @@ void toSerializable(DistributedStorage storage)
...
+      } else {
+        resultsEntry = null;
+      }
+
+      // Compare the scan results with the latest scan results
+      while (latestResult.isPresent() && resultsEntry != null) {
+        TransactionResult latestTxResult = new TransactionResult(latestResult.get());
+        Key key = new Key(scan, latestTxResult);
+
+        if (latestTxResult.getId() != null && latestTxResult.getId().equals(id)) {
```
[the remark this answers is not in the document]
brfrn169: This occurs when importing an existing table, in which case all the metadata columns are null.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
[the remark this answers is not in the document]
komamitsu: > So you’re talking about cross-partition scans or scans using secondary indexes, right?

Ah, sorry for the confusion. I had the wrong idea on the order of clustering keys in the partition scan and should've not used it as an example. So, as you said, please read the example cases above by replacing with `scans using secondary indexes`.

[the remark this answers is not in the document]
komamitsu: > please read the example cases above by replacing with scans using secondary indexes.

FYI: Updated the example cases above.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -556,11 +511,149 @@ void toSerializable(DistributedStorage storage)
...
+      //
+      Optional<Result> latestResult = scanner.one();
+
+      Iterator<Entry<Key, TransactionResult>> resultsIterator = results.entrySet().iterator();
+      Entry<Key, TransactionResult> resultsEntry;
+      if (resultsIterator.hasNext()) {
+        resultsEntry = resultsIterator.next();
+      } else {
+        resultsEntry = null;
+      }
```
Torch3333, line 562: [minor]
This block of code, which retrieves the next iterator value, is present several times in this method. To slightly improve readability, what about replacing it with a utility method, such as Guava [Iterators.getNext()](https://guava.dev/releases/17.0/api/docs/com/google/common/collect/Iterators.html#getNext(java.util.Iterator,%20T))
 ```suggestion
      Entry<Key, TransactionResult> resultsEntry = Iterators.getNext(resultsIterator, null);
```

Comment (brfrn169) [reply core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java]: Thanks! Fixed in 8bab0b5e2639601a0e6f8e0b481d88bc25cc6814.

**Comment by komamitsu** [comment]
The change itself looks good to me 👍

But when I read the following [comment](https://github.com/scalar-labs/scalardb/pull/2621/files#r2081089703), I only thought of RDBMS.

> In rare situations where the same scan returns results in a different order, the transaction will be rejected during validation. But since this is not common, I think that’s acceptable.

Honestly speaking, in case of NoSQL, I'm not sure how rare the same scan returns records in different order. I checked only Cassandra using 5 nodes and it seemed rare (👍), but it would depend on the implementation.

Do you think it's really okay to think NoSQL basically returns records in the same order even from different nodes? My concern is the new validation might often fail if the underlying NoSQL collects and returns scan results as follows:

```
result = []
target_nodes(partition_key).async_each { node ->
  result.add_all(node.scan(query))
}
return result
```

**Comment by komamitsu** [comment]
@brfrn169 Thanks for the answer. Let me clarify a few points.

1.

>  I think the only safe way to validate is to do a scan without a limit, get all the records, and then check.

The validation seems to scan records without limits, so you mentioned the original scan?

https://github.com/scalar-labs/scalardb/blob/8bab0b5e2639601a0e6f8e0b481d88bc25cc6814/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java#L549

2.

> For example, if we have data like record1 to record3 and we do a scan with limit 2, the result could be:
>
> - record1 and record2
> - record1 and record3
> - record2 and record3

My understanding is this issue could occur only with LIMIT and without ORDER. If so, how about either of the following options when LIMIT is specified and ORDER isn't specified?

A. Original scan and validation scan don't use LIMIT to fetch all target records (as you suggested)
B. Original scan and validation scan implicitly use ORDER. With this option, in the above example case assuming record1 < record2 < record3, the original and validation scans always read record1 and record2

**Comment by komamitsu** [comment]
@brfrn169 Thanks for the online discussion. Let me wrap up my questions and so on.

> The validation seems to scan records without limits, so you mentioned the original scan?

The original scan also needs scan without limits. As for the validation scan, it stops the iteration when reaching the limit. So it may be more efficient, although whether additional resource consumption occurs in the underlying database depends on its design and implementation.

> how about either of the following options when LIMIT is specified and ORDER isn't specified?

I had the wrong idea that we can use ORDER with the secondary indexes...

Regarding the potential random order issue, it could occur with the current implementation. So, we can treat it separately from this PR.

**Review by komamitsu — APPROVED** [review]
LGTM, thank you!

As I mentioned, my potential concern is that the validation scan without limits might affect the resource consumption and/or performance inside the underlying database, although it would depend on its design and implementation. Please go ahead if it doesn't matter for now.

File: core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java [hunk 4]
```diff
@@ -556,11 +512,135 @@ void toSerializable(DistributedStorage storage)
...
+        latestResult = scanner.one();
+        resultsEntry = Iterators.getNext(resultsIterator, null);
+      }
+
+      if (resultsEntry != null) {
+        // Some of the records of the scan results are deleted by another transaction
+        throwExceptionDueToAntiDependency();
+      }
+
+      if (scan.getLimit() != 0 && results.size() == scan.getLimit()) {
```
feeblefakie, line 608: I probably misunderstand something. But this always seems to hold when `scan.getLimit() != 0`, which is the first condition, and the latter condition is not changed after the above processing, so I'm not sure what this is checking. Can you elaborate on this?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -556,11 +512,135 @@ void toSerializable(DistributedStorage storage)
...
+      Iterator<Entry<Key, TransactionResult>> resultsIterator = results.entrySet().iterator();
+      Entry<Key, TransactionResult> resultsEntry = Iterators.getNext(resultsIterator, null);
+
+      // Compare the scan results with the latest scan results
+      while (latestResult.isPresent() && resultsEntry != null) {
+        TransactionResult latestTxResult = new TransactionResult(latestResult.get());
+        Key key = new Key(scan, latestTxResult);
+
+        if (latestTxResult.getId() != null && latestTxResult.getId().equals(id)) {
+          // The record is inserted/deleted/updated by this transaction
```
feeblefakie, line 566: I might miss something, but the result has been retrieved, so it cannot be deleted by this transaction?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -556,11 +512,135 @@ void toSerializable(DistributedStorage storage)
...
+        throwExceptionDueToAntiDependency();
+      }
+
+      if (scan.getLimit() != 0 && results.size() == scan.getLimit()) {
+        // We’ve already checked up to the limit, so no further checks are needed
+        return;
+      }
+
+      // Check if there are any remaining records in the latest scan results
+      while (latestResult.isPresent()) {
```
feeblefakie, line 614: This while block can be removed without violating the correctness?
Or, there might be a case where it returns wrong results or causes anomalies without this?

feeblefakie (commented):
Overall, looking good.
Left several questions to better understand the logic. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
feeblefakie: Thank you for the explanation offline! I understand the logic.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
feeblefakie: Thank you for the explanation offline! I understand the logic.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
feeblefakie: Thank you for the explanation offline! I understand the logic.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -556,11 +512,135 @@ void toSerializable(DistributedStorage storage)
...
+   * same order, provided that the underlying data remains unchanged.
+   *
+   * @param storage a distributed storage
+   * @param scan the scan to be validated
+   * @param results the results of the scan
+   * @throws ExecutionException if a storage operation fails
+   * @throws ValidationConflictException if the scan results are changed by another transaction
+   */
+  private void validateScanResults(
+      DistributedStorage storage, Scan scan, LinkedHashMap<Key, TransactionResult> results)
```
feeblefakie, line 536: How about using `original` and `latest` to clearly differentiate the first scan results before the read validation and the second scan results during the read validation?
`original` and `rescanned` can be another option.

brfrn169: Fixed in bcfcf068e78243934344f94ba6cbb93e45e00c57. Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #2696, opened 2025-05-26
Add util method to check type of ScalarDB connection

inv-jishnu:
## Description

This PR adds a basic database configuration check to see if scalardb cluster is set via the configuration properties.

## Related issues and/or PRs

NA

## Changes made

I have added a util method which checks `transaction manager` value and returns if the value is set to `cluster` or not

## Checklist

NA

## Release notes

Added a util method to check if cluster is enabled in scalardb configuration file

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
@@ -28,6 +28,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


Torch3333 (approved):
LGTM, thank you!

ypeckstadt (approved):
LGTM. Thank you.

brfrn169 (commented):
The update looks good to me, but why do we need it?

inv-jishnu:
@brfrn169 san,

Currently, the data loader CLI always creates an instance for DistributedStorage even when we don't use it when the user is running the CLI in transaction mode. This is fine for direct database access via the ScalarDB API as it does not cause any problems.

However, when connecting to a ScalarDB cluster, this causes problems as the cluster does not support the usage of DistributedStorage at all. Even just instantiating it already causes the data loader CLI to crash.

So the first step for us to handle this problem and add proper ScalarDB cluster support is being able to check whether we are connecting to a cluster or not.   Because the codebase for scalardb-dataloader-cli and scalardb-cluster-dataloader-cli is going to be the same, I like to add it here.

ypeckstadt:
@komamitsu Actually while the schema loader does have the folder in the cluster repo, it's just Gradle replacing the dependencies for the ScalarDB API. The code is the same. This we will do for the data loader as well. (add the same in the cluster repo)

Basically the problem is the following:

The data loader supports storage and transactional mode for direct to database access. We will offer both for when connecting to a ScalarDB cluster.  For cluster transactional mode, it's straightforward as all we need to do is replace the API dependencies like the schema loader is doing.

However, storage mode is a different matter. ScalarDB cluster only supports this when the cluster is configured for single transaction mode + when the SingleTransactionCrudManager is used from the API.   Currently the data loader uses either DistributedStorage or DistributedTransaction for the data retrieval or manipulation. For cluster - storage(single transaction) mode we need to use the other manager.

Ideally, like you said, it would be better to find a way that we can add that code difference in the cluster repo data loader folder but I wonder if that is going to be possible without a lot of the code having to be in both repositories and hence being duplicated.

Therefore, we thought to just add that simple check in the main repo to avoid having to split the code. The check being `if connected to a cluster and the data-loader was started with --mode storage use the SingleCrudOperationTransactionManager instead of DistributedStorage.

If its preferred to exactly handle it via the 2 repo's we can try to find a way for that. For example in the main code create some kind of wrapper class around the DistributedStorage class and replace that one in the cluster repo code with one that uses the SingleCrudOperationTransactionManager.

brfrn169:
@ypeckstadt (CC: @komamitsu)
Thanks for the explanation! Let me first discuss it with @feeblefakie.

ypeckstadt:
As discussed today we will not be adding cluster support in the 3.16 release and will handle this problem later in 3.17. Probably this kind of check will not be put here.

PR #2698, opened 2025-05-26
Add getScanner() method to transaction interfaces

brfrn169:
## Description

This PR adds the getScanner() method to the transaction interfaces.

We are working on this feature in the `add-scanner-api-to-transaction-abstraction` feature branch, as this change may cause compile errors in dependent projects.

## Related issues and/or PRs

N/A

## Changes made

- 

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

> Provide any additional information or notes that may be relevant to the reviewers or stakeholders.

## Release notes

N/A

PR #2711, opened 2025-05-29
Implement scanner API for Consensus Commit

brfrn169:
## Description

This PR adds a scanner API implementation for Consensus Commit transactions.

Note that we are working on this feature in the `add-scanner-api-to-transaction-abstraction` feature branch.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/2698

## Changes made

- Added a scanner API implementation for Consensus Commit transactions.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
@@ -24,8 +24,10 @@ import com.scalar.db.exception.transaction.UnknownTransactionStatusException;
 import com.scalar.db.exception.transaction.UnsatisfiedConditionException;
 import com.scalar.db.util.ScalarDbUtils;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 import org.slf4j.Logger;
@@ -98,7 +100,41 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
 
   @Override
   public Scanner getScanner(Scan scan) throws CrudException {
-    throw new UnsupportedOperationException("Implement later");
+    scan = copyAndSetTargetToIfNot(scan);
+    Scanner scanner = crud.getScanner(scan);
+
+    return new Scanner() {
+      @Override
+      public Optional<Result> one() throws CrudException {
+        try {
+          return scanner.one();
+        } catch (UncommittedRecordException e) {
+          lazyRecovery(e);
+          throw e;
+        }
+      }
+
+      @Override
+      public List<Result> all() throws CrudException {
+        try {
+          return scanner.all();
+        } catch (UncommittedRecordException e) {
+          lazyRecovery(e);
+          throw e;
+        }
+      }
+
+      @Override
+      public void close() throws CrudException {
+        scanner.close();
+      }
+
+      @Nonnull
+      @Override
+      public Iterator<Result> iterator() {
+        return scanner.iterator();
+      }
+    };
   }
 
   /** @deprecated As of release 3.13.0. Will be removed in release 5.0.0. */
@@ -213,6 +249,10 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
 
   @Override
   public void commit() throws CommitException, UnknownTransactionStatusException {
+    if (!crud.areAllScannersClosed()) {
+      throw new IllegalStateException(CoreError.CONSENSUS_COMMIT_SCANNER_NOT_CLOSED.buildMessage());
+    }
+
     // Execute implicit pre-read
     try {
       crud.readIfImplicitPreReadEnabled();
@@ -234,6 +274,12 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
 
   @Override
   public void rollback() {
+    try {
+      crud.closeScanners();
+    } catch (CrudException e) {
+      logger.warn("Failed to close the scanner", e);
+    }
+
     if (groupCommitter != null) {
       groupCommitter.remove(crud.getSnapshot().getId());
     }
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
@@ -488,7 +488,7 @@ public class CrudHandler {
         // into the scan set
         snapshot.putIntoScanSet(scan, results);
       } else {
-        // If the scan is not fully scanned, put the results into the scanner set
+        // If the scanner is not fully scanned, put the results into the scanner set
         snapshot.putIntoScannerSet(scan, results);
       }
 
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
@@ -31,6 +31,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
@@ -410,9 +411,8 @@ public class CrudHandler {
     private final Scanner scanner;
 
     private final LinkedHashMap<Snapshot.Key, TransactionResult> results = new LinkedHashMap<>();
-    private boolean fullyScanned;
-    private boolean failed;
-    private boolean closed;
+    private AtomicBoolean fullyScanned = new AtomicBoolean();
+    private AtomicBoolean closed = new AtomicBoolean();
 
     public ConsensusCommitStorageScanner(Scan scan, List<String> originalProjections)
         throws CrudException {
@@ -427,7 +427,7 @@ public class CrudHandler {
         Optional<Result> r = scanner.one();
 
         if (!r.isPresent()) {
-          fullyScanned = true;
+          fullyScanned.set(true);
           return Optional.empty();
         }
 
@@ -440,13 +440,13 @@ public class CrudHandler {
         return Optional.of(
             new FilteredResult(result, originalProjections, metadata, isIncludeMetadataEnabled));
       } catch (ExecutionException e) {
-        failed = true;
+        closeScanner();
         throw new CrudException(
             CoreError.CONSENSUS_COMMIT_SCANNING_RECORDS_FROM_STORAGE_FAILED.buildMessage(),
             e,
             snapshot.getId());
       } catch (CrudException e) {
-        failed = true;
+        closeScanner();
         throw e;
       }
     }
@@ -468,22 +468,13 @@ public class CrudHandler {
 
     @Override
     public void close() {
-      if (closed) {
-        return;
-      }
-      closed = true;
-
-      try {
-        scanner.close();
-      } catch (IOException e) {
-        logger.warn("Failed to close the scanner", e);
-      }
-
-      if (failed) {
+      if (closed.get()) {
         return;
       }
 
-      if (fullyScanned) {
+      closeScanner();
+
+      if (fullyScanned.get()) {
         // If the scanner is fully scanned, we can treat it as a normal scan, and put the results
         // into the scan set
         snapshot.putIntoScanSet(scan, results);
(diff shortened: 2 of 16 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -213,6 +249,10 @@ public void mutate(List<? extends Mutation> mutations) throws CrudException {
 
   @Override
   public void commit() throws CommitException, UnknownTransactionStatusException {
+    if (!crud.areAllScannersClosed()) {
```
brfrn169, line 252: Check if all the scanners are closed before committing a transaction.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -211,6 +216,37 @@ private void processScanResult(Snapshot.Key key, Scan scan, TransactionResult re
...
+    ConsensusCommitScanner scanner;
+
+    Optional<LinkedHashMap<Snapshot.Key, TransactionResult>> resultsInSnapshot =
+        snapshot.getResults(scan);
+    if (resultsInSnapshot.isPresent()) {
+      scanner =
+          new ConsensusCommitSnapshotScanner(scan, originalProjections, resultsInSnapshot.get());
+    } else {
+      scanner = new ConsensusCommitStorageScanner(scan, originalProjections);
+    }
```
brfrn169, line 232: If the scan result of the same scan is in the scanSet in the snapshot, we use it. If not, we scan the storage directly.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -360,4 +396,170 @@ private TableMetadata getTableMetadata(Operation operation) throws CrudException
   public Snapshot getSnapshot() {
     return snapshot;
   }
+
+  private interface ConsensusCommitScanner extends TransactionCrudOperable.Scanner {
+    boolean isClosed();
+  }
+
+  @NotThreadSafe
+  private class ConsensusCommitStorageScanner extends AbstractTransactionCrudOperableScanner
```
brfrn169, line 405: The scanner implementation for scanning the storage.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -360,4 +396,170 @@ private TableMetadata getTableMetadata(Operation operation) throws CrudException
...
+    }
+
+    @Override
+    public boolean isClosed() {
+      return closed;
+    }
+  }
+
+  @NotThreadSafe
+  private class ConsensusCommitSnapshotScanner extends AbstractTransactionCrudOperableScanner
```
brfrn169, line 505: The scanner implementation for scanning the result in the snapshot.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -360,4 +396,170 @@ private TableMetadata getTableMetadata(Operation operation) throws CrudException
...
+      }
+
+      if (fullyScanned) {
+        // If the scanner is fully scanned, we can treat it as a normal scan, and put the results
+        // into the scan set
+        snapshot.putIntoScanSet(scan, results);
+      } else {
+        // If the scanner is not fully scanned, put the results into the scanner set
+        snapshot.putIntoScannerSet(scan, results);
+      }
```
brfrn169, line 493: This is a tricky part. The behavior depends on whether the storage scanner has been fully scanned.

If the storage scanner has been fully scanned, we can treat it as a normal scan and put the result into the scan set, just like a regular scan.

If the storage scanner has not been fully scanned, we put the results into the scanner set.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -64,6 +64,9 @@ public class Snapshot {
   private final Map<Key, Put> writeSet;
   private final Map<Key, Delete> deleteSet;
 
+  // The scanner set used to store information about scanners that are not fully scanned
+  private final List<ScannerInfo> scannerSet;
```
brfrn169, line 68: Added scannerSet that's used to store information about scanners that are not fully scanned.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -485,7 +495,12 @@ void toSerializable(DistributedStorage storage)
...
     // Scan set is re-validated to check if there is no anti-dependency
     for (Map.Entry<Scan, LinkedHashMap<Key, TransactionResult>> entry : scanSet.entrySet()) {
-      tasks.add(() -> validateScanResults(storage, entry.getKey(), entry.getValue()));
+      tasks.add(() -> validateScanResults(storage, entry.getKey(), entry.getValue(), false));
+    }
+
+    // Scanner set is re-validated to check if there is no anti-dependency
+    for (ScannerInfo scannerInfo : scannerSet) {
+      tasks.add(() -> validateScanResults(storage, scannerInfo.scan, scannerInfo.results, true));
     }
```
brfrn169, line 504: Add serializable validation for the scanner set.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -604,6 +624,11 @@ private void validateScanResults(
         return;
       }
 
+      if (notFullyScannedScanner) {
+        // If the scanner is not fully scanned, no further checks are needed
+        return;
+      }
```
brfrn169, line 630: If the scanner has been not fully scanned, we don't need a further check. This is similar to the limit case:
https://github.com/scalar-labs/scalardb/blob/1717cfb7842cf674580490aad4b0c445f2211377/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java#L622-L625

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: > The reason why we need to differently handle the not-fully-scanned scan is that there can be remaining records?

Yes, because of that, the not-fully-scanned scan should need special handling in the serializable validation, similar to scan with limit.

> If so, how about reading the remaining records in this method and adding them to results so that we can treat the scan as a fully scanned one?

In that case, if the remaining records are large, we would consume a lot of memory. One of the advantages of the scanner API is that it only consumes memory as records are scanned. So, we should avoid doing that.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -98,7 +100,41 @@ public List<Result> scan(Scan scan) throws CrudException {
...
+    Scanner scanner = crud.getScanner(scan);
+
+    return new Scanner() {
+      @Override
+      public Optional<Result> one() throws CrudException {
+        try {
+          return scanner.one();
+        } catch (UncommittedRecordException e) {
+          lazyRecovery(e);
+          throw e;
```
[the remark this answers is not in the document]
brfrn169: We don't close the scanner here, but we close here:
https://github.com/scalar-labs/scalardb/pull/2711/files#diff-acc93f1366b03168017001646c7217f7dd1f3c7cbe8a41b6a9a8484dc5ef0905R406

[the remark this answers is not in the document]
feeblefakie: The reason is probably `ConsensusCommitManager` commits after one scan, whereas `ConsensusCommit` commits after doing several operations?

feeblefakie (commented):
Overall, looking good!
I will take another look.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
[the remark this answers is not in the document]
komamitsu: (Oops, sorry I thought here was in `catch (CrudException e) { ... }` block).

I just thought we can immediately close the failed scanner as well as `ConsensusCommitManager.getScanner()`, for consistency and a bit faster resource release. It's just out of curiosity.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
komamitsu: I see. That sounds reasonable.

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -360,4 +396,170 @@ private TableMetadata getTableMetadata(Operation operation) throws CrudException
...
+            snapshot.getId());
+      } catch (CrudException e) {
+        failed = true;
+        throw e;
+      }
+    }
+
+    @Override
+    public List<Result> all() throws CrudException {
+      List<Result> results = new ArrayList<>();
```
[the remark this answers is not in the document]
brfrn169: You’re referring to the following class variable `results`, right?
https://github.com/scalar-labs/scalardb/pull/2711/files#diff-acc93f1366b03168017001646c7217f7dd1f3c7cbe8a41b6a9a8484dc5ef0905R413

If so, the class variable `results` contains all the fetched results, but the `all()` method is supposed to return only the remaining results.

For example, suppose the scan result is `[1, 2, 3, 4, 5]`. In the following case, the `all()` method should return `[3, 4, 5]`:

```java
Result result1 = scanner.one(); // returns 1
Result result2 = scanner.one(); // returns 2

List<Result> results = scanner.all(); // returns [3, 4, 5]
```

That’s why we can’t simply return something like `ImmutableList.copyOf(this.results)` in the `all()` method. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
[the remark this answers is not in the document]
brfrn169: Ah, sorry for the lack of explanation.

This anonymous `Scanner` class wraps the scanner instance returned by `CrudHandler` here:
https://github.com/scalar-labs/scalardb/pull/2711/files#diff-ec7fb790c921b21b0690f23d557c721585b0c83ef32b729e7dada7009d6142d0R104

The `CrudHandler.getScanner()` method returns an instance of `ConsensusCommitStorageScanner` if the scan result is not present in the snapshot.

The `ConsensusCommitStorageScanner.one()` method immediately closes the underlying scanner when a failure occurs:
https://github.com/scalar-labs/scalardb/pull/2711/files#diff-acc93f1366b03168017001646c7217f7dd1f3c7cbe8a41b6a9a8484dc5ef0905R424-R452

So, the behavior is consistent with `ConsensusCommitManager.getScanner()`.

By the way, the anonymous `Scanner` class simply executes lazy recovery when an `UncommittedRecordException` is caught.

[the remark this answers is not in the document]
komamitsu: Thanks for the explanation! Sounds good :100:

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -360,4 +397,169 @@ private TableMetadata getTableMetadata(Operation operation) throws CrudException
...
+  private class ConsensusCommitStorageScanner extends AbstractTransactionCrudOperableScanner
+      implements ConsensusCommitScanner {
+
+    private final Scan scan;
+    private final List<String> originalProjections;
+    private final Scanner scanner;
+
+    private final LinkedHashMap<Snapshot.Key, TransactionResult> results = new LinkedHashMap<>();
+    private AtomicBoolean fullyScanned = new AtomicBoolean();
+    private AtomicBoolean closed = new AtomicBoolean();
```
Torch3333:
```suggestion
    private final AtomicBoolean fullyScanned = new AtomicBoolean();
    private final AtomicBoolean closed = new AtomicBoolean();
```

Torch3333 (approved):
I left a minor comment, PTAL. 
Besides this, LGTM, thank you.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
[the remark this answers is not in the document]
feeblefakie: Oh, OK, I misunderstood the behavior. 
I don't think we can change it for backward compatibility, but the method name might be confusing since `all()` sounds like returning all results, not remaining. (it's probably my mistake though)

PR #2728, opened 2025-06-03
Move parallelism level for the importing process from the record level to the data chunk level

thongdk8:
## Description

The current import process parallelizes at the record level, which creates inefficiencies in thread pool utilization. The main thread must wait for all record processing threads to complete before moving to the next data chunk, causing the thread pool to be underutilized when waiting for the slowest batch to finish. Also the current implementation is sensitive to the params of `tx_size`, `chunk size` and number of `threads`, for example, if `tx_size=10`, `chunk_size=30`, and `threads=4` then we dont utilized all the threads as the chunk split into 3 tasks and run it in the pool, it also happend if the the `chunk_size` is not diviable by `tx_size * threads`

This PR moves parallelism from the record level to the data chunk level, allowing continuous processing of data chunks through the same thread pool using a blocking queue. Improved overall import performance. PTAL. Thank you.

Before:
- Data chunks are split into multiple record batches
- Thread pool processes record batches in parallel
- Main thread blocks until all record batches complete
- Thread pool sits idle while waiting for the slowest batch

After:
- Thread pool processes entire data chunks in parallel
- Continuous chunk processing via blocking queue
- Improved thread pool utilization
- Eliminated synchronization bottlenecks between chunks
- Reduced sensitivity to parameter constraints

## Related issues and/or PRs

NA

## Changes made

- Move parallelism level for the importing process from the record level to the data chunk level

## Checklist

NA

## Release notes

NA


data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataimport/processor/ImportProcessor.java
```diff
@@ -62,41 +62,77 @@ public abstract class ImportProcessor {
...
+    Phaser phaser = new Phaser(1);
+
     try {
       CompletableFuture<Void> readerFuture =
           CompletableFuture.runAsync(
-              () -> readDataChunks(reader, dataChunkSize, dataChunkQueue), dataChunkExecutor);
+              () -> readDataChunks(reader, dataChunkSize, dataChunkQueue), dataChunkReaderExecutor);
 
       while (!(dataChunkQueue.isEmpty() && readerFuture.isDone())) {
         ImportDataChunk dataChunk = dataChunkQueue.poll(100, TimeUnit.MILLISECONDS);
```
komamitsu, line 82: This is not related to this PR, but why we need to wake up every 100 ms? I just think longer timeout or using `take()` might be better.

thongdk8:
@komamitsu Thank you for reviewing. Regarding the case you mentioned, I agree that in the mentioned case the behavior will be like you mentioned, but it is a rare case yes, and it happens only when the chunk size is large and close to the number of total records, and for the last batch of chunks processed by the pool. So I think we can ignore such concern as you said.

data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataimport/processor/ImportProcessor.java
thongdk8: The queue is buffered, so the timeout of 100ms would be fine I think.
We can replace it with `take`, but it seems a bit unsafe since the `take` will block the loop (deadlock) and you would never get a chance to recheck the `while` conditions unless a new item arrives. For example, if the queue is empty but the reader is on going for exiting its job (the reader future is going to be done), although this case is hard to be happenned, but still has a posibility. WDYT?

komamitsu: Yeah, you're correct. I didn't consider the exit condition. Let's keep it as is.

data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataimport/processor/ImportProcessor.java
```diff
@@ -440,32 +456,14 @@ private ImportDataChunkStatus processDataChunkWithoutTransactions(ImportDataChun
...
-    } finally {
-      recordExecutor.shutdown();
+
+    for (ImportRow importRow : dataChunk.getSourceData()) {
+      ImportTaskResult result = processStorageRecord(dataChunk.getDataChunkId(), importRow);
+      boolean allSaved =
+          result.getTargets().stream()
+              .allMatch(t -> t.getStatus().equals(ImportTargetResultStatus.SAVED));
+      if (allSaved) successCount.incrementAndGet();
+      else failureCount.incrementAndGet();
```
brfrn169:
```suggestion
      if (allSaved) {
        successCount.incrementAndGet();
      } else {
        failureCount.incrementAndGet();
      }
```

data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataimport/processor/ImportProcessor.java
```diff
@@ -62,41 +62,77 @@ public abstract class ImportProcessor {
...
   public void process(int dataChunkSize, int transactionBatchSize, BufferedReader reader) {
-    ExecutorService dataChunkExecutor = Executors.newSingleThreadExecutor();
+    ExecutorService dataChunkReaderExecutor = Executors.newSingleThreadExecutor();
+    ExecutorService dataChunkProcessorExecutor =
+        Executors.newFixedThreadPool(params.getImportOptions().getMaxThreads());
     BlockingQueue<ImportDataChunk> dataChunkQueue =
         new LinkedBlockingQueue<>(params.getImportOptions().getDataChunkQueueSize());
 
+    // Semaphore controls concurrent task submissions, small buffer to be two times of threads
+    Semaphore taskSemaphore = new Semaphore(params.getImportOptions().getMaxThreads() * 2);
```
brfrn169, line 72: Instead of using a `Semaphore`, I think we can use an `ExecutorService` with a bounded blocking queue, as shown below:

```java
    ExecutorService dataChunkProcessorExecutor =
        new ThreadPoolExecutor(
            params.getImportOptions().getMaxThreads(),
            params.getImportOptions().getMaxThreads(),
            0L,
            TimeUnit.MILLISECONDS,
            new ArrayBlockingQueue<>(params.getImportOptions().getMaxThreads() * 2),
            Executors.defaultThreadFactory(),
            new ThreadPoolExecutor.CallerRunsPolicy());
```

What do you think?

thongdk8: Ah yes, we can use bounded blocking queue like you mentioned, but as it is using `CallerRunsPolicy` so when the thread pool is completely saturated (all threads busy + queue full), `CallerRunsPolicy` runs the task in the `calling thread` (main thread in our case) instead of the pool threads, that is not what we expect as all the processing tasks are executed in the pool is more natural I think.
There are other policies as well, but seems there is no suitable one unless we create a custom one, so I think using semaphore with a fixed pool here is ok. WDYT?

brfrn169 (approved):
LGTM! Thank you!

ypeckstadt (approved):
LGTM. Thank you.

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #2740, opened 2025-06-05
Add transaction mode in export on dataloader core and CLI

inv-jishnu:
## Description

In this PR I have added transaction mode in export (to use transaction for data export) on dataloader core and added necessary changes in CLI to use it

## Related issues and/or PRs
NA

## Changes made

On core
* Added  changes to support transaction for export in TableMetadataService, ExportManager and ScalarDBDao
* Added/Updated unit tests.
On CLI
* Added a parameter to input Scalardb mode.
* Added checks to use transaction based on input for metadata retrieval and export.

## Checklist

NA

## Release notes

Added transaction mode in data export on data loader core and CLI

### DIFF
--- a/data-loader/cli/src/main/java/com/scalar/db/dataloader/cli/command/dataexport/ExportCommand.java
+++ b/data-loader/cli/src/main/java/com/scalar/db/dataloader/cli/command/dataexport/ExportCommand.java
@@ -4,6 +4,7 @@ import static java.nio.file.StandardOpenOption.APPEND;
 import static java.nio.file.StandardOpenOption.CREATE;
 
 import com.scalar.db.api.DistributedStorage;
+import com.scalar.db.api.DistributedTransactionManager;
 import com.scalar.db.api.TableMetadata;
 import com.scalar.db.common.error.CoreError;
 import com.scalar.db.dataloader.cli.exception.DirectoryValidationException;
@@ -12,6 +13,7 @@ import com.scalar.db.dataloader.cli.util.FileUtils;
 import com.scalar.db.dataloader.cli.util.InvalidFilePathException;
 import com.scalar.db.dataloader.core.ColumnKeyValue;
 import com.scalar.db.dataloader.core.FileFormat;
+import com.scalar.db.dataloader.core.ScalarDbMode;
 import com.scalar.db.dataloader.core.ScanRange;
 import com.scalar.db.dataloader.core.dataexport.CsvExportManager;
 import com.scalar.db.dataloader.core.dataexport.ExportManager;
@@ -27,7 +29,9 @@ import com.scalar.db.dataloader.core.tablemetadata.TableMetadataService;
 import com.scalar.db.dataloader.core.util.KeyUtils;
 import com.scalar.db.io.Key;
 import com.scalar.db.service.StorageFactory;
+import com.scalar.db.service.TransactionFactory;
 import java.io.BufferedWriter;
+import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Paths;
@@ -56,15 +60,14 @@ public class ExportCommand extends ExportCommandOptions implements Callable<Inte
     try {
       validateOutputDirectory();
       FileUtils.validateFilePath(scalarDbPropertiesFilePath);
-
-      StorageFactory storageFactory = StorageFactory.create(scalarDbPropertiesFilePath);
-      TableMetadataService metaDataService =
-          new TableMetadataService(storageFactory.getStorageAdmin());
+      TableMetadataService tableMetadataService =
+          createTableMetadataService(scalarDbMode, scalarDbPropertiesFilePath);
       ScalarDbDao scalarDbDao = new ScalarDbDao();
 
-      ExportManager exportManager = createExportManager(storageFactory, scalarDbDao, outputFormat);
+      ExportManager exportManager =
+          createExportManager(scalarDbMode, scalarDbDao, outputFormat, scalarDbPropertiesFilePath);
 
-      TableMetadata tableMetadata = metaDataService.getTableMetadata(namespace, table);
+      TableMetadata tableMetadata = tableMetadataService.getTableMetadata(namespace, table);
 
       Key partitionKey =
           partitionKeyValue != null ? getKeysFromList(partitionKeyValue, tableMetadata) : null;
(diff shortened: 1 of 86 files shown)


data-loader/core/src/main/java/com/scalar/db/dataloader/core/tablemetadata/TableMetadataService.java
```diff
@@ -18,7 +19,17 @@
 public class TableMetadataService {
 
   private final DistributedStorageAdmin storageAdmin;
+  private final DistributedTransactionAdmin transactionAdmin;
```
komamitsu: This class seems the union of `TableMetadataService` for storage and for transaction with some internal `if` branches. How about making `TableMetadataService` abstract and creating 2 child classes that return the TableMetadata for storage or transaction from something like `getTableMetadataInternal()`?

komamitsu:
@inv-jishnu I think there are many `if` conditions to check whether it's storage or transaction. This is just an idea, but `ScalarDbDao` might be a good wrapper of the storage and transaction. How about making `ScalarDbDao` an interface (or abstract class) and adding concrete classes something like `ScalarDbStorageDao` and `ScalarDbTransactionDao` which contain a storage or a transaction, respectively, instead of accepting either of storage or transaction at method level (e.g., `get()`) ? I'm not sure it's a reasonable suggestion in terms of release schedule and so on, though.

data-loader/core/src/main/java/com/scalar/db/dataloader/core/tablemetadata/TableMetadataService.java
inv-jishnu: @komamitsu san, 
I have split up table metadata service as suggested.
Thank you.

inv-jishnu:
> @inv-jishnu I think there are many `if` conditions to check whether it's storage or transaction. This is just an idea, but `ScalarDbDao` might be a good wrapper of the storage and transaction. How about making `ScalarDbDao` an interface (or abstract class) and adding concrete classes something like `ScalarDbStorageDao` and `ScalarDbTransactionDao` which contain a storage or a transaction, respectively, instead of accepting either of storage or transaction at method level (e.g., `get()`) ? I'm not sure it's a reasonable suggestion in terms of release schedule and so on, though.

@komamitsu san,
Thank you for the suggestion.
I will work on this in another PR (as you mentioned about the release schedule). I will add the necessary changes in that one.

data-loader/cli/src/main/java/com/scalar/db/dataloader/cli/command/dataexport/ExportCommandOptions.java
```diff
@@ -144,4 +145,11 @@ public class ExportCommandOptions {
       description = "Size of the data chunk to process in a single task (default: 200)",
       defaultValue = "200")
   protected int dataChunkSize;
+
+  @CommandLine.Option(
+      names = {"--mode", "-sm"},
```
brfrn169: Why `-sm`?

data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataimport/dao/ScalarDbDao.java
```diff
@@ -248,6 +249,34 @@ public Scanner createScanner(
...
+      String namespace,
+      String table,
+      List<String> projectionColumns,
+      int limit,
+      DistributedTransaction transaction)
+      throws ScalarDbDaoException {
+    Scan scan =
+        createScan(namespace, table, null, null, new ArrayList<>(), projectionColumns, limit);
+    try {
+      return transaction.getScanner(scan);
```
brfrn169, line 271: Please use `DistributedTransactionManager.getScanner()`. You don’t need to begin or commit a transaction when using it.

data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataexport/ExportManager.java
```diff
@@ -68,59 +93,146 @@ abstract void processFooter(
...
+      try (BufferedWriter bufferedWriter = new BufferedWriter(writer)) {
+        processHeader(exportOptions, tableMetadata, bufferedWriter);
 
-      BufferedWriter bufferedWriter = new BufferedWriter(writer);
-      boolean isJson = exportOptions.getOutputFileFormat() == FileFormat.JSON;
+        int threadCount =
+            exportOptions.getMaxThreadCount() > 0
+                ? exportOptions.getMaxThreadCount()
+                : Runtime.getRuntime().availableProcessors();
+        executorService = Executors.newFixedThreadPool(threadCount);
```
brfrn169, line 129: My understanding is as follows:

- The main thread (a single thread) scans the table.
- Multiple threads process the scan results in batches and write them through a single writer (a single file).

Is that correct?

ypeckstadt:
@brfrn169 Hiro would like is to focus on the migration of existing features only for the 3.16 release. So no need to handle/merge this PR yet this week.

Is it best that I add this to the 3.16.1 project when available instead? (and remove it from 3.16 for now)?

data-loader/cli/src/main/java/com/scalar/db/dataloader/cli/command/dataexport/ExportCommandOptions.java
inv-jishnu: @brfrn169 san,
`-m` is already  used by the `--include-matadata` input as this was already present
```
@CommandLine.Option(
      names = {"--include-metadata", "-m"},
      description = "Include transaction metadata in the exported data (default: false)",
      defaultValue = "false")
  protected boolean includeTransactionMetadata;
  ```
  I didn't want to change the existing value as it has been there from the always  and there may be users who use it.
  Then also the `--mode in import is `-m` (which also was added in latest change)
  Should I change the mode to `-m` and include metadata to something else like  `-im` ?

File: data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataexport/ExportManager.java (hunk 4 above)
Comment (inv-jishnu) [reply data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataexport/ExportManager.java:129]: Yes, that is correct.

File: data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataimport/dao/ScalarDbDao.java (hunk 3 above)
Comment (inv-jishnu) [reply data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataimport/dao/ScalarDbDao.java:271]: @brfrn169 san,
I had initially used `DistributedTransactionManager` directly but then was a bit confused whether I can use it like that (in import `transaction` was used in the Dao). So I used `transaction`. I have updated this as suggested.
Thank you.

**Comment by inv-jishnu** [comment]
@brfrn169 san
I have made changes as suggested.
Please take a look at this again when you get a chance.
Thank you.

**Comment by brfrn169** [comment]
> Is it best that I add this to the 3.16.1 project when available instead? (and remove it from 3.16 for now)?

@ypeckstadt Basically, we don’t include new features in patch versions.

As discussed with @feeblefakie, we’ve decided not to release 3.16 this week but to release it next week instead. So I think we can include this PR in that release. Thanks.

File: data-loader/cli/src/main/java/com/scalar/db/dataloader/cli/command/dataexport/ExportCommandOptions.java (hunk 2 above)
Comment (inv-jishnu) [reply data-loader/cli/src/main/java/com/scalar/db/dataloader/cli/command/dataexport/ExportCommandOptions.java]: @brfrn169 san,
After discussing with Yves san, I have set `-m` for mode and `-im` to inlcude-metadata to make it consistent with import.
Thank you.

**Comment by thongdk8** [comment]
FYI, as discussed in another place [Slack conv](https://scalar-labs.slack.com/archives/CDSFWGSTZ/p1750067086740499). With the default config of transaction manager `consensus-commit` when exporting data in transaction mode, the OMM will occur if the export data exceeds the available memory. So we are expecting to run exporting in transaction mode with `scalar.db.transaction_manager=single-crud-operation`, with `single-crud-operation` transaction manager the memory usage will be safe as exporting in `storage` mode.

**Comment by inv-jishnu** [comment]
@brfrn169 san, @komamitsu san,
I have made updates based on feedback.
Please take a look at this PR when you get a chance.
Thank you

File: data-loader/cli/src/main/java/com/scalar/db/dataloader/cli/command/dataexport/ExportCommand.java [hunk 5]
```diff
@@ -126,11 +131,100 @@ private void validateOutputDirectory() throws DirectoryValidationException {
...
+      String scalarDbPropertiesFilePath)
+      throws IOException {
     ProducerTaskFactory taskFactory =
         new ProducerTaskFactory(delimiter, includeTransactionMetadata, prettyPrintJson);
-    DistributedStorage storage = storageFactory.getStorage();
+    if (scalarDbMode.equals(ScalarDbMode.STORAGE)) {
+      DistributedStorage storage = StorageFactory.create(scalarDbPropertiesFilePath).getStorage();
+      return createExportManagerWithStorage(storage, scalarDbDao, fileFormat, taskFactory);
+    } else {
+      DistributedTransactionManager distributedTransactionManager =
```
komamitsu, line 179: Don't need to consider https://github.com/scalar-labs/scalardb/pull/2740#issuecomment-2982959912 ?

inv-jishnu:
@komamitsu san, @brfrn169 san (cc @ypeckstadt )

Based on comment from @thongdk8 , for export operations in transaction mode, is it acceptable to explicitly set scalar.db.transaction_manager to single-crud-operation, even if the user hasn’t specified this value in the database configuration file?

komamitsu:
@inv-jishnu I think it's a reasonable and safe option. But it depends on the requirement (e.g., users can use `consensus-commit` to export snapshot data even it has the risk of OOM). Probably, you can talk with @ypeckstadt to clarify the requirement.

brfrn169:
> Based on comment from @thongdk8 , for export operations in transaction mode, is it acceptable to explicitly set scalar.db.transaction_manager to single-crud-operation, even if the user hasn’t specified this value in the database configuration file?

@inv-jishnu Let me discuss it with @feeblefakie.

ypeckstadt:
Based on our discussion, this code and PR is necessary and will represent the consensus-commit approach for the export. (which needs to be supported). At the same time, the single CRUD operation will be handled via another PR. That PR will basically remove all usage of DistributedStorage and replace it with usage of the Single CRUD interface.

ypeckstadt:
After discussion in the team., it's clear that this PR is not required anymore and we the export will just be done using the DistributedTransactionManager instead.

@inv-jishnu FYI

PR #2752, opened 2025-06-10
Resolve minor bugs in data loader core

inv-jishnu:
## Description

This PR addresses several bugs encountered while testing the CLI application for data import using transactions.

 **Incorrect success/failure count in transaction import**

There was a mismatch in success and failure counts when an import failed within a transaction batch. Although all entries were initially marked as saved, if any transaction failed during the pre-commit check, their statuses were changed to aborted. However, the success/failure counts were calculated before this status update, leading to incorrect reporting.

Updated the logic to determine if any transaction in the batch failed:
- If so, the full batch size is counted as failures.
- Otherwise, it is counted as successes.

**Missing namespace and table Names in column validation errors**

Column validation error messages were showing null for the namespace and table name. This happened because ColumnInfo was being constructed without these values.

Updated ColumnInfo initialization to include namespace and table names. Also added the invalid value in the error message to improve clarity.

**Failure and summary logs not generated on import failure**

When an import failed during a transaction batch, failure and summary logs were not being created. This was caused by an attempt to update a singleton list used to store transaction batch results, which threw a java.lang.UnsupportedOperationException.

Changed the list structure to a modifiable list, allowing proper updates and enabling correct logging.

## Related issues and/or PRs

NA

## Changes made
- Updated the logic to correctly calculate success and failure counts for each data chunk.
- Added table and namespace name parameters to ensure they are not null and are properly included in error messages.
  - Also updated the error message to include the value that caused the error for better context.
  - Included the import source record in the failure log to provide additional information.
- Changed the creation of the status object list to use an ArrayList instead of a singleton list, as the list may need to be modified later in case of specific errors.

## Checklist

NA

## Release notes

Fixes for minor bugs and issues in data loader core


komamitsu (approved):
LGTM! 👍

data-loader/core/src/main/java/com/scalar/db/dataloader/core/dataimport/task/ImportTask.java
```diff
@@ -73,7 +73,7 @@ public ImportTaskResult execute() {
       return ImportTaskResult.builder()
           .rawRecord(params.getSourceRecord())
           .rowNumber(params.getRowNumber())
-          .targets(Collections.singletonList(singleTargetResult))
+          .targets(new ArrayList<>(Collections.singletonList(singleTargetResult)))
```
Torch3333, line 76: What is the motivation behind this change?

inv-jishnu: @Torch3333 san,
Earlier I used the `Collections.singletonList(singleTargetResult)` which caused an exception(java.lang.UnsupportedOperationException) when trying to update the value (which is done in case of an exception) as it be immutable. So to fix this issue I wrapped it with `ArrayList<>()`.

Torch3333: Understood, thanks!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #2756, opened 2025-06-11
Introduce StorageInfo

brfrn169:
## Description

This PR introduces `StorageInfo`, which provides metadata about the storage.

Currently, it includes the following information:
- Storage name
- Mutation atomicity unit: Defines the scope of atomicity for mutations.
- Maximum atomic mutation count: The maximum number of mutations that can be performed atomically in the storage.

For more details, refer to the `StorageInfo` interface.

You can get a `StorageInfo` instance via the `DistributedStorageAdmin.getStorageInfo()` method.

This PR also extends the behavior of the `DistributedStorage.mutate()` method based on the storage’s mutation atomicity unit and maximum atomic mutation count. For details, see the Javadoc for the `DistributedStorage.mutate()` method.

## Related issues and/or PRs

N/A

## Changes made

Added some inline comments. Please take a look for the details.

## Checklist

N/A

## Release notes

Introduced `StorageInfo`, which provides metadata about the storage. You can obtain a `StorageInfo` instance via the `DistributedStorageAdmin.getStorageInfo()` method.

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedStorage.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedStorage.java
@@ -185,7 +185,7 @@ public interface DistributedStorage extends AutoCloseable {
    * StorageInfo#getMutationAtomicityUnit()}. For example, if the atomicity unit of the storage is
    * {@link StorageInfo.MutationAtomicityUnit#PARTITION}, the mutations must occur within the same
    * partition. Also note that the maximum number of mutations that can be performed atomically is
-   * defined by {@link StorageInfo#getMaxAtomicMutationCount()}.
+   * defined by {@link StorageInfo#getMaxAtomicMutationsCount()}.
    *
    * <p>To retrieve storage information, use {@link DistributedStorageAdmin#getStorageInfo(String)}.
    *
--- a/core/src/main/java/com/scalar/db/api/DistributedStorage.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedStorage.java
@@ -181,6 +181,14 @@ public interface DistributedStorage extends AutoCloseable {
   /**
    * Mutates entries of the underlying storage with the specified list of {@link Mutation} commands.
    *
+   * <p>Note that this method only supports mutations within the atomicity unit specified by {@link
+   * StorageInfo#getMutationAtomicityUnit()}. For example, if the atomicity unit of the storage is
+   * {@link StorageInfo.MutationAtomicityUnit#PARTITION}, the mutations must occur within the same
+   * partition. Also note that the maximum number of mutations that can be performed atomically is
+   * defined by {@link StorageInfo#getMaxAtomicMutationCount()}.
+   *
+   * <p>To retrieve storage information, use {@link DistributedStorageAdmin#getStorageInfo(String)}.
+   *
    * @param mutations a list of {@code Mutation} commands
    * @throws ExecutionException if the operation fails
    */
--- a/core/src/main/java/com/scalar/db/api/DistributedStorageAdmin.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedStorageAdmin.java
@@ -88,6 +88,15 @@ public interface DistributedStorageAdmin extends Admin, AutoCloseable {
   void addRawColumnToTable(String namespace, String table, String columnName, DataType columnType)
       throws ExecutionException;
 
+  /**
+   * Returns the storage information.
+   *
+   * @param namespace the namespace to get the storage information for
+   * @return the storage information
+   * @throws ExecutionException if the operation fails
+   */
+  StorageInfo getStorageInfo(String namespace) throws ExecutionException;
+
   /** Closes connections to the storage. */
   @Override
   void close();
--- a/core/src/main/java/com/scalar/db/api/StorageInfo.java
+++ b/core/src/main/java/com/scalar/db/api/StorageInfo.java
@@ -1,0 +1,62 @@
+package com.scalar.db.api;
+
+public interface StorageInfo {
+  /**
+   * Returns the storage name.
+   *
+   * @return the storage name
+   */
+  String getStorageName();
+
+  /**
+   * Returns the mutation atomicity unit of the storage.
+   *
+   * @return the mutation atomicity unit of the storage
+   */
+  MutationAtomicityUnit getMutationAtomicityUnit();
+
+  /**
+   * Returns the maximum number of mutations that can be performed atomically in the storage.
+   *
+   * @return the maximum number of mutations that can be performed atomically in the storage
+   */
+  int getMaxAtomicMutationCount();
+
+  /**
+   * The mutation atomicity unit of the storage.
+   *
+   * <p>This enum defines the atomicity unit for mutations in the storage. It determines the scope
+   * of atomicity for mutations such as put and delete.
+   */
+  enum MutationAtomicityUnit {
+    /**
+     * The atomicity unit is at the record level, meaning that mutations are performed atomically
+     * for each record.
+     */
+    RECORD,
+
+    /**
+     * The atomicity unit is at the partition level, meaning that mutations are performed atomically
+     * for each partition.
+     */
+    PARTITION,
+
+    /**
+     * The atomicity unit is at the table level, meaning that mutations are performed atomically for
+     * each table.
+     */
+    TABLE,
+
+    /**
+     * The atomicity unit is at the namespace level, meaning that mutations are performed atomically
+     * for each namespace.
+     */
+    NAMESPACE,
+
+    /**
+     * The atomicity unit is at the storage level, meaning that mutations are performed atomically
+     * for the entire storage.
+     */
+    STORAGE
+  }
+}
--- a/core/src/main/java/com/scalar/db/api/StorageInfo.java
+++ b/core/src/main/java/com/scalar/db/api/StorageInfo.java
@@ -20,7 +20,7 @@ public interface StorageInfo {
    *
    * @return the maximum number of mutations that can be performed atomically in the storage
    */
-  int getMaxAtomicMutationCount();
+  int getMaxAtomicMutationsCount();
 
   /**
    * The mutation atomicity unit of the storage.
--- a/core/src/main/java/com/scalar/db/common/StorageInfoProvider.java
+++ b/core/src/main/java/com/scalar/db/common/StorageInfoProvider.java
@@ -1,0 +1,42 @@
+package com.scalar.db.common;
+
+import com.scalar.db.api.DistributedStorageAdmin;
+import com.scalar.db.api.StorageInfo;
+import com.scalar.db.exception.storage.ExecutionException;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import javax.annotation.concurrent.ThreadSafe;
+
+@ThreadSafe
+public class StorageInfoProvider {
+
+  private final DistributedStorageAdmin admin;
+
+  // Cache to store storage information. A map namespace to StorageInfo.
+  private final ConcurrentMap<String, StorageInfo> storageInfoCache = new ConcurrentHashMap<>();
+
+  @SuppressFBWarnings("EI_EXPOSE_REP2")
+  public StorageInfoProvider(DistributedStorageAdmin admin) {
+    this.admin = admin;
+  }
+
+  public StorageInfo getStorageInfo(String namespace) throws ExecutionException {
+    try {
+      return storageInfoCache.computeIfAbsent(namespace, this::getStorageInfoInternal);
+    } catch (RuntimeException e) {
+      if (e.getCause() instanceof ExecutionException) {
+        throw (ExecutionException) e.getCause();
+      }
+      throw e;
+    }
+  }
+
+  private StorageInfo getStorageInfoInternal(String namespace) {
+    try {
+      return admin.getStorageInfo(namespace);
+    } catch (ExecutionException e) {
+      throw new RuntimeException(e);
+    }
+  }
+}
--- a/core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
+++ b/core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
@@ -11,7 +11,9 @@ import com.scalar.db.api.Scan;
 import com.scalar.db.api.ScanAll;
 import com.scalar.db.api.Selection;
 import com.scalar.db.api.Selection.Conjunction;
+import com.scalar.db.api.StorageInfo;
 import com.scalar.db.api.TableMetadata;
+import com.scalar.db.common.StorageInfoProvider;
 import com.scalar.db.common.TableMetadataManager;
 import com.scalar.db.common.error.CoreError;
 import com.scalar.db.config.DatabaseConfig;
@@ -31,10 +33,15 @@ public class OperationChecker {
 
   private final DatabaseConfig config;
   private final TableMetadataManager tableMetadataManager;
+  private final StorageInfoProvider storageInfoProvider;
 
-  public OperationChecker(DatabaseConfig config, TableMetadataManager tableMetadataManager) {
+  public OperationChecker(
+      DatabaseConfig config,
+      TableMetadataManager tableMetadataManager,
+      StorageInfoProvider storageInfoProvider) {
     this.config = config;
     this.tableMetadataManager = tableMetadataManager;
+    this.storageInfoProvider = storageInfoProvider;
   }
 
   public void check(Get get) throws ExecutionException {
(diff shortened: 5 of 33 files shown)


core/src/main/java/com/scalar/db/api/DistributedStorage.java
```diff
@@ -181,6 +181,14 @@ public interface DistributedStorage extends AutoCloseable {
   /**
    * Mutates entries of the underlying storage with the specified list of {@link Mutation} commands.
    *
+   * <p>Note that this method only supports mutations within the atomicity unit specified by {@link
+   * StorageInfo#getMutationAtomicityUnit()}. For example, if the atomicity unit of the storage is
+   * {@link StorageInfo.MutationAtomicityUnit#PARTITION}, the mutations must occur within the same
+   * partition. Also note that the maximum number of mutations that can be performed atomically is
+   * defined by {@link StorageInfo#getMaxAtomicMutationCount()}.
+   *
+   * <p>To retrieve storage information, use {@link DistributedStorageAdmin#getStorageInfo(String)}.
```
brfrn169: Extended the behavior of the `DistributedStorage.mutate()`. See this Javadoc for the details.

core/src/main/java/com/scalar/db/api/DistributedStorageAdmin.java
```diff
@@ -88,6 +88,15 @@ TableMetadata getImportTableMetadata(
...
       throws ExecutionException;
 
+  /**
+   * Returns the storage information.
+   *
+   * @param namespace the namespace to get the storage information for
+   * @return the storage information
+   * @throws ExecutionException if the operation fails
+   */
+  StorageInfo getStorageInfo(String namespace) throws ExecutionException;
```
brfrn169, line 98: You can get the `StorageInfo` instance of the storage through this method.

core/src/main/java/com/scalar/db/common/StorageInfoProvider.java
```diff
@@ -0,0 +1,42 @@
...
+import com.scalar.db.api.DistributedStorageAdmin;
+import com.scalar.db.api.StorageInfo;
+import com.scalar.db.exception.storage.ExecutionException;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import javax.annotation.concurrent.ThreadSafe;
+
+@ThreadSafe
+public class StorageInfoProvider {
```
brfrn169, line 12: `StorageInfoProvider` provides caching functionality for `StorageInfo` instances.

core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
```diff
@@ -334,16 +341,18 @@ public void check(List<? extends Mutation> mutations) throws ExecutionException
...
 
-      // Check if all mutations are for the same partition
-      if (!mutation.forNamespace().equals(first.forNamespace())
-          || !mutation.forTable().equals(first.forTable())
-          || !mutation.getPartitionKey().equals(first.getPartitionKey())) {
+      // Check if the mutations are within the atomicity unit of the storage
+      if (isOutOfAtomicityUnit(first, storageInfoForFirst, mutation)) {
         throw new IllegalArgumentException(
-            CoreError.OPERATION_CHECK_ERROR_MULTI_PARTITION_MUTATION.buildMessage(mutations));
+            getErrorMessageForOutOfAtomicityUnit(storageInfoForFirst, mutations));
```
brfrn169, line 355: Changed the check logic based on the mutation atomicity unit of the storage.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -47,6 +49,13 @@ public class CassandraAdmin implements DistributedStorageAdmin {
   public static final String NAMESPACES_NAME_COL = "name";
   private static final String DEFAULT_REPLICATION_FACTOR = "3";
   @VisibleForTesting static final String INDEX_NAME_PREFIX = "index";
+  private static final StorageInfo STORAGE_INFO =
+      new StorageInfoImpl(
+          "cassandra",
+          StorageInfo.MutationAtomicityUnit.PARTITION,
+          // No limit on the number of mutations
+          Integer.MAX_VALUE);
```
brfrn169, line 57: This is the `StorageInfo` instance for Cassandra.

core/src/main/java/com/scalar/db/storage/cosmos/CosmosAdmin.java
```diff
@@ -64,6 +66,12 @@ public class CosmosAdmin implements DistributedStorageAdmin {
   @VisibleForTesting public static final String STORED_PROCEDURE_FILE_NAME = "mutate.js";
   private static final String STORED_PROCEDURE_PATH =
       "cosmosdb_stored_procedure/" + STORED_PROCEDURE_FILE_NAME;
+  private static final StorageInfo STORAGE_INFO =
+      new StorageInfoImpl(
+          "cosmos",
+          StorageInfo.MutationAtomicityUnit.PARTITION,
+          // No limit on the number of mutations
+          Integer.MAX_VALUE);
```
brfrn169, line 74: This is the `StorageInfo` instance for Cosmos DB.

core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
```diff
@@ -153,6 +155,12 @@ public class DynamoAdmin implements DistributedStorageAdmin {
           .put(SCALING_TYPE_INDEX_READ, MetricType.DYNAMO_DB_READ_CAPACITY_UTILIZATION)
           .put(SCALING_TYPE_INDEX_WRITE, MetricType.DYNAMO_DB_WRITE_CAPACITY_UTILIZATION)
           .build();
+  private static final StorageInfo STORAGE_INFO =
+      new StorageInfoImpl(
+          "dynamo",
+          StorageInfo.MutationAtomicityUnit.STORAGE,
+          // DynamoDB has a limit of 100 items per transactional batch write operation
+          100);
```
brfrn169, line 163: This is the `StorageInfo` instance for DynamoDB.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -59,6 +61,12 @@ public class JdbcAdmin implements DistributedStorageAdmin {
   @VisibleForTesting static final String NAMESPACE_COL_NAMESPACE_NAME = "namespace_name";
   private static final Logger logger = LoggerFactory.getLogger(JdbcAdmin.class);
   private static final String INDEX_NAME_PREFIX = "index";
+  private static final StorageInfo STORAGE_INFO =
+      new StorageInfoImpl(
+          "jdbc",
+          StorageInfo.MutationAtomicityUnit.STORAGE,
+          // No limit on the number of mutations
+          Integer.MAX_VALUE);
```
brfrn169, line 69: This is the `StorageInfo` instance for JDBC databases.

core/src/main/java/com/scalar/db/api/StorageInfo.java
```diff
@@ -0,0 +1,62 @@
...
+   * @return the storage name
+   */
+  String getStorageName();
+
+  /**
+   * Returns the mutation atomicity unit of the storage.
+   *
+   * @return the mutation atomicity unit of the storage
+   */
+  MutationAtomicityUnit getMutationAtomicityUnit();
```
feeblefakie, line 16:
```suggestion
  MutationAtomicityGranularity getMutationAtomicityGranularity();
```
After some thought, I feel it's probably more proper to call it `granularity`, although it's a bit long.
What do you think?

core/src/main/java/com/scalar/db/api/StorageInfo.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Returns the maximum number of mutations that can be performed atomically in the storage.
+   *
+   * @return the maximum number of mutations that can be performed atomically in the storage
+   */
+  int getMaxAtomicMutationCount();
```
feeblefakie:
```suggestion
  int getMaxAtomicMutationsCount();
```

core/src/main/java/com/scalar/db/api/StorageInfo.java
brfrn169: Gemini says `MutationAtomicityUnit` is more appropriate, but I'm not sure.

```
In the context of transactions, when referring to the scope within which updates can be performed atomically, **"MutationAtomicityUnit"** is the better choice.

Here's why:

* **Unit**: "Unit" directly implies a distinct, defined **scope** or **entity**. The example you provided with `RECORD`, `PARTITION`, `TABLE`, etc., clearly defines the atomic "unit" of operation. It naturally conveys the idea that mutations are performed atomically for *each* of these specified units.
* **Granularity**: "Granularity" refers to the **fineness or coarseness** of something. While related, it describes the *degree* of detail or division rather than the specific scope itself. For instance, you might discuss whether the atomicity has "fine-grained" or "coarse-grained" granularity. When you want to name the actual *thing* that is atomic, "Unit" is more precise and less ambiguous.

Therefore, "MutationAtomicityUnit" more accurately and clearly communicates the intended meaning of the atomic **scope** for mutations.
```

What do you think? 🙇

core/src/main/java/com/scalar/db/common/checker/OperationChecker.java
```diff
@@ -357,6 +366,72 @@ public void check(List<? extends Mutation> mutations) throws ExecutionException
...
+  private boolean isOutOfAtomicityUnit(
+      Mutation mutation1, StorageInfo storageInfo1, Mutation mutation2) throws ExecutionException {
+    assert mutation1.forNamespace().isPresent()
+        && mutation1.forTable().isPresent()
+        && mutation2.forNamespace().isPresent()
+        && mutation2.forTable().isPresent();
+
+    switch (storageInfo1.getMutationAtomicityUnit()) {
+      case RECORD:
+        if (!mutation1.getClusteringKey().equals(mutation2.getClusteringKey())) {
```
komamitsu, line 378: I might be missing something, but this seems only checking the clustering keys. Is it possible the 2 mutations have different namespace, table, or partition keys here?

komamitsu: Ah, I thought this case had `break`... Looks good 👍

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
```diff
@@ -288,4 +324,15 @@ public void close() {
       admin.close();
     }
   }
+
+  @VisibleForTesting
+  static class AdminHolder {
+    public final String storageName;
```
komamitsu: [minor] These don't need to be public.

core/src/main/java/com/scalar/db/api/StorageInfo.java
feeblefakie: I also asked Gemini. It suggests Granularity.... I think it depends on how we ask.

**Question:**
I'm considering how to name a method, and I would appreciate your advice.

We provide a transaction manager for various kinds of databases. The transaction manager pushes down write operations into the underlying databases based on the size that the databases can atomically write.

To make each database provide the size information, I created a `StorageInfo` interface, created a class that implements the interface for each database, and created a method that returns the size information.
The size information is an Enum class that has the following values: STORAGE, NAMESPACE, TABLE, PARTITION, and RECORD.

I have two candidate names for the method.
One is `getMutationAtomicityUnit()` and the other is `getMutationAtomicityGranularity()`.
Which do you think is better?
Do you have any other better ideas?

**Answer:**
I think getMutationAtomicityGranularity() is the better choice.

Here’s a breakdown of why, along with some alternative ideas.

Comparison of Your Candidates
Both of your proposed names are very good because they are descriptive and unambiguous. However, getMutationAtomicityGranularity() has a slight edge.

getMutationAtomicityGranularity(): Slightly better. 👍
In database and systems design, granularity is the standard term for referring to the scale or level at which an operation occurs (e.g., fine-grained vs. coarse-grained). Since your Enum defines a scale from RECORD (fine-grained) to STORAGE (coarse-grained), this name is a perfect technical fit.

getMutationAtomicityUnit(): Also very good.
The word unit is also accurate. It correctly describes that the method returns the single, indivisible "chunk" for atomic writes. It's simple and clear. The only minor drawback is that "granularity" is more idiomatic in this specific technical domain.

Other Ideas
If you're open to other suggestions, here are a couple of alternatives that aim for simplicity while remaining clear.

getAtomicityLevel()
This is a strong contender. It's shorter and replaces the more technical "Granularity" with the simpler "Level," which is easily understood by any developer. The context of your transaction manager implies that the atomicity relates to mutations.

getAtomicWriteScope()
Using Scope is another excellent option. It clearly communicates that the Enum return value defines the boundary or scope (e.g., the scope of a table, the scope of a partition) within which writes are atomic.

Final Recommendation
Go with getMutationAtomicityGranularity(). It's the most precise and idiomatic term for your use case.

If you feel it's too much jargon, getAtomicityLevel() is a fantastic, simpler alternative that conveys the same meaning effectively.

feeblefakie: Can you share the question as well?

Anyway, Unit seems also correct, so I'm fine with Unit, though.

brfrn169: My question was as follows:

In the context of transactions, which term is more appropriate to express the range in which updates can be performed atomically: MutationAtomicityUnit or MutationAtomicityGranularity?



Below is an example using MutationAtomicityUnit:

```java
  /**
   * The mutation atomicity unit of the storage.
   *
   * <p>This enum defines the atomicity unit for mutations in the storage. It determines the scope
   * of atomicity for mutations such as put and delete.
   */
  enum MutationAtomicityUnit {
    /**
     * The atomicity unit is at the record level, meaning that mutations are performed atomically
     * for each record.
     */
    RECORD,

    /**
     * The atomicity unit is at the partition level, meaning that mutations are performed atomically
     * for each partition.
     */
    PARTITION,

    /**
     * The atomicity unit is at the table level, meaning that mutations are performed atomically for
     * each table.
     */
    TABLE,

    /**
     * The atomicity unit is at the namespace level, meaning that mutations are performed atomically
     * for each namespace.
     */
    NAMESPACE,

    /**
     * The atomicity unit is at the storage level, meaning that mutations are performed atomically
     * for the entire storage.
     */
    STORAGE
  }
```

core/src/main/java/com/scalar/db/storage/multistorage/MultiStorageAdmin.java
brfrn169: You mean the following right?

```java
@VisibleForTesting
  static class AdminHolder {
    final String storageName;
    final DistributedStorageAdmin admin;

    AdminHolder(String storageName, DistributedStorageAdmin admin) {
      this.storageName = storageName;
      this.admin = admin;
    }
  }
```

BTW, considering that the class is already package-private, is there any benefit to removing public from the fields or constructor?

komamitsu: Disclaimer: I'm a bit paranoia on this kind of minor things, and you don't need to follow the comment 😁 
 
 I was thinking of making the fields private:
```java
  static class AdminHolder {
    private final String storageName;
    private final DistributedStorageAdmin admin;

    AdminHolder(String storageName, DistributedStorageAdmin admin) {
      this.storageName = storageName;
      this.admin = admin;
    }
  }
```

This should be stronger encapsulation. If the fields are package-private, I guess that these fields are intended to be accessed from other class in the same package and it requires more attention during code review. (Also, I basically follow Principle of Least Privilege if possible.)

brfrn169: > I basically follow Principle of Least Privilege if possible.

Yeah, I totally agree with you. The downside I’ve noticed recently is that, in such cases, we sometimes have to change the access level when we need to use the class outside the package or in a different context, which can be a bit of a hassle. But I guess it’s a trade-off we have to accept.

Fixed in bd542b7ae2a75d125cbf19a44c102b4744338166. Thanks!

core/src/main/java/com/scalar/db/api/StorageInfo.java
brfrn169: Fixed in 0a32de8e350a05fc52aaa4d8cec6471da83788d9. Thanks!

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/api/StorageInfo.java
feeblefakie: Based on the discussion, since `Unit` is also correct and simpler, let's go with as is.

PR #2764, opened 2025-06-13
Add further optimizations for Consensus Commit

brfrn169:
## Description

This PR adds further optimizations for Consensus Commit.

It introduces a `single-operation` mode, a mode for transactions executed without explicitly beginning a transaction, such as `DistributedTransactionManager.get()`, `DistributedTransactionManager.scan()`, and `DistributedTransactionManager.insert()`.

Details of the optimizations are provided in the inline comments.

## Related issues and/or PRs

N/A

## Changes made

Added some inline comments. Please take a look for the details.

## Checklist

N/A

## Release notes

N/A


core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -51,6 +51,11 @@ public class CrudHandler {
   // Whether the transaction is in read-only mode or not.
   private final boolean readOnly;
 
+  // Whether the transaction is in single-operation mode or not. Single-operation mode refers to
+  // executing a CRUD operation directly through `DistributedTransactionManager` without explicitly
+  // beginning a transaction.
+  private final boolean singleOperation;
```
brfrn169: Introduced `single-operation` mode, a mode for transactions executed without explicitly beginning a transaction, such as `DistributedTransactionManager.get()`, `DistributedTransactionManager.scan()`, and `DistributedTransactionManager.insert()`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -263,9 +288,43 @@ private void putIntoReadSetInSnapshot(Snapshot.Key key, Optional<TransactionResu
...
+  }
+
   private void verifyNoOverlap(Scan scan, Map<Snapshot.Key, TransactionResult> results) {
-    // In read-only mode, we don't need to verify the overlap
-    if (!readOnly) {
+    // In either read-only mode or single-operation mode, we don't need to verify the overlap
+    if (!readOnly && !singleOperation) {
       snapshot.verifyNoOverlap(scan, results);
     }
   }
```
brfrn169: Here is the main part of the optimizations.

The optimizations can be summarized as follows:

- If snapshot reads and serializable validation are not required, we can avoid adding records to the `getSet` and `scanSet`, thereby reducing memory usage. This is because they are only used for snapshot reads and serializable validation.
- If serializable validation is not required, we can skip adding ScannerInfo to the `scannerSet`, since it is used solely for that purpose.
- In addition to read-only mode, we can also skip verifying overlaps between the scan and the records in the `writeSet` and `deleteSet` in single-operation mode.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -432,7 +491,7 @@ private class ConsensusCommitStorageScanner extends AbstractTransactionCrudOpera
     private final List<String> originalProjections;
     private final Scanner scanner;
 
-    private final LinkedHashMap<Snapshot.Key, TransactionResult> results = new LinkedHashMap<>();
+    private final LinkedHashMap<Snapshot.Key, TransactionResult> results;
```
komamitsu: This should have `@Nullable`?

komamitsu (commented):
@brfrn169 We don't need to confirm the number of operations in a transaction is only 1 at the commit phase, if the single operation feature is enabled on the transaction?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -102,11 +111,27 @@ public Optional<Result> get(Get originalGet) throws CrudException {
...
+          .map(r -> new FilteredResult(r, originalProjections, metadata, isIncludeMetadataEnabled));
+    } else {
+      Optional<TransactionResult> result = read(key, get);
+      if (get.getConjunctions().isEmpty()) {
+        return snapshot
+            .mergeResult(key, result)
+            .map(
+                r ->
+                    new FilteredResult(r, originalProjections, metadata, isIncludeMetadataEnabled));
+      } else {
```
feeblefakie: How about handling null/empty conjunction in the `mergeResult` method to unify the operations in the if/else clauses?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
feeblefakie: If a single-crud operation is a self-join reading the same record more than twice, is it OK not to use snapshot read?

feeblefakie (commented):
Left some comments and questions. PTAL!

brfrn169:
> We don't need to confirm the number of operations in a transaction is only 1 at the commit phase, if the single operation feature is enabled on the transaction?

@komamitsu The single-operation flag is set to `true` only for transactions that are executed without explicitly beginning a transaction, such as the following methods:

```java
DistributedTransactionManger transactionManager = ...;

Optional<Result> result = transactionManager.get(Get.newBuilder()...);

List<Result> results = transactionManager.scan(Scan.newBuilder()...);

Scanner scanner = transactionManager.getScanner(Scan.newBuilder()...);

transactionManager.insert(Insert.newBuilder()...);

transactionManager.upsert(Upsert.newBuilder()...);

transactionManager.update(Update.newBuilder()...);

transactionManager.delete(Delete.newBuilder()...);

transactionManager.mutate(Arrays.asList(Delete.newBuilder()..., Update.newBuilder()...));
```

In these cases, the number of operations in a transaction is always 1 at the commit phase (except for `mutate`, but we can handle it similarly). So we don’t need to explicitly check the number of operations.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Fixed in 61b28602ebe7758be455ae11b4dadc2896cd482c. Thanks!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: The single-operation mode here means the mode for transactions that are executed without explicitly beginning a transaction, such as the following methods:
 
```java
DistributedTransactionManger transactionManager = ...;

Optional<Result> result = transactionManager.get(Get.newBuilder()...);

List<Result> results = transactionManager.scan(Scan.newBuilder()...);

Scanner scanner = transactionManager.getScanner(Scan.newBuilder()...);

transactionManager.insert(Insert.newBuilder()...);

transactionManager.upsert(Upsert.newBuilder()...);

transactionManager.update(Update.newBuilder()...);

transactionManager.delete(Delete.newBuilder()...);

transactionManager.mutate(Arrays.asList(Delete.newBuilder()..., Update.newBuilder()...));
```

In these cases, the transaction is committed immediately after the operation is done. So we don't need snapshot reads in such cases. Does this answer your question?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Fixed in cefe83b4c8fb7d9d8ec3ca29740ec396b107d770. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
feeblefakie: @brfrn169 Probably, I am confused with single-crud operation transaction and single-operation mode.
If they are different things, it's pretty confusing so I think we should change the name.

brfrn169: Ah, okay. This isn’t related to single CRUD operation transactions — it’s specific to Consensus Commit.

Do you have any suggestions for what we should call transactions that are executed without explicitly beginning a transaction?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -238,7 +239,8 @@ DistributedTransaction begin(String txId, Isolation isolation, boolean readOnly)
             tableMetadataManager,
             isIncludeMetadataEnabled,
             parallelExecutor,
-            readOnly);
+            readOnly,
+            singleOperation);
```
feeblefakie:
```suggestion
            oneOperation);
```

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
brfrn169: Fixed in 042d1e1ed2d4d7468a515d4844e57076901fec16. Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #2765, opened 2025-06-13
Skip commit-state for read-only transactions in Consensus Commit

brfrn169:
## Description

This PR updates the code to omit writing the commit state for read-only transactions in Consensus Commit to improve performance. This behavior is enabled by default but can be disabled by setting the property `scalar.db.consensus_commit.coordinator.write_omission_on_read_only.enabled` to `false`.

## Related issues and/or PRs

N/A

## Changes made

Add inline comments. Please take a look for the details.

## Checklist

N/A

## Release notes

Changed to omit commit-state for read-only transactions in Consensus Commit to improve performance. This behavior is enabled by default but can be disabled by setting the property `scalar.db.consensus_commit.coordinator.write_omission_on_read_only.enabled` to `false`.


core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -106,42 +106,52 @@ private void waitBeforePreparationSnapshotHookFuture(
...
+        throw new CommitException(e.getMessage(), e, e.getTransactionId().orElse(null));
+      } catch (Exception e) {
+        safelyCallOnFailureBeforeCommit(snapshot);
+        throw e;
       }
-      throw new CommitException(e.getMessage(), e, e.getTransactionId().orElse(null));
-    } catch (Exception e) {
-      safelyCallOnFailureBeforeCommit(snapshot);
-      throw e;
     }
```
brfrn169: If no writes and deletes, skip prepare-records.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -106,42 +106,52 @@ private void waitBeforePreparationSnapshotHookFuture(
...
```
brfrn169: If no reads, skip validate-record.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -106,42 +106,52 @@ private void waitBeforePreparationSnapshotHookFuture(
...
 
     waitBeforePreparationSnapshotHookFuture(snapshot, snapshotHookFuture.orElse(null));
 
-    commitState(snapshot);
-    commitRecords(snapshot);
+    if (!hasNoWritesAndDeletesInSnapshot) {
+      commitState(snapshot);
+      commitRecords(snapshot);
+    }
```
brfrn169: If no writes and deletes, skip commit-state and commit-records.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -221,7 +221,7 @@ DistributedTransaction beginReadOnly(Isolation isolation) {
   DistributedTransaction begin(String txId, Isolation isolation, boolean readOnly) {
     checkArgument(!Strings.isNullOrEmpty(txId));
     checkNotNull(isolation);
-    if (isGroupCommitEnabled()) {
+    if (!readOnly && isGroupCommitEnabled()) {
       assert groupCommitter != null;
       txId = groupCommitter.reserve(txId);
     }
```
brfrn169, line 238: In read-only mode, skip reserving a slot in the group commit.

jnmt (commented):
Thanks for the heads up.

I checked the Auditor's recovery behavior. It first tries to abort a lock holder transaction regardless of read/write types. If a read-only transaction does not put the commit state by this PR, this abort process will always result in creating `aborted` status record and make the Auditor release locks.

Releasing itself is OK for read locks, but I think it's pretty confusing since the created status record is not just garbage but shows the wrong status (aborted) even if the read-only transaction is actually committed.

So, from the ScalarDL perspective, it would be better to configure disabling this optimization. The default `enabled` would be acceptable since we can disable it when creating the transaction manager on the ScalarDL side.

@feeblefakie Please correct me if I miss something and let us know your opinion.

jnmt:
@brfrn169 I missed one point.

> this abort process will always result in creating aborted status record

This is caused by `DistributedTransactionManager.abort()` of the read-only transaction. Should we also skip putting the state here if the transaction was read-only? If skipped, my concern will be resolved, and the configuration option is not necessary.

https://github.com/scalar-labs/scalardb/blob/19aac3d79d57ceb2fd2bbbf34c385db507394a5e/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java#L486

brfrn169:
@jnmt Thank you for reviewing this PR and checking the behavior of the Auditor!

> This is caused by `DistributedTransactionManager.abort()` of the read-only transaction. Should we also skip putting the state here if the transaction was read-only? If skipped, my concern will be resolved, and the configuration option is not necessary.

Actually, we cannot determine whether the transaction associated with the specified transaction ID is read-only or not. So I don’t think we can handle that on the ScalarDB side.
How about on the Auditor side? If the Auditor can detect that the transaction is read-only, could it skip calling `DistributedTransactionManager.abort()`?

jnmt:
> How about on the Auditor side? If the Auditor can detect that the transaction is read-only, could it skip calling DistributedTransactionManager.abort()?

Unfortunately, I don’t think we can easily determine if a transaction is read-only in the recovery context. It might be possible by re-executing the transaction based on the request log, but it’s too much and not feasible for this purpose, I think.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -106,42 +106,58 @@ private void waitBeforePreparationSnapshotHookFuture(
...
-    try {
-      prepare(snapshot);
-    } catch (PreparationException e) {
-      safelyCallOnFailureBeforeCommit(snapshot);
-      abortState(snapshot.getId());
-      rollbackRecords(snapshot);
-      if (e instanceof PreparationConflictException) {
-        throw new CommitConflictException(e.getMessage(), e, e.getTransactionId().orElse(null));
+
+    if (!hasNoWritesAndDeletesInSnapshot) {
```
feeblefakie: I think double negative is a bit hard to read.
Since all the usage of `hasNoWritesAndDeletesInSnapshot` is with negation (!), how about making the method like `hasSomeWritesAndDeletesInSnapshot` so that we don't have to negate all the if clauses?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -106,42 +106,58 @@ private void waitBeforePreparationSnapshotHookFuture(
...
 
-    try {
-      validate(snapshot);
-    } catch (ValidationException e) {
-      safelyCallOnFailureBeforeCommit(snapshot);
-      abortState(snapshot.getId());
-      rollbackRecords(snapshot);
-      if (e instanceof ValidationConflictException) {
-        throw new CommitConflictException(e.getMessage(), e, e.getTransactionId().orElse(null));
+    if (!snapshot.hasNoReads()) {
```
feeblefakie: Ditto.

feeblefakie (commented):
Overall, looking good. 
Left some suggestions. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: Fixed in 6c6299ddb11bc046d23766ef6b544dcbd7a3dc0c. Thanks!

brfrn169:
@feeblefakie What do you think about @jnmt's comment?
https://github.com/scalar-labs/scalardb/pull/2765#pullrequestreview-2924313851

Should we go with making the skipping commit-state behavior configurable?

feeblefakie:
@brfrn169 As @jnmt mentioned, I think we need a coordinator write even for read-only operations.
So, can you make it configurable?

@jnmt 
If we use its own transaction state management mechanism (enabled by `tx_state_management`), we don't need to rely on the coordinator behavior changes.
If the underlying database provides atomicity for the entire storage (STORAGE), we don't have to double the writes for the transaction coordination.

brfrn169:
@feeblefakie @jnmt Added the `scalar.db.consensus_commit.coordinator.write_omission_on_read_only.enabled` property to control the omission behavior. Please take a look when you have time!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -106,42 +109,62 @@ private void waitBeforePreparationSnapshotHookFuture(
     }
   }
 
-  public void commit(Snapshot snapshot) throws CommitException, UnknownTransactionStatusException {
+  public void commit(Snapshot snapshot, boolean readOnly)
+      throws CommitException, UnknownTransactionStatusException {
+    boolean hasSomeWritesOrDeletesInSnapshot = !readOnly && snapshot.hasSomeWritesOrDeletes();
```
jnmt: Super minor. I feel "Some" is a bit redundant if you don't have any special intention.

```suggestion
    boolean hasWritesOrDeletesInSnapshot = !readOnly && snapshot.hasWritesOrDeletes();
```

brfrn169: Fixed in e783b136e64f5fafe88a183c8e88b754fcba100c. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -106,42 +109,62 @@ private void waitBeforePreparationSnapshotHookFuture(
     }
   }
 
-  public void commit(Snapshot snapshot) throws CommitException, UnknownTransactionStatusException {
+  public void commit(Snapshot snapshot, boolean readOnly)
+      throws CommitException, UnknownTransactionStatusException {
+    boolean hasSomeWritesOrDeletesInSnapshot = !readOnly && snapshot.hasWritesOrDeletes();
```
feeblefakie:
```suggestion
    boolean hasWritesOrDeletesInSnapshot = !readOnly && snapshot.hasWritesOrDeletes();
```
If `Some` is removed, should it be consistent?

feeblefakie (approved):
LGTM! Thank you!
Just one suggestion for the naming.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: Good catch! Fixed in 2c13b1cc50509e8850b506c50fc98883ebafe6ab. Thanks!

PR #2766, opened 2025-06-13
Fix potential connection leak when using `jdbc` storage and Scan operation fails because the target table doesn't exist

komamitsu:
## Description

I found the current `com.scalar.db.storage.jdbc.JdbcDatabase#scan` doesn't close the connection when the target table doesn't exist. This PR fixes the issue.

## Related issues and/or PRs

None

## Changes made

- Close the connection for Scan operation in `JdbcDatabase.scan()` whenever an exception is thrown not only when SQLException is thrown, since `IllegalArgumentException` can be thrown when the target table doesn't exist

## Checklist

None

## Release notes

Fixed potential connection leak when using `jdbc` storage and Scan operation fails because the target table doesn't exist

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
@@ -105,7 +105,7 @@ public class JdbcDatabase extends AbstractDistributedStorage {
       connection.setAutoCommit(false);
       rdbEngine.setConnectionToReadOnly(connection, true);
       return jdbcService.getScanner(scan, connection);
-    } catch (SQLException e) {
+    } catch (Exception e) {
       try {
         if (connection != null) {
           connection.rollback();
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
@@ -105,7 +105,7 @@ public class JdbcDatabase extends AbstractDistributedStorage {
       connection.setAutoCommit(false);
       rdbEngine.setConnectionToReadOnly(connection, true);
       return jdbcService.getScanner(scan, connection);
-    } catch (Exception e) {
+    } catch (SQLException e) {
       try {
         if (connection != null) {
           connection.rollback();
@@ -117,6 +117,11 @@ public class JdbcDatabase extends AbstractDistributedStorage {
       close(connection);
       throw new ExecutionException(
           CoreError.JDBC_ERROR_OCCURRED_IN_SELECTION.buildMessage(e.getMessage()), e);
+    } catch (Exception e) {
+      if (connection != null) {
+        close(connection);
+      }
+      throw e;
     }
   }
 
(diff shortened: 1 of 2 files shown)


komamitsu:
~The current fix changes the bahavior a bit. Let me update this PR.~

Fixed it.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -117,6 +117,11 @@ public Scanner scan(Scan scan) throws ExecutionException {
       close(connection);
       throw new ExecutionException(
           CoreError.JDBC_ERROR_OCCURRED_IN_SELECTION.buildMessage(e.getMessage()), e);
+    } catch (Exception e) {
+      if (connection != null) {
+        close(connection);
+      }
```
brfrn169: The null handling is already done in the `close()` method, so we don't need to do that here:
```suggestion
      close(connection);
```

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -117,6 +117,11 @@ public Scanner scan(Scan scan) throws ExecutionException {
       close(connection);
       throw new ExecutionException(
           CoreError.JDBC_ERROR_OCCURRED_IN_SELECTION.buildMessage(e.getMessage()), e);
+    } catch (Exception e) {
+      if (connection != null) {
```
brfrn169: Don't we need to `connection.rollback()` here?

brfrn169 (commented):
Good catch! Left a few comments. Please take a look when you have time!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
komamitsu: Thanks! I focused on the connection leak, but I think the transaction should be rolled-back immediately to avoid potential future lazy-recovery. Addressed it in https://github.com/scalar-labs/scalardb/pull/2766/commits/e96888e673ea7f9aea35c6622e4937a2b448bf61.

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcDatabase.java
```diff
@@ -197,6 +197,9 @@ public void mutate(List<? extends Mutation> mutations) throws ExecutionException
       close(connection);
       throw new ExecutionException(
           CoreError.JDBC_ERROR_OCCURRED_IN_MUTATION.buildMessage(e.getMessage()), e);
+    } catch (Exception e) {
```
komamitsu, line 200: @brfrn169 I took care of this similar potential leak in https://github.com/scalar-labs/scalardb/pull/2766/commits/f18d40f6765a0a61b6e73c521d6967d0551ca830.

brfrn169: Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #2786, opened 2025-06-18
Handle Get and Scan with conjunctions correctly in Consensus Commit

brfrn169:
## Description

This PR updates the code to handle `Get` and `Scan` with conjunctions correctly in Consensus Commit.

When users specify conjunctions for Get and Scan operations, the same conjunctions also need to be applied to the columns in the before images. The reason is explained in the following Javadoc—please refer to it for details:

https://github.com/scalar-labs/scalardb/pull/2786/files#diff-acc93f1366b03168017001646c7217f7dd1f3c7cbe8a41b6a9a8484dc5ef0905R482-R546

## Related issues and/or PRs

N/A

## Changes made

- Update the code to convert the conjunctions when users specify conjunctions for Get and Scan operations.

## Checklist

N/A

## Release notes

Fixed a bug where records could be missed when executing `Get` or `Scan` with conjunctions in Consensus Commit.

### DIFF
--- a/integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
+++ b/integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
@@ -3126,7 +3126,7 @@ public abstract class ConsensusCommitSpecificIntegrationTestBase {
   @EnabledIf("isGroupCommitEnabled")
   void put_WhenTheOtherTransactionsFails_ShouldBeCommittedWithoutBlocked() throws Exception {
     // Arrange
-    doThrow(PreparationConflictException.class).when(commit).prepare(any());
+    doThrow(PreparationConflictException.class).when(commit).prepareRecords(any());
 
     // Act
     DistributedTransaction failingTxn = manager.begin();
@@ -3206,7 +3206,7 @@ public abstract class ConsensusCommitSpecificIntegrationTestBase {
     DistributedTransaction failingTxn1 = manager.begin(Isolation.SERIALIZABLE);
     DistributedTransaction failingTxn2 = manager.begin(Isolation.SERIALIZABLE);
 
-    doThrow(PreparationConflictException.class).when(commit).prepare(any());
+    doThrow(PreparationConflictException.class).when(commit).prepareRecords(any());
 
     failingTxn1.put(preparePut(0, 0, namespace1, TABLE_1));
     failingTxn2.put(preparePut(1, 0, namespace1, TABLE_1));
--- a/.github/workflows/data-loader-ci.yaml
+++ b/.github/workflows/data-loader-ci.yaml
@@ -1,0 +1,26 @@
+name: CI for Data Loader CLI
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  data_loader_ci:
+    name: Build Data Loader CLI
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up JDK 8
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 8
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Gradle Data Loader CLI build
+        run: ./gradlew :data-loader:cli:build --no-daemon --stacktrace
+
--- a/core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/dynamo/DynamoAdmin.java
@@ -97,6 +97,7 @@ public class DynamoAdmin implements DistributedStorageAdmin {
   public static final String DEFAULT_NO_BACKUP = "false";
   public static final String DEFAULT_REQUEST_UNIT = "10";
   private static final int DEFAULT_WAITING_DURATION_SECS = 3;
+  @VisibleForTesting static final int MAX_RETRY_COUNT = 10;
 
   @VisibleForTesting static final String PARTITION_KEY = "concatenatedPartitionKey";
   @VisibleForTesting static final String CLUSTERING_KEY = "concatenatedClusteringKey";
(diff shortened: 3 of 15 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -417,6 +479,119 @@ private Scanner scanFromStorage(Scan scan) throws CrudException {
...
+   * records in storage:
+   *
+   * <pre>
+   *   | partition_key | clustering_key | column | status    | before_column | before_status  |
+   *   |---------------|----------------|--------|-----------|---------------|----------------|
+   *   | 0             | 0              | 1000   | COMMITTED |               |                |
+   *   | 0             | 1              | 200    | PREPARED  | 1000          | COMMITTED      |
+   * </pre>
+   *
+   * If we scan records with the condition `column = 1000`, we will only retrieve the first record,
```
feeblefakie: The description is a bit weird to me.
How about something like below?

If we scan records with the condition "column = 1000" without converting the condition (conjunction), we only get the first record, not the second one, because the condition does not match. However, the second record has not been committed yet, so we should still retrieve it, considering the possibility that the record will be rolled back.

brfrn169: Fixed in 683337b45881a7512f419015e25235889e25556f. Thanks!

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -5273,6 +5276,657 @@ public void getScanner_InReadOnlyMode_WithSerializable_ShouldNotThrowAnyExceptio
...
+
+    assertThat(actual).isPresent();
+    assertThat(actual.get().getInt(ACCOUNT_ID)).isEqualTo(0);
+    assertThat(actual.get().getInt(ACCOUNT_TYPE)).isEqualTo(0);
+    assertThat(actual.get().getInt(BALANCE)).isEqualTo(INITIAL_BALANCE);
+  }
+
+  @Test
+  public void
+      get_WithConjunction_ForCommittedRecordWhoseBeforeImageMatchesConjunction_ShouldNotReturnRecordAfterLazyRecovery()
```
komamitsu: IIUC, the final state is already COMMITTED and no lazy recovery occurs?

brfrn169: Good catch! Let me fix it.

brfrn169: Fixed in 6c16dcd530b4d023abbbae6047032ed7db0e038d. Thanks!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -137,34 +145,44 @@ void readUnread(@Nullable Snapshot.Key key, Get get) throws CrudException {
...
+      if (key != null) {
+        putIntoReadSetInSnapshot(key, result);
+      } else {
+        // Only for a Get with index, the argument `key` is null
+
+        if (result.isPresent()) {
+          // Only when we can get the record with the Get with index, we can put it into the read
+          // set
+          key = new Snapshot.Key(get, result.get());
+          putIntoReadSetInSnapshot(key, result);
```
jnmt, line 181: Confirmation. The issue we discussed before (unexpected snapshot overwrite) will be resolved in another PR, or am I missing the point?

jnmt (commented):
Thank you for finding the bug. Overall looks good to me. I just left a confirmation. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: We will address that issue in another PR. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -137,34 +145,44 @@ void readUnread(@Nullable Snapshot.Key key, Get get) throws CrudException {
...
+    }
 
-          if (result.isPresent()) {
-            // Only when we can get the record with the Get with index, we can put it into the read
-            // set
-            key = new Snapshot.Key(get, result.get());
-            putIntoReadSetInSnapshot(key, result);
-          }
+    if (!get.getConjunctions().isEmpty()) {
+      // Check if the result matches the conjunctions
```
jnmt: It's likely this check exists because we get records based on not only the current image but also before image using OR conditions. We would like to filter the case where the second record is committed in the example of the Javadoc comment. If my understanding is correct, it might be nice to have a comment like this for future readers to easily grasp the reasoning.

```suggestion
      // Because we also get records whose before images match the conjunctions, we need to check if
      // the current status of the records actually match the conjunctions.
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
jnmt: Got it. Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -234,10 +264,24 @@ private void processScanResult(Snapshot.Key key, Scan scan, TransactionResult re
           snapshot.getId());
     }
 
-    // We always update the read set to create before image by using the latest record (result)
-    // because another conflicting transaction might have updated the record after this
-    // transaction read it first.
-    putIntoReadSetInSnapshot(key, Optional.of(result));
+    Optional<TransactionResult> ret = Optional.of(result);
+    if (!scan.getConjunctions().isEmpty()) {
+      // Check if the result matches the conjunctions
```
jnmt: Ditto.

```suggestion
      // Because we also get records whose before images match the conjunctions, we need to check if
      // the current status of the records actually match the conjunctions.
```

jnmt (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Fixed in 4ec25d631eab49f358a7068b7af68ac854602648. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Fixed in 4ec25d631eab49f358a7068b7af68ac854602648. Thanks!

brfrn169:
@Torch3333 I'll merge this. Please take a look when you have time!

PR #2797, opened 2025-06-19
Prevent overwriting read set in Consensus Commit

brfrn169:
## Description

Currently, we always update the read set when the same record is read multiple times from storage. However, this behavior can lead to anomalies like lost updates.

You can reproduce the issue with the following test:
https://github.com/scalar-labs/scalardb/pull/2797/files#diff-7d8b7163e906546d45d9a7b662f97eb671e39030217ab77ef22325b0f74c2b5eR5898-R5981

In the test, two slightly different Get operations are used to force reading the same record multiple times from storage and bypass snapshot reads.

## Related issues and/or PRs

N/A

## Changes made

- Updated the code to prevent overwriting read set.

## Checklist

N/A

## Release notes

Fixed an issue in Consensus Commit where reading the same record multiple times from storage could cause anomalies like lost updates. The read set is no longer updated on repeated reads to resolve this issue.

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
@@ -164,11 +164,10 @@ public class CrudHandler {
     }
 
     if (result.isPresent() || get.getConjunctions().isEmpty()) {
-      // Keep the read set latest to create before image by using the latest record (result)
-      // because another conflicting transaction might have updated the record after this
-      // transaction read it first. However, we update it only if a get operation has no
-      // conjunction or the result exists. This is because we don’t know whether the record
-      // actually exists or not due to the conjunction.
+      // We put the result into the read set only if a get operation has no conjunction or the
+      // result exists. This is because we don’t know whether the record actually exists or not
+      // due to the conjunction.
+
       if (key != null) {
         putIntoReadSetInSnapshot(key, result);
       } else {
@@ -277,9 +276,6 @@ public class CrudHandler {
     }
 
     if (ret.isPresent()) {
-      // We always update the read set to create before image by using the latest record (result)
-      // because another conflicting transaction might have updated the record after this
-      // transaction read it first.
       putIntoReadSetInSnapshot(key, ret);
     }
 
@@ -319,7 +315,7 @@ public class CrudHandler {
 
   private void putIntoReadSetInSnapshot(Snapshot.Key key, Optional<TransactionResult> result) {
     // In read-only mode, we don't need to put the result into the read set
-    if (!readOnly) {
+    if (!readOnly && !snapshot.containsKeyInReadSet(key)) {
       snapshot.putIntoReadSet(key, result);
     }
   }
--- a/integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
+++ b/integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
@@ -1,6 +1,7 @@
 package com.scalar.db.transaction.consensuscommit;
 
 import static com.scalar.db.api.ConditionBuilder.column;
+import static com.scalar.db.api.ConditionBuilder.updateIf;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -164,11 +164,10 @@ Optional<TransactionResult> read(@Nullable Snapshot.Key key, Get get) throws Cru
     }
 
     if (result.isPresent() || get.getConjunctions().isEmpty()) {
-      // Keep the read set latest to create before image by using the latest record (result)
-      // because another conflicting transaction might have updated the record after this
-      // transaction read it first. However, we update it only if a get operation has no
```
jnmt, line 169: Sorry, my bad. This part was a completely wrong idea. Probably, I was trying to avoid aborts as much as possible, but paid little attention to the case for reading a record multiple times. I need to think about how we can avoid these kinds of concurrency bugs…

jnmt (approved):
Thanks again for finding a bug. LGTM, thank you!

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -5893,6 +5895,91 @@ public void getScanner_InReadOnlyMode_WithSerializable_ShouldNotThrowAnyExceptio
...
+    manager.update(
+        Update.newBuilder()
+            .namespace(namespace1)
+            .table(TABLE_1)
+            .partitionKey(Key.ofInt(ACCOUNT_ID, 0))
+            .clusteringKey(Key.ofInt(ACCOUNT_TYPE, 0))
+            .intValue(BALANCE, 200)
+            .build());
+
+    // Retrieve the record again, but use a different Get object (with a where clause)
```
komamitsu, line 5951: 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -319,7 +315,7 @@ public void closeScanners() throws CrudException {
 
   private void putIntoReadSetInSnapshot(Snapshot.Key key, Optional<TransactionResult> result) {
     // In read-only mode, we don't need to put the result into the read set
-    if (!readOnly) {
+    if (!readOnly && !snapshot.containsKeyInReadSet(key)) {
```
komamitsu, line 318: IIUC, we can use the latest result if read-modify-write hasn't occurred on the key, like as follows?

```java
    if (!readOnly && !(snapshot.containsKeyInReadSet(key) && snapshot.containsKeyInWriteOrDeleteSet(key)) {
```

This question is mainly for confirmation and the current code looks good to me, though.

brfrn169: Sorry, what do you mean by `read-modify-write`? You mean CAS operations in the prepare-records phase, right?

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
komamitsu: Sorry for the unclear comment. My question was as follows:
- With this PR, the read-set isn't updated if the key is read
- I think we can update the read-set even if the key is read, as long as the key isn't written or deleted in the transaction
- For instance
  - Record A has the key = 100 and value = "one"
  - Transaction 1 reads the key and the value is "one", and puts it into the read-set
  - Transaction 2 updates the value of the record to "two" and commits it
  - Transaction 1 reads the key with different GET and the value is "two", **and puts it into the read-set since there is no write operation on the key in Transaction 1**
  - Transaction 1 updates the value of the record to "three" based on the read value "two"
  - In this case, no anomaly doesn't occur even if we use the latest result as a read-set
    - If Transaction 1 updates the record before the second GET, the second GET's result shouldn't be put into the read-set to avoid lost update anomaly

jnmt: Good point. We might be able to rescue such a transaction, but I'm really not sure if it's good for our snapshot definition. One approach is to make it a safer side (like the current PR) for now, and then improve it after the deep discussion and confirmation.

brfrn169: @komamitsu Indeed, we might be able to rescue such a transaction. However, since we don’t have enough time to test it, let’s make it the safer side for now as @jnmt suggested. Thanks!

komamitsu (approved):
LGTM! :+1:

PR #2787, opened 2025-06-18
Remove before images after committing or rolling back records in Consensus Commit

brfrn169:
## Description

This PR updates the code to remove before images after committing or rolling back records in Consensus Commit to reduce disk usage.

## Related issues and/or PRs

N/A

## Changes made

- Updated `CommitMutationComposer` and `RollbackMutationComposer` to set the before image columns to null.

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes.
- [ ] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [ ] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [ ] Tests (unit, integration, etc.) have been added for the changes.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

N/A

## Release notes

Updated the code to remove before images after committing or rolling back records in Consensus Commit to reduce disk usage.

PR #2798, opened 2025-06-19
Improve read process in Consensus Commit

brfrn169:
## Description

This PR improves the read process in Consensus Commit.

The improved read algorithm is as follows:

1. If reading a record with status `COMMITTED`, return the record as before.  
2. If reading a record with status `PREPARED` or `DELETED`, then check the Coordinator table:
   - If the transaction status is `COMMITTED`, start lazy recovery and return the after image.
   - If the transaction status is `ABORTED`, start lazy recovery and return the before image.
   - If there is no status record for the transaction in the Coordinator table:
     - If the transaction is expired, start lazy recovery and return the before image.
     - If the transaction is not expired, throw an `UncommittedRecordException`.

Additionally, this PR updates the code to perform lazy recovery in background threads. In cases where the recovered record needs to be written in the transaction, or where serializable validation (under SERIALIZABLE isolation level) is required, we need to wait for all lazy recoveries to complete before committing the transaction.

## Related issues and/or PRs

N/A

## Changes made

- 

## Checklist

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
@@ -24,10 +24,8 @@ import com.scalar.db.exception.transaction.UnknownTransactionStatusException;
 import com.scalar.db.exception.transaction.UnsatisfiedConditionException;
 import com.scalar.db.util.ScalarDbUtils;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 import org.slf4j.Logger;
@@ -51,24 +49,19 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
   private static final Logger logger = LoggerFactory.getLogger(ConsensusCommit.class);
   private final CrudHandler crud;
   private final CommitHandler commit;
-  private final RecoveryHandler recovery;
   private final ConsensusCommitMutationOperationChecker mutationOperationChecker;
   @Nullable private final CoordinatorGroupCommitter groupCommitter;
-  private Runnable beforeRecoveryHook;
 
   @SuppressFBWarnings("EI_EXPOSE_REP2")
   public ConsensusCommit(
       CrudHandler crud,
       CommitHandler commit,
-      RecoveryHandler recovery,
       ConsensusCommitMutationOperationChecker mutationOperationChecker,
       @Nullable CoordinatorGroupCommitter groupCommitter) {
     this.crud = checkNotNull(crud);
     this.commit = checkNotNull(commit);
-    this.recovery = checkNotNull(recovery);
     this.mutationOperationChecker = mutationOperationChecker;
     this.groupCommitter = groupCommitter;
-    this.beforeRecoveryHook = () -> {};
   }
 
   @Override
@@ -78,63 +71,18 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
 
   @Override
   public Optional<Result> get(Get get) throws CrudException {
-    get = copyAndSetTargetToIfNot(get);
-    try {
-      return crud.get(get);
-    } catch (UncommittedRecordException e) {
-      lazyRecovery(e);
-      throw e;
-    }
+    return crud.get(copyAndSetTargetToIfNot(get));
   }
 
   @Override
   public List<Result> scan(Scan scan) throws CrudException {
-    scan = copyAndSetTargetToIfNot(scan);
-    try {
-      return crud.scan(scan);
-    } catch (UncommittedRecordException e) {
-      lazyRecovery(e);
-      throw e;
-    }
+    return crud.scan(copyAndSetTargetToIfNot(scan));
   }
 
   @Override
   public Scanner getScanner(Scan scan) throws CrudException {
     scan = copyAndSetTargetToIfNot(scan);
-    Scanner scanner = crud.getScanner(scan);
-
-    return new Scanner() {
-      @Override
-      public Optional<Result> one() throws CrudException {
-        try {
-          return scanner.one();
-        } catch (UncommittedRecordException e) {
-          lazyRecovery(e);
-          throw e;
-        }
-      }
-
-      @Override
-      public List<Result> all() throws CrudException {
-        try {
-          return scanner.all();
-        } catch (UncommittedRecordException e) {
-          lazyRecovery(e);
-          throw e;
-        }
-      }
-
-      @Override
-      public void close() throws CrudException {
-        scanner.close();
-      }
-
-      @Nonnull
-      @Override
-      public Iterator<Result> iterator() {
-        return scanner.iterator();
-      }
-    };
+    return crud.getScanner(scan);
   }
 
   /** @deprecated As of release 3.13.0. Will be removed in release 5.0.0. */
@@ -143,12 +91,7 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
   public void put(Put put) throws CrudException {
     put = copyAndSetTargetToIfNot(put);
     checkMutation(put);
-    try {
-      crud.put(put);
-    } catch (UncommittedRecordException e) {
-      lazyRecovery(e);
-      throw e;
-    }
+    crud.put(put);
   }
 
   /** @deprecated As of release 3.13.0. Will be removed in release 5.0.0. */
@@ -165,12 +108,7 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
   public void delete(Delete delete) throws CrudException {
     delete = copyAndSetTargetToIfNot(delete);
     checkMutation(delete);
-    try {
-      crud.delete(delete);
-    } catch (UncommittedRecordException e) {
-      lazyRecovery(e);
-      throw e;
-    }
+    crud.delete(delete);
   }
 
   /** @deprecated As of release 3.13.0. Will be removed in release 5.0.0. */
@@ -196,12 +134,7 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
     upsert = copyAndSetTargetToIfNot(upsert);
     Put put = ConsensusCommitUtils.createPutForUpsert(upsert);
     checkMutation(put);
-    try {
-      crud.put(put);
-    } catch (UncommittedRecordException e) {
-      lazyRecovery(e);
-      throw e;
-    }
+    crud.put(put);
   }
 
   @Override
@@ -222,9 +155,6 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
 
       // If the condition is not specified, it means that the record does not exist. In this case,
       // we do nothing
-    } catch (UncommittedRecordException e) {
-      lazyRecovery(e);
-      throw e;
     }
   }
 
@@ -257,9 +187,6 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
     try {
       crud.readIfImplicitPreReadEnabled();
     } catch (CrudConflictException e) {
-      if (e instanceof UncommittedRecordException) {
-        lazyRecovery((UncommittedRecordException) e);
-      }
       throw new CommitConflictException(
           CoreError.CONSENSUS_COMMIT_CONFLICT_OCCURRED_WHILE_IMPLICIT_PRE_READ.buildMessage(),
           e,
@@ -267,6 +194,12 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
     } catch (CrudException e) {
       throw new CommitException(
           CoreError.CONSENSUS_COMMIT_EXECUTING_IMPLICIT_PRE_READ_FAILED.buildMessage(), e, getId());
+    }
+
+    try {
+      crud.waitForRecoveryCompletionIfNecessary();
+    } catch (CrudException e) {
+      throw new CommitException(e.getMessage(), e, getId());
     }
 
     commit.commit(crud.getSnapshot(), crud.isReadOnly());
@@ -293,22 +226,6 @@ public class ConsensusCommit extends AbstractDistributedTransaction {
   @VisibleForTesting
   CommitHandler getCommitHandler() {
     return commit;
-  }
-
-  @VisibleForTesting
-  RecoveryHandler getRecoveryHandler() {
-    return recovery;
-  }
-
-  @VisibleForTesting
-  void setBeforeRecoveryHook(Runnable beforeRecoveryHook) {
-    this.beforeRecoveryHook = beforeRecoveryHook;
-  }
-
-  private void lazyRecovery(UncommittedRecordException e) {
-    logger.debug("Recover uncommitted records: {}", e.getResults());
-    beforeRecoveryHook.run();
-    e.getResults().forEach(r -> recovery.recover(e.getSelection(), r));
   }
 
   private void checkMutation(Mutation mutation) throws CrudException {
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableMap;
 import com.scalar.db.api.ConditionBuilder;
 import com.scalar.db.api.Insert;
 import com.scalar.db.api.MutationCondition;
+import com.scalar.db.api.Operation;
 import com.scalar.db.api.Put;
 import com.scalar.db.api.PutBuilder;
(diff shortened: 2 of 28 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -269,6 +196,12 @@ public void commit() throws CommitException, UnknownTransactionStatusException {
           CoreError.CONSENSUS_COMMIT_EXECUTING_IMPLICIT_PRE_READ_FAILED.buildMessage(), e, getId());
     }
 
+    try {
+      crud.waitForRecoveryCompletionIfNecessary();
```
brfrn169, line 200: Wait lazy recoveries before committing the transaction if necessary.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -146,11 +151,15 @@ void readUnread(@Nullable Snapshot.Key key, Get get) throws CrudException {
...
-          snapshot.getId());
+      // Lazy recovery
+
+      if (key == null) {
+        // Only for a Get with index, the argument `key` is null. In that case, create a key from
+        // the result
+        key = new Snapshot.Key(get, result.get());
+      }
+
+      result = executeRecovery(key, get, result.get());
```
brfrn169, line 162: If reading an uncommitted record, the improved read process starts.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -257,15 +274,14 @@ private LinkedHashMap<Snapshot.Key, TransactionResult> scanInternal(Scan scan)
...
       Snapshot.Key key, Scan scan, TransactionResult result) throws CrudException {
+    Optional<TransactionResult> ret;
     if (!result.isCommitted()) {
-      throw new UncommittedRecordException(
-          scan,
-          result,
-          CoreError.CONSENSUS_COMMIT_READ_UNCOMMITTED_RECORD.buildMessage(),
-          snapshot.getId());
+      // Lazy recovery
+      ret = executeRecovery(key, scan, result);
```
brfrn169, line 279: Ditto. If reading an uncommitted record, the improved read process starts.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -438,6 +454,49 @@ private Get createGet(Snapshot.Key key) throws CrudException {
     return (Get) prepareStorageSelection(buildableGet.build());
   }
 
+  public void waitForRecoveryCompletionIfNecessary() throws CrudException {
+    for (RecoveryExecutor.Result recoveryResult : recoveryResults) {
+      try {
+        if (snapshot.containsKeyInWriteSet(recoveryResult.key)
+            || snapshot.containsKeyInDeleteSet(recoveryResult.key)
+            || snapshot.isValidationRequired()) {
+          recoveryResult.recoveryFuture.get();
```
brfrn169, line 487: If the recovered records are in writeSet or deleteSet, or if serializable validation is required, wait for the lazy recoveries to complete.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -0,0 +1,277 @@
...
+    // Recover the record
+    Future<Void> future =
+        executorService.submit(
+            () -> {
+              recovery.recover(selection, result, state);
+              return null;
+            });
+
+    return new Result(key, recoveredResult, future);
+  }
```
brfrn169, line 83: Here is the main part of the improved read logic.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -594,51 +653,25 @@ private Set<AndConditionSet> convertConjunctions(
...
-  private Selection prepareStorageSelection(Selection selection) throws CrudException {
+  private Selection prepareStorageSelection(Selection selection) {
     selection.clearProjections();
-    // Retrieve only the after images columns when including the metadata is disabled, otherwise
-    // retrieve all the columns
-    if (!isIncludeMetadataEnabled) {
-      LinkedHashSet<String> afterImageColumnNames =
-          getTransactionTableMetadata(selection).getAfterImageColumnNames();
-      selection.withProjections(afterImageColumnNames);
-    }
```
brfrn169, line 601: Stop specifying projections for the after images columns because before images are removed after https://github.com/scalar-labs/scalardb/pull/2787.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -0,0 +1,277 @@
...
+      Coordinator coordinator,
+      RecoveryHandler recovery,
+      TransactionTableMetadataManager tableMetadataManager,
+      int threadPoolSize) {
+    this.coordinator = Objects.requireNonNull(coordinator);
+    this.recovery = Objects.requireNonNull(recovery);
+    this.tableMetadataManager = Objects.requireNonNull(tableMetadataManager);
+    executorService =
+        Executors.newFixedThreadPool(
+            threadPoolSize,
```
feeblefakie: Separating configurations for the thread pool sizes for parallel execution and recovery execution might be pretty tricky for users. The inflexibility might also result in unexpected stalling since sometimes parallel execution threads are more required than recovery execution threads.
Having one thread pool and allocating threads to whichever needed is more flexible.
What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -0,0 +1,277 @@
...
+    if (!state.isPresent() && !recovery.isTransactionExpired(result)) {
+      throw new UncommittedRecordException(
+          selection,
+          result,
+          CoreError.CONSENSUS_COMMIT_READ_UNCOMMITTED_RECORD.buildMessage(),
+          transactionId);
+    }
+  }
+
+  private Optional<TransactionResult> createRolledBackRecord(
```
feeblefakie: `createRolledBackRecord` sounds a bit unnatural and not explicit to me.
How about this?
```suggestion
  private Optional<TransactionResult> createRecordFromBeforeImage(
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -0,0 +1,277 @@
...
+
+    TransactionTableMetadata transactionTableMetadata =
+        getTransactionTableMetadata(selection, transactionId);
+    LinkedHashSet<String> beforeImageColumnNames =
+        transactionTableMetadata.getBeforeImageColumnNames();
+    TableMetadata tableMetadata = transactionTableMetadata.getTableMetadata();
+
+    Map<String, Column<?>> columns = new HashMap<>();
+
+    extractAfterImageColumnsFromBeforeImage(columns, result, beforeImageColumnNames);
```
feeblefakie: Should it be `create` instead of `extract` since `extract` sounds like before image includes after image and contradicting.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -0,0 +1,277 @@
...
+
+    Optional<Key> clusteringKey = ScalarDbUtils.getClusteringKey(result, tableMetadata);
+    clusteringKey.ifPresent(k -> k.getColumns().forEach(c -> columns.put(c.getName(), c)));
+
+    addNullBeforeImageColumns(columns, beforeImageColumnNames, tableMetadata);
+
+    return Optional.of(new TransactionResult(new ResultImpl(columns, tableMetadata)));
+  }
+
+  private Optional<TransactionResult> createRolledForwardResult(
```
feeblefakie:
```suggestion
  private Optional<TransactionResult> createResultFromAfterImage(
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
feeblefakie: Let me clarify one thing.
Is this wait required for correctness?
Or, the recovered record will be used for sure in such cases, so it waits here to avoid unnecessary aborts?

feeblefakie (commented):
Left some questions. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
brfrn169: Renamed in 309e61fdfff36dca18f78e0b4f73386d6e54dcf3. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
brfrn169: Renamed in 309e61fdfff36dca18f78e0b4f73386d6e54dcf3. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
brfrn169: Renamed in 309e61fdfff36dca18f78e0b4f73386d6e54dcf3. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: I think for records in writeSet or deleteSet, it's required for correctness. And in case where serializable validation is required, it's to avoid unnecessary aborts.

For records in the writeSet or deleteSet, if we don’t wait for lazy recovery to complete, we might end up preparing records whose status is still PREPARED or DELETED, which is not the intended behavior.

Similarly, when serializable validation is required, if we don’t wait for lazy recovery, the validation could fail if it reads records with PREPARED or DELETED status. That’s why waiting is necessary in these cases.

komamitsu: > I think for records in writeSet or deleteSet, it's required for correctness.

@brfrn169 Can you give me a simple example for me?

brfrn169: > I think for records in writeSet or deleteSet, it's required for correctness.

@komamitsu Sorry, it was incorrect. As discussed, it's also required to avoid unnecessary aborts.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -314,4 +318,38 @@ public static int getNextTxVersion(@Nullable Integer currentTxVersion) {
       return currentTxVersion + 1;
     }
   }
+
+  static void createAfterImageColumnsFromBeforeImage(
```
komamitsu, line 322: [minor] This method name sounds to create a new collection. Something like `copy` or `extract` might be better?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
feeblefakie: Based on the offline discussion, I understand that it is not for correctness, but a nice-to-have for unnecessary aborts.
Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
brfrn169: @feeblefakie What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
brfrn169: As discussed, we've decided to use a cached thread pool for it. Fixed in 24c4822e3e3b4347ea83571a70ea56d46fb0fcdc. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -49,14 +49,12 @@ public class RecoveryExecutor implements AutoCloseable {
...
-      TransactionTableMetadataManager tableMetadataManager,
-      int threadPoolSize) {
+      TransactionTableMetadataManager tableMetadataManager) {
     this.coordinator = Objects.requireNonNull(coordinator);
     this.recovery = Objects.requireNonNull(recovery);
     this.tableMetadataManager = Objects.requireNonNull(tableMetadataManager);
     executorService =
-        Executors.newFixedThreadPool(
-            threadPoolSize,
+        Executors.newCachedThreadPool(
```
feeblefakie, line 57: So, it cannot set the max size?

feeblefakie (approved):
LGTM! Thank you!
So, the cache sizes is not configurable in `newCachedThreadPool` ?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
feeblefakie: `create` sounds fine to me. (because I proposed it and it's more like a behavior from external.)
I feel `copy` and `extract` are more like explaining the internal behavior.
`extract` also sounds to me like after image is a subset of before image.
We also discussed `build`, but the upper-level method uses `create`, so using a different action name seems to intentionally mean something.

komamitsu: I see. I'm okay with the current name. But, how about `createAfterImageColumnsFromBeforeImageOn(Map<String, Column<?>> columns, ...)` ?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
komamitsu: @brfrn169 Can you add a short comment about the background reason?

brfrn169: > Can you add a short comment about the background reason?

@komamitsu Yes.

For records in the writeSet or deleteSet, if we don’t wait for lazy recovery to complete, we might attempt to perform prepare-records on records whose status is still PREPARED or DELETED.

If we perform prepare-records on records that should be rolled forward, the prepare will actually succeed. However, this will create a PREPARED-state before image, which is unexpected. I don’t think it affects correctness, but it’s something we should avoid.

On the other hand, if we perform prepare-records on records that should be rolled back, the prepare will always fail, causing the transaction to abort.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
brfrn169: If we use the Java standard API `Executors.newCachedThreadPool()`, we cannot set a maximum pool size.

The implementation of `Executors.newCachedThreadPool()` is as follows:
```java
    public static ExecutorService newCachedThreadPool(ThreadFactory threadFactory) {
        return new ThreadPoolExecutor(0, Integer.MAX_VALUE,
                                      60L, TimeUnit.SECONDS,
                                      new SynchronousQueue<Runnable>(),
                                      threadFactory);
    }
```

If we want to set a maximum size, we can do something like this:
```java
                   new ThreadPoolExecutor(0, <MAX_SIZE>,
                                      60L, TimeUnit.SECONDS,
                                      new SynchronousQueue<Runnable>(),
                                      threadFactory);
```

But for now, as discussed, we’ve decided to use `Executors.newCachedThreadPool()` and see how it performs during benchmarking.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
komamitsu: Thanks! Looks good! Let's put it as a Java comment 🙇

brfrn169: Added Javadoc in 1305546679bcbe6ce028e0316a3374e4c2e0723e. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
brfrn169: Let's discuss this in a separate PR. Thanks!

PR #2811, opened 2025-06-20
Support one-phase commit optimization in Consensus Commit

brfrn169:
## Description

This PR adds support for one-phase commit optimization in Consensus Commit. Users can enable it by setting the property `scalar.db.consensus_commit.one_phase_commit.enabled` to `true`. The default is `false`.

The basic idea is that if the storage can execute all mutations in a transaction atomically, we don’t need two-phase commit mechanisms such as prepare-records, commit-state, and then commit-records. Instead, we can simply perform commit-records in a single phase.

## Related issues and/or PRs

N/A

## Changes made

Added inline comments. Please take a look for the details.

## Checklist

N/A

## Release notes

Added support for one-phase commit optimization in Consensus Commit, significantly improving performance by skipping prepare-records and commit-state when all mutations can be committed atomically. Enable this optimization with the `scalar.db.consensus_commit.one_phase_commit.enabled` property.

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -222,9 +222,9 @@ public class CommitHandler {
     }
 
     try {
-      // If the mutations can be grouped together, the mutations can be done in a single mutate API
-      // call, so we can one-phase commit the transaction
-      return mutationsGrouper.canBeGroupedTogether(
+      // If the mutations can be grouped altogether, the mutations can be done in a single mutate
+      // API call, so we can one-phase commit the transaction
+      return mutationsGrouper.canBeGroupedAltogether(
           Stream.concat(snapshot.getPutsInWriteSet().stream(), deletesInDeleteSet.stream())
               .collect(Collectors.toList()));
     } catch (ExecutionException e) {
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -2,7 +2,9 @@ package com.scalar.db.transaction.consensuscommit;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.errorprone.annotations.concurrent.LazyInit;
+import com.scalar.db.api.Delete;
 import com.scalar.db.api.DistributedStorage;
 import com.scalar.db.api.Mutation;
 import com.scalar.db.api.TransactionState;
@@ -24,6 +26,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Future;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 import org.slf4j.Logger;
@@ -38,6 +42,7 @@ public class CommitHandler {
   private final ParallelExecutor parallelExecutor;
   private final MutationsGrouper mutationsGrouper;
   protected final boolean coordinatorWriteOmissionOnReadOnlyEnabled;
+  private final boolean onePhaseCommitEnabled;
 
   @LazyInit @Nullable private BeforePreparationSnapshotHook beforePreparationSnapshotHook;
 
@@ -48,13 +53,15 @@ public class CommitHandler {
       TransactionTableMetadataManager tableMetadataManager,
       ParallelExecutor parallelExecutor,
       MutationsGrouper mutationsGrouper,
-      boolean coordinatorWriteOmissionOnReadOnlyEnabled) {
+      boolean coordinatorWriteOmissionOnReadOnlyEnabled,
+      boolean onePhaseCommitEnabled) {
     this.storage = checkNotNull(storage);
     this.coordinator = checkNotNull(coordinator);
     this.tableMetadataManager = checkNotNull(tableMetadataManager);
     this.parallelExecutor = checkNotNull(parallelExecutor);
     this.mutationsGrouper = checkNotNull(mutationsGrouper);
     this.coordinatorWriteOmissionOnReadOnlyEnabled = coordinatorWriteOmissionOnReadOnlyEnabled;
+    this.onePhaseCommitEnabled = onePhaseCommitEnabled;
   }
 
   /**
@@ -118,6 +125,16 @@ public class CommitHandler {
 
     Optional<Future<Void>> snapshotHookFuture = invokeBeforePreparationSnapshotHook(snapshot);
 
+    if (canOnePhaseCommit(snapshot)) {
+      try {
+        onePhaseCommitRecords(snapshot);
+        return;
+      } catch (Exception e) {
+        safelyCallOnFailureBeforeCommit(snapshot);
+        throw e;
+      }
+    }
+
     if (hasWritesOrDeletesInSnapshot) {
       try {
         prepareRecords(snapshot);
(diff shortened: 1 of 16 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -170,6 +187,52 @@ public void commit(Snapshot snapshot, boolean readOnly)
...
+      // If the mutations can be grouped together, the mutations can be done in a single mutate API
+      // call, so we can one-phase commit the transaction
+      return mutationsGrouper.canBeGroupedTogether(
+          Stream.concat(snapshot.getPutsInWriteSet().stream(), deletesInDeleteSet.stream())
+              .collect(Collectors.toList()));
+    } catch (ExecutionException e) {
+      throw new CommitException(
+          CoreError.CONSENSUS_COMMIT_COMMITTING_RECORDS_FAILED.buildMessage(), e, snapshot.getId());
+    }
+  }
```
brfrn169: These are the conditions for the one-phase commit optimization. Please see the comments for the details.

core/src/main/java/com/scalar/db/transaction/consensuscommit/MutationsGrouper.java
```diff
@@ -50,6 +51,41 @@ public List<List<Mutation>> groupMutations(Collection<Mutation> mutations)
     return groupToBatches.values().stream().flatMap(List::stream).collect(Collectors.toList());
   }
 
+  public boolean canBeGroupedTogether(Collection<Mutation> mutations) throws ExecutionException {
```
brfrn169: This method determines whether all the mutations can be grouped into a single group.

core/src/main/java/com/scalar/db/transaction/consensuscommit/OnePhaseCommitMutationComposer.java
```diff
@@ -0,0 +1,132 @@
...
+import com.scalar.db.api.PutBuilder;
+import com.scalar.db.api.TableMetadata;
+import com.scalar.db.api.TransactionState;
+import com.scalar.db.exception.storage.ExecutionException;
+import java.util.LinkedHashSet;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+@NotThreadSafe
+public class OnePhaseCommitMutationComposer extends AbstractMutationComposer {
```
brfrn169, line 24: `OnePhaseCommitMutationComposer` creates mutations for one-phase commit.

core/src/main/java/com/scalar/db/transaction/consensuscommit/MutationsGrouper.java
```diff
@@ -50,6 +51,41 @@ public List<List<Mutation>> groupMutations(Collection<Mutation> mutations)
...
+    if (mutations.isEmpty()) {
```
komamitsu:
```suggestion
    if (mutations.size() <= 1) {
```

This small optimization works?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -118,6 +125,16 @@ public void commit(Snapshot snapshot, boolean readOnly)
 
     Optional<Future<Void>> snapshotHookFuture = invokeBeforePreparationSnapshotHook(snapshot);
 
+    if (canOnePhaseCommit(snapshot)) {
+      try {
+        onePhaseCommitRecords(snapshot);
+        return;
```
komamitsu, line 131: Without the commit state, I think SSR will treat the write operations as an aborted transaction's ones and ignore them...

core/src/main/java/com/scalar/db/transaction/consensuscommit/MutationsGrouper.java
brfrn169: Yes. Fixed in 6a0b9c405c4b2c279681e35bc2f756da6c1e5c46. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: Yes, if SSR is enabled, one-phase commit must not be enabled. We should consider adding validation for this on the cluster side later.

komamitsu: Sounds good!

core/src/main/java/com/scalar/db/transaction/consensuscommit/MutationsGrouper.java
feeblefakie:
```suggestion
  public boolean canBeGroupedAltogether(Collection<Mutation> mutations) throws ExecutionException {
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -170,6 +187,52 @@ public void commit(Snapshot snapshot, boolean readOnly)
...
+      // set. So the result should always exist in the read set.
+      assert result != null;
+
+      if (!result.isPresent()) {
+        return false;
+      }
+    }
+
+    try {
+      // If the mutations can be grouped together, the mutations can be done in a single mutate API
```
feeblefakie:
```suggestion
      // If the mutations can be grouped altogether, the mutations can be done in a single mutate API
```

feeblefakie (approved):
I left minor naming suggestions, but LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -170,6 +187,52 @@ public void commit(Snapshot snapshot, boolean readOnly)
...
+    if (!snapshot.hasWritesOrDeletes()) {
+      return false;
+    }
+
+    List<Delete> deletesInDeleteSet = snapshot.getDeletesInDeleteSet();
+
+    // If a record corresponding to a delete in the delete set does not exist in the storage,　we
+    // cannot one-phase commit the transaction. This is because the storage does not support
+    // delete-if-not-exists semantics, so we cannot detect conflicts with other transactions.
+    for (Delete delete : deletesInDeleteSet) {
```
komamitsu, line 212: I wanted to know what would happen if I removed this block, so I removed the block and executed `ConsensusCommitNullMetadataIntegrationTestBase`. But all the tests passed unexpectedly. Is it difficult to write a test case where this block is necessary? (or I'm missing something....)

core/src/main/java/com/scalar/db/transaction/consensuscommit/MutationsGrouper.java
brfrn169: Fixed in 24c314ff6bc97eedf925c2fea259438b416555b6. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: Fixed in 24c314ff6bc97eedf925c2fea259438b416555b6. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169: For example, suppose a user tries to get record A but it doesn’t exist, and then the user attempts to delete that non-existing record.

In this case, under one-phase commit, we would need to perform a delete-if-not-exists operation for that record to detect conflicts—since another transaction might insert the record. However, because we don’t have delete-if-not-exists semantics in ScalarDB, we can’t do that. As a result, we cannot apply one-phase commit in this case.

brfrn169: By the way, one possible solution is to introduce a logical-delete state. In that case, we would perform a put-if-not-exists operation for the record representing that logical-delete state.

komamitsu: I see. Thanks for the explanation! It would be great if we add a test case for the situation later.

brfrn169: I think we have a test for it:
https://github.com/scalar-labs/scalardb/pull/2811/files#diff-23bc469b473150d216e1205dae9897c9ff85492d4846055cb5dccf6a60fefacfR1008-R1026

In this case, we can't perform one-phase commit.

komamitsu: Thanks, but unfortunately I wanted to find a proper integration test case to know actual use case.

PR #2827, opened 2025-07-06
Fix error handling for mutations in Cassandra

brfrn169:
## Description

This PR fixes error handling for mutations in Cassandra.

## Related issues and/or PRs

N/A

## Changes made

- 

## Checklist

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cassandra/BatchHandler.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/BatchHandler.java
@@ -11,8 +11,8 @@ import com.datastax.driver.core.exceptions.WriteTimeoutException;
 import com.google.common.annotations.VisibleForTesting;
 import com.scalar.db.api.Mutation;
 import com.scalar.db.common.CoreError;
+import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.exception.storage.NoMutationException;
-import com.scalar.db.exception.storage.RetriableExecutionException;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.List;
 import javax.annotation.concurrent.ThreadSafe;
@@ -48,12 +48,11 @@ public class BatchHandler {
    * must be for the same partition.
    *
    * @param mutations a list of {@code Mutation}s to execute
-   * @throws RetriableExecutionException if it fails, but it can be retried
    * @throws NoMutationException if at least one of conditional {@code Mutation}s fails because it
    *     didn't meet the condition
+   * @throws ExecutionException if the execution fails
    */
-  public void handle(List<? extends Mutation> mutations)
-      throws RetriableExecutionException, NoMutationException {
+  public void handle(List<? extends Mutation> mutations) throws ExecutionException {
     try {
       ResultSet results = execute(mutations);
       // it's for conditional update. non-conditional update always return true
@@ -64,16 +63,17 @@ public class BatchHandler {
       logger.warn("Write timeout happened during batch mutate operation", e);
       WriteType writeType = e.getWriteType();
       if (writeType == WriteType.BATCH_LOG) {
-        throw new RetriableExecutionException(
-            CoreError.CASSANDRA_LOGGING_FAILED_IN_BATCH.buildMessage(), e);
+        // A timeout occurred while the coordinator was waiting for the batch log replicas to
+        // acknowledge the log. Thus, the batch may or may not be applied.
+        throw new ExecutionException(CoreError.CASSANDRA_LOGGING_FAILED_IN_BATCH.buildMessage(), e);
       } else if (writeType == WriteType.BATCH) {
         logger.warn("Logging succeeded, but mutations in the batch partially failed", e);
       } else {
-        throw new RetriableExecutionException(
+        throw new ExecutionException(
             CoreError.CASSANDRA_OPERATION_FAILED_IN_BATCH.buildMessage(writeType), e);
       }
     } catch (RuntimeException e) {
-      throw new RetriableExecutionException(
+      throw new ExecutionException(
           CoreError.CASSANDRA_ERROR_OCCURRED_IN_BATCH.buildMessage(e.getMessage()), e);
     }
   }
--- a/core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
@@ -12,7 +12,6 @@ import com.scalar.db.api.Operation;
 import com.scalar.db.common.CoreError;
 import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.exception.storage.NoMutationException;
-import com.scalar.db.exception.storage.RetriableExecutionException;
 import javax.annotation.Nonnull;
 import javax.annotation.concurrent.ThreadSafe;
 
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/scalar/db/storage/cassandra/BatchHandler.java
```diff
@@ -64,16 +63,17 @@ public void handle(List<? extends Mutation> mutations)
       logger.warn("Write timeout happened during batch mutate operation", e);
       WriteType writeType = e.getWriteType();
       if (writeType == WriteType.BATCH_LOG) {
-        throw new RetriableExecutionException(
-            CoreError.CASSANDRA_LOGGING_FAILED_IN_BATCH.buildMessage(), e);
+        // A timeout occurred while the coordinator was waiting for the batch log replicas to
+        // acknowledge the log. Thus, the batch may or may not be applied.
+        throw new ExecutionException(CoreError.CASSANDRA_LOGGING_FAILED_IN_BATCH.buildMessage(), e);
```
brfrn169, line 68: `RetriableExecutionException` should be thrown when a mutation fails due to a conflict error and it is certain that the mutation was not applied.
If it is unknown whether the mutation was applied, an `ExecutionException` should be thrown instead.

core/src/main/java/com/scalar/db/storage/cassandra/BatchHandler.java
```diff
@@ -64,16 +63,17 @@ public void handle(List<? extends Mutation> mutations)
...
       } else if (writeType == WriteType.BATCH) {
         logger.warn("Logging succeeded, but mutations in the batch partially failed", e);
       } else {
-        throw new RetriableExecutionException(
+        throw new ExecutionException(
```
brfrn169, line 72: Ditto.

core/src/main/java/com/scalar/db/storage/cassandra/BatchHandler.java
```diff
@@ -64,16 +63,17 @@ public void handle(List<? extends Mutation> mutations)
...
             CoreError.CASSANDRA_OPERATION_FAILED_IN_BATCH.buildMessage(writeType), e);
       }
     } catch (RuntimeException e) {
-      throw new RetriableExecutionException(
+      throw new ExecutionException(
```
brfrn169, line 76: Ditto.

core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
```diff
@@ -46,7 +45,7 @@ public ResultSet handle(Operation operation) throws ExecutionException {
     } catch (WriteTimeoutException e) {
       if (e.getWriteType() == WriteType.CAS) {
         // retry needs to be done if applications need to do the operation exactly
-        throw new RetriableExecutionException(
+        throw new ExecutionException(
```
brfrn169, line 47: Ditto.

core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
```diff
@@ -56,7 +55,7 @@ public ResultSet handle(Operation operation) throws ExecutionException {
               CoreError.CASSANDRA_WRITE_TIMEOUT_IN_LEARN_PHASE_IN_MUTATION.buildMessage(), e);
         } else {
           // retry needs to be done if applications need to do the operation exactly
-          throw new RetriableExecutionException(
+          throw new ExecutionException(
```
brfrn169, line 56: Ditto.

core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
```diff
@@ -66,7 +65,7 @@ public ResultSet handle(Operation operation) throws ExecutionException {
             CoreError.CASSANDRA_WRITE_TIMEOUT_WITH_OTHER_WRITE_TYPE_IN_MUTATION.buildMessage(), e);
       }
     } catch (RuntimeException e) {
-      throw new RetriableExecutionException(
+      throw new ExecutionException(
```
brfrn169, line 66: Ditto.

core/src/main/java/com/scalar/db/storage/cassandra/MutateStatementHandler.java
```diff
@@ -55,8 +53,7 @@ public ResultSet handle(Operation operation) throws ExecutionException {
           throw new ReadRepairableExecutionException(
```
komamitsu, line 53: This is thrown when the write type is SIMPLE and the mutation has conditions. But, I think it's possible the cause of the failure is not related to the condition, and throwing `ExecutionException` might be safer? Sorry, I don't correctly understand what ReadRepairableExecutionException means and how to handle it, and I might be missing something.

brfrn169: `ReadRepairableExecutionException` is located in the `com.scalar.db.storage.cassandra` package and is specific to the Cassandra adaptor.
Currently, we don’t handle this exception explicitly in the transaction layer (e.g., Consensus Commit).

The Cassandra adaptor throws this exception when the error can be repaired by a following read. I think it might be useful for users who use the Cassandra adaptor directly through the Storage API, so I decided to keep throwing this exception here.

What do you think?

komamitsu: Thanks for the explanation about `ReadRepairableExecutionException`. Understood.

As for my question above, I'm still wondering if the following case could occur (note: the users are using Storage API) :
- User A tries to insert a record with `pk1` using condition `IF_NOT_EXISTS`
  - The operation is applied on Cassandra side, but the response is delayed
- User B reads the record `pk1` and deletes it
- The first insert operation's response is timed out
- User A catches `ReadRepairableExecutionException` since it has the condition
- User A reads the record `pk1` and confirms it doesn't exist
- User A re-inserts the record `pk1`, although it's supposed to be deleted by User B

Hmm. I'm not sure this is a ReadRepairableExecutionException specific issue, though...

brfrn169: I think that case could occur with the the Storage API, too.

The Storage API guarantees linearizability for individual operations but does not ensure serializability across multiple operations. Therefore, it’s the responsibility of application developers to handle such cases appropriately.

komamitsu: Ah, sorry. I was confused... Thanks for pointing it out.

> The Cassandra adaptor throws this exception when the error can be repaired by a following read.

You expect the client will re-read the record and decide the next operation, when catching the exception? If so, it sounds to be similar to the error handling of `ExecutionException` to me. Is there any significant difference?

BTW, I couldn't find any document and guidance for `ReadRepairableExecutionException`, so I guess the users don't know what to do when catching the exception...

brfrn169: > You expect the client will re-read the record and decide the next operation, when catching the exception? If so, it sounds to be similar to the error handling of ExecutionException to me. Is there any significant difference?

I think it depends on the application. If the application needs to guarantee that the record is repaired, it should re-read the record automatically. If not, the record will eventually be repaired the next time it is read.

> BTW, I couldn't find any document and guidance for ReadRepairableExecutionException, so I guess the users don't know what to do when catching the exception...

Yes, that’s correct… Maybe we should mention it when we add a dedicated Cassandra adaptor page in the future.

komamitsu (approved):
LGTM! 👍

PR #2897, opened 2025-07-22
Fix Cassandra case insensitive issue

KodaiD:
## Description

This PR fixes an issue where `CassandraAdmin` fails to alter a table due to case insensitivity in the table name.

The current implementation of `CassandraAdmin` does not handle case sensitivity correctly when altering a table, leading to a not found error. For example, if a table is created with the name `Test_table`, attempting to alter it with the name `test_table` will result in an error because the table name is not found.

The fix is to enclose the table name in `""` when altering the table, ensuring that the case is preserved. Also, this PR updated the integration tests to cover the case.

## Related issues and/or PRs

N/A

## Changes made

- Fixed to use `quoteIfNecessary` when altering a table in `CassandraAdmin`.
- Fixed to remove `quoteIfNecessary` where it is not necessary in `CassandraAdmin`.
- Updated the namespace and table names in the integration tests to cover the case sensitivity.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
+++ b/core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
@@ -139,7 +139,7 @@ public class CassandraAdmin implements DistributedStorageAdmin {
     String insertQuery =
         QueryBuilder.insertInto(
                 quoteIfNecessary(metadataKeyspace), quoteIfNecessary(NAMESPACES_TABLE))
-            .value(NAMESPACES_NAME_COL, quoteIfNecessary(keyspace))
+            .value(NAMESPACES_NAME_COL, keyspace)
             .toString();
     clusterManager.getSession().execute(insertQuery);
   }
@@ -180,7 +180,7 @@ public class CassandraAdmin implements DistributedStorageAdmin {
     String deleteQuery =
         QueryBuilder.delete()
             .from(quoteIfNecessary(metadataKeyspace), quoteIfNecessary(NAMESPACES_TABLE))
-            .where(QueryBuilder.eq(NAMESPACES_NAME_COL, quoteIfNecessary(keyspace)))
+            .where(QueryBuilder.eq(NAMESPACES_NAME_COL, keyspace))
             .toString();
     clusterManager.getSession().execute(deleteQuery);
   }
@@ -349,7 +349,7 @@ public class CassandraAdmin implements DistributedStorageAdmin {
       String query =
           QueryBuilder.select(NAMESPACES_NAME_COL)
               .from(quoteIfNecessary(metadataKeyspace), quoteIfNecessary(NAMESPACES_TABLE))
-              .where(QueryBuilder.eq(NAMESPACES_NAME_COL, quoteIfNecessary(namespace)))
+              .where(QueryBuilder.eq(NAMESPACES_NAME_COL, namespace))
               .toString();
       ResultSet resultSet = clusterManager.getSession().execute(query);
 
@@ -396,7 +396,7 @@ public class CassandraAdmin implements DistributedStorageAdmin {
     }
     try {
       String alterTableQuery =
-          SchemaBuilder.alterTable(namespace, table)
+          SchemaBuilder.alterTable(quoteIfNecessary(namespace), quoteIfNecessary(table))
               .addColumn(columnName)
               .type(toCassandraDataType(columnType))
               .getQueryString();
--- a/integration-test/src/main/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
+++ b/integration-test/src/main/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
@@ -36,13 +36,13 @@ public abstract class DistributedStorageAdminIntegrationTestBase {
       LoggerFactory.getLogger(DistributedStorageAdminIntegrationTestBase.class);
 
   private static final String TEST_NAME = "storage_admin";
-  private static final String NAMESPACE1 = "int_test_" + TEST_NAME + "1";
-  private static final String NAMESPACE2 = "int_test_" + TEST_NAME + "2";
-  private static final String NAMESPACE3 = "int_test_" + TEST_NAME + "3";
-  private static final String TABLE1 = "test_table1";
-  private static final String TABLE2 = "test_table2";
-  private static final String TABLE3 = "test_table3";
-  private static final String TABLE4 = "test_table4";
+  private static final String NAMESPACE1 = "Int_test_" + TEST_NAME + "1";
+  private static final String NAMESPACE2 = "Int_test_" + TEST_NAME + "2";
+  private static final String NAMESPACE3 = "Int_test_" + TEST_NAME + "3";
+  private static final String TABLE1 = "Test_table1";
+  private static final String TABLE2 = "Test_table2";
+  private static final String TABLE3 = "Test_table3";
+  private static final String TABLE4 = "Test_table4";
   private static final String COL_NAME1 = "c1";
   private static final String COL_NAME2 = "c2";
   private static final String COL_NAME3 = "c3";
--- a/core/src/test/java/com/scalar/db/storage/cassandra/CassandraAdminTest.java
+++ b/core/src/test/java/com/scalar/db/storage/cassandra/CassandraAdminTest.java
@@ -229,7 +229,7 @@ public class CassandraAdminTest {
         QueryBuilder.insertInto(
                 quoteIfNecessary(METADATA_KEYSPACE),
                 quoteIfNecessary(CassandraAdmin.NAMESPACES_TABLE))
-            .value(CassandraAdmin.NAMESPACES_NAME_COL, quoteIfNecessary(keyspace))
+            .value(CassandraAdmin.NAMESPACES_NAME_COL, keyspace)
             .toString();
     verify(cassandraSession).execute(query);
   }
@@ -569,7 +569,7 @@ public class CassandraAdminTest {
             .from(
                 quoteIfNecessary(METADATA_KEYSPACE),
                 quoteIfNecessary(CassandraAdmin.NAMESPACES_TABLE))
-            .where(QueryBuilder.eq(CassandraAdmin.NAMESPACES_NAME_COL, quoteIfNecessary(keyspace)))
+            .where(QueryBuilder.eq(CassandraAdmin.NAMESPACES_NAME_COL, keyspace))
             .toString();
     verify(cassandraSession).execute(query);
   }


core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -139,7 +139,7 @@ private void upsertIntoNamespacesTable(String keyspace) {
     String insertQuery =
         QueryBuilder.insertInto(
                 quoteIfNecessary(metadataKeyspace), quoteIfNecessary(NAMESPACES_TABLE))
-            .value(NAMESPACES_NAME_COL, quoteIfNecessary(keyspace))
+            .value(NAMESPACES_NAME_COL, keyspace)
```
KodaiD, line 142: Remove `quoteIfNecessary` because it is not necessary for values that are not identifiers. Also, we want to handle ScalarDB namespaces without quotes. For example, we want to handle `scalar_table` as `scalar_table` and not `"scalar_table"`.

KodaiD: Without this fix, some integration test cases fails as follows:

```
java.lang.AssertionError: 
Expecting ArrayList:
  ["scalardb_storage_admin",
    ""Int_test_storage_admin1"",
    ""Int_test_storage_admin2""]
to contain only:
  ["Int_test_storage_admin1", "Int_test_storage_admin2", "scalardb_storage_admin"]
element(s) not found:
  ["Int_test_storage_admin1", "Int_test_storage_admin2"]
and element(s) not expected:
  [""Int_test_storage_admin1"", ""Int_test_storage_admin2""]
	at com.scalar.db.api.DistributedStorageAdminIntegrationTestBase.getNamespaceNames_ShouldReturnCreatedNamespaces(DistributedStorageAdminIntegrationTestBase.java:878)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -396,7 +396,7 @@ public void addNewColumnToTable(
     }
     try {
       String alterTableQuery =
-          SchemaBuilder.alterTable(namespace, table)
+          SchemaBuilder.alterTable(quoteIfNecessary(namespace), quoteIfNecessary(table))
```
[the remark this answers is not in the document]
KodaiD: Fixed in [13d2178](https://github.com/scalar-labs/scalardb/pull/2897/commits/13d2178e1bd1bea0166236120e50b736c7c5fbaa).

integration-test/src/main/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
```diff
@@ -36,28 +36,28 @@ public abstract class DistributedStorageAdminIntegrationTestBase {
...
-  private static final String COL_NAME7 = "c7";
-  private static final String COL_NAME8 = "c8";
-  private static final String COL_NAME9 = "c9";
-  private static final String COL_NAME10 = "c10";
-  private static final String COL_NAME11 = "c11";
-  private static final String COL_NAME12 = "c12";
-  private static final String COL_NAME13 = "c13";
-  private static final String COL_NAME14 = "c14";
-  private static final String COL_NAME15 = "c15";
+  private static final String NAMESPACE1 = "Int_test_" + TEST_NAME + "1";
```
komamitsu: CQL seems to have [the tricky identifier matching rule](https://cassandra.apache.org/doc/latest/cassandra/developing/cql/cql_singlefile.html#identifiers). I agree that capitalizing namespace/table/column names improves the test cases. But how about leaving or adding some non-capitalized identifiers to cover a wider range of cases (just in case)?

KodaiD: @komamitsu Thanks for your feedback! I understand your concerns about the test coverage. I believe using the object names that include both upper and lower case letters (like "Column1") can help us verify case-sensitivity more thoroughly. That's why I renamed the test column names to "Column1" instead of "C1". What do you think about this approach?

komamitsu: Actually, I'm not familiar with CQL, so I might be missing something, but I found this:

> A fully lowercase quoted identifier that matches [a-zA-Z0-9_]* is equivalent to the unquoted identifier obtained by removing the double-quote (so "myid" is equivalent to myid and to myId but different from "myId").

So, it seems fully lowercase identifiers might be treated specially (if it's quoted) and it might be worth using such identifies in the test cases just in case.

KodaiD: @komamitsu Finally, I get your point. Then, how about preparing two `DistributedStorageAdminIntegrationTestBase` implementations, one having the test cases with upper and lower case letters in object names, and the other having the test cases with only lower case letters in object names? For example, we can have `CassandraAdminIntegrationTestWithLowerCaseObjectNames` in addition to `CassandraAdminIntegrationTest`.

komamitsu: Yeah, that's an option. But I guess, to prevent `CassandraAdminIntegrationTestWithLowerCaseObjectNames` (or `DistributedStorageAdminIntegrationTestWithLowerCaseObjectNames`?) from having duplicated test cases, probably we need to access the column names via methods like `protected String getColumnName1()` instead of directly using `COL_NAME1`. It might be a bit too much, so I think it's easier to mix full lower case identifier with the capitalized identifiers, although it's a bit ugly. But, either is fine with me!

KodaiD: @komamitsu Because the object names used in the test cases depend on each test case (e.g., some of the test cases use only `NAMESPACE1`), it might be better to prepare another `DistributedStorageAdminIntegrationTestBase` implementation that uses only lower case letters in object names. 

@brfrn169 I'm planning to make a different PR for that. Is it OK? For now, I will keep the test cases to use object names including both upper and lower case letters.

komamitsu: Good catch! Agreed.

> so I think it's easier to mix full lower case identifier with the capitalized identifiers

I meant to mix full lower case identifier and capitalized identifier by `mix`. It may have confused you.

Torch3333: Depending on the option you choose in the end, the Junit annotation `@ParametrizedClass` may be useful. 
https://scalar-labs.slack.com/archives/CDRAHRX6X/p1742970297385049

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
```diff
@@ -271,15 +271,21 @@ private TableMetadata createTableMetadata(com.datastax.driver.core.TableMetadata
...
-      String clusteringColumnName = metadata.getClusteringColumns().get(i).getName();
+      String clusteringColumnName =
+          unquoteIfNecessary(metadata.getClusteringColumns().get(i).getName());
       ClusteringOrder clusteringOrder = metadata.getClusteringOrder().get(i);
       builder.addClusteringKey(clusteringColumnName, convertOrder(clusteringOrder));
     }
-    metadata.getIndexes().forEach(i -> builder.addSecondaryIndex(i.getTarget()));
+    metadata
+        .getIndexes()
+        .forEach(i -> builder.addSecondaryIndex(unquoteIfNecessary(i.getTarget())));
```
brfrn169: Why do we need to unquote here?

integration-test/src/main/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
brfrn169: Since `DistributedStorageAdminIntegrationTestBase` is used not only for the Cassandra adapter but also for other adapters, we should add separate test cases for capitalized namespace, table, and column names instead of modifying the existing ones.

As for how to implement that, we have a few options, as suggested by @komamitsu and @Torch3333.

BTW, shouldn’t we also check the same for the operations in `DistributedStorage`?

KodaiD: @brfrn169 
> Since DistributedStorageAdminIntegrationTestBase is used not only for the Cassandra adapter but also for other adapters, we should add separate test cases for capitalized namespace, table, and column names instead of modifying the existing ones.

Other databases such as Oracle and PostgreSQL also can have the case-insensitive issue (they treat object names as case-insensitive without double quotes). Therefore, I think it is better to have more complex object names that include both upper and lower case letters in `DistributedStorageAdminIntegrationTestBase`. In addition to that, because Cassandra has the tricky rule @komamitsu mentioned, we should also test the Cassandra adapter with object names that include only lower case letters. What do you think?

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
KodaiD: @brfrn169  It because the table metadata returned by `getTableMetadata` can have the quoted object names without this fix as follows (in secondaryIndexNames):

```
TableMetadata{columnNames=[Column2, Column1, Column4, Column3, Column10, Column11, Column12, Column13, Column14, Column5, Column6, Column7, Column8, Column9], columnDataTypes={Column2=TEXT, Column1=INT, Column4=INT, Column3=TEXT, Column10=BOOLEAN, Column11=BLOB, Column12=DATE, Column13=TIME, Column14=TIMESTAMPTZ, Column5=INT, Column6=TEXT, Column7=BIGINT, Column8=FLOAT, Column9=DOUBLE}, partitionKeyNames=[Column2, Column1], clusteringKeyNames=[Column4, Column3], clusteringOrders={Column3=DESC, Column4=ASC}, secondaryIndexNames=["Column6", "Column5"], encryptedColumnNames=[]}
```

But it seems that except for the secondary index names, the other object names are not quoted. So, I fixed the code to unquote only the secondary index names in dfaabf1. It seems that `i.getTarget()` returns the quoted object names.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
KodaiD: > BTW, shouldn’t we also check the same for the operations in DistributedStorage?

Thanks for pointing it out. I fixed the object names in DistributedStorage test as well in [5f747ea](https://github.com/scalar-labs/scalardb/pull/2897/commits/5f747eaf4d218e515f424aad826334e66dba7207).

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
brfrn169: Oh, ok. Umm... it looks a bug on Cassandra.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageAdminIntegrationTestBase.java
KodaiD: @komamitsu 
I discussed with @brfrn169, and we agreed that it is better to prepare another test that extends `DistributedStorageAdminIntegrationTestBase` to use more complex object names that include both upper and lower case letters. I'll create a new PR for that. For now, I will keep the object names in the test cases to use only lower case letters.

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
komamitsu: Your change looks good. But let me confirm one thing: this change doesn't break any backward compatibility, right? (Because without the fix, namespace starting from an uppercase letter can't be accessed by users)

brfrn169 (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/cassandra/CassandraAdmin.java
KodaiD: @komamitsu I don't fully get your point, but this fix is one of the bug fixes for the case-insensitive issue. This change isn't because of a bug fix for ALTER. It's one of the changes we need to fix the case-insensitive issue.

KodaiD: Sorry, I get your point. Without this fix, users cannot access the namespace having an upper case letter due to the bug. Also, this fix is for the namespace table in v4. So, I think this change does not break backward compatibility.

komamitsu: Sorry for my lack of explanation. I should've said something like "Let's say a user created a namespace name `Abc` before this fix was applied, and ...". And thanks for the explanation. Glad to hear that!

komamitsu (approved):
LGTM, thank you!

PR #3001, opened 2025-09-04
Support TiDB

Torch3333:
## Description

Adds support for TiDB.

## Related issues and/or PRs

> If this PR addresses or references any issues and/or other PRs, list them here. For example, "Fixes #123", "Related to #456", or "Addresses #789".

## Changes made

> Outline the specific changes made in this pull request. Include relevant details, such as added features, bug fixes, code refactoring, or improvements.

## Checklist

> Provide any additional information or notes that may be relevant to the reviewers or stakeholders.

## Release notes

> Provide release note text for this PR based on the guidelines at [Guidelines for Writing Release Notes](https://developers.scalar-labs.com/docs/style-guide/release-notes/).
> 
> For example:
> 
> Added validation for the Consensus Commit mutation operation.


core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcEnv.java
```diff
@@ -2,13 +2,15 @@
 
 import com.scalar.db.config.DatabaseConfig;
 import java.util.Properties;
+import org.assertj.core.util.Strings;
```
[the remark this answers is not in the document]
Torch3333: We use AssertJ heavily in test code so I think it's fine as is

core/src/integration-test/java/com/scalar/db/storage/jdbc/JdbcAdminImportTestUtils.java
```diff
@@ -763,6 +770,15 @@ private boolean isMariaDB() {
     }
   }
 
+  private boolean isTiDB() {
+    try (Connection connection = dataSource.getConnection()) {
+      String version = connection.getMetaData().getDatabaseProductVersion();
+      return version.contains("TiDB");
+    } catch (SQLException e) {
+      throw new RuntimeException("Get database product version failed");
```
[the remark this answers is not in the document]
Torch3333: Revised in https://github.com/scalar-labs/scalardb/pull/3001/commits/9213310356797612202c8bb841dbd61f109e632d

brfrn169:
> Since the ScalarDB transaction API sets a SERIALIZABLE isolation level by default not supported by TiDB

@Torch3333 I don’t think we set the SERIALIZABLE isolation level by default. We only set the isolation level when users specify the `scalar.db.jdbc.isolation_level` property. What do you think?

feeblefakie:
@Torch3333 What would happen if users don't specify `scalardb.jdbc.isolation_level=REPEATABLE_READ` or READ-COMMITTED?

BTW, TiDB seems to be linearizable, independent of isolation levels. So, any isolation levels can probaly be fine for correctness. (cc: @brfrn169 )
https://docs.pingcap.com/tidb/stable/system-variables/#tidb_guarantee_linearizability-new-in-v50

Torch3333:
@brfrn169 @feeblefakie @komamitsu 

> @Torch3333 I don’t think we set the SERIALIZABLE isolation level by default. We only set the isolation level when users specify the scalar.db.jdbc.isolation_level property. What do you think?

You are right. I mistakenly thought ScalarDB was setting the JDBC transaction isolation level for all the transaction managers, but in fact, it is only set for the JDBC transaction manager, not for Consensus Commit. 
As discussed with @brfrn169  and @feeblefakie, we decided that setting it for the JDBC transaction manager was not required. From now on, the storage default isolation level will be used. For better clarity for our users, I made these changes in a separate PR #3076  

> What would happen if users don't specify `scalardb.jdbc.isolation_level=REPEATABLE_READ` or READ-COMMITTED?

TiDB uses the REPEATABLE-READ as the default level. READ-COMMITED is also an accepted value. Setting SERIALIZABLE will cause the driver to throw this error. The documentation explains it [here](https://docs.pingcap.com/tidbcloud/dev-guide-third-party-tools-compatibility/#the-read-uncommitted-and-serializable-isolation-levels-are-not-supported).
```
The isolation level 'SERIALIZABLE' is not supported. Set tidb_skip_isolation_level_check=1 to skip this error
```

Other than that, I finally found a [document](https://docs.pingcap.com/tidbcloud/dev-guide-third-party-tools-compatibility/#compatibility-with-mysql-jdbc) listing the features of the MySQL Connector/J driver that TiDB does not support. As far as my understanding goes, I don't think ScalarDB uses any of the incompatibilities, so this confirms that using this driver instead of the TiDB modified version of Connector/J is fine.

Torch3333:
I will put this PR in draft until this blocking dependent PR is merged. 
https://github.com/scalar-labs/scalardb/pull/3076

Torch3333:
@komamitsu You raised a point during the demo meeting more than 2 weeks ago about how the driver handles load balancing. There seems to be a [library](https://docs.pingcap.com/tidb/stable/dev-guide-choose-driver-or-orm/#java-client-load-balancing) to handle just that, but I will investigate it once other higher-priority tasks are completed.

feeblefakie:
@Torch3333 This can be opened again now?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineTidb.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.scalar.db.common.CoreError;
+import com.scalar.db.io.DataType;
+
+/**
+ * This implements a RdbEngine for TiDB that extends MySQL one. TiDB is MySQL compatible and uses
+ * the same connection string, so special handling is needed to instantiate it, cf. {@link
+ * RdbEngineFactory#create(JdbcConfig)}
+ */
+public class RdbEngineTidb extends RdbEngineMysql {
```
Torch3333, line 11: Altering the column type from BLOB to TEXT is not supported in TiDB, so I needed to add an `RdbEngine` for TiDB

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineFactory.java
```diff
@@ -32,4 +35,30 @@ public static RdbEngineStrategy create(JdbcConfig config) {
...
+      } else {
+        return mysqlEngine;
+      }
+    } catch (SQLException e) {
+      // We can't throw a checked exception here because it would break backward compatibility since
+      // the calling method is executed in constructor of JdbcAdmin or JdbcService
+      throw new RuntimeException(
+          CoreError.JDBC_MYSQL_GETTING_CONNECTION_METADATA_FAILED.buildMessage(e), e);
+    }
+  }
```
Torch3333, line 63: Creating a `RdbEngine` for TiDB and MySQL needed special handling since TiDB uses the same connection string as MySQL, we can't distinguish between them without establishing a connection to check the metadata.

Torch3333 (commented):
The PR is ready for review again. 
Please review everything once again, since there are numerous new files since your last review.
See my explanation below about the changes.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -808,6 +808,12 @@ public enum CoreError implements ScalarDbError {
       "Db2 does not support column type conversion from %s to %s",
       "",
       ""),
+  JDBC_TIDB_UNSUPPORTED_COLUMN_TYPE_CONVERSION(
+      Category.USER_ERROR,
+      "0241",
+      "TiDB does not support column type conversion from %s to %s",
```
komamitsu, line 825: [Future improvement idea] This is not related to this PR at all, but just a memo for future improvements. I think we can probably generalize these kinds of error messages to parameterize the RDBMS name like "%s doesn't support ...." so that we don't need to add new error messages for each database.

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineFactory.java
```diff
@@ -32,4 +35,30 @@ public static RdbEngineStrategy create(JdbcConfig config) {
...
+   *
+   * @param config the config
+   * @return a {@link RdbEngineMysql} or {@link RdbEngineTidb}.
+   */
+  private static RdbEngineStrategy createMysqlOrTidbEngine(JdbcConfig config) {
+    RdbEngineMysql mysqlEngine = new RdbEngineMysql(config);
+    try (BasicDataSource dataSource = JdbcUtils.initDataSourceForAdmin(config, mysqlEngine);
+        Connection connection = dataSource.getConnection()) {
+      String version = connection.getMetaData().getDatabaseProductVersion();
+      if (version.contains("TiDB")) {
```
komamitsu, line 52: Determining the actual database based on the version string in the production code feels a bit naive approach to me, since I'm not sure the output string is guaranteed to remain unchanged. BTW, introducing a specific JDBC contact type (e.g., `tidb:`) might be a more straightforward approach, similar to how we handle MariaDB or YugabyteDB. What do you think?

Torch3333: Contrary to MariaDB and YugabyteDB, which use a dedicated connection string prefix such as `mariadb:` or `yugabytedb:`. TiDB continues to use `mysql:` even for its in-house driver. So if we want to use a `tidb:`prefix, before establishing a connection to TiDB, ScalarDB needs to replace the `tidb:` prefix with `mysql:` as this is the only prefix that the JDBC driver accepts. Replacing the prefix itself is not an issue. But I am worried about misconfiguration, since the user would need to explicitly check the ScalarDB docs to know about such a prefix. The likelihood that a user will configure their TiDB connection to start with `mysql` is substantial. 

On the other hand, `connection.getMetaData().getDatabaseProductVersion()` returns the value of the system variable [version](https://docs.pingcap.com/tidb/stable/system-variables/#version), which is documented to always contain `TiDB`. 
 
An alternative to parsing the version is to check whether [TiDB-specific system variables](https://docs.pingcap.com/tidb/stable/system-variables) exist. This may be more robust. 
```
SHOW VARIABLES LIKE 'tidb%';
```
image
 
What do you think?

komamitsu: Thanks for the detailed background. I understand your concerns about the `tidb:` JDBC protocol name. I think there are pros and cons to both approaches (w/ `mysql:` and w/ `tidb:`). I'm okay with the current approach for now. Probably, we can revisit this topic if/when TiDB JDBC driver supports the `tidb:` prefix in the future.

PR #3133, opened 2025-11-08
Fix secondary index behavior in Consensus Commit

brfrn169:
## Description

Currently, when users specify an index key for Get or Scan operations with the Consensus Commit transaction manager, we do not add any additional conditions for the before columns. However, as described in the following Javadoc, we need to add the same conditions for the before columns to ensure correct results:

https://github.com/scalar-labs/scalardb/blob/cc720dbf9bc49e4a4db01cc42bc254cf8f2415b8/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java#L407-L472

However, if we simply add the same conditions to the before columns, the secondary index will no longer be used, which can significantly degrade performance.

Therefore, after further discussion, we decided to redefine Get/Scan operations using secondary indexes as eventually consistent operations. This means they:

- Do NOT include conditions on before columns
- Can efficiently use secondary indexes
- May return slightly stale data

Since SERIALIZABLE isolation level requires the strongest consistency guarantees and cannot tolerate eventually consistent reads, we now prohibit secondary index operations when using SERIALIZABLE isolation.

**ADDED**:
I’ve summarized the behavior here:
https://github.com/scalar-labs/scalardb/pull/3133#discussion_r2517397932

## Related issues and/or PRs

N/A

## Changes made

1. New Operation Checker (`ConsensusCommitOperationChecker`)
- Renamed and expanded `ConsensusCommitMutationOperationChecker` to check all operations (Get, Scan, and Mutations)
- Added validation for Get operations:
  - SERIALIZABLE check: Rejects Get operations using secondary indexes (unless the index column is part of the primary key)
- Added validation for Scan operations:
  - SERIALIZABLE check: Rejects Scan operations using secondary indexes (unless the index column is part of the primary key)
  - SERIALIZABLE check: For cross-partition scans (ScanAll), rejects conditions on any indexed columns

2. Updated Conjunction Conversion Logic
- Modified `ConsensusCommitUtils.convertConjunctions()`:
  - Added convertIndexedColumns parameter to control whether conditions on indexed columns should be converted to include before-column conditions
  - For cross-partition scans (ScanAll), indexed column conditions are NOT converted (passed as-is)
  - For partition scans and Gets, all conditions including indexed columns ARE converted

## Checklist

N/A

## Release notes

Prohibited Get and Scan operations using secondary indexes when the isolation level is SERIALIZABLE in the Consensus Commit transaction manager, as these operations are now defined as eventually consistent and cannot guarantee the strict consistency required by SERIALIZABLE isolation.

Applications that previously issued the following operations under SERIALIZABLE isolation will now be rejected with `IllegalArgumentException`. Such operations must either be reissued under a non-SERIALIZABLE isolation level (under which they are now eventually consistent and may return slightly stale data) or be rewritten to use the primary key. Applications that depend on strict consistency for these reads should be reviewed before upgrading.


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -0,0 +1,253 @@
...
+      TableMetadata tableMetadata = metadata.getTableMetadata();
+      if (ScalarDbUtils.isSecondaryIndexSpecified(get, tableMetadata)) {
+        // If the index column is part of the primary key, it's allowed
+        String indexKeyColumnName = get.getPartitionKey().getColumns().get(0).getName();
+        if (!tableMetadata.getPartitionKeyNames().contains(indexKeyColumnName)
+            && !tableMetadata.getClusteringKeyNames().contains(indexKeyColumnName)) {
+          throw new IllegalArgumentException(
+              CoreError.CONSENSUS_COMMIT_INDEX_GET_NOT_ALLOWED_IN_SERIALIZABLE.buildMessage());
+        }
+      }
```
brfrn169, line 90: Disallow using indexes in the SERIALIZABLE isolation.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -0,0 +1,253 @@
...
+            if (metadata.getSecondaryIndexNames().contains(column)) {
+              throw new IllegalArgumentException(
+                  CoreError
+                      .CONSENSUS_COMMIT_CONDITION_ON_INDEXED_COLUMNS_NOT_ALLOWED_IN_CROSS_PARTITION_SCAN_IN_SERIALIZABLE
+                      .buildMessage());
+            }
+          }
+        }
+      }
+    }
```
brfrn169, line 174: Ditto. Disallow using indexes in the SERIALIZABLE isolation.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -460,10 +467,13 @@ static Scan prepareScanForStorage(Scan scan, TableMetadata metadata) {
    *
    * @param conjunctions the conjunctions to convert
    * @param metadata the table metadata of the target table
+   * @param convertIndexedColumns whether to convert conditions on indexed columns
    * @return the converted conjunctions
    */
   private static Set<AndConditionSet> convertConjunctions(
```
brfrn169, line 473: Modified ConsensusCommitUtils.convertConjunctions():
 - Added convertIndexedColumns parameter to control whether conditions on indexed columns should be converted to include before-column conditions
- For cross-partition scans (ScanAll), indexed column conditions are NOT converted (passed as-is)
- For partition scans and Gets, all conditions including indexed columns ARE converted

core/src/main/java/com/scalar/db/common/CoreError.java
brfrn169, line 1: @josh-wong Could you please check the error messages when you have time?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -889,6 +889,36 @@ public enum CoreError implements ScalarDbError {
...
+  CONSENSUS_COMMIT_SPECIFYING_TRANSACTION_METADATA_COLUMNS_IN_ORDERING_NOT_ALLOWED(
+      Category.USER_ERROR,
+      "0257",
+      "Specifying transaction metadata columns in the ordering is not allowed. Table: %s; Column: %s",
+      "",
+      ""),
+  CONSENSUS_COMMIT_INDEX_GET_NOT_ALLOWED_IN_SERIALIZABLE(
+      Category.USER_ERROR,
+      "0258",
+      "Get operations using an index is not allowed in the SERIALIZABLE isolation level",
```
josh-wong:
```suggestion
      "Get operations by using an index is not allowed in the SERIALIZABLE isolation level",
```

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -889,6 +889,36 @@ public enum CoreError implements ScalarDbError {
...
+      "",
+      ""),
+  CONSENSUS_COMMIT_INDEX_SCAN_NOT_ALLOWED_IN_SERIALIZABLE(
+      Category.USER_ERROR,
+      "0259",
+      "Scan operations using an index is not allowed in the SERIALIZABLE isolation level",
```
josh-wong:
```suggestion
      "Scan operations by using an index is not allowed in the SERIALIZABLE isolation level",
```

josh-wong (approved):
I've left a couple of minor suggestions. Other than that, LGTM. Thank you!🙇🏻‍♂️

komamitsu:
> However, if we simply add the same conditions to the before columns, the secondary index will no longer be used, which can significantly degrade performance.

Just out of curiosity, but using indexes for a normal column and the before column separately and merging them as follows don't work? 

```
  result1 = Scan.withIndexOn(col = 100)
  result2 = Scan.withIndexOn(before_col = 100)
  result = mergeAndDeduplicateResults(result1, result2)
```

brfrn169:
> > However, if we simply add the same conditions to the before columns, the secondary index will no longer be used, which can significantly degrade performance.
> 
> Just out of curiosity, but using indexes for a normal column and the before column separately and merging them as follows don't work?
> 
> ```
>   result1 = Scan.withIndexOn(col = 100)
>   result2 = Scan.withIndexOn(before_col = 100)
>   result = mergeAndDeduplicateResults(result1, result2)
> ```

@komamitsu Yes, that manual merge index approach should work.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  /**
+   * Checks the mutation validity
+   *
+   * @param mutation a mutation operation
+   * @throws ExecutionException when retrieving the table metadata fails
+   * @throws IllegalArgumentException when the mutation is invalid
+   */
+  public void check(Mutation mutation) throws ExecutionException {
```
feeblefakie, line 184: Just a confirmation, but from here, nothing has been changed in this file?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -483,6 +493,11 @@ private static Set<AndConditionSet> convertConjunctions(
           continue;
         }
 
+        if (!convertIndexedColumns && metadata.getSecondaryIndexNames().contains(columnName)) {
+          conditions.add(condition);
```
feeblefakie, line 497: I probably miss something, but is this needed? It looks like the condition is already added in [line 481](https://github.com/scalar-labs/scalardb/pull/3133/files#diff-dc46e8c1f8f1acf4fdbecd44b717a7e4d41abe637f1a2310a61cc8c552c40b91R481)?
(but [line 492](https://github.com/scalar-labs/scalardb/pull/3133/files#diff-dc46e8c1f8f1acf4fdbecd44b717a7e4d41abe637f1a2310a61cc8c552c40b91R492) also does the same thing, so I understand it's probably correct.
Can you clarify?

feeblefakie (commented):
Overall, looking good!
Left some questions for clarification. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
brfrn169: That’s right. I also renamed the file, so the diff looks a bit confusing. Sorry for that.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -390,7 +391,13 @@ static Scan prepareScanForStorage(Scan scan, TableMetadata metadata) {
...
       // If there are conjunctions, we need to convert them to include conditions on the before
       // image
-      Set<AndConditionSet> converted = convertConjunctions(scan.getConjunctions(), metadata);
+
+      // If the scan is a cross-partition scan (ScanAll), we don't convert conditions on indexed
+      // columns
+      boolean convertIndexedColumns = !(scan instanceof ScanAll);
+
+      Set<AndConditionSet> converted =
+          convertConjunctions(scan.getConjunctions(), metadata, convertIndexedColumns);
```
jnmt, line 400: Just a confirmation. The new validation for get/scan is added only for serializable isolation, and `convertConjunctions` is always called for partition Scan/Get and index Scan/Get.

> Therefore, after further discussion, we decided to redefine Get/Scan operations using secondary indexes as eventually consistent operations

So, does the above actually mean "ScanAll using secondary indexes under SI becomes eventually consistent"? Since users can scan using index columns (which are not part of the primary key) under SI with strong consistency, thanks to the conjunction conversion, I think we are going to relax (redefine) consistency only for ScanAll. I'm just curious about this for a kind of documentation matter.

jnmt (commented):
Thank you for the careful thought on this. I left a clarification question. PTAL!

jnmt:
@brfrn169 @feeblefakie  This is off topic, but could we handle scan using index columns well even under serializable isolation if we had "after" image as transaction metadata instead of before image? Rollforward and rollback mechanisms must be completely changed, though.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
brfrn169: Let’s say we have a conjunction `A = 1 and B = 1`, and `A` is an index key column.  

In that case, we need to convert the condition to `(A = 1 and B = 1) or (A = 1 and before_B = 1)`.  

[Line 481](https://github.com/scalar-labs/scalardb/pull/3133/files#diff-dc46e8c1f8f1acf4fdbecd44b717a7e4d41abe637f1a2310a61cc8c552c40b91R481) adds the first condition `(A = 1 and B = 1)` (which is the original conjunction).  
Then, here we create the second condition — but we don’t convert `A = 1` since `A` is an index key column.  
So we simply call `conditions.add(condition)` without any conversion.

The same applies to primary key columns, as shown in [line 492](https://github.com/scalar-labs/scalardb/pull/3133/files#diff-dc46e8c1f8f1acf4fdbecd44b717a7e4d41abe637f1a2310a61cc8c552c40b91R492).  

Does that make sense?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
brfrn169: Let me summarize the behavior:

We have seven query patterns related to the secondary index:

1. Get with conjunctions on an index column:
```java
Get get =
    Get.newBuilder()
        .namespace("ns")
        .table("tbl")
        .partitionKey(partitionKey)
        .clusteringKey(clusteringKey)
        .where(ConditionBuilder.column("index_col").isEqualToInt(1))
        .build();
```

2. Get with an index key specified:
```java
Get get =
    Get.newBuilder()
        .namespace("ns")
        .table("tbl")
        .indexKey(Key.ofInt("index_col", 1))
        .build();
```

3. Get with an index key specified and conjunctions on another index column:
```java
Get get =
    Get.newBuilder()
        .namespace("ns")
        .table("tbl")
        .indexKey(Key.ofInt("index_col", 1))
        .where(ConditionBuilder.column("another_index_col").isEqualToInt(10))
        .build();
```

4. Partition scan with conjunctions on an index column:
```java
Scan scan =
    Scan.newBuilder()
        .namespace("ns")
        .table("tbl")
        .partitionKey(partitionKey)
        .where(ConditionBuilder.column("index_col").isEqualToInt(1))
        .build();
```

5. Scan with an index key specified:
```java
Scan scan =
    Scan.newBuilder()
        .namespace("ns")
        .table("tbl")
        .indexKey(Key.ofInt("index_col", 1))
        .build();
```

6. Scan with an index key specified and conjunctions on another index column:
```java
Scan scan =
    Scan.newBuilder()
        .namespace("ns")
        .table("tbl")
        .indexKey(Key.ofInt("index_col", 1))
        .where(ConditionBuilder.column("another_index_col").isEqualToInt(10))
        .build();
```

7. Cross-partition scan with conjunctions on an index column:
```java
Scan scan =
    Scan.newBuilder()
        .namespace("ns")
        .table("tbl")
        .all()
        .where(ConditionBuilder.column("index_col").isEqualToInt(1))
        .build();
```

Now, the behavior for each pattern:

1: The Get already specifies the primary key. We convert the conjunctions, so it’s strongly consistent.
2: We don’t add any condition for the index key before column to take advantage of the index, so it’s eventually consistent.
3: Same as above — we don’t add any condition for the index key before column, so it’s eventually consistent. The conjunctions are converted in this case.
4: The Scan already specifies the partition key. We convert the conjunctions, so it’s strongly consistent.
5: We don’t add any condition for the index key before column to take advantage of the index, so it’s eventually consistent.
6: Same as above — we don’t add any condition for the index key before column, so it’s eventually consistent. The conjunctions are converted in this case.
7: We don’t convert the conjunctions to maintain index usage, so it’s eventually consistent.

In `SERIALIZABLE`, we don't allow the eventual consistent patterns: 2, 3, 5, 6, and 7.
We convert the conjunctions for the pattens: 1, 3, 4, and 6.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
feeblefakie: Ah, OK. It makes sense now. Thank you for the explanation!

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -0,0 +1,253 @@
...
+        // If the index column is part of the primary key, it's allowed
+        String indexKeyColumnName = scan.getPartitionKey().getColumns().get(0).getName();
+        if (!tableMetadata.getPartitionKeyNames().contains(indexKeyColumnName)
+            && !tableMetadata.getClusteringKeyNames().contains(indexKeyColumnName)) {
+          throw new IllegalArgumentException(
+              CoreError.CONSENSUS_COMMIT_INDEX_SCAN_NOT_ALLOWED_IN_SERIALIZABLE.buildMessage());
+        }
+      }
+
+      // If the scan is a cross-partition scan (ScanAll), don't allow conditions on indexed columns
```
komamitsu, line 157: Sorry, can you elaborate why this is prohibited?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
jnmt: @brfrn169 Thank you for taking the time to elaborate on this. I understood that only the partition get/scan is still strongly consistent, and others are not under SI. It would be great if we could describe this kind of concept or specification in the docs later.

jnmt (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
brfrn169: That's Case 7 in the comment:
https://github.com/scalar-labs/scalardb/pull/3133#discussion_r2517397932

For Case 7, we don’t convert the conjunctions to take advantage of the index, so it's an eventually consistent operation. That's why the case is prohibited in `SERIALIZABLE`.

komamitsu: I didn't understand the performance benefit from the combination of cross-partition scan and WHERE condition on the secondary index. Thanks to another discussion, I understand it now and I think this validation is reasonable.

On the other hand, I personally have a slightly different idea as follows:
- My understanding is that ScalarDB is good at partition queries while the performance of cross-partition scans isn't guaranteed. So, I think users can't expect that cross-partition scans even with WHERE condition on the secondary index will be fast.
- If users who want to use the secondary index column as a regular column for filtering encounter this exception, they might be negatively surprised.

I guess the change introduced in this PR is a difficult trade off and I'm not against it. The above is just personal feedback.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -0,0 +1,253 @@
...
+      if (scan instanceof ScanAll) {
+        for (Selection.Conjunction conjunction : scan.getConjunctions()) {
+          for (ConditionalExpression condition : conjunction.getConditions()) {
+            String column = condition.getColumn().getName();
+            if (metadata.getSecondaryIndexNames().contains(column)) {
```
komamitsu: Sorry, one more small question. Is this prohibited even if this column is the first primary key column?

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
brfrn169: If the index column is part of the primary key, we don't need to reject it. I'll fix it. Thanks!

brfrn169: Fixed in c65d1c3c14092e273cb49483e3881d914f496b50. Thanks!

komamitsu (approved):
LGTM, thank you!

PR #3193, opened 2025-11-20
Fix to use default credential providers

KodaiD:
## Description

This PR updates the Object Storage adapter to use default credentials providers, which are more secure.

## Related issues and/or PRs
 
- #3141 
- #3179 

## Changes made

- Updated S3Wrapper to use [Default credentials provider](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html).
- Updated CloudStorageWrapper to use [Application Default Credentials](https://docs.cloud.google.com/docs/authentication/provide-credentials-adc).
- Updated the workflow.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/.github/workflows/object-storage-adapter-check.yaml
+++ b/.github/workflows/object-storage-adapter-check.yaml
@@ -91,6 +91,7 @@ jobs:
         run: |
           container_id=$(docker create "container-registry.oracle.com/java/jdk:${{ env.INT_TEST_JAVA_RUNTIME_VERSION }}")
           docker cp -L "$container_id:/usr/java/default" /usr/lib/jvm/oracle-jdk && docker rm "$container_id"
+
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v5
 
@@ -145,11 +146,17 @@ jobs:
         run: |
           container_id=$(docker create "container-registry.oracle.com/java/jdk:${{ env.INT_TEST_JAVA_RUNTIME_VERSION }}")
           docker cp -L "$container_id:/usr/java/default" /usr/lib/jvm/oracle-jdk && docker rm "$container_id"
+
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v5
 
+      - name: Prepare Google Cloud Credentials
+        run: |
+          echo '${{ env.CLOUD_STORAGE_SERVICE_ACCOUNT_KEY }}' > $HOME/gcloud_service_account.json
+          export GOOGLE_APPLICATION_CREDENTIALS=$HOME/gcloud_service_account.json
+
       - name: Execute Gradle 'integrationTestObjectStorage' task
-        run: ./gradlew integrationTestObjectStorage -Dscalardb.object_storage.storage=cloud-storage -Dscalardb.object_storage.endpoint=scalardb-test-bucket -Dscalardb.object_storage.username=${{ env.CLOUD_STORAGE_PROJECT_ID }} -Dscalardb.object_storage.password=${{ env.CLOUD_STORAGE_SERVICE_ACCOUNT_KEY }} ${{ matrix.mode.group_commit_enabled && env.INT_TEST_GRADLE_OPTIONS_FOR_GROUP_COMMIT || '' }}
+        run: ./gradlew integrationTestObjectStorage -Dscalardb.object_storage.storage=cloud-storage -Dscalardb.object_storage.endpoint=scalardb-test-bucket -Dscalardb.object_storage.username=${{ env.CLOUD_STORAGE_PROJECT_ID }} ${{ matrix.mode.group_commit_enabled && env.INT_TEST_GRADLE_OPTIONS_FOR_GROUP_COMMIT || '' }}
 
       - name: Upload Gradle test reports
         if: always()
--- a/.github/workflows/object-storage-adapter-check.yaml
+++ b/.github/workflows/object-storage-adapter-check.yaml
@@ -152,10 +152,11 @@ jobs:
 
       - name: Prepare Google Cloud Credentials
         run: |
-          echo '${{ env.CLOUD_STORAGE_SERVICE_ACCOUNT_KEY }}' > $HOME/gcloud_service_account.json
-          export GOOGLE_APPLICATION_CREDENTIALS=$HOME/gcloud_service_account.json
+          echo '${{ env.CLOUD_STORAGE_SERVICE_ACCOUNT_KEY }}' > ${{ runner.temp }}/gcloud_service_account.json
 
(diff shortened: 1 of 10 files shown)


.github/workflows/object-storage-adapter-check.yaml
KodaiD, line 44: The provider checks these environment variables.

.github/workflows/object-storage-adapter-check.yaml
```diff
@@ -145,11 +146,17 @@ jobs:
...
           docker cp -L "$container_id:/usr/java/default" /usr/lib/jvm/oracle-jdk && docker rm "$container_id"
+
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v5
 
+      - name: Prepare Google Cloud Credentials
+        run: |
+          echo '${{ env.CLOUD_STORAGE_SERVICE_ACCOUNT_KEY }}' > $HOME/gcloud_service_account.json
+          export GOOGLE_APPLICATION_CREDENTIALS=$HOME/gcloud_service_account.json
+
```
KodaiD, line 156: The provider checks GOOGLE_APPLICATION_CREDENTIALS to get the path to the key.

I will replace this part with [more secure ways](https://docs.cloud.google.com/iam/docs/workload-identity-federation-with-deployment-pipelines) when I have the chance.

core/src/main/java/com/scalar/db/storage/objectstorage/cloudstorage/CloudStorageWrapper.java
```diff
@@ -34,12 +34,7 @@ public class CloudStorageWrapper implements ObjectStorageWrapper {
   private final Integer parallelUploadBlockSizeInBytes;
 
   public CloudStorageWrapper(CloudStorageConfig config) {
-    storage =
-        StorageOptions.newBuilder()
-            .setProjectId(config.getProjectId())
-            .setCredentials(config.getCredentials())
-            .build()
-            .getService();
+    storage = StorageOptions.newBuilder().setProjectId(config.getProjectId()).build().getService();
```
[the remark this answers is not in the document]
KodaiD: We should follow the result of google-java-format.

.github/workflows/object-storage-adapter-check.yaml
```diff
@@ -145,11 +146,17 @@ jobs:
...
       - name: Execute Gradle 'integrationTestObjectStorage' task
```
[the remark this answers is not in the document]
KodaiD: Fixed in [a4002be](https://github.com/scalar-labs/scalardb/pull/3193/commits/a4002bef38323946e68a1982a181162f25e1de01).

.github/workflows/object-storage-adapter-check.yaml
```diff
@@ -145,11 +146,18 @@ jobs:
         run: |
           container_id=$(docker create "container-registry.oracle.com/java/jdk:${{ env.INT_TEST_JAVA_RUNTIME_VERSION }}")
           docker cp -L "$container_id:/usr/java/default" /usr/lib/jvm/oracle-jdk && docker rm "$container_id"
+
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v5
 
+      - name: Prepare Google Cloud Credentials
```
feeblefakie, line 153: It's a question for the following. 
I'm not very familiar with IAM role, is it still regarded as using a secret key, and IAM roles are not used?
https://github.com/scalar-labs/scalardb/pull/3193/files#diff-11fd3465a765a999577e6b617ec3aafeab4b8690954bd0b8504236ff35ba98f6R43-R44

I'm asking because we should use IAM roles since using secret keys is not recommended in AWS. 
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html

core/src/main/java/com/scalar/db/storage/objectstorage/s3/S3Wrapper.java
```diff
@@ -61,9 +59,6 @@ public S3Wrapper(S3Config config) {
     this.client =
         S3AsyncClient.builder()
             .region(Region.of(config.getRegion()))
-            .credentialsProvider(
-                StaticCredentialsProvider.create(
-                    AwsBasicCredentials.create(config.getUsername(), config.getPassword())))
```
feeblefakie, line 66: So, does it change the way to deploy ScalarDB in case the underlying databases include S3?
(If so, do we need to change the Helm chart?)

.github/workflows/object-storage-adapter-check.yaml
KodaiD: It's for Google Cloud, not for AWS. As written in [here](https://github.com/scalar-labs/scalardb/pull/3193#discussion_r2544198510), we should use a more secure way, like workload identity, to authenticate from outside Google Cloud. I'll address it after the release.

core/src/main/java/com/scalar/db/storage/objectstorage/s3/S3Wrapper.java
KodaiD: I don't think we need to change the Helm chart. This change just eliminates the need to enter credentials in a custom values ​​file. ScalarDB Cluster users can assign an IAM Role to the Kubernetes service account by using EKS Pod Identity.

feeblefakie: So, do we need a particular section for configuring credentials for S3?
The current doc basically passes passwords to Cluster through K8s secrets.

.github/workflows/object-storage-adapter-check.yaml
KodaiD: The main goal of this PR is to allow users to choose authentication without using secret access keys or service account keys. Securing CI will be addressed in a separate PR.

core/src/main/java/com/scalar/db/storage/objectstorage/s3/S3Wrapper.java
KodaiD: Yes, we do.  I'll prepare it.

KodaiD: @feeblefakie The current Helm chart has a feature that mounts the service account. So, we don't need to update the Helm chart. But users have to specify those values.

Ref. https://github.com/scalar-labs/helm-charts/blob/scalardb-cluster-1.9.0/charts/scalardb-cluster/values.yaml#L290-L294

KodaiD:
Sorry reviewers, since we should do more verifications, I'm closing this PR. I will address this after the release.

PR #3403, opened 2026-03-06
Add solutions to errors 10000 to 10093 in `CoreError.java`

josh-wong:
## Description

This PR adds error solutions to errors 10000 to 10093 in `CoreError.java`. These solutions help users understand how to resolve or further investigate the errors, improving debuggability and user support.

As requested, these error solutions were initially generated by using GitHub Copilot.

> [!NOTE]
>
> As previously discussed, since we have a lot of errors within the five classes in `CoreError.java`, I've included only a portion of the solutions in this PR. Later PRs will include similar portions, with the last error solution in each PR ending on a topic (in this case, before the Consensus Commit–related error messages).

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/3402

> [!NOTE]
>
> To avoid having the batches of error solutions become a blocker if we need to release a minor or patch version, this branch targets a parent branch for the `CoreError.java` error solutions. After all PRs for error solutions for `CoreError.java` have been merged, we can merge that branch into the master branch.

## Changes made

- Added solutions to errors 10000 to 10093 in `CoreError.java`.
- Applied Spotless formatting to those error solutions.

## Checklist

Since this PR covers only a portion of the error solutions, the parent PR for the `CoreError.java` solutions mentioned above will include a release note.

## Release notes

N/A


core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "0009",
       "The specified projection is not found. Projection: %s, Operation: %s; Table metadata: %s",
       "",
-      ""),
+      "Verify that the projection column exists in the table metadata and correct the column name."),
   OPERATION_CHECK_ERROR_CLUSTERING_KEY_BOUNDARY(
       Category.USER_ERROR,
       "0010",
-      "The clustering key boundary is not properly specified. Operation: %s; Table metadata: %s",
+      "The clustering-key boundary is not properly specified. Operation: %s; Table metadata: %s",
```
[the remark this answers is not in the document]
josh-wong: Since "clustering key" doesn't modify a noun in those cases, we don't need to include a hyphen.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Add the column definition for the clustering-key column."),
   TRANSACTION_STATE_INSTANTIATION_ERROR_INVALID_ID(
-      Category.USER_ERROR, "0042", "Invalid ID specified. ID: %d", "", ""),
+      Category.USER_ERROR,
+      "0042",
+      "Invalid ID specified. ID: %d",
+      "",
+      "Use a valid transaction state ID."),
```
komamitsu, line 265: @brfrn169 I'm wondering how users can handle this error. Maybe this should not be a USER_ERROR?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
+      "The mutations are empty",
+      "",
+      "Provide at least one mutation (Put, Delete, Insert, Update, or Upsert) in your request."),
   OPERATION_CHECK_ERROR_MULTI_PARTITION_MUTATION(
       Category.USER_ERROR,
       "0019",
       "The storage does not support mutations across multiple partitions. Storage: %s; Mutations: %s",
       "",
-      ""),
+      "Group mutations by partition key and execute them in separate transactions, or use a storage that supports multi-partition mutations."),
```
Torch3333, line 130: @brfrn169 I don't think we document the atomicity unit we apply for each storage. It is helpful for performance optimization, so documenting it might be nice. What do you think?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Enable authentication and authorization in your ScalarDB Enterprise configuration, or remove authentication-related operations from your code."),
   CONDITION_BUILD_ERROR_CONDITION_NOT_ALLOWED_FOR_PUT_IF(
       Category.USER_ERROR,
       "0023",
       "This condition is not allowed for the PutIf operation. Condition: %s",
       "",
-      ""),
+      "Use only supported conditions for PutIf operations (PutIfExists, PutIfNotExists, or column-based conditions)."),
```
Torch3333: Currently, this error is only thrown when a `PutIf` condition uses the `LIKE` or `NOT_LIKE` operator, which are column-based conditions. So we could leave the message as that?
```suggestion
      "Use only supported conditions for PutIf operations."),
```

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
   CONDITION_BUILD_ERROR_CONDITION_NOT_ALLOWED_FOR_DELETE_IF(
       Category.USER_ERROR,
       "0024",
       "This condition is not allowed for the DeleteIf operation. Condition: %s",
       "",
-      ""),
+      "Use only supported conditions for DeleteIf operations (DeleteIfExists or column-based conditions)."),
```
Torch3333: Likewise here.
```suggestion
      "Use only supported conditions for DeleteIf operations."),
```

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
-      ""),
+      "Provide at least one contact point in the 'scalar.db.contact_points' configuration property."),
   COSMOS_CONDITION_OPERATION_NOT_SUPPORTED_FOR_BLOB_TYPE(
       Category.USER_ERROR,
       "0076",
       "Cosmos DB supports only EQ, NE, IS_NULL, and IS_NOT_NULL operations for the BLOB type in conditions. Mutation: %s",
       "",
-      ""),
+      "Use only EQ, NE, IS_NULL, or IS_NOT_NULL operators for BLOB columns in Cosmos DB conditions."),
   INVALID_CONSISTENCY_LEVEL(
```
Torch3333, line 460: This error message is only used for CosmosDB, and since the solution is only applicable to Cosmos. I think we might as well rename the error to be Cosmos DB-specific rather than storage-agnostic.
```suggestion
  COSMOS_INVALID_CONSISTENCY_LEVEL(
```

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
-      ""),
+      "Avoid nesting multi-storage configurations. Define each storage directly."),
   MULTI_STORAGE_STORAGE_NOT_FOUND(
-      Category.USER_ERROR, "0084", "Storage not found. Storage: %s", "", ""),
+      Category.USER_ERROR,
+      "0084",
+      "Storage not found. Storage: %s",
+      "",
+      "Verify the storage name is defined in your multi-storage configuration."),
   JDBC_SQLITE_NAMESPACE_NAME_NOT_ACCEPTABLE(
```
Torch3333, line 502: @brfrn169 
The error and the one below are raised when the table or namespace names contain a `$` character, which is used internally to separate the namespace from the table in the table name, such as  `MY_NAMESPACE$MY_TABLE`
Should we mention directly in the error message that this `$` is not allowed ? Otherwise, I think the error message is not very helpful.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
+      "The %s table must have a primary key",
+      "",
+      "Add a primary key to the table before importing it."),
   JDBC_RDB_ENGINE_NOT_SUPPORTED(
       Category.USER_ERROR,
       "0089",
       "The RDB engine is not supported. JDBC connection URL: %s",
       "",
-      ""),
+      "Use a supported JDBC database (MySQL, PostgreSQL, Oracle, SQL Server, SQLite, Db2, or TiDB)."),
```
Torch3333: Since we are frequently adding new storage, for ease of maintenance, should we leave the message as that?
```suggestion
      "Use a supported JDBC database."),
```

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: Sounds good to me! I've revised this in https://github.com/scalar-labs/scalardb/pull/3403/changes/20d7e4f24d27e7ccb2d8d5f6f389c9b1d294b9c1.

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: Just to confirm, are we OK to make this change here now? Or will changing it affect other source code?

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: Good idea! I've revised this in https://github.com/scalar-labs/scalardb/pull/3403/changes/20d7e4f24d27e7ccb2d8d5f6f389c9b1d294b9c1.

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: Sounds good to me! I've revised this in https://github.com/scalar-labs/scalardb/pull/3403/changes/20d7e4f24d27e7ccb2d8d5f6f389c9b1d294b9c1.

core/src/main/java/com/scalar/db/common/CoreError.java
Torch3333: It will affect another Java file. Since it is a small change, I think it's fine to make it here.

josh-wong: Understood, thanks! I've addressed it in https://github.com/scalar-labs/scalardb/pull/3403/changes/a28db844e2fbd31d6ffc4ea343b60e19f79616f0.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
   OPERATION_CHECK_ERROR_CLUSTERING_KEY_BOUNDARY(
       Category.USER_ERROR,
       "0010",
-      "The clustering key boundary is not properly specified. Operation: %s; Table metadata: %s",
+      "The clustering-key boundary is not properly specified. Operation: %s; Table metadata: %s",
       "",
-      ""),
+      "Ensure the clustering-key boundary values match the table's clustering-key definition and are specified in the correct order."),
   OPERATION_CHECK_ERROR_START_CLUSTERING_KEY(
       Category.USER_ERROR,
```
feeblefakie, line 78: Message-text edit `clustering key` → `clustering-key` is unrelated to adding solutions. Consider dropping or splitting.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Enable authentication and authorization in your ScalarDB Enterprise configuration, or remove authentication-related operations from your code."),
   CONDITION_BUILD_ERROR_CONDITION_NOT_ALLOWED_FOR_PUT_IF(
       Category.USER_ERROR,
       "0023",
       "This condition is not allowed for the PutIf operation. Condition: %s",
       "",
-      ""),
+      "Use only supported conditions for PutIf operations."),
```
feeblefakie, line 155: This is tautological with the message ("Only supported conditions can be specified in a PutIf operation"). Doesn't tell the user which conditions are supported. Better to leave `""` than restate the message.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Use only supported conditions for DeleteIf operations."),
   LIKE_CHECK_ERROR_OPERATOR_MUST_BE_LIKE_OR_NOT_LIKE(
-      Category.USER_ERROR, "0025", "Operator must be LIKE or NOT_LIKE. Operator: %s", "", ""),
+      Category.USER_ERROR,
+      "0025",
+      "Operator must be LIKE or NOT_LIKE. Operator: %s",
+      "",
+      "Use LIKE or NOT_LIKE operators for pattern matching."),
```
feeblefakie, line 167: Tautological with the message. 0023/0024/0025/0027/0042/0092/0093 all restate the message with no added guidance — `""` would be more honest.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Set 'scalar.db.cross_partition_scan.enabled' to 'true' in your configuration."),
   OUT_OF_RANGE_COLUMN_VALUE_FOR_BIGINT(
       Category.USER_ERROR,
       "0063",
       "This column value is out of range for BigInt. Value: %s",
       "",
-      ""),
+      "Use a value within the valid range for BigInt ('Long.MIN_VALUE' to 'Long.MAX_VALUE')."),
```
feeblefakie, line 393: This is wrong. `OUT_OF_RANGE_COLUMN_VALUE_FOR_BIGINT` is only thrown from Cosmos/ObjectStorage where the limit is ±2^53 (`BIGINT_MAX_VALUE = 9007199254740992L` in both `CosmosOperationChecker` and `ObjectStorageOperationChecker`), not `Long.MIN_VALUE`..`Long.MAX_VALUE`.

Also worth noting: the unprefixed `OUT_OF_RANGE_COLUMN_VALUE_FOR_BIGINT` (0063) has no callers — only the `COSMOS_` / `OBJECT_STORAGE_` variants are used — so this entry may be dead.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "0071",
       "The property 'scalar.db.contact_port' must be greater than or equal to zero",
       "",
-      ""),
+      "Set 'scalar.db.contact_port' to a valid port number (0-65535)."),
   COSMOS_CLUSTERING_KEY_BLOB_TYPE_NOT_SUPPORTED(
       Category.USER_ERROR,
       "0073",
       "The BLOB type is not supported for clustering keys in Cosmos DB. Column: %s",
       "",
```
feeblefakie, line 440: Message-text edit `must be set to true` → `must be set to 'true'` is unrelated to adding solutions. Consider dropping or splitting.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
+      "Provide at least one contact point in the 'scalar.db.contact_points' configuration property."),
   COSMOS_CONDITION_OPERATION_NOT_SUPPORTED_FOR_BLOB_TYPE(
       Category.USER_ERROR,
       "0076",
       "Cosmos DB supports only EQ, NE, IS_NULL, and IS_NOT_NULL operations for the BLOB type in conditions. Mutation: %s",
       "",
-      ""),
-  INVALID_CONSISTENCY_LEVEL(
+      "Use only EQ, NE, IS_NULL, or IS_NOT_NULL operators for BLOB columns in Cosmos DB conditions."),
+  COSMOS_INVALID_CONSISTENCY_LEVEL(
```
feeblefakie: This rename breaks the build — `CosmosUtils.java:39` still calls `CoreError.INVALID_CONSISTENCY_LEVEL.buildMessage(consistencyLevel)`. Either update the caller in this PR or keep the old name.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       Category.USER_ERROR,
       "0077",
       "The specified consistency level is not supported. Consistency level: %s",
       "",
-      ""),
+      "Use a supported consistency level (SEQUENTIAL, EVENTUAL, LINEARIZABLE, or BOUNDED_STALENESS)."),
```
feeblefakie: This solution is wrong. `CosmosUtils.getConsistencyLevel` only accepts `STRONG` and `BOUNDED_STALENESS` — the other three levels (`SEQUENTIAL`, `EVENTUAL`, `LINEARIZABLE`) reproduce the same error, and `SEQUENTIAL`/`LINEARIZABLE` aren't even valid Azure Cosmos consistency levels. Suggest something like ``Use `STRONG` or `BOUNDED_STALENESS`; Cosmos DB in ScalarDB does not support other consistency levels``.

feeblefakie (commented):
Left some comments. PTAL!

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Remove the ordering specification from your operation when using an index."),
   OPERATION_CHECK_ERROR_LIMIT(
-      Category.USER_ERROR, "0005", "The limit cannot be negative. Operation: %s", "", ""),
+      Category.USER_ERROR,
+      "0005",
+      "The limit cannot be negative. Operation: %s",
+      "",
+      "Specify a non-negative value for the limit (0 or greater)."),
```
feeblefakie, line 46: All 90 added solutions end with `.`, but existing messages in this file follow the CLAUDE.md rule of no trailing punctuation (191 without vs 4 with). Worth deciding the convention explicitly since this PR sets the precedent for later batches.

core/src/main/java/com/scalar/db/common/CoreError.java
feeblefakie: Also: `CoreError` is a `public` enum, so renaming this constant is a source-incompat change for downstream consumers (Enterprise, custom storage adapters) — they'll get `NoSuchFieldError` at runtime if only core is bumped. Consider keeping the old name as a `@Deprecated` alias, or splitting the rename into its own PR so it's easy to find in the log.

feeblefakie (commented):
A couple more. PTAL!

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: If we remove this message, this error message will not have a corresponding solution message. Could you please confirm if we're OK with not including solutions in some error messages?

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: If we remove this message, this error message will not have a corresponding solution message. Could you please confirm if we're OK with not including solutions in some error messages?

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: OK, I'm unsure what to do here in that case. If this error has no callers, should we delete this error altogether?

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: Is this comment referring to the error message `"The property 'scalar.db.cross_partition_scan.enabled' must be set to 'true' "`? If so, I added the single quotation marks because, if we have single quotation marks around the configurations, then we should have them around the values as well.

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: @Torch3333 According to @feeblefakie's [comment](https://github.com/scalar-labs/scalardb/pull/3403#discussion_r3650453362), making this change would be be source-incompatible for downstream users. To avoid that, I've reverted the change to use the original name (`INVALID_CONSISTENCY_LEVEL`).

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: As @Torch3333 [suggested in another comment](https://github.com/scalar-labs/scalardb/pull/3403#discussion_r2979824691), I renamed `INVALID_CONSISTENCY_LEVEL` to `COSMOS_INVALID_CONSISTENCY_LEVEL`. However, since it causes source-incompatible changes downstream, I've reverted the changes.

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: Grammatically, we should be ending sentences with a period `.` to avoid inconsistencies. If a message has multiple sentences in it or if a message includes a variable placeholder, like `"The limit cannot be negative. Operation: %s",` where the sentence ends in a period, then the rule mentioned in CLAUDE.md needs to be broken.

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: Thanks. I applied the suggestion in 1efd70aa4d99c6a6d64ce6304e151fa0a32e14f7.

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: Since this change is source incompatible for downstream users, (according to [this comment](https://github.com/scalar-labs/scalardb/pull/3403#discussion_r3650453362)), I've reverted the change to keep the old name.

core/src/main/java/com/scalar/db/common/CoreError.java
josh-wong: True; we can address this later. Addressed in b0fc69f191ad236abc5abedfa7b5f342d35369da.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Remove the call to get a transaction state when using JDBC transactions."),
   JDBC_TRANSACTION_ROLLING_BACK_TRANSACTION_NOT_SUPPORTED(
       Category.USER_ERROR,
       "0093",
       "Rolling back a transaction is not supported in JDBC transactions",
       "",
-      ""),
+      "Remove the rollback call. JDBC transactions in ScalarDB do not support manual rollback."),
```
feeblefakie, line 555: I think this one is not just tautological but wrong. `JdbcTransaction.rollback()` works fine and is the normal way to abort — `JdbcTransactionManager` calls it itself (`rollbackTransaction()` at `JdbcTransactionManager.java:426`). Only the manager's ID-based `rollback(String txId)` throws this (`JdbcTransactionManager.java:439-441`). Someone taking this literally would delete working rollback calls from their error handling.

How about something like: "Call rollback() on the transaction object instead. Rolling back by transaction ID is not supported in JDBC transactions."?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
+      "Storage not found. Storage: %s",
+      "",
+      "Verify the storage name is defined in your multi-storage configuration."),
   JDBC_SQLITE_NAMESPACE_NAME_NOT_ACCEPTABLE(
       Category.USER_ERROR,
       "0085",
       "The namespace name is not acceptable in SQLite. Namespace: %s",
       "",
-      ""),
+      "Use a valid SQLite database name that follows SQLite naming conventions."),
```
feeblefakie, line 507: I don't think there is a SQLite naming convention behind this — the check is just `contains("$")` (`RdbEngineSqlite.java:174`), because we use `$` to separate the namespace from the table. Anyone who goes looking at the SQLite docs will find nothing. This also calls a namespace a "database name".

How about: "Remove the '$' character from the name. ScalarDB uses '$' as the namespace separator in SQLite."? Same for 0086 below (`RdbEngineSqlite.java:182`).

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Use a supported JDBC database."),
   JDBC_IMPORT_DATA_TYPE_WITH_SIZE_NOT_SUPPORTED(
-      Category.USER_ERROR, "0090", "Data type %s(%d) is not supported: %s", "", ""),
+      Category.USER_ERROR,
+      "0090",
+      "Data type %s(%d) is not supported: %s",
+      "",
+      "Use a supported ScalarDB data type or check the import-type mapping configuration."),
```
feeblefakie, line 537: I don't think we have an "import-type mapping configuration" — the actual name is `override-columns-type` in the import schema JSON (`ImportTableSchema.java:15`). It also reads backwards: on import the user isn't picking a ScalarDB type, it's the source column type that has no mapping.

How about: "Change the column to a type ScalarDB supports, exclude it from the import, or map it explicitly with 'override-columns-type' in the import schema."? Same for 0091 below.

feeblefakie (commented):
Left some comments, PTAL!

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
+      "Cannot encode a Text value that contains '\\u0000'",
+      "",
+      "Remove null characters (\\u0000) from the text value."),
   DYNAMO_INDEX_COLUMN_CANNOT_BE_SET_TO_EMPTY(
       Category.USER_ERROR,
       "0081",
       "An index column cannot be set to an empty value for Text or Blob in DynamoDB. Operation: %s",
       "",
-      ""),
+      "Provide a non-null, non-empty value for indexed Text or Blob columns in DynamoDB."),
```
feeblefakie, line 483: Null looks allowed here — `ColumnChecker` only rejects empty non-null values (`ColumnChecker.java:139`), and the message itself says "empty". How about dropping "non-null"?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
+      "The %s network strategy does not exist",
+      "",
+      "Use a valid Cassandra network strategy ('SimpleStrategy' or 'NetworkTopologyStrategy')."),
   INVALID_CONTACT_PORT(
       Category.USER_ERROR,
       "0071",
       "The property 'scalar.db.contact_port' must be greater than or equal to zero",
       "",
-      ""),
+      "Set 'scalar.db.contact_port' to a valid port number (0-65535)."),
```
feeblefakie, line 435: We only check `contactPort >= 0` (`DatabaseConfig.java:104`), and the message says the same. The 65535 upper bound isn't enforced anywhere, so stating it as a rule might be misleading. How about "Set 'scalar.db.contact_port' to a non-negative value."?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
       "",
-      ""),
+      "Begin a new transaction if you need to perform additional operations."),
   TRANSACTION_NOT_PREPARED(
-      Category.USER_ERROR, "0045", "The transaction has not been prepared. Status: %s", "", ""),
+      Category.USER_ERROR,
+      "0045",
+      "The transaction has not been prepared. Status: %s",
+      "",
+      "Call 'prepare()' before committing a two-phase transaction."),
```
feeblefakie, line 283: This is thrown from `validate()`, not commit — `StateManagedTwoPhaseCommitTransactionManager.java:161-163` is the only throw site. Someone who hits this never called commit; 0046 below is the commit one.

How about "Call prepare() before calling validate()."?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
+      "The transaction is not active. Status: %s",
+      "",
+      "Begin a new transaction before performing operations."),
   TRANSACTION_ALREADY_COMMITTED(
       Category.USER_ERROR,
       "0044",
       "The transaction has already been committed. Status: %s",
       "",
-      ""),
+      "Begin a new transaction if you need to perform additional operations."),
```
feeblefakie, line 277: All four throw sites are `rollback()`/`abort()` guards after a successful commit (`StateManagedDistributedTransactionManager.java:158,174` and `StateManagedTwoPhaseCommitTransactionManager.java:195,211`). Wanting to do more work isn't really the mistake here — it's usually an unconditional `rollback()` in a finally block.

How about "Do not call rollback() or abort() after the transaction has been committed."?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -13,408 +13,546 @@ public enum CoreError implements ScalarDbError {
...
+      "The mutations are empty",
+      "",
+      "Provide at least one mutation (Put, Delete, Insert, Update, or Upsert) in your request."),
   OPERATION_CHECK_ERROR_MULTI_PARTITION_MUTATION(
       Category.USER_ERROR,
       "0019",
       "The storage does not support mutations across multiple partitions. Storage: %s; Mutations: %s",
       "",
-      ""),
+      "Group mutations by partition key and execute them in separate transactions, or use a storage that supports multi-partition mutations."),
```
feeblefakie, line 130: "separate transactions" might be the wrong layer — this is only reachable from `OperationChecker.check(List<? extends Mutation>)` (`OperationChecker.java:346,427`), i.e. the storage-level `mutate()`. Consensus Commit checks user mutations one at a time (`ConsensusCommit.java:256`) and groups its internal ones by atomicity unit before calling `mutate()`, so there is no transaction in the picture when this fires.

How about "...issue a separate mutate() call for each group..."?

core/src/main/java/com/scalar/db/common/CoreError.java
feeblefakie: While we are on conventions, two more inconsistencies in the added solutions, if you want to sweep them in the same pass:

- Quoting is uneven: `'prepare()'` / `'scan()'` are quoted but `clearConditions()` is bare; `'STRONG'` is quoted but `EQ` / `LIKE` are bare.
- The "...is not properly specified" family drifts between "Ensure" and "Verify that", and between "in the correct order" and "in the proper order".

PR #3419, opened 2026-03-12
Fix index-based Get and Scan to not miss PREPARED/DELETED records by adding before-image index check

brfrn169:
## Description

In Consensus Commit, `Get`/`Scan` operations using secondary indexes can miss records that are in PREPARED or DELETED state when the indexed column value has been changed. For example, if a record's `indexed_column` is being updated from 10 to 20 (PREPARED state), a query for `indexed_column = 10` will not find the record because the storage layer looks up the current column value, which no longer matches the committed value. For details on this issue, see https://github.com/scalar-labs/scalardb/pull/3133.

This PR fixes this issue by introducing before-image secondary indexes (`before_<col>`) and using them in `CrudHandler` to detect and recover such records, ensuring correctness for index-based reads.

### Algorithm

When an index-based `Get` or `Scan` is executed (e.g., `BALANCE = 10`), the following steps are performed:

1. **Normal index lookup**: The storage layer returns records where the current `BALANCE` value is 10.
2. **Before-image index check**: An additional scan is performed on `before_BALANCE = 10` to find PREPARED/DELETED records whose committed (before-image) value matches the query but whose current value does not.
3. **Recovery and retry**: If such records are found, recovery is executed for each record:
   - If the record is **rolled back** (the before-image value is restored), the record now matches the original query again, so the original operation is **retried**.
   - If the record is **rolled forward** (the new value is committed), the record's committed value no longer matches the query, so **no retry is needed**.

### Example

Consider three records with `BALANCE = 10`:

| ID | BALANCE | State |
|--------|---------|-------|
| 0 | 10 | COMMITTED |
| 1 | 20 (before: 10) | PREPARED |
| 2 | 10 | COMMITTED |

A Scan with `BALANCE = 10` returns only `record0` and `record2` from the normal index lookup because `record1` has `BALANCE = 20` in its current value. The before-image index check then scans `before_BALANCE = 10` and finds `record1` in PREPARED state.

- If `record1` is rolled back: `BALANCE` reverts to 10. The operation is retried and now returns all three records.
- If `record1` is rolled forward: `BALANCE` is committed as 20. The operation does not need a retry and correctly returns `record0` and `record2`.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/3133

### Before-image secondary index schema

- In `ConsensusCommitUtils.buildTransactionTableMetadata()` and `buildTransactionMetadataTableMetadata()`, a `before_<col>` secondary index is now automatically created for each user-defined secondary index on a non-primary-key column.
- Added `hasBeforeImageSecondaryIndex(String columnName)` to `TransactionTableMetadata` to check whether a column has a corresponding before-image index at runtime.

### `CrudHandler` before-index check and recovery

- For `Index Get`, `Index Scan`, and `ScanAll` (with indexed column conditions), when a before-image secondary index exists, an additional scan on `before_<col>` is performed to find PREPARED/DELETED records whose committed values match the original query.
- When such records are found, recovery is executed. If the recovery results in a rollback (the before-image value is restored), the original operation is retried (up to `MAX_BEFORE_INDEX_CHECK_RETRIES`). If it results in a roll-forward, no retry is needed.
- In `ConsensusCommitStorageScanner.close()`, a before-index check is performed; if PREPARED/DELETED records needing recovery are found, a `CrudConflictException` is thrown to prompt a transaction-level retry.

### RecoveryExecutor.Result enhancement

- Added `rolledBack` field to `RecoveryExecutor.Result` to distinguish rollback from roll-forward outcomes, enabling `CrudHandler` to decide whether a retry is necessary.

### ConsensusCommitAdmin updates

- `createIndex()` and `dropIndex()` now automatically create/drop the corresponding `before_<col>` index when the before-image column exists.

### Helper methods in ConsensusCommitUtils

- `createBeforeIndexScan(Selection)`: Creates a `ScanWithIndex` using the `before_<col>` index key.
- `createBeforeIndexScanAll(ScanAll, TableMetadata)`: Creates a `ScanAll` with before-image column conditions.
- `requiresBeforeIndexCheck(Selection, TransactionTableMetadata):` Determines whether a before-index check is needed for a given operation.
- `warnIfBeforeImageIndexesAreMissing(DistributedStorageAdmin, ConsensusCommitConfig)`: Checks all tables at startup and logs a warning for tables with secondary indexes but missing before-image indexes.

### Startup warning for missing before-image indexes

- At startup, `ConsensusCommitManager` and `TwoPhaseConsensusCommitManager` check all transaction tables for missing before-image indexes and log a warning suggesting repairTable(). This check is only performed when the isolation level is SNAPSHOT or READ_COMMITTED and index eventual consistent read is disabled, as these are the configurations where missing before-image indexes can cause records to be missed.

## Checklist

### Backward compatibility

Tables created before this change will not have `before_<col>` secondary indexes. In that case, `hasBeforeImageSecondaryIndex()` returns false and the before-index check is skipped entirely, preserving existing behavior. Running `repairTable()` on existing tables will automatically add the `before_<col> `indexes. At startup, a warning is logged for tables missing before-image indexes to guide users.

### SERIALIZABLE support

This PR doesn’t include SERIALIZABLE support. I’ll create a separate PR to handle it.

## Release notes

Fixed a correctness issue in index-based `Get`, `Scan`, and `ScanAll` operations in Consensus Commit where records whose indexed column was being concurrently updated by another transaction could be missed. ScalarDB now maintains a companion before-image secondary index for each user-defined secondary index on a non-primary-key column and uses it to recover PREPARED/DELETED records during index reads.

When upgrading from an earlier version, you must run `repairTable()` on each existing table to create the companion before-image secondary indexes that the new check requires. Until `repairTable()` is run, ScalarDB logs a warning at startup and index-based reads fall back to the pre-check behavior, which may miss records whose indexed column is being concurrently updated. A new configuration option `scalar.db.consensus_commit.index.eventually_consistent_read.enabled` (default `false`) is available to opt out of the new check; this is a backward-compatibility option and is not recommended for new workloads.

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
@@ -41,6 +41,7 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
   private final DistributedStorageAdmin admin;
   private final String coordinatorNamespace;
   private final boolean isIncludeMetadataEnabled;
+  private final boolean isIndexEventuallyConsistentReadEnabled;
 
   @SuppressFBWarnings("EI_EXPOSE_REP2")
   @Inject
@@ -49,6 +50,7 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
     ConsensusCommitConfig config = new ConsensusCommitConfig(databaseConfig);
     coordinatorNamespace = config.getCoordinatorNamespace().orElse(Coordinator.NAMESPACE);
     isIncludeMetadataEnabled = config.isIncludeMetadataEnabled();
+    isIndexEventuallyConsistentReadEnabled = config.isIndexEventuallyConsistentReadEnabled();
   }
 
   public ConsensusCommitAdmin(DatabaseConfig databaseConfig) {
@@ -58,6 +60,7 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
     ConsensusCommitConfig config = new ConsensusCommitConfig(databaseConfig);
     coordinatorNamespace = config.getCoordinatorNamespace().orElse(Coordinator.NAMESPACE);
     isIncludeMetadataEnabled = config.isIncludeMetadataEnabled();
+    isIndexEventuallyConsistentReadEnabled = config.isIndexEventuallyConsistentReadEnabled();
   }
 
   @VisibleForTesting
@@ -68,6 +71,7 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
     this.admin = admin;
     coordinatorNamespace = config.getCoordinatorNamespace().orElse(Coordinator.NAMESPACE);
     this.isIncludeMetadataEnabled = isIncludeMetadataEnabled;
+    isIndexEventuallyConsistentReadEnabled = config.isIndexEventuallyConsistentReadEnabled();
   }
 
   @Override
@@ -147,7 +151,8 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
       admin.createTable(
           namespace,
           txMetadataTableName,
-          ConsensusCommitUtils.buildTransactionMetadataTableMetadata(metadata),
+          ConsensusCommitUtils.buildTransactionMetadataTableMetadata(
+              metadata, isIndexEventuallyConsistentReadEnabled),
           options);
 
       // Create a virtual table based on the data table and the transaction metadata table
@@ -163,7 +168,11 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
       return;
     }
 
-    admin.createTable(namespace, table, buildTransactionTableMetadata(metadata), options);
+    admin.createTable(
+        namespace,
+        table,
+        buildTransactionTableMetadata(metadata, isIndexEventuallyConsistentReadEnabled),
+        options);
   }
 
   @Override
@@ -223,6 +232,17 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
     checkNamespace(namespace);
 
     admin.createIndex(namespace, table, columnName, options);
+
+    if (!isIndexEventuallyConsistentReadEnabled) {
+      // Also create an index on the before image column if it exists
+      TableMetadata rawMetadata = admin.getTableMetadata(namespace, table);
+      if (rawMetadata != null) {
+        String beforeColumnName = Attribute.BEFORE_PREFIX + columnName;
+        if (rawMetadata.getColumnNames().contains(beforeColumnName)) {
+          admin.createIndex(namespace, table, beforeColumnName, options);
+        }
+      }
+    }
   }
 
   @Override
@@ -231,6 +251,17 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
     checkNamespace(namespace);
 
     admin.dropIndex(namespace, table, columnName);
+
+    if (!isIndexEventuallyConsistentReadEnabled) {
+      // Also drop the index on the before image column if it exists
+      TableMetadata rawMetadata = admin.getTableMetadata(namespace, table);
+      if (rawMetadata != null) {
+        String beforeColumnName = Attribute.BEFORE_PREFIX + columnName;
+        if (rawMetadata.getColumnNames().contains(beforeColumnName)) {
+          admin.dropIndex(namespace, table, beforeColumnName, true);
+        }
+      }
+    }
   }
 
   @Override
@@ -293,7 +324,11 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
     checkNamespace(namespace);
     throwIfTransactionMetadataDecouplingApplied(namespace, table, "repairTable()");
 
-    admin.repairTable(namespace, table, buildTransactionTableMetadata(metadata), options);
+    admin.repairTable(
+        namespace,
+        table,
+        buildTransactionTableMetadata(metadata, isIndexEventuallyConsistentReadEnabled),
+        options);
   }
 
   @Override
@@ -438,7 +473,8 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
       admin.createTable(
           namespace,
           txMetadataTableName,
-          ConsensusCommitUtils.buildTransactionMetadataTableMetadata(dataTableMetadata),
+          ConsensusCommitUtils.buildTransactionMetadataTableMetadata(
+              dataTableMetadata, isIndexEventuallyConsistentReadEnabled),
           options);
 
       // create a virtual table based on the data table and the transaction metadata table
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
@@ -39,6 +39,8 @@ public class ConsensusCommitConfig {
   public static final String PARALLEL_IMPLICIT_PRE_READ =
       PREFIX + "parallel_implicit_pre_read.enabled";
   public static final String INCLUDE_METADATA_ENABLED = PREFIX + "include_metadata.enabled";
+  public static final String INDEX_EVENTUALLY_CONSISTENT_READ_ENABLED =
+      PREFIX + "index.eventually_consistent_read.enabled";
 
   public static final String COORDINATOR_GROUP_COMMIT_PREFIX = PREFIX + "coordinator.group_commit.";
   public static final String COORDINATOR_GROUP_COMMIT_ENABLED =
@@ -79,6 +81,7 @@ public class ConsensusCommitConfig {
   private final boolean onePhaseCommitEnabled;
   private final boolean parallelImplicitPreReadEnabled;
   private final boolean includeMetadataEnabled;
+  private final boolean indexEventuallyConsistentReadEnabled;
 
   private final boolean coordinatorGroupCommitEnabled;
   private final int coordinatorGroupCommitSlotCapacity;
@@ -152,6 +155,8 @@ public class ConsensusCommitConfig {
     parallelImplicitPreReadEnabled = getBoolean(properties, PARALLEL_IMPLICIT_PRE_READ, true);
 
     includeMetadataEnabled = getBoolean(properties, INCLUDE_METADATA_ENABLED, false);
+    indexEventuallyConsistentReadEnabled =
+        getBoolean(properties, INDEX_EVENTUALLY_CONSISTENT_READ_ENABLED, false);
 
     coordinatorGroupCommitEnabled = getBoolean(properties, COORDINATOR_GROUP_COMMIT_ENABLED, false);
     coordinatorGroupCommitSlotCapacity =
@@ -233,6 +238,10 @@ public class ConsensusCommitConfig {
 
   public boolean isIncludeMetadataEnabled() {
     return includeMetadataEnabled;
+  }
+
+  public boolean isIndexEventuallyConsistentReadEnabled() {
+    return indexEventuallyConsistentReadEnabled;
   }
 
   public boolean isCoordinatorGroupCommitEnabled() {
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
@@ -714,8 +714,12 @@ public final class ConsensusCommitUtils {
           }
 
           Set<String> secondaryIndexNames = metadata.getSecondaryIndexNames();
+          Set<String> partitionKeyNames = metadata.getPartitionKeyNames();
+          Set<String> clusteringKeyNames = metadata.getClusteringKeyNames();
           for (String indexName : secondaryIndexNames) {
             if (!indexName.startsWith(Attribute.BEFORE_PREFIX)
+                && !partitionKeyNames.contains(indexName)
+                && !clusteringKeyNames.contains(indexName)
                 && !secondaryIndexNames.contains(Attribute.BEFORE_PREFIX + indexName)) {
               tablesWithMissingIndexes.add(ScalarDbUtils.getFullTableName(namespace, table));
               break;
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
@@ -687,11 +687,22 @@ public final class ConsensusCommitUtils {
   }
 
   /**
-   * Checks all tables for missing before-image secondary indexes and logs warnings. Tables that
-   * have secondary indexes but lack corresponding before-image indexes should be repaired using
-   * {@code repairTable()}. This check is only relevant when the isolation level is SNAPSHOT or
-   * READ_COMMITTED and the index eventually consistent read is disabled, because in those cases
-   * index-based operations can miss PREPARED/DELETED records without before-image indexes.
+   * Checks all tables for missing before-image secondary indexes and logs warnings at startup.
+   * Tables that have secondary indexes but lack corresponding before-image indexes should be
+   * repaired using {@code repairTable()}.
+   *
+   * <p>This check is skipped in the following cases:
+   *
+   * <ul>
+   *   <li>SERIALIZABLE isolation: Operations that require before-image index checks will fail with
+   *       an error prompting the user to run {@code repairTable()}, so a startup warning is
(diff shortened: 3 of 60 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
```diff
@@ -223,6 +232,17 @@ public void createIndex(
...
+    if (!isIndexEventuallyConsistentReadEnabled) {
+      // Also create an index on the before image column if it exists
+      TableMetadata rawMetadata = admin.getTableMetadata(namespace, table);
+      if (rawMetadata != null) {
+        String beforeColumnName = Attribute.BEFORE_PREFIX + columnName;
+        if (rawMetadata.getColumnNames().contains(beforeColumnName)) {
+          admin.createIndex(namespace, table, beforeColumnName, options);
+        }
+      }
+    }
```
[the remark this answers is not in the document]
brfrn169: Let's keep it as-is.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -517,44 +570,168 @@ private static Set<AndConditionSet> convertConjunctions(
...
+          }
+
+          Set<String> secondaryIndexNames = metadata.getSecondaryIndexNames();
+          for (String indexName : secondaryIndexNames) {
+            if (!indexName.startsWith(Attribute.BEFORE_PREFIX)
+                && !secondaryIndexNames.contains(Attribute.BEFORE_PREFIX + indexName)) {
+              tablesWithMissingIndexes.add(ScalarDbUtils.getFullTableName(namespace, table));
+              break;
+            }
+          }
```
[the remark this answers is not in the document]
brfrn169: Fixed in 42fcdb0fb8b58e3daf3594ace5d7e0949a0ba931.

core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationCheckerTest.java
```diff
@@ -683,9 +682,12 @@ public void checkForScan_WithIndexKeyUsingClusteringKeyInSerializable_ShouldNotT
...
             .addSecondaryIndex("idx_col")
             .build();
-    Set<String> secondaryIndexNames = new LinkedHashSet<>(Collections.singletonList("idx_col"));
     when(tableMetadata.getTableMetadata()).thenReturn(metadata);
-    when(tableMetadata.getSecondaryIndexNames()).thenReturn(secondaryIndexNames);
+    TableMetadata mockTableMetadata = mock(TableMetadata.class);
+    when(mockTableMetadata.getPartitionKeyNames()).thenReturn(new LinkedHashSet<>());
+    when(mockTableMetadata.getClusteringKeyNames()).thenReturn(new LinkedHashSet<>());
+    Set<String> secondaryIndexNames = new LinkedHashSet<>(Collections.singletonList("idx_col"));
+    when(mockTableMetadata.getSecondaryIndexNames()).thenReturn(secondaryIndexNames);
```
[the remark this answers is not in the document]
brfrn169: Fixed in 42fcdb0fb8b58e3daf3594ace5d7e0949a0ba931.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
```diff
@@ -231,6 +251,17 @@ public void dropIndex(String namespace, String table, String columnName)
...
     admin.dropIndex(namespace, table, columnName);
+
+    if (!isIndexEventuallyConsistentReadEnabled) {
+      // Also drop the index on the before image column if it exists
+      TableMetadata rawMetadata = admin.getTableMetadata(namespace, table);
+      if (rawMetadata != null) {
+        String beforeColumnName = Attribute.BEFORE_PREFIX + columnName;
+        if (rawMetadata.getColumnNames().contains(beforeColumnName)) {
+          admin.dropIndex(namespace, table, beforeColumnName, true);
+        }
```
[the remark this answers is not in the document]
brfrn169: @gemini-code-assist

Regarding point 1: The `dropIndex(String, String, String, boolean)` method does exist as a default method in the `Admin` interface (see Admin.java:298), so this is not an issue.

Regarding point 2: Since we're already passing `ifExists=true`, the call is safe even if the index doesn't exist. So the current code works correctly.

integration-test/src/main/java/com/scalar/db/api/DistributedStorageAdminCaseSensitivityIntegrationTestBase.java
```diff
@@ -7,10 +7,10 @@ public abstract class DistributedStorageAdminCaseSensitivityIntegrationTestBase
   private static final String NAMESPACE1 = "Int_test_" + TEST_NAME + "1";
   private static final String NAMESPACE2 = "Int_test_" + TEST_NAME + "2";
   private static final String NAMESPACE3 = "Int_test_" + TEST_NAME + "3";
-  private static final String TABLE1 = "Test_table1";
-  private static final String TABLE2 = "Test_table2";
-  private static final String TABLE3 = "Test_table3";
-  private static final String TABLE4 = "Test_table4";
+  private static final String TABLE1 = "Tbl1";
```
brfrn169, line 10: Changed table names in integration tests to shorter ones to avoid index name length limitations in the underlying databases.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
```diff
@@ -231,6 +251,17 @@ public void dropIndex(String namespace, String table, String columnName)
...
 
     admin.dropIndex(namespace, table, columnName);
+
+    if (!isIndexEventuallyConsistentReadEnabled) {
+      // Also drop the index on the before image column if it exists
+      TableMetadata rawMetadata = admin.getTableMetadata(namespace, table);
+      if (rawMetadata != null) {
+        String beforeColumnName = Attribute.BEFORE_PREFIX + columnName;
+        if (rawMetadata.getColumnNames().contains(beforeColumnName)) {
+          admin.dropIndex(namespace, table, beforeColumnName, true);
```
komamitsu, line 261: I think an index on a before image column created by `createIndex()` with `isIndexEventuallyConsistentReadEnabled=false` won't be dropped with `dropIndex()` with `isIndexEventuallyConsistentReadEnabled=true`. Is this expected?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -112,65 +121,91 @@ public Optional<Result> get(Get get, TransactionContext context) throws CrudExce
...
+        if (key != null) {
           putIntoReadSetInSnapshot(key, result, context);
+        } else {
+          // Only for a Get with index, the argument `key` is null
+
+          if (result.isPresent()) {
+            // Only when we can get the record with the Get with index, we can put it into the read
+            // set
+            key = new Snapshot.Key(get, result.get(), metadata);
+            putIntoReadSetInSnapshot(key, result, context);
```
komamitsu, line 200: I'm wondering about the following case:
- The argument `key` is null
- In the first iteration
  - a record which has `key:A` is scanned. `key` is set to `A`
  - `checkAndRecoverBeforeIndexRecords()` scans another uncommitted record. It's rolled back and the loop will be continued
- In the second iteration
  - another record which has `key:B` is scanned. `key` is not null and still `A`. But `result` is set to `B`
  - `putIntoReadSetInSnapshot(key:A, result:B)` is called. Inconsistent?

Is my understanding correct? If so, it doesn't matter?

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -112,65 +121,91 @@ public Optional<Result> get(Get get, TransactionContext context) throws CrudExce
...
-      // Because we also get records whose before images match the conjunctions, we need to check if
-      // the current status of the records actually match the conjunctions.
-      result =
-          result.filter(
-              r ->
-                  ScalarDbUtils.columnsMatchAnyOfConjunctions(
-                      r.getColumns(), get.getConjunctions()));
-    }
+        RecoveryExecutor.Result recoveryResult = executeRecovery(key, get, result.get(), context);
+        context.recoveryResults.add(recoveryResult);
```
komamitsu, line 158: This method is called by `readIfImplicitPreReadEnabled()` concurrently with shared `context`. `context.recoveryResults` is actually an `ArrayList`, so it could be a potential race condition issue?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
brfrn169: Yes, that's expected. 

`isIndexEventuallyConsistentReadEnabled` (`scalar.db.consensus_commit.index.eventually_consistent_read.enabled`) is a parameter for backward compatibility and to prevent performance degradation. When it is true, the previous behavior (without before-image indexes) is maintained.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Good catch! Fixed in 3359bd164d0a8b7ef5f861a750526e378f24db95.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Ah, it looks like this is an existing bug... Good catch! Fixed in b9f8e844bdb1566708d9b6d49f0fa1aeca1e5265.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -112,65 +121,93 @@ public Optional<Result> get(Get get, TransactionContext context) throws CrudExce
...
+      }
 
-        if (result.isPresent()) {
-          // Only when we can get the record with the Get with index, we can put it into the read
-          // set
-          key = new Snapshot.Key(get, result.get(), metadata);
+      // Check if there are PREPARED/DELETED records whose committed (before-image) values match
+      // the query. If any were rolled back, the storage read result may be stale, so retry from
+      // the beginning before caching into the snapshot.
+      if (beforeIndexCheckRequired && checkAndRecoverBeforeIndexRecords(get, context, txMetadata)) {
```
komamitsu, line 175: I'm wondering if we can skip this when `result` is already present, since we can return the `result` regardless of uncommitted records that will eventually match the condition.

brfrn169: Good question! I think we can't skip the before-image index check even when `result` is already present because the retry after recovery could change the number of matching records.

For example, consider the following state:

| Record | BALANCE | State |
|--------|---------|-------|
| A | 10 | COMMITTED |
| B | 20 (before: 10) | PREPARED |

When executing `Get with index (BALANCE = 10)`:

1. The normal index lookup returns record A (1 hit, `result` is present)
2. The before-image index check finds record B via `before_BALANCE = 10` and recovers it. If B is rolled back, BALANCE reverts to 10
3. On retry, the normal index lookup now finds both A and B (2 hits), which should result in an error since `Get` expects at most one result

If we skip the before-image index check when `result` is present, we would return A without detecting this situation.

What do you think?

komamitsu: > which should result in an error since Get expects at most one result

Ah, I see. I completely missed it. Let's go with as is.

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -517,44 +570,172 @@ private static Set<AndConditionSet> convertConjunctions(
...
+   * {@code repairTable()}. This check is only relevant when the isolation level is SNAPSHOT or
+   * READ_COMMITTED and the index eventually consistent read is disabled, because in those cases
+   * index-based operations can miss PREPARED/DELETED records without before-image indexes.
+   *
+   * @param admin a distributed storage admin
+   * @param config a consensus commit config
+   */
+  static void warnIfBeforeImageIndexesAreMissing(
+      DistributedStorageAdmin admin, ConsensusCommitConfig config) {
+    if (config.getIsolation() == Isolation.SERIALIZABLE
```
feeblefakie, line 712: I probably miss something, but why don't we need to warn in case the isolation is SERIALIZABLE?
I want the reason to be noted in the comment.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -562,13 +611,131 @@ private TableMetadata getTableMetadata(Operation operation, String transactionId
...
+          // Always use RETURN_LATEST_RESULT_AND_RECOVER regardless of isolation level because
+          // recovery must actually execute; otherwise, the PREPARED/DELETED record remains in
+          // storage and retrying would be useless
+          RecoveryExecutor.Result recoveryResult =
+              recoveryExecutor.execute(
+                  key,
+                  beforeIndexScan,
+                  result,
+                  context.transactionId,
+                  RecoveryExecutor.RecoveryType.RETURN_LATEST_RESULT_AND_RECOVER);
```
feeblefakie, line 703: Just a question or confirmation.
When target records are prepared states and the coordinator entry for the transaction does not exist yet, and it has not expired yet, then the recovery will throw UncommittedRecordException, then this processing will also end with an exception, right?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
```diff
@@ -152,6 +155,8 @@ public ConsensusCommitConfig(DatabaseConfig databaseConfig) {
     parallelImplicitPreReadEnabled = getBoolean(properties, PARALLEL_IMPLICIT_PRE_READ, true);
 
     includeMetadataEnabled = getBoolean(properties, INCLUDE_METADATA_ENABLED, false);
+    indexEventuallyConsistentReadEnabled =
+        getBoolean(properties, INDEX_EVENTUALLY_CONSISTENT_READ_ENABLED, false);
```
feeblefakie, line 159: We discussed it, but I forgot the reason why it is `false` by default.
Users who already run ScalarDB in production use secondary indexes with eventual consistency, so setting it to `true` might be more appropriate, since they don't need to perform a repair during the upgrade.

feeblefakie (commented):

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
brfrn169: The default is `false` (i.e., the before-image index check is enabled) because we basically want users to get the correct behavior out of the box without additional configuration.

For existing users upgrading, they can either set `scalar.db.consensus_commit.index.eventually_consistent_read.enabled=true` to maintain the previous behavior, or run `repairTable()` to add the before-image indexes and benefit from the improved consistency. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Yes, that's correct. When the coordinator entry doesn't exist and the transaction hasn't expired, `RecoveryExecutor` throws `UncommittedRecordException`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
brfrn169: In SERIALIZABLE isolation, operations that require before-image index checks will fail with an error prompting the user to run `repairTable()`, so a startup warning is unnecessary.

I've updated the Javadoc in 5e561e03be6b17014e8269970166ff89e4482937.

Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitConfig.java
feeblefakie: We were probably discussing that we can let users set `scalar.db.consensus_commit.index.eventually_consistent_read.enabled=true` for backward compatibility...
Anyway, I got it. Thanks.

feeblefakie (approved):
LGTM! Thank you!

PR #3441, opened 2026-03-28
Close BufferedReader in process_withSingleCrudMode

tejas-ae:
There might be a small issue in process_withSingleCrudMode_shouldProcessAllDataChunks: The resource opened there can leak if process_withSingleCrudMode_shouldProcessAllDataChunks exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path. Let me know if I’m missing context.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


tejas-ae:
Updated — wrapped all remaining `BufferedReader` instances in try-with-resources across all test methods in the file.

PR #3463, opened 2026-03-31
Support SERIALIZABLE isolation for index-based operations with before-image index validation

brfrn169:
## Description

PR #3419 introduced before-image secondary index checks to fix index-based Get/Scan missing PREPARED/DELETED records, but excluded SERIALIZABLE isolation support (noted in the "Additional notes" section). This PR extends that work to support SERIALIZABLE isolation for index-based operations (Get with index, Scan with index, and ScanAll with indexed column conditions) by leveraging before-image indexes for anti-dependency validation during the SERIALIZABLE extra-read phase.

## Related issues and/or PRs

- #3419

### Lifted restriction on index-based operations in SERIALIZABLE

- Updated `ConsensusCommitOperationChecker` to allow index-based Get, Scan, and ScanAll with indexed column conditions in SERIALIZABLE isolation when before-image indexes exist for the table. Previously, these operations unconditionally threw `IllegalArgumentException` in SERIALIZABLE.
- Updated error messages in `CoreError` to inform users they can run `repairTable()` to create before-image indexes to enable these operations.

### Before-image index validation in Snapshot (SERIALIZABLE extra-read)

- Added `validateBeforeIndex()` method to `Snapshot` that checks for uncommitted records from other transactions on the before-image index during SERIALIZABLE validation. If any uncommitted record is found, validation fails with a conflict exception, preventing phantom reads.
- Updated SERIALIZABLE validation for scan set, scanner set, and get set to call `validateBeforeIndex()` after the standard validation.

### Refactoring

- Extracted the `requiresBeforeIndexCheck()` logic from `CrudHandler` into `ConsensusCommitUtils` so it can be reused by both `CrudHandler` and `Snapshot`.
- Refactored `validateScanResults()` in Snapshot to accept `TableMetadata` as a parameter to avoid redundant metadata lookups, and modernized resource handling to use try-with-resources.

### Tests

- Added unit tests for `ConsensusCommitOperationChecker`, `ConsensusCommitUtils.requiresBeforeIndexCheck()`, and `Snapshot.validateBeforeIndex()`.
- Updated integration tests to verify that index-based operations now succeed in SERIALIZABLE when before-image indexes are present, and that commit conflicts are correctly detected when uncommitted records exist on the before-image index.

## Checklist

- Tables without before-image indexes will continue to reject index-based operations in SERIALIZABLE, with updated error messages guiding users to run `repairTable()`.
- The before-image index validation in SERIALIZABLE is conservative: it fails if any uncommitted record from another transaction is found on the before-image index, which may cause false positives (e.g., when the record will be rolled forward and its committed index value won't actually match the scan condition). On retry, the record should be committed, so the retry will succeed.

## Release notes

Same as #3419


core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -610,20 +631,19 @@ void toSerializable(DistributedStorage storage)
    * @param results the results of the scan
    * @param notFullyScannedScanner if this is a validation for a scanner that has not been fully
    *     scanned
+   * @param metadata a
```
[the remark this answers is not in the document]
brfrn169: Fixed in c6ceafbf30b901d72daed5b4b1cfecfa3beacc78.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -914,19 +914,22 @@ public enum CoreError implements ScalarDbError {
...
+      "Scan operations by using an index are not allowed in the SERIALIZABLE isolation level without before-image indexes. "
+          + "Run repairTable() to create before-image indexes for the table, which will enable index-based Scan operations in the SERIALIZABLE isolation level",
       "",
       ""),
   CONSENSUS_COMMIT_CONDITION_ON_INDEXED_COLUMNS_NOT_ALLOWED_IN_CROSS_PARTITION_SCAN_IN_SERIALIZABLE(
       Category.USER_ERROR,
       "0262",
-      "Conditions on indexed columns in cross-partition scan operations are not allowed in the SERIALIZABLE isolation level",
+      "Conditions on indexed columns in cross-partition scan operations are not allowed in the SERIALIZABLE isolation level without before-image indexes. "
+          + "Run repairTable() to create before-image indexes for the table, which will enable conditions on indexed columns in cross-partition scan operations in the SERIALIZABLE isolation level",
```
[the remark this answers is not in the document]
brfrn169: Fixed in c6ceafbf30b901d72daed5b4b1cfecfa3beacc78.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -810,10 +824,58 @@ private void validateGetResult(
...
+      for (Result result : scanner) {
+        TransactionResult txResult = new TransactionResult(result);
+        // Conservatively fail if any uncommitted record from another transaction is found on the
+        // before-image index. This may cause false positives (e.g., when the record will be
+        // rolled forward and its committed index value won't actually match the scan condition),
+        // but it guarantees correctness. On retry, the record should be committed, so the retry
+        // will succeed.
+        if (!txResult.isCommitted() && !id.equals(txResult.getId())) {
+          throwExceptionDueToAntiDependency();
+        }
```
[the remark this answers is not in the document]
brfrn169: Fixed in 9eed1c1dd376b9481d55e905741d4bf086eddfc1.

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -625,6 +625,37 @@ private static AndConditionSet convertConjunctionToBeforeImage(
...
+    if (selection instanceof ScanAll) {
+      for (Selection.Conjunction conjunction : selection.getConjunctions()) {
+        for (ConditionalExpression condition : conjunction.getConditions()) {
+          if (metadata.hasBeforeImageSecondaryIndex(condition.getColumn().getName())) {
+            return true;
+          }
+        }
+      }
+      return false;
+    }
```
feeblefakie: It looks like, without proper secondary indexes for the before images of the columns used in conditions, it will treat the selection as not requiring before-index checking.
Is it really OK for correctness?
Usually, there are no secondary indexes for the before columns, so it seems to always return `false` unless users create indexes explicitly?
Since what secondary indexes are needed is dependent on selection conditions, it seems hard to pre-build indexes for those.

I'm probably confused, but please clarify this.

brfrn169: The before-image secondary indexes are automatically created during table creation in `buildTransactionTableMetadata`. Specifically, when a column has a secondary index, the corresponding `before_<column>` secondary index is also automatically added. So users don't need to create them explicitly.

However, tables created before this feature was introduced may not have before-image secondary indexes. In that case, running `repairTable()` will add them.

The reason `requiresBeforeIndexCheck` only checks for the existence of the before-image index is that before-image indexes are always a subset of the original column indexes — if a before-image index exists, the original column's index is guaranteed to exist as well.

As for the behavior when before-image indexes don't exist:

- SERIALIZABLE: The operation is rejected by `ConsensusCommitOperationChecker` with an error message guiding the user to run `repairTable()`, so `requiresBeforeIndexCheck` is never reached.
- SNAPSHOT / READ_COMMITTED: The before-image index check is skipped, but a warning is logged at startup via `warnIfBeforeImageIndexesAreMissing`.

Does that answer your question?

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
feeblefakie: Thank you for the explanation.
I understand it.

But I think the location of the operation checker's validation and the location of this checking are too far apart, so we should leave some comments noting the following things:

- The non-existence of the before-image index doesn't occur in the SERIALIZABLE case because the operation checker prevents executing the selection without the index.
- Therefore, the existence of before-imange index means the selection requires the before-image index check.

Also, another question.
In case of SNAPSHOT and READ_COMMITTED, is the non-existence of the before-image index fine? Could selection return eventually consistent data in that case?

brfrn169: Thank you for the feedback!

I added comments as you suggested, but I also refactored the code to make the separation of concerns clearer in 512c202b7a6c59ef7fb2ead548f6aef2566fc62a.

The key issue was that `ConsensusCommitUtils.requiresBeforeIndexCheck()` was shared between two call sites with different semantics:

1. `CrudHandler` (all isolation levels): Needs to check both whether the operation is index-based AND whether the before-image index exists, because in SNAPSHOT/READ_COMMITTED the before-image index may not exist.
2. `Snapshot.validateBeforeIndex()` (SERIALIZABLE only): Only needs to check whether the operation is index-based, because `ConsensusCommitOperationChecker` guarantees that before-image indexes exist in SERIALIZABLE.

So instead of adding comments to explain this shared method, I separated the logic:

- `Snapshot`: Replaced the `requiresBeforeIndexCheck()` call with a new `isIndexBasedOperation()` method that only checks whether the selection is index-based, with a Javadoc comment explaining that in SERIALIZABLE, the existence of before-image indexes is guaranteed by the operation checker.
- `CrudHandler`: Moved `requiresBeforeIndexCheck()` from `ConsensusCommitUtils` back into `CrudHandler` (since it's the only caller now), with a Javadoc comment explaining that when before-image indexes don't exist, the check is skipped and index-based reads may return eventually consistent results in SNAPSHOT/READ_COMMITTED (a known limitation).
- `ConsensusCommitUtils`: Removed `requiresBeforeIndexCheck()`.

Regarding your question about SNAPSHOT/READ_COMMITTED: yes, if before-image indexes don't exist, index-based reads may return eventually consistent results. This is a known limitation, and a warning is logged at startup via `warnIfBeforeImageIndexesAreMissing`.

brfrn169: As discussed, I added a clarifying comment to `isIndexBasedOperation()` explaining that for ScanAll, whether the underlying storage actually uses the index depends on the storage implementation, but this method considers it an index-based operation if the conditions reference indexed columns (03a3c805f57272d2a4a15e0ebfef07caa41000e4). Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #3486, opened 2026-04-09
Add dependency license check and third-party license notices

kota2and3kan:
## Description

This PR adds a new library `com.github.jk1.dependency-license-report` for the following two purposes:

1. Generate the `THIRD-PARTY-NOTICES.txt` file, which shows third-party license information.
    - I put this `THIRD-PARTY-NOTICES.txt` file under the `/licenses` directory.
    - The CI `.github/workflows/update-third-party-notices.yaml` automatically updates this file once a quarter.
1. Check the licenses that are used in ScalarDB Core. This can check whether using libraries that are provided under an unexpected license (i.e., guarantee using allowed license libraries only).
    - The CI `.github/workflows/license-check.yaml` checks per PRs.

Also, this PR updates the existing license information as follows:

- Move the `/LICENSE` file to `/licenses/LICENSE`.
- Update the `License` section of README to refer the `/licenses` directory.

Please take a look!

## Related issues and/or PRs

- This PR applies almost same the changes of the following PRs.
  - https://github.com/scalar-labs/scalardb-cluster/pull/1682
  - https://github.com/scalar-labs/scalardb-cluster/pull/1697

## Changes made

- Add `com.github.jk1.dependency-license-report`.
  - You can run the two Gradle tasks `./gradlew generateLicenseReport` and `./gradlew checkLicense`.
    - `./gradlew generateLicenseReport` generates `THIRD-PARTY-NOTICES.txt` file. We place this file under the `/licenses` directory.
    - `./gradlew checkLicense` checks whether ScalarDB Core uses only the allowed license or not.
- Add CI to run the `./gradlew checkLicense` command per PRs.
- Add `buildSrc/src/main/java/com/scalar/db/PomOnlyFilter.java` to avoid an issue.
  - I will add more details as a review comment.

## Checklist

N/A

## Release notes

N/A


buildSrc/src/main/java/com/scalar/db/PomOnlyFilter.java
kota2and3kan, line 1: This `buildSrc/src/main/java/com/scalar/db/PomOnlyFilter.java` is necessary for the following reason:

# Issue of the license detection feature

This tool collects license information from three places:

1. POM
    - https://github.com/jk1/Gradle-License-Report/blob/v3.1.1/src/main/groovy/com/github/jk1/license/reader/PomReader.groovy
2. Manifest File
    - https://github.com/jk1/Gradle-License-Report/blob/v3.1.1/src/main/groovy/com/github/jk1/license/reader/ManifestReader.groovy
3. LICENSE File
    - https://github.com/jk1/Gradle-License-Report/blob/v3.1.1/src/main/groovy/com/github/jk1/license/reader/LicenseFilesReader.groovy

And, this library merges all license information of the above three sources.

However, sometimes, these three sources and merging behavior cause an issue. It misjudges the license of a library.

This issue occurs due to the following two problems:

1. In the `LicenseFilesReader.groovy`, it just does the text matching in the `LICENSE` file.
    - https://github.com/jk1/Gradle-License-Report/blob/v3.1.1/src/main/groovy/com/github/jk1/license/reader/LicenseFilesReader.groovy#L103-L129
2. Some `LICENSE` files include not only their own license but also the license information of dependency libraries.
    - For example, the LICENSE file of MySQL Connector J includes both its own license and the dependency libraries' license information.

For example, MySQL Connector J is NOT provided under the `Apache License, Version 2.0`. However, this tool shows that MySQL Connector J is provided under `Apache License, Version 2.0` as follows:

```json
{
    "moduleName": "com.mysql:mysql-connector-j",
    "moduleVersion": "8.4.0",
    "moduleUrls": [
        "http://dev.mysql.com/doc/connector-j/en/"
    ],
    "moduleLicenses": [
        {
            "moduleLicense": "Apache License, Version 2.0",
            "moduleLicenseUrl": "https://www.apache.org/licenses/LICENSE-2.0"
        },
        {
            "moduleLicense": "GNU GENERAL PUBLIC LICENSE, Version 1",
            "moduleLicenseUrl": "https://www.gnu.org/licenses/gpl-1.0"
        },
        {
            "moduleLicense": "GNU GENERAL PUBLIC LICENSE, Version 2",
            "moduleLicenseUrl": "https://www.gnu.org/licenses/gpl-2.0"
        },
        {
            "moduleLicense": "GNU GENERAL PUBLIC LICENSE, Version 2 + Universal FOSS Exception, Version 1.0",
            "moduleLicenseUrl": "https://www.mysql.com/about/legal/licensing/oem/"
        }
    ]
}
```

This is because the LICENSE file of MySQL Connector J includes the following part (license information of `opentelemetry-java` that is a dependency of MySQL Connector J).

```console
opentelemetry-java

You may be receiving a copy of the opentelemetry-java library with this
MySQL product.

The terms of the Oracle license do NOT apply to the opentelemetry-java library;
it is licensed under the following license, separately from the Oracle programs
you receive.

If you do not wish to install this program, you may delete its files but the
Oracle program might not operate properly or at all without it.

/*
 * Copyright The OpenTelemetry Authors
 * SPDX-License-Identifier: Apache-2.0
 */

See Apache License v2.0, January 2004 in the 'Standard Licenses' section.
```

This behavior causes the misjudging issue in the `./gradlew checkLicense` command. It cannot detect that MySQL Connector J is not allowed to be used in this project, even if we don't allow the GPL v2 in the `allowed-licenses.json` file.

# How to resolve the issue of the license detection feature

To resolve the above issue, we can use only the POM information. I think the POM is the most reliable source of the license information without ambiguity (Although it depends on the author's description...).

We can achieve it by defining our own `DependencyFilter` that removes license information provided by the manifest file and the license file.

- https://github.com/jk1/Gradle-License-Report/blob/v3.1.1/src/main/groovy/com/github/jk1/license/filter/DependencyFilter.groovy

The `DependencyFilter` has one method that takes `ProjectData` as an argument and returns `ProjectData`.

Also, in the `ProjectData`, the license information is stored in `Set`.

- https://github.com/jk1/Gradle-License-Report/blob/v3.1.1/src/main/groovy/com/github/jk1/license/Model.groovy#L65-L67

    ```java
    class ModuleData {
        String group, name, version
        boolean hasArtifactFile
        Set<ManifestData> manifests = new TreeSet<ManifestData>()
        Set<LicenseFileData> licenseFiles = new TreeSet<LicenseFileData>()
        Set<PomData> poms = new TreeSet<PomData>()
    
        boolean isEmpty() { manifests.isEmpty() && poms.isEmpty() && licenseFiles.isEmpty() }
    }
    ```

Therefore, we can make these `Set` of license file and manifest file empty by using `Set.clear()`.

This is what I implemented in the `PomOnlyFilter` class.

# Result of the `PomOnlyFilter` class

After we apply the `PomOnlyFilter`, this library detects the license information of MySQL Connector J properly based on the only POM information as follows:

```json
{
    "moduleName": "com.mysql:mysql-connector-j",
    "moduleVersion": "8.4.0",
    "moduleUrls": [
        "http://dev.mysql.com/doc/connector-j/en/"
    ],
    "moduleLicenses": [
        {
            "moduleLicense": "GNU GENERAL PUBLIC LICENSE, Version 1",
            "moduleLicenseUrl": "https://www.gnu.org/licenses/gpl-1.0"
        },
        {
            "moduleLicense": "GNU GENERAL PUBLIC LICENSE, Version 2",
            "moduleLicenseUrl": "https://www.gnu.org/licenses/gpl-2.0"
        },
        {
            "moduleLicense": "GNU GENERAL PUBLIC LICENSE, Version 2 + Universal FOSS Exception, Version 1.0",
            "moduleLicenseUrl": "https://www.mysql.com/about/legal/licensing/oem/"
        }
    ]
}
```

Now, it does not include `Apache License, Version 2.0` (it does not have misjudgment).

This is not a completely safe workaround because it depends on the POM information that is managed by the authors of each dependency library. However, at least, we can avoid misjudgment caused by the `LICENSE` file that includes all dependency libraries' information.

.github/workflows/update-third-party-notices.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Create pull request
+        if: steps.diff.outputs.changed == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          BRANCH="update-third-party-notices-$(date +%Y%m%d)"
+
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
kota2and3kan, line 61: We can use these `user.name` and `user.email` to commit with the GitHub Actions Bot user.

- https://github.com/orgs/community/discussions/119597

licenses/THIRD-PARTY-NOTICES.txt
kota2and3kan, line 1: This file is automatically generated by the `./gradlew generateLicenseReport` command.

Basically, you don't need to check the details of this file in this PR.

README.md
```diff
@@ -80,7 +80,4 @@ When contributing to this project, please follow these guidelines.
 
 ## License
 
-ScalarDB is dual-licensed under both the Apache 2.0 License (found in the LICENSE file in the root directory) and a commercial license.
-You may select, at your option, one of the above-listed licenses.
-The commercial license includes several enterprise-grade features such as ScalarDB Cluster, management tools, and declarative query interfaces like GraphQL and SQL interfaces.
-For more information about the commercial license, please [contact us](https://www.scalar-labs.com/contact).
+See [licenses](./licenses/) directory.
```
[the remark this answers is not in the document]
kota2and3kan: For now, the enterprise features that are provided under the commercial license are completely separated from this repository into other repositories.

Therefore, I think we don't need to mention the commercial license here.

@feeblefakie 
What do you think?

.github/workflows/update-third-party-notices.yaml
kota2and3kan, line 1: We have a plan that applies the same updates to other repositories (e.g., ScalarDL Ledger).

So, it might be better to use the reusable workflow and maintain it on the https://github.com/scalar-labs/actions side.

However, there is no concrete schedule for now.

Therefore, I added this workflow to this repository directly.

I will create a reusable workflow on the https://github.com/scalar-labs/actions side and update this workflow in the future, if needed.

buildSrc/src/main/java/com/scalar/db/PomOnlyFilter.java
feeblefakie: Since the `scalardb-cluster` is a private repo, can you note it here as well?
(I think you can overwrite the current comment if possible)

README.md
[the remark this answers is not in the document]
komamitsu: I think putting LICENSE file in the project root directory is a kind of OSS standard way. There might be a tool automatically checks LICENSE files only in the project root. How about keeping a copy in the root rather than moving it?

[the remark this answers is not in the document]
feeblefakie: I think putting LICENSE in the root and putting the third-party licenses in the `licenses` directory seems to be common.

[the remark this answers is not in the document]
feeblefakie: > For now, the enterprise features that are provided under the commercial license are completely separated from this repository into other repositories.
> 
> Therefore, I think we don't need to mention the commercial license here.

Yeah, we can delete the commercial part for now.

Torch3333 (approved):
LGTM, thank you!

README.md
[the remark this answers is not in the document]
kota2and3kan: Thank you for your review!

> I think putting LICENSE file in the project root directory is a kind of OSS standard way. There might be a tool automatically checks LICENSE files only in the project root. How about keeping a copy in the root rather than moving it?

> I think putting LICENSE in the root and putting the third-party licenses in the licenses directory seems to be common.

Ah, I see. Indeed, I think your suggested pattern (putting the `LICENSE` file in the project root) is standard.

Actually, putting the `LICENSE` file under the `/licenses` directory is a requirement of Red Hat Ecosystem Catalog. And, we've already satisfied on it on the ScalarDB Cluster side. In other words, we don't need to satisfy the requirement in this (OSS) repository.

Therefore, I will move the `LICENSE` file back to the project root.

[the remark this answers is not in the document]
kota2and3kan: Fixed in 18a3db32ac5c9cf4ab60a1b7978a286f86713302.

komamitsu (approved):
LGTM, thank you!

.github/workflows/update-third-party-notices.yaml
```diff
@@ -0,0 +1,75 @@
+name: Update THIRD-PARTY-NOTICES.txt
+
+on:
+  schedule:
+    # Run quarterly: 01:00 UTC (10:00 JST) on the 1st of January, April, July, and October.
+    - cron: "0 1 1 1,4,7,10 *"
```
brfrn169, line 6: I’m wondering whether a quarterly update is really the right frequency. Maybe we should run it via `workflow_dispatch` before releasing new versions, rather than on a fixed schedule?

brfrn169 (approved):
Left a comment. Other than that, LGTM! Thank you!

BTW, it looks like `License Compliance` in the checks failed. Is that okay?

.github/workflows/update-third-party-notices.yaml
kota2and3kan: Honestly, I don't have a strong opinion or clear reason about when we should run this workflow (when we update the third-party license information).

Ideally, if we add a new third-party library, we should always update the third-party license information. However, I think it doesn't occur so frequently.

Also, I think forcing the check and updating the third-party license information in all PR is a bit troublesome for developers.

Therefore, to avoid forgetting to update (to keep license information correct in an eventual consistent way), I set the schedule update, just in case.

It's like a guardrail. I think we shouldn't rely on this guardrail completely.

Therefore, as you mentioned, I think it's better to update the third-party license information per release by using `workflow_dispatch`.

It's not a high-priority thing, but let me discuss the concrete operation for this later.

kota2and3kan (commented):
@brfrn169 
Thank you for your review!

> BTW, it looks like `License Compliance` in the checks failed. Is that okay?

FOSSA detects the GPLv2 (without Classpath Exception) based on the text in the line `7823` of `/licenses/THIRD-PARTY-NOTICES.txt` as follows:
(I guess FOSSA checks strings simply.)

```console
The GNU General Public License (GPL) Version 2, June 1991
```

However, in addition to the above strings, there is an article of Classpath Exception as follows:

```console
"CLASSPATH" EXCEPTION TO THE GPL VERSION 2
```

Also, from the POM information, we can see that the license is `GNU GENERAL PUBLIC LICENSE, Version 2 + Classpath Exception` as well.

```console
144. Group: javax.xml.bind  Name: jaxb-api  Version: 2.3.1

Manifest Project URL: http://www.oracle.com/

Manifest license URL: https://oss.oracle.com/licenses/CDDL-1.1

POM License: COMMON DEVELOPMENT AND DISTRIBUTION LICENSE (CDDL) Version 1.1 - https://oss.oracle.com/licenses/CDDL-1.1

POM License: GNU GENERAL PUBLIC LICENSE, Version 2 + Classpath Exception - https://openjdk.java.net/legal/gplv2+ce.html
```

Therefore, from the license perspective, there is no issue.

---

As mentioned above, there is no issue from the license perspective, but the CI fails.

And, we (@feeblefakie and I) checked the configuration of FOSSA. As a result, in the current plan, unfortunately, it seems that we cannot address CI failure (misdetection)...

We need to continue discussing it, but we can ignore the failure of `License Compliance`, for now.

feeblefakie (approved):
LGTM! Thank you!

PR #3488, opened 2026-04-10
Fix index-based Get/Scan to filter out results whose index key no longer matches after lazy recovery rollback

brfrn169:
## Description

When a PREPARED record is found via the normal secondary index during an index-based Get/Scan, and lazy recovery rolls back the transaction, the index column value reverts to its before-image value. If this reverted value no longer matches the queried index key, the result should be filtered out. Previously, such results were returned incorrectly.

For example:
1. Transaction T1 updates col_idx from "B" to "A" (PREPARED state)
2. Transaction T2 queries col_idx = "A" via index-based Get/Scan
3. The PREPARED record is found via the normal index (col_idx = "A")
4. Lazy recovery rolls back T1, reverting col_idx to "B"
5. Bug: The record with col_idx = "B" was returned for the col_idx = "A" query

This PR addresses this issue.

## Related issues and/or PRs

N/A

## Changes made

- Added index key validation after lazy recovery rollback in `CrudHandler.read()` (Get) and `CrudHandler.processScanResult()` (Scan). When `recoveryResult.rolledBack` is true and the operation uses a secondary index, the recovered result is filtered using `resultMatchesIndexKey()`.
- Added `resultMatchesIndexKey()` private method in CrudHandler that compares the result's index column value against the queried index key.
- Added unit tests in `CrudHandlerTest` covering rollback with matching/non-matching index keys for both `Get` and `Scan` operations.
- Added integration tests in `ConsensusCommitSpecificIntegrationTestBase` for `Get`, `Scan`, and `getScanner` that verify rolled-back PREPARED records are correctly filtered out when the index key matches the after-image.

## Checklist

- The filtering is only applied when `rolledBack` is true because roll-forward (commit) preserves the after-image value, which already matches the queried index key by definition — the record was found via the normal index because its after-image matched the query.

## Release notes

Fixed a bug where index-based Get and Scan operations could return incorrect results after lazy recovery rolled back a PREPARED record whose after-image index value matched the query but whose before-image (restored) value did not.


core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -658,6 +676,21 @@ boolean requiresBeforeIndexCheck(Selection selection, TransactionTableMetadata m
...
+   * Checks if the result's index column value matches the queried index key value. This is needed
+   * because after lazy recovery (e.g., rollback), the index column value may revert to its original
+   * value, which might not match the queried index key.
+   *
+   * @param selection the index-based selection operation
+   * @param result the result to check
+   * @return true if the result's index column matches the queried index key
+   */
+  private boolean resultMatchesIndexKey(Selection selection, TransactionResult result) {
+    Column<?> indexColumn = selection.getPartitionKey().getColumns().get(0);
```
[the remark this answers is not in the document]
brfrn169: Fixed in 186f6a86bd7359bc01c09ca0c6b82faa7a3a2b31.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -156,12 +157,19 @@ Optional<TransactionResult> read(
...
         context.recoveryResults.add(recoveryResult);
-        result = recoveryResult.recoveredResult;
+
+        // After recovery (e.g., rollback), the index column value may have changed back to its
+        // original value, which might not match the queried index key. Filter out such results.
+        if (recoveryResult.rolledBack && ScalarDbUtils.isSecondaryIndexSpecified(get, metadata)) {
+          result = recoveryResult.recoveredResult.filter(r -> resultMatchesIndexKey(get, r));
+        } else {
+          result = recoveryResult.recoveredResult;
+        }
```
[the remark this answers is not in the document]
brfrn169: Fixed in 186f6a86bd7359bc01c09ca0c6b82faa7a3a2b31.

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -2411,6 +2411,162 @@ public void get_GetWithIndexForDeletedWhenCoordinatorStateAborted_ShouldRollBack
...
+    // After roll-back, the PREPARED record (0,1) has BALANCE reverted to INITIAL_BALANCE, which
+    // does not match the queried index key BALANCE=NEW_BALANCE. Only the 2 COMMITTED records
+    // (0,0) and (0,2) with BALANCE=NEW_BALANCE should remain.
+    assertThat(results).hasSize(2);
+    assertThat(results.get(0).getInt(ACCOUNT_ID)).isEqualTo(0);
+    assertThat(results.get(0).getInt(ACCOUNT_TYPE)).isEqualTo(0);
+    assertThat(results.get(0).getInt(BALANCE)).isEqualTo(NEW_BALANCE);
+    assertThat(results.get(1).getInt(ACCOUNT_ID)).isEqualTo(0);
+    assertThat(results.get(1).getInt(ACCOUNT_TYPE)).isEqualTo(2);
+    assertThat(results.get(1).getInt(BALANCE)).isEqualTo(NEW_BALANCE);
```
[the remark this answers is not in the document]
brfrn169: Fixed in aeff947f5137976d3c3f7ba68648bae65ce1b2eb.

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -2411,6 +2411,163 @@ public void get_GetWithIndexForDeletedWhenCoordinatorStateAborted_ShouldRollBack
...
+   * after-image index. Creates:
+   *
+   * <ul>
+   *   <li>(0,0): COMMITTED with BALANCE=NEW_BALANCE
+   *   <li>(0,2): COMMITTED with BALANCE=NEW_BALANCE
+   *   <li>(0,1): PREPARED/DELETED with BALANCE changed from INITIAL_BALANCE to NEW_BALANCE
+   * </ul>
+   *
+   * When querying BALANCE=NEW_BALANCE, all 3 records match the normal index. After rollback of
+   * (0,1), its BALANCE reverts to INITIAL_BALANCE, so it should be filtered out, leaving only the 2
```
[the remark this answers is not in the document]
brfrn169: Fixed in 9309986ed10aab8d2d4c9df0213176271d9af57c.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -183,10 +196,12 @@ Optional<TransactionResult> read(
       }
 
       // Put the result in the snapshot
-      if (result.isPresent() || get.getConjunctions().isEmpty()) {
+      if (result.isPresent() || (get.getConjunctions().isEmpty() && !indexKeyFilteredOut)) {
         // We put the result into the read set only if a get operation has no conjunction or the
         // result exists. This is because we don’t know whether the record actually exists or not
-        // due to the conjunction.
+        // due to the conjunction. Additionally, when a result was filtered out because the index
```
feeblefakie: I found it difficult to understand the meaning of the second condition `get.getConjunctions().isEmpty() && !indexKeyFilteredOut)`. Can you elaborate on that?
I felt it was partly from the fact that the condition is `!indexKeyFilteredOut` but the description explains a case of ` when a result was filtered out`.

brfrn169: Thanks for the feedback! The original comment was confusing because the condition uses `!indexKeyFilteredOut` (negated) while the explanation described the positive case ("when a result was filtered out"), forcing the reader to mentally invert it.

In f63d78adf5697ba5c18c9529d80728b512aac838, I've rewritten the comment to:
- Match the polarity of the condition by framing it as "when we cache `Optional.empty()`"
- Explicitly explain what each sub-condition (`get.getConjunctions().isEmpty()` and `!indexKeyFilteredOut`) means, and why `Optional.empty()` must not be cached if either does not hold

PTAL!

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -2411,6 +2411,170 @@ public void get_GetWithIndexForDeletedWhenCoordinatorStateAborted_ShouldRollBack
     verify(recovery).rollbackRecord(any(Selection.class), any(TransactionResult.class));
   }
 
+  @ParameterizedTest
+  @EnumSource(Isolation.class)
+  public void
+      get_GetWithIndexForPreparedWhenCoordinatorStateAbortedAndIndexKeyMatchesAfterImage_ShouldRollBackAndFilterOutResult(
```
komamitsu, line 2417: [minor] Don't we need to add a test for DELETED state to see if a rolled-back record isn't filtered out?

brfrn169: Thanks for checking! The scenario where a DELETED record is rolled back and should NOT be filtered out (because its restored value still matches the queried index key) is already covered by [`get_GetWithIndexForDeletedWhenCoordinatorStateAborted_ShouldRollBackAndReturnResult`](https://github.com/scalar-labs/scalardb/blob/f63d78adf5697ba5c18c9529d80728b512aac838/integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java#L2379-L2412). That test queries by INITIAL_BALANCE, which matches the DELETED record's restored value after rollback, so the new filter logic's `resultMatchesIndexKey()` returns true and the record is correctly not filtered out. What do you think?

komamitsu: That's true from the implementation perspective. But in terms of integration tests, I think it might be better to add a test for the DELETED case to explicitly verify the rolled-back record isn't filtered out by the new logic.

I don't insist on it, so it's up to you.

brfrn169: Just to make sure I understand — are you suggesting a new test method that's essentially identical to `get_GetWithIndexForDeletedWhenCoordinatorStateAborted_ShouldRollBackAndReturnResult` in content, but named to explicitly reflect the new filter logic (i.e., "not filtered out")? Or did you have a different scenario in mind?

komamitsu: Ah, sorry. I misread the prevous comment. I don't think we need to add a new test for DELETED.

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

KodaiD (approved):
LGTM. Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #3510, opened 2026-04-15
Support Google Spanner

Torch3333:
## Description

Adds support for **Google Cloud Spanner** as a JDBC storage in ScalarDB. The adapter targets Spanner's PostgreSQL-compatible dialect via the official `com.google.cloud.spanner:google-cloud-spanner-jdbc` driver.

A new configuration is introduced: 
- `scalar.db.jdbc.spanner.time_column.default_date_component`: Since TIME data is encoded into a Spanner `TIMESTAMP WITH TIME ZONE` data type. Time value is appended to a static date value. This config is only relevant for the import table use case. A similar config can be found with Oracle and Db2.

The adapter has the following limitations:
- FLOAT column cannot be used as partition key, clustering key or secondary index
- Renaming a table or a column is not supported
- Only altering the column type from BLOB to TEXT is supported
- `\` (backslash) is used as the default escape character for LIKE expression and cannot be configured or disabled

Regarding authentication, Spanner credentials (Google Cloud service account key in JSON format) are passed via the `scalardb.jdbc.password` ScalarDB property. The `scalardb.jdbc.username` property is never used. Authentication through the JDBC driver is done via the `credentialsProvider` JDBC property with requires setting a JVM wide system property `ENABLE_CREDENTIALS_PROVIDER`. 

## Related issues and/or PRs

N/A

## Changes made
- Add `RdbEngineSpanner` extending `RdbEnginePostgreSQL`
- Spanner is not added to the current CI because some tests take more than 2 hours to complete. For now, a workflow dispatch action for Spanner has only been created. This is only temporary.
- Update the integration test to handle the limitations explained above.

## Checklist

- The Spanner adapter has been validated against both the Spanner emulator (via the new GitHub Actions workflow) and a real Cloud Spanner instance

cf. the [Spanner CI run](https://github.com/scalar-labs/scalardb/actions/runs/25051357006) 

## Release notes

Added support for Google Cloud Spanner as a JDBC storage, via its PostgreSQL-compatible dialect.

### DIFF
--- a/.github/workflows/spanner-adapter-check.yaml
+++ b/.github/workflows/spanner-adapter-check.yaml
@@ -1,0 +1,336 @@
+# Do not include Spanner integration test in the CI because some tests require more than 2 hours
+# to complete.
+name: CI for Spanner Adapter
+run-name: "CI for Spanner Adapter : JDK ${{ inputs.INT_TEST_JAVA_RUNTIME_VERSION || '8' }} (${{ inputs.INT_TEST_JAVA_RUNTIME_VENDOR || 'temurin' }})"
+
+on:
+  workflow_dispatch:
+    inputs:
+      INT_TEST_JAVA_RUNTIME_VERSION:
+        description: JDK version used to run the integration test
+        type: choice
+        default: '8'
+        options:
+          - '8'
+          - '11'
+          - '17'
+          - '21'
+      INT_TEST_JAVA_RUNTIME_VENDOR:
+        description: Vendor of the JDK used to run the integration test
+        type: choice
+        default: 'temurin'
+        options:
+          - 'corretto'
+          - 'microsoft'
+          - 'oracle'
+          - 'temurin'
+
+env:
+  ARTIFACT_SUFFIX: ${{ inputs.INT_TEST_JAVA_RUNTIME_VERSION || '8' }}_${{ inputs.INT_TEST_JAVA_RUNTIME_VENDOR || 'temurin' }}
+  INT_TEST_JAVA_RUNTIME_VERSION: ${{ inputs.INT_TEST_JAVA_RUNTIME_VERSION || '8' }}
+  INT_TEST_JAVA_RUNTIME_VENDOR: ${{ inputs.INT_TEST_JAVA_RUNTIME_VENDOR || 'temurin' }}
+  SPANNER_JDBC_URL: "jdbc:cloudspanner://localhost:9010/projects/emulator-project/instances/test-instance/databases/test-db;autoConfigEmulator=true;dialect=POSTGRESQL"
+
+# Group commit only applies to **.ConsensusCommit** tests.
+# Groups without ConsensusCommit tests have group_commit_tests are skipped in the group_commit
+# flavor via the job-level `if` condition.
+
+jobs:
+  spanner-integration-test:
+    name: "${{ matrix.test_group.name }} (${{ matrix.group_commit_enabled == 'true' && 'group_commit' || 'default' }})"
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        group_commit_enabled:
+          - 'false'
+          - 'true'
+        test_group:
+          - name: ConsensusCommitSpecific
+            label: spanner_cc_specific
+            tests: "ConsensusCommitSpecificIntegrationTestWithJdbcDatabase"
+
+          - name: ConsensusCommitSpecific (highest isolation)
+            label: spanner_cc_specific_highest
+            tests: "ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation"
+
+          - name: ConsensusCommitSpecific (metadata decoupling)
+            label: spanner_cc_specific_md
+            tests: "ConsensusCommitSpecificWithMetadataDecouplingIntegrationTestWithJdbcDatabase"
+
+          - name: MultipleClusteringKeyScan
+            label: spanner_multi_ck_scan
+            tests: "JdbcDatabaseMultipleClusteringKeyScanIntegrationTestWithSpanner"
+
+          - name: ConditionalMutation
+            label: spanner_cond_mutation
+            tests: "JdbcDatabaseConditionalMutationIntegrationTestWithSpanner"
+
+          - name: "Admin & Import"
+            label: spanner_group1
+            tests: >-
+              ConsensusCommitAdminIntegrationTestWithJdbcDatabase
+              ConsensusCommitAdminRepairIntegrationTestWithJdbcDatabase
+              ConsensusCommitAdminImportTableIntegrationTestWithJdbcDatabase
+              ConsensusCommitAdminImportTableWithMetadataDecouplingIntegrationTestWithJdbcDatabase
+              ConsensusCommitImportTableIntegrationTestWithJdbcDatabase
+              ConsensusCommitImportTableWithMetadataDecouplingIntegrationTestWithJdbcDatabase
+              JdbcAdminImportTableIntegrationTest
+              JdbcAdminRepairIntegrationTest
+              JdbcSchemaLoaderImportIntegrationTest
+              JdbcSchemaLoaderImportWithMetadataDecouplingIntegrationTest
+              JdbcAdminIntegrationTest
+              JdbcTransactionAdminIntegrationTest
+
+          - name: "Core JDBC"
+            label: spanner_group2
+            tests: >-
+              JdbcDatabaseIntegrationTest
+              JdbcDatabaseColumnValueIntegrationTest
+              JdbcDatabaseCrossPartitionScanIntegrationTest
+              JdbcDatabaseSingleClusteringKeyScanIntegrationTest
+              JdbcDatabaseSinglePartitionKeyIntegrationTest
+              JdbcDatabaseMultiplePartitionKeyIntegrationTest
+              JdbcDatabaseSecondaryIndexIntegrationTest
+
+          - name: "Schema & i18n"
+            label: spanner_group3
+            tests: >-
+              JdbcDatabaseJapaneseIntegrationTest
+              JdbcDatabaseCaseSensitivityIntegrationTest
+              JdbcDatabaseWithReservedKeywordIntegrationTest
+              JdbcDatabaseVirtualTablesIntegrationTest
+              JdbcDatabaseMutationAtomicityUnitIntegrationTest
+              JdbcAdminCaseSensitivityIntegrationTest
+              JdbcSchemaLoaderIntegrationTest
+              JdbcSchemaLoaderWithMetadataDecouplingIntegrationTest
+
+          - name: "CC base & TwoPhase & SingleCrud & JDBC transaction"
+            label: spanner_group4
+            tests: >-
+              ConsensusCommitIntegrationTestWithJdbcDatabase
+              ConsensusCommitIntegrationTestWithJdbcDatabaseInHighestIsolation
+              ConsensusCommitCrossPartitionScanIntegrationTestWithJdbcDatabase
+              ConsensusCommitNullMetadataIntegrationTestWithJdbcDatabase
+              ConsensusCommitWithIncludeMetadataEnabledIntegrationTestWithJdbcDatabase
+              ConsensusCommitWithMetadataDecouplingIntegrationTestWithJdbcDatabase
+              TwoPhaseConsensusCommitIntegrationTestWithJdbcDatabase
+              TwoPhaseConsensusCommitSpecificIntegrationTestWithJdbcDatabase
+              TwoPhaseConsensusCommitCrossPartitionScanIntegrationTestWithJdbcDatabase
+              TwoPhaseConsensusCommitWithIncludeMetadataEnabledIntegrationTestWithJdbcDatabase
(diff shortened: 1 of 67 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSpanner.java
```diff
@@ -0,0 +1,304 @@
...
+      String indexName =
+          shortenIndexNameIfNeeded(
+              CLUSTERING_ORDER_INDEX_NAME_PREFIX + schema + "_" + table,
+              CLUSTERING_ORDER_INDEX_NAME_PREFIX);
+      sqls.add(dropIndexSql(schema, table, indexName));
+    }
+    sqls.add("DROP TABLE " + encloseFullTableName(schema, table));
+
+    return sqls.toArray(new String[0]);
+  }
```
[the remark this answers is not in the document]
Torch3333: Fixed in https://github.com/scalar-labs/scalardb/pull/3510/changes/21c0360a229a25f50c6fdb9527e8161172dcab10

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSpanner.java
```diff
@@ -0,0 +1,304 @@
...
+  @Override
+  public DateColumn parseDateColumn(ResultSet resultSet, String columnName) throws SQLException {
+    String dateStr = resultSet.getString(columnName);
+    if (dateStr == null) {
+      return DateColumn.ofNull(columnName);
+    } else {
+      LocalDate date = RdbEngineTimeTypeSpanner.DATE_FORMATTER.parse(dateStr, LocalDate::from);
+      return DateColumn.of(columnName, date);
+    }
+  }
```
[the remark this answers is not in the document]
Torch3333: Added a comment in https://github.com/scalar-labs/scalardb/pull/3510/commits/a1dc38734e7ca082b2014ab2d8c98a258a782eb5

core/src/main/java/com/scalar/db/storage/jdbc/SpannerCredentialsProvider.java
```diff
@@ -0,0 +1,60 @@
...
+        // Lost the race — fall through to the equality check against the winner
+        existing = CREDENTIALS.get();
+      } else {
+        return;
+      }
+    }
+    if (!existing.equals(credentials)) {
+      throw new IllegalStateException(
+          CoreError.JDBC_SPANNER_CREDENTIALS_ALREADY_REGISTERED.buildMessage());
+    }
```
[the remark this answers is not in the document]
Torch3333: It now compares the credentials using the JSON file set to the `scalardb.jdbc.password` value.
Also, up to 5 different credentials can be used.
[scalardb.jdbc.password](https://github.com/scalar-labs/scalardb/pull/3510/commits/dd3f9785d801090925479f13f362fc3afca3e60b)

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -315,8 +315,10 @@ public void dropTable(String namespace, String table) throws ExecutionException
 
   private void dropTableInternal(Connection connection, String schema, String table)
       throws SQLException {
-    String dropTableStatement = "DROP TABLE " + encloseFullTableName(schema, table);
-    execute(connection, dropTableStatement, requiresExplicitCommit);
+    TableMetadata tableMetadata = tableMetadataService.getTableMetadata(connection, schema, table);
+    assert tableMetadata != null;
+    String[] dropTableStatements = rdbEngine.dropTableSql(tableMetadata, schema, table);
+    execute(connection, dropTableStatements, requiresExplicitCommit);
```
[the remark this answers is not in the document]
Torch3333: Metadata existence is checked upstream through this method call chain, so using `assert` is enough.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSpanner.java
```diff
@@ -0,0 +1,317 @@
...
+
+  @Override
+  public boolean requiresExplicitDropIndexBeforeDropColumn() {
+    return true;
+  }
+
+  @Override
+  public Map<String, String> getConnectionProperties(JdbcConfig config) {
+    return ImmutableMap.of("dialect", "POSTGRESQL");
```
feeblefakie: Sorry, I'm asking without knowing Spanner very much.
But, what does this mean?

I understand there are mainly two ways to interact with Spanner: use [com.google.cloud.spanner.jdbc.JdbcDriver ](https://docs.cloud.google.com/spanner/docs/use-oss-jdbc)and use [org.postgresql.Driver](https://docs.cloud.google.com/spanner/docs/pg-jdbc-connect),
and I saw it uses `com.google.cloud.spanner.jdbc.JdbcDriver` above.

But, this sounds like it uses the latter.
Can you clarify?

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
```diff
@@ -118,6 +120,11 @@ public class JdbcConfig {
   // is 1970-01-01.
   public static final String DEFAULT_DB2_TIME_COLUMN_DEFAULT_DATE_COMPONENT = "1970-01-01";
 
+  // In Spanner PostgreSQL, there is no native TIME type, so TIME values are stored as TIMESTAMPTZ
```
feeblefakie, line 123: This is related to my other question.
Which one are we using?
- https://docs.cloud.google.com/spanner/docs/use-oss-jdbc
- https://docs.cloud.google.com/spanner/docs/pg-jdbc-connect

I expect the former, and it seems it does uses it, but `Spanner PostgreSQL` sounds like the latter.
Please clarify.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSpanner.java
Torch3333: The `com.google.cloud.spanner.jdbc.JdbcDriver` supports both Google SQL and PostgreSQL dialects; this is the driver we use. 

Also, setting `dialect=POSTGRESQL` in the connection string appears not to be necessary, so I will remove it. Thank you for pointing it out.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcConfig.java
Torch3333: We use the driver created by Google `google-cloud-spanner-jdbc` cf. this [page](https://docs.cloud.google.com/spanner/docs/pg-jdbc-connect). 
So, in the comment, `Spanner PostgreSQL` means `Spanner with PostgreSQL dialect.`

feeblefakie: @Torch3333 The page you specified is https://docs.cloud.google.com/spanner/docs/pg-jdbc-connect, which uses `org.postgresql.Driver` but that's wrong?

Torch3333: I think it's wrong.
There is a sample on the Google Spanner JDBC repository that uses `com.google.cloud.spanner.jdbc.JdbcDriver` as the driver when using the PostgreSQL dialect.
https://github.com/googleapis/google-cloud-java/blob/66ded327ed12fb64acbe076d957f882d35ba25e3/java-spanner-jdbc/samples/spring-data-mybatis/postgresql/src/main/resources/application-cs.properties#L20

Google's support AI bot also confirmed using `com.google.cloud.spanner.jdbc.JdbcDriver` is correct.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSpanner.java
Torch3333: Fixed in https://github.com/scalar-labs/scalardb/pull/3510/commits/bf28adf673dadb7e104bfebf24fe98e0441a7e5c

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -315,8 +315,10 @@ public void dropTable(String namespace, String table) throws ExecutionException
 
   private void dropTableInternal(Connection connection, String schema, String table)
       throws SQLException {
-    String dropTableStatement = "DROP TABLE " + encloseFullTableName(schema, table);
-    execute(connection, dropTableStatement, requiresExplicitCommit);
+    TableMetadata tableMetadata = tableMetadataService.getTableMetadata(connection, schema, table);
+    assert tableMetadata != null;
+    String[] dropTableStatements = rdbEngine.dropTableSql(tableMetadata, schema, table);
```
brfrn169: For consistency, I think we could add `requiresExplicitDropIndexBeforeDropTable()` to `RdbEngineStrategy`, similar to `requiresExplicitDropIndexBeforeDropColumn()`, and explicitly drop indexes when `requiresExplicitDropIndexBeforeDropTable()` returns `true`.

What do you think?

brfrn169 (commented):
Left a minor comment. PTAL!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
Torch3333: I considered adding `requiresExplicitDropIndexBeforeDropTable()` too, but because the clustering index order needs to be dropped and this index is not present on all storages, it would have required adding another RdbEngineStrategy method to generate the drop index syntax by storage.

https://github.com/scalar-labs/scalardb/blob/1a9b6c4f8ae59e13c2354a094bcc3eb9ba9cbc98/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineSpanner.java#L295-L301

That's why I chose this design for simplicity. 
Though I understand your point of wanting to separate the drop indexes from the drop table call. We could add a `dropTableInternalSqlsBeforeDropTable` similarly to the `createTableInternalSqlsAfterCreateTable` method. 
What do you think?

komamitsu (approved):
LGTM! 👍 

I've not thoroughly looked into the test code, though.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
brfrn169: Sounds good. Let’s proceed with that. Thanks.

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
Torch3333: Revised in https://github.com/scalar-labs/scalardb/pull/3510/commits/553e944fbe5589e6255cdb8ee301f66d2e99a82f
PTAL.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -315,8 +315,14 @@ public void dropTable(String namespace, String table) throws ExecutionException
...
   private void dropTableInternal(Connection connection, String schema, String table)
       throws SQLException {
-    String dropTableStatement = "DROP TABLE " + encloseFullTableName(schema, table);
-    execute(connection, dropTableStatement, requiresExplicitCommit);
+    TableMetadata tableMetadata = tableMetadataService.getTableMetadata(connection, schema, table);
+    assert tableMetadata != null;
+    execute(
+        connection,
+        rdbEngine.dropTableInternalSqlsBeforeDropTable(schema, table, tableMetadata),
+        requiresExplicitCommit);
```
brfrn169, line 323: Ah, one more thing I noticed: when dropping indexes, we need to handle the legacy case where the original long index name was used before the shortening logic was introduced. For that case, we need fallback logic like this:

https://github.com/scalar-labs/scalardb/blob/826c6d6b2638b27808e5da1b92bdba8b11e3dba9/core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java#L1134-L1154

If we don’t have time to address this now, we can leave a `FIXME` comment for the time being and fix it later.

Torch3333: To confirm: Spanner is added in the same release as the new index naming policy, so we don't need to handle legacy naming, correct?

.github/workflows/spanner-adapter-check.yaml
```diff
@@ -0,0 +1,336 @@
+# Do not include Spanner integration test in the CI because some tests require more than 2 hours
+# to complete.
+name: CI for Spanner Adapter
```
Torch3333, line 3: This is a comment to block the PR from being merged. 
(I need to run the Spanner CI)

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
brfrn169: Thanks for pointing this out! Since this PR is the first to add Spanner support, there are no pre-existing indexes created before the shortening logic was introduced—so the legacy case doesn’t apply to Spanner. The current logic should be fine.

brfrn169 (approved):
LGTM! Thank you!

PR #3567, opened 2026-05-18
Add DistributedTransactionManager#finishTransaction

brfrn169:
## Description

`DistributedTransactionManager` currently relies on lazy recovery to clean up Coordinator state rows and to roll forward / roll back per-record `PREPARED` images that survived the commit / abort path. The cleanup is correct but eventual — operators have no way to trigger it eagerly, and `tx_write_set` rows accumulate until lazy recovery touches them.

This PR adds `finishTransaction(String txId)` as a low-level operational API that performs the cleanup synchronously: it runs per-record recovery using the persisted write set on the Coordinator state row, then deletes the state row. The API is idempotent on already-finished transactions and safe under races with concurrent recovery.

## Related issues and/or PRs

- Stacked on top of #3564 (populates the `tx_write_set` column on the Coordinator table — required for this feature to have a write set to read)

### API

- Add `boolean finishTransaction(String txId) throws TransactionException` to `DistributedTransactionManager`. Javadoc covers applicability (only `DistributedTransaction#commit()`-terminated transactions), idempotency, group-commit semantics, and return-value semantics: returns `true` when the transaction was finished (or was already finished), and `false` when the transaction is not applicable because it carries no write set (terminated via `rollback()` / `abort()`, aborted by lazy recovery, or originated from a pre-feature binary). The not-applicable case is an expected outcome, not an error — retrying with the same transaction ID would never succeed — so it is reported via the return value rather than an exception. A `TransactionException` is thrown when finishing fails.
- Add `@throws UnsupportedOperationException` to the existing `getState` / `rollback` / `abort` Javadocs so the four operational APIs document consistent contracts.
- Add `CoreError` entries: `0068` (finishing the transaction failed), `0289` / `0290` (unsupported on `JdbcTransactionManager` / `SingleCrudOperationTransactionManager`).

### Implementation

- `ConsensusCommitManager#finishTransaction`: read state row → return `true` on absent row (idempotent no-op) → `AssertionError` on non-terminal state → return `false` on no-write-set (not applicable) → walk every `EntryGroup` / `Entry` and run per-record recovery → delete state row → return `true`. Records already finalized or overwritten by another transaction are filtered out via `shouldRecover` to avoid wasted conditional mutations.
- `Coordinator#deleteState`: unconditional, retry-free delete. Storage errors surface as `CoordinatorException`.
- `RecoveryExecutor#executeSynchronously`: new package-private delegate that exposes the per-record recovery primitive without leaking `RecoveryHandler` itself. Synchronous counterpart of the existing async `execute`.
- `WriteSetDecoder`: new package-private class — inverse of `WriteSetEncoder`. `toGet(Entry)` is the only production-facing entry point; rebuilds a `Get` (namespace / table / partition key / clustering key / LINEARIZABLE consistency) from a persisted proto `Entry`.
- `DecoratedDistributedTransactionManager` / `TransactionService`: one-line delegate to the wrapped manager.
- `JdbcTransactionManager` / `SingleCrudOperationTransactionManager`: throw `UnsupportedOperationException` — these backends have no Coordinator state row to clean up.
- `core/build.gradle`: exclude protobuf-generated sources (`build/generated/sources/proto/**`) from Error Prone. The generator emits patterns (reference-equality checks against `DEFAULT_INSTANCE`, unused builder parameters, deprecated API references) we can't fix locally and that drown the real signal at compile time.

### Tests

- `ConsensusCommitManagerTest`: new `finishTransaction*` unit tests covering the full flow (1PC commit / abort happy paths, 3-child group commit, missing state row returning `true`, no-write-set returning `false`, non-terminal state assertion, `getState` failure, transient storage failure, missing record, read-only commit, record already finalized, record overwritten by an unrelated transaction, deleteState failure, `DELETED`-state record recovery for both COMMITTED and ABORTED coordinators).
- `WriteSetDecoderTest`: 16 round-trip tests covering 11 column types, composite partition + clustering keys, `Get` construction, and `AssertionError` invariants on null markers / `VALUE_NOT_SET`.
- `RecoveryExecutorTest`: new delegate test for `executeSynchronously`.
- `JdbcTransactionManagerTest` / `SingleCrudOperationTransactionManagerTest`: verify `UnsupportedOperationException` is thrown with the new `CoreError` code prefixes.
- `CoordinatorTest`: new tests for `Coordinator.deleteState`. Existing tests refactored to use `assertThat(state).isPresent()` before `.get()` chaining.
- Integration tests in `ConsensusCommitSpecificIntegrationTestBase` cover recovery scenarios on real storage:
  1. Partial-COMMIT rollforward — Coordinator says `COMMITTED`, one record is still `PREPARED`; `finishTransaction` rolls it forward, deletes the state row, and returns `true`.
  2. Partial-PREPARE `ABORTED` rollback — `Coordinator` says `ABORTED` with a write set listing two records, one was prepared and one was never written; `finishTransaction` rolls the prepared record back to its before-image, deletes the state row, and returns `true`.
  3. Group-commit rollforward — group-committed siblings with one `PREPARED` record; `finishTransaction` on either child id recovers the remaining record, deletes the shared parent state row, and returns `true`.
  4. No-write-set not-applicable — a transaction terminated via `rollback()` leaves an `ABORTED` state row with no write set; `finishTransaction` returns `false` and leaves the state row in place for lazy recovery.

## Checklist

This is a low-level operational API. The Javadoc explicitly notes that most applications should not call it directly — it is intended for advanced use cases such as operational tooling, where the caller understands the underlying transaction lifecycle.

For group-commit transactions, `state.getId()` returns the parent id, so `Coordinator#deleteState` correctly targets the shared parent state row. Sibling `finishTransaction` calls observe an absent state row and return `true` via the idempotency contract.

`finishTransaction` is added as a non-default abstract method on the `DistributedTransactionManager` interface. All in-tree implementations are updated; downstream code that implements the interface directly will need to add an explicit override (or extend `AbstractDistributedTransactionManager`, which inherits the unsupported behavior via the existing decorator chain).

## Release notes

Added the `DistributedTransactionManager#finishTransaction(String)` API, which completes per-record post-commit recovery for a committed transaction and removes its Coordinator state row. It returns `true` when the transaction was finished (or was already finished) and `false` when the transaction is not applicable because it carries no write set (for example, a transaction terminated via `rollback()` / `abort()` rather than `commit()`); a `TransactionException` is thrown when finishing fails. Note that this is a low-level operational API specific to the Consensus Commit transaction manager. Most applications should not call it directly; it is intended for advanced use cases. Callers are expected to understand the underlying transaction lifecycle and the implications of invoking this method directly.


core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
brfrn169, line 1: This PR should be merged after https://github.com/scalar-labs/scalardb/pull/3564 is merged.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -561,6 +569,134 @@ public TransactionState rollback(String txId) {
...
+    // Only COMMITTED or ABORTED states are expected here. Anything else indicates an internal
+    // invariant violation, not a runtime condition.
+    TransactionState txState = state.getState();
+    if (txState != TransactionState.COMMITTED && txState != TransactionState.ABORTED) {
+      throw new AssertionError(
+          "Unexpected non-terminal transaction state at finishTransaction. State: "
+              + txState
+              + ", Transaction ID: "
+              + txId);
+    }
```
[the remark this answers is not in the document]
brfrn169: @gemini-code-assist This `AssertionError` is intentional. Our convention is to use `AssertionError` for internal invariant violations (i.e., bugs in the state machine) and `TransactionException` (or other checked/runtime exceptions) for failures caused by external factors (storage I/O, races, etc.). A non-terminal state on a persisted Coordinator state row would mean `commitState` / `abortState` failed to enforce the terminal-state invariant — that is an internal bug, not a recoverable runtime condition.

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
```diff
@@ -431,32 +431,88 @@ default DistributedTransaction join(String txId) throws TransactionNotFoundExcep
...
+   * underlying transaction lifecycle and the implications of invoking this method directly.
+   *
+   * <p><b>Applicability:</b> only transactions committed via {@link
+   * DistributedTransaction#commit()} are eligible — they are the ones that persist a write set
+   * alongside the Coordinator state row. Transactions that did not commit through {@link
+   * DistributedTransaction#commit()} (for example, transactions terminated via {@link
+   * #rollback(String)}, transactions aborted by lazy recovery, or transactions originated from
+   * older binaries that pre-date the write-set column) do not carry a write set, and calling this
+   * method on their transaction ID results in a {@link TransactionException} indicating that no
+   * write set is recorded. Those state rows are left for lazy recovery to handle.
```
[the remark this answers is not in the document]
brfrn169: Good catch — updated the Javadoc to describe both COMMITTED and ABORTED terminal states (rollforward for committed, rollback for aborted).

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -561,6 +569,134 @@ public TransactionState rollback(String txId) {
...
+          recoveryExecutor.executeSynchronously(get, txResult, Optional.of(state));
+        } catch (ExecutionException | CoordinatorException e) {
+          throw new TransactionException(
+              CoreError.CONSENSUS_COMMIT_FINISHING_TRANSACTION_FAILED.buildMessage(
+                  txId, e.getMessage()),
+              e,
+              txId);
+        }
+      }
+    }
```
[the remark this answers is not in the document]
brfrn169: @gemini-code-assist Thanks for the suggestion. We are not addressing this in the current PR — the current happy-path use case (calling finishTransaction eagerly right after a transaction terminates) typically has small write sets, and the synchronous loop keeps the implementation simple and easy to reason about for idempotency / retry. We may revisit parallelization in the future if a use case with large write sets emerges.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -315,6 +315,36 @@ private TransactionTableMetadata getTransactionTableMetadata(
...
+   * @param selection the selection that identifies the user record
+   * @param result the latest known TransactionResult for the record
+   * @param state the coordinator state for the transaction that wrote the record
+   * @throws ExecutionException if the underlying storage read or mutation fails
+   */
+  void executeSynchronously(
+      Selection selection, TransactionResult result, Optional<Coordinator.State> state)
+      throws ExecutionException {
+    try {
+      recovery.recover(selection, result, state);
```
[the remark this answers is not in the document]
brfrn169: @gemini-code-assist Good call. Tightened `RecoveryExecutor#executeSynchronously` to take `Coordinator.State` directly so the invariant is enforced at the API level, and updated the caller in `ConsensusCommitManager#finishTransaction` to pass `state` without wrapping. The `CoordinatorException` → `AssertionError` block stays as a defensive guard against future changes to `RecoveryHandler#recover`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -561,6 +569,135 @@ public TransactionState rollback(String txId) {
...
+          // been written). Nothing to recover.
+          continue;
+        }
+        TransactionResult txResult = new TransactionResult(resultOpt.get());
+        if (!shouldRecover(txResult, expectedTxId)) {
+          // Already finalized, or overwritten by an unrelated transaction. Recovery would no-op.
+          continue;
+        }
+        try {
+          recoveryExecutor.executeSynchronously(get, txResult, Optional.of(state));
```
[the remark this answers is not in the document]
brfrn169: @gemini-code-assist Done — call site now passes `state` directly, in tandem with the `executeSynchronously` signature change.

komamitsu:
@brfrn169 `finishTransaction()` could sound to terminate a transaction like as `abort()` or `commit()`. How about `finalizeTransaction()`, `cleanupTransaction()`, or something? `finalize` is actually used in the comment, so it might be consistent.

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
```diff
@@ -431,32 +431,93 @@ default DistributedTransaction join(String txId) throws TransactionNotFoundExcep
...
     return rollback(txId);
   }
 
+  /**
+   * Finishes a given terminated transaction by performing any remaining post-termination work and
+   * cleaning up the Coordinator state row. The transaction must already be in a terminal state
+   * ({@code COMMITTED} or {@code ABORTED}); this method completes the per-record work that was
+   * otherwise deferred to lazy recovery — rolling forward {@code PREPARED} records of a committed
+   * transaction, or rolling back {@code PREPARED} records of an aborted one — and then removes the
+   * Coordinator state row.
```
KodaiD: [minor] Should we also cover the case of deleting DELETED records here?

KodaiD (approved):
LGTM. Thank you!

brfrn169:
@komamitsu Thanks! We actually went back and forth between `finishTransaction` and `finalizeTransaction`, and chose `finishTransaction` because it was the more common term in prior art (similar APIs in other products).

Regarding the "could sound like terminating a transaction" concern — I think `finalizeTransaction` has the same issue: "finalize" is just as much a completion verb as "finish" (arguably even closer to commit in connotation), so renaming to finalize wouldn't really resolve that point.

Given that, I'd lean toward keeping `finishTransaction`. That said, if you feel the termination connotation is a real concern, `cleanupTransaction` would avoid it — we just hadn't considered it during the prior-art comparison. Happy to discuss if you'd prefer that direction.

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
brfrn169: Thanks! You are right. Fixed in 066dd8044d917e76910884a5995dcb5487fc9aca.

komamitsu:
> Given that, I'd lean toward keeping `finishTransaction`. That said, if you feel the termination connotation is a real concern, `cleanupTransaction` would avoid it — we just hadn't considered it during the prior-art comparison. Happy to discuss if you'd prefer that direction.

I see. Although I still prefer `cleanupTransaction`, I don't insist on it. I'm okay with the current name.

komamitsu (approved):
LGTM! 👍

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
```diff
@@ -431,32 +431,93 @@ default DistributedTransaction join(String txId) throws TransactionNotFoundExcep
...
     return rollback(txId);
   }
 
+  /**
+   * Finishes a given terminated transaction by performing any remaining post-termination work and
+   * cleaning up the Coordinator state row. The transaction must already be in a terminal state
+   * ({@code COMMITTED} or {@code ABORTED}); this method completes the per-record work that was
+   * otherwise deferred to lazy recovery — rolling forward {@code PREPARED} or {@code DELETED}
+   * records of a committed transaction, or rolling back {@code PREPARED} or {@code DELETED} records
+   * of an aborted one — and then removes the Coordinator state row.
```
jnmt, line 487: This `finishTransaction` Javadoc describes Consensus Commit implementation details, i.e., `Coordinator` state row, `write set`, `PREPARED`/`DELETED` record rollforward/rollback, etc. On the other hand, `ConsensusCommitManager#finishTransaction` currently has no Javadoc at all. So, should we move these kinds of details to the concrete class and keep this implementation-agnostic?

KodaiD:
@brfrn169 If we rename primary key column names, write set entries in the coordinator table will be stale. In such a case, `finishTransaction`s always fail. Is my understanding correct?

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
```diff
@@ -431,32 +431,97 @@ default DistributedTransaction join(String txId) throws TransactionNotFoundExcep
...
+   *
+   * <p><b>Applicability:</b> only transactions terminated via {@link
+   * DistributedTransaction#commit()} are eligible — they are the ones that persist a write set
+   * alongside the Coordinator state row, regardless of whether the commit succeeded ({@code
+   * COMMITTED}) or failed via a conflict during preparation ({@code ABORTED}). Transactions that
+   * did not go through {@link DistributedTransaction#commit()} (for example, transactions
+   * terminated via {@link #rollback(String)}, transactions aborted by lazy recovery, or
+   * transactions originated from older binaries that pre-date the write-set column) do not carry a
+   * write set, and calling this method on their transaction ID results in a {@link
+   * TransactionException} indicating that no write set is recorded. Those state rows are left for
```
jnmt: An application, i.e., ScalarDL, might want to distinguish between `TransactionException` due to no write set and the one due to other retryable reasons (e.g., temporal database failure). We can skip both cases and try to finish them again next time, but it will keep failing in the former case. Or, we can just ignore the `finishTransaction` result in both cases and finish the purge process on the ScalarDL side, but in those cases, we would like ScalarDB to purge state records eventually (regardless of the transaction state). Let's discuss details offline later.

jnmt (commented):
Overall looks good to me, thank you! Left a few comments. PTAL 🙇‍♂️

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
brfrn169: Thanks — this is a fair point, but it's actually intentional. This whole family of operational APIs (`getState`, `rollback`/`abort`, and now `finishTransaction`) is already declared as "a low-level operational API specific to the Consensus Commit transaction manager" in the interface Javadoc, and they all throw `UnsupportedOperationException` on the other managers (`JdbcTransactionManager`, `SingleCrudOperationTransactionManager`). So unlike the core CRUD APIs, we deliberately don't attempt to keep these implementation-agnostic — they only have meaningful semantics under Consensus Commit.

Given that, I think we can keep the lifecycle details (Coordinator state row, write set, PREPARED/DELETED rollforward/rollback) on the interface method: callers depend on the interface, and these details are the operational contract they need to understand before calling.

What do you think?

jnmt: Ahh, I missed this change: [07c885b](https://github.com/scalar-labs/scalardb/pull/3567/commits/07c885b1466e42cc714aaa8abc0f8dd776ad27b5). I understood that keeping it here helps users.

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
brfrn169: Thanks for raising this. I see two ways to let callers distinguish the "no write set" case from a retryable failure:
  
1. Exception types — keep void and throw a dedicated exception subtype for the no-write-set case, while retryable failures (e.g. a transient database failure) throw the base `TransactionException`. Callers branch on the type.
2. Boolean return value — return a boolean from `finishTransaction`: `true` when the cleanup completed (or the state row was already gone), `false` when the transaction carries no write set and is therefore not applicable (retrying with the same `txId` would never succeed). Only retryable failures throw `TransactionException`.

I'm currently leaning toward option 2 (the boolean return value). The no-write-set case isn't really an error — callers (e.g. ScalarDL) will routinely hit transactions without a write set (rolled back, lazy-recovered, or pre-feature binaries), so it feels more like a normal, expected outcome than an exceptional one. With the boolean, the contract stays simple: if it throws, retry later; otherwise this `txId` is done and the caller can move on with its purge.

What do you think?

jnmt: As discussed, I also prefer option 2. Thank you!

brfrn169: Thank you for the confirmation! I’ve updated the code in e3c2169d0d6e739f258f75556692d1ba862c44bb. PTAL!

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
```diff
@@ -431,32 +431,101 @@ default DistributedTransaction join(String txId) throws TransactionNotFoundExcep
...
+   * alongside the Coordinator state row, regardless of whether the commit succeeded ({@code
+   * COMMITTED}) or failed via a conflict during preparation ({@code ABORTED}). For an eligible
+   * transaction, this method completes the cleanup and returns {@code true}. Transactions that did
+   * not go through {@link DistributedTransaction#commit()} (for example, transactions terminated
+   * via {@link #rollback(String)} or {@link #abort(String)}, transactions aborted by lazy recovery,
+   * or transactions originated from older binaries that pre-date the write-set column) do not carry
+   * a write set; they are not applicable to this method, and calling it on their transaction ID
+   * returns {@code false} without doing any work. This is an expected outcome rather than an error
+   * — retrying with the same transaction ID would never succeed — and those state rows are left for
+   * lazy recovery to handle.
```
jnmt: Minor nitpick. Do these two mentions of "lazy recovery" refer to the same thing?

If the latter one is actually referring to the periodic garbage collection that we'll likely add in the future, then reusing the term "lazy recovery" here might be a bit confusing.

Conversely, if it does mean the same lazy recovery, then I find the wording slightly off: a write-set-less ABORTED state row isn't really kept for lazy recovery. I think it's retained as a safeguard so that a still-active transaction (if it exists) can't be mistakenly committed; deleting it wouldn't change the behavior of lazy recovery itself.

jnmt (approved):
LGTM, thank you! I left a minor comment for the Javadoc description for clarification.

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
brfrn169: Thanks for catching this. Explaining exactly how those state rows are subsequently handled (lazy recovery touches the per-record work rather than the state row itself, an aborted row stays as a safeguard against a mistaken commit, etc.) would be too much information here, so I just removed the trailing clause in f62e2738a4472005d1123e2c8a668744bf9659e9.

PR #3607, opened 2026-06-06
Fix ConsensusCommit secondary-index Get failing on transient duplicate rows during delete + insert

brfrn169:
## Description

Under Consensus Commit, a secondary-index `Get` could fail with `IllegalArgumentException` (`DB-CORE-10068: Please use scan() for non-exact match selection`) when the queried index value was concurrently affected by a delete + insert.

When a transaction deletes one record and inserts another that **has a different primary key but the same secondary-index value**, two physical rows momentarily co-exist in storage during/after the writer's commit: a `DELETED` row (the deleted record) and a `PREPARED`/committed row (the inserted record). The current-image index read used a single-row storage `Get`, which rejects multiple matching rows **before** Consensus Commit's lazy recovery can resolve the uncommitted records. As a result, a separate transaction's index `Get` could intermittently throw, even though only one logical record should be visible. (This is not specific to records sharing a partition key — any two distinct primary keys with the same indexed value can trigger it.)

**Example of the problem**

```java
// Table: tbl(account_id [PK], account_type [CK], balance), with a secondary index on `balance`.

// Tx1 moves a record to a new key while keeping the same indexed value (balance=2000):
tx1.delete(Delete.partitionKey(account_id=0).clusteringKey(account_type=0));         // old row
tx1.put(Put.partitionKey(account_id=0).clusteringKey(account_type=1).balance(2000)); // new row
tx1.commit();
//   During/after commit, two physical rows transiently share balance=2000:
//     (0,0): DELETED   (uncommitted, awaiting cleanup)
//     (0,1): PREPARED  -> committed

// Tx2 reads by the secondary index around the same time:
Optional<Result> r = tx2.get(Get.indexKey(balance=2000));
//   BEFORE this fix: throws IllegalArgumentException
//       "DB-CORE-10068: Please use scan() for non-exact match selection"
//     because the index matched two physical rows before lazy recovery ran.
//   AFTER this fix: lazy recovery resolves the transient duplicate and the Get
//     returns the single correct record.
```

## Related issues and/or PRs

N/A

## Changes made

- **`CrudHandler#read`**: for a `Get` with a secondary index, read via a `Scan` with index instead of a single-row `Get`. Each matched row is run through lazy recovery, and the result is asserted to be at most one surviving record. If more than one *genuinely committed* record still matches, the same `GET_OPERATION_USED_FOR_NON_EXACT_MATCH_SELECTION` error is raised (now from `CrudHandler`), so the exact-match contract for `Get` is preserved and users are still told to use a `Scan`.
- Extracted a shared `recoverAndFilter` helper (per-row lazy recovery + post-rollback index-key filter + conjunction filter, no caching) used by `read`, the new `resolveIndexGet`, and `processScanResult`.
- Moved recovery-future registration into `executeRecovery` (it now executes recovery and registers the future).
- Extracted the `Get`-with-index -> `Scan`-with-index conversion into `ConsensusCommitUtils#createScanWithIndexFromGet`, shared by `CrudHandler` and `Snapshot#validateGetWithIndexResult` (which previously inlined it).
- The internal scan does **not** run `verifyNoOverlap`, so a same-transaction write followed by an index `Get` behaves exactly as a non-index read; snapshot get-set/read-set keying and the before-image index recovery retry loop are unchanged. The fix lives in Consensus Commit, so it covers all storage backends.
- Added unit tests (`CrudHandlerTest`) and parameterized integration tests (`ConsensusCommitSpecificIntegrationTestBase`), including a case verified to fail on the pre-fix code with `DB-CORE-10068`, and a case showing the record is still returned in all isolation levels (including `READ_COMMITTED`).

## Checklist

- Behavior by isolation is unchanged from normal Consensus Commit semantics: `SNAPSHOT`/`SERIALIZABLE` roll forward and return the latest record; `READ_COMMITTED` returns the committed (before-image) record. In no isolation does the transient duplicate cause `DB-CORE-10068` — that error now fires only when two genuinely committed records share the index value (a real "use `Scan`" case).
- The integration test that reproduces the transient duplicate was confirmed to fail on the pre-fix code with `DB-CORE-10068`.

## Release notes

Fixed an issue in Consensus Commit where a `Get` operation using a secondary index could fail with an `IllegalArgumentException` when another record with the same indexed value was concurrently being deleted and inserted.


Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -752,9 +876,11 @@ boolean checkAndRecoverBeforeIndexRecords(
         TransactionResult result = new TransactionResult(r);
         if (!result.isCommitted()) {
           Snapshot.Key key = new Snapshot.Key(beforeIndexScan, r, metadata.getTableMetadata());
-          // Always use RETURN_LATEST_RESULT_AND_RECOVER regardless of isolation level because
-          // recovery must actually execute; otherwise, the PREPARED/DELETED record remains in
-          // storage and retrying would be useless
+          // This intentionally bypasses executeRecovery(): it always uses
+          // RETURN_LATEST_RESULT_AND_RECOVER regardless of isolation level (recovery must actually
```
komamitsu, line 880: I don't think this PR changes the behavior, so probably this is a bit off-topic. But I'm wondering if we don't need to consider `context.readOnly` here. My concern is if the behavior might conflict with [the following document](https://github.com/scalar-labs/docs-internal-scalardb/blob/main/docs/en-us/scalardb-cluster/remote-replication.mdx?plain=1#L149-L153):

> Only [transactions in read-only mode](../api-guide.mdx#begin-or-start-a-transaction-in-read-only-mode) with the [read-committed isolation level](../consensus-commit.mdx#isolation-levels) are permitted on backup sites until failover.

What do you think?

brfrn169: Good catch. To summarize the current state and the root cause:

The new index-Get resolution added in this PR goes through `executeRecovery`, which does honor read-only mode (READ_COMMITTED + read-only → RETURN_COMMITTED_RESULT_AND_NOT_RECOVER, no write). So the code this PR adds is fine.

The path you pointed at, `checkAndRecoverBeforeIndexRecords`, is the pre-existing before-image index retry loop, and it must force RETURN_LATEST_RESULT_AND_RECOVER: the retry only converges if the rollback is physically applied to storage. Switching it to a non-recovering type in read-only mode would make the loop never converge (it would hit the retry limit and throw), so we can't simply make this path read-only-aware.

The deeper issue is that the DELETED state is overloaded. Its original meaning is "a delete that is prepared but not yet confirmed" (an in-flight tombstone, for which `isCommitted()` is false and recovery is therefore required). Remote replication, however, uses DELETED to mean a confirmed/committed deletion. Because the two meanings share one state, we can't distinguish them and ends up triggering recovery writes on what is logically a settled record.

The proper fix is to introduce a distinct state for a confirmed deletion (e.g. COMMITTED_DELETED) so that recovery is not triggered on it. We should file a follow-up task for it.

komamitsu (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -232,6 +222,159 @@ Optional<TransactionResult> read(
...
+          survivor = recovered.result;
+          survivorKey = key;
+        }
+      }
+    } catch (IllegalArgumentException e) {
+      // The >1-survivor exact-match contract violation thrown above is a RuntimeException too;
+      // rethrow it explicitly so it is not handled by the ExecutionException-unwrapping branch
+      // below.
+      throw e;
+    } catch (RuntimeException e) {
```
feeblefakie, line 281: The scan-failure handling here (RuntimeException -> ExecutionException unwrap, catch ordering, and the IOException-on-close swallow) isn't covered by any unit test. Could you add one that makes the scanner throw a RuntimeException(ExecutionException) and asserts the resulting CrudException?

feeblefakie (approved):
LGTM! Thank you!
Left one comment, so PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Added in cb30e9edd4547145a19f49d1500875b05151c187. Thanks!

PR #3611, opened 2026-06-08
Add RdbEngineStrategyExceptionIntegrationTest

Torch3333:
## Description

The `RdbEngineStrategy` implementations expose several methods that interpret backend-specific `SQLException`s (e.g. `isDuplicateTableError`, `isDuplicateKeyError`, `isDuplicateSchemaError`, `isDuplicateIndexError`, `isUndefinedIndexError`, and the Db2-specific `throwIfDuplicatedIndexWarning`). These methods rely on vendor-specific error codes / SQL states and warnings, which are easy to get wrong and are not covered by the existing unit tests against mocked exceptions. This PR adds an integration test that provokes the real error conditions against each JDBC backend and asserts that the corresponding detection method behaves correctly.

The `RdbEngineStrategy.isConflict` is true because it is difficult to reproduce the conflict error across all storage.

## Related issues and/or PRs

> N/A

## Changes made

- Added `RdbEngineStrategyExceptionIntegrationTest` (`core/src/integration-test/java/com/scalar/db/storage/jdbc/`), which, against the configured JDBC backend:
  - Creates a duplicate table and verifies `isDuplicateTableError`.
  - Inserts a duplicate primary key and verifies `isDuplicateKeyError`.
  - Creates a duplicate schema and verifies `isDuplicateSchemaError`.
  - Creates a duplicate index and verifies `isDuplicateIndexError`.
  - Drops a non-existent index and verifies `isUndefinedIndexError`.
  - Verifies the Db2 warning-based duplicate-index path via `throwIfDuplicatedIndexWarning`.
- Engine-specific behavior is gated with `@DisabledIf`/`@EnabledIf` predicates so each assertion runs only on backends where it is meaningful (e.g. tests that rely on duplicate-error detection are skipped for engines that use `CREATE ... IF NOT EXISTS` syntax, including MySQL, PostgreSQL, and Db2).

## Checklist

This is a test-only change; no production code is modified.

## Release notes

N/A


komamitsu (approved):
LGTM, thank you!

KodaiD (approved):
LGTM. Thank you!

core/src/integration-test/java/com/scalar/db/storage/jdbc/RdbEngineStrategyExceptionIntegrationTest.java
```diff
@@ -0,0 +1,339 @@
...
+  public void tearDownAll() {
+    if (dataSource != null) {
+      executeIgnoringError(rdbEngine.dropNamespaceSql(TEST_SCHEMA));
+      dataSource.close();
+    }
+  }
+
+  @Test
+  @DisabledIf("isCreateTableIfNotExistsSyntaxSupported")
+  public void isDuplicateTableError_WhenDuplicateTable_ShouldReturnTrue() throws SQLException {
```
brfrn169, line 63: It seems like there is inconsistency in `isDuplicateTableError` across engines

The "table already exists" case is suppressed in `JdbcAdmin#createTable` through one of two mechanisms:
1. adding `IF NOT EXISTS` via `tryAddIfNotExistsToCreateTableSql`
2. the catch block swallowing the error when `isDuplicateTableError(e)` returns true.

Today the engines are split:
- Oracle / SQL Server — no `IF NOT EXISTS` support, so `2` is load-bearing → `isDuplicateTableError` returns the real code.
- MySQL / PostgreSQL / SQLite — add `IF NOT EXISTS` (`1`) and implement `2` with the real code.
- Db2 / Spanner — add `IF NOT EXISTS` (`1`) but `isDuplicateTableError` returns false unconditionally.

I'd suggest making this consistent: either drop `2` from MySQL/PostgreSQL/SQLite, or implement `2` properly for Db2/Spanner.

My concern with dropping `2`: `CREATE TABLE IF NOT EXISTS` is not necessarily atomic under concurrent DDL (PostgreSQL, for instance, can still raise `42P07` when two sessions create the same table concurrently). Since ScalarDB instances may call `createTable(ifNotExists=true)` concurrently on startup, `1` acts as a real race safety net. If that's right, the safer direction is to implement `2` for Db2/Spanner rather than remove it from the others — and it also means this test could then cover Db2/Spanner.

Could you confirm whether the false for Db2/Spanner is an intentional "we always rely on IF NOT EXISTS" decision, and whether CREATE TABLE IF NOT EXISTS is race-safe on Db2/Spanner? If it isn't, a concurrent createTable there would re-throw the duplicate error today.

(Out of scope for this test PR — fine as a follow-up. Flagging since the test gating is shaped by exactly this asymmetry.)

core/src/integration-test/java/com/scalar/db/storage/jdbc/RdbEngineStrategyExceptionIntegrationTest.java
```diff
@@ -0,0 +1,339 @@
...
+                      thrown.set(e);
+                    }
+                    warning = warning.getNextWarning();
+                  }
+                }
+              });
+
+      assertThat((Object) thrown.get())
+          .as("Expected throwIfDuplicatedIndexWarning to throw for DB2 warning code 605")
+          .isNotNull();
```
brfrn169: The current `assertThat(thrown.get()).isNotNull()` passes for any `SQLException`. Maybe we should assert that the error is the DB2 605 warning, like this:
```java
assertThat(thrown.get().getErrorCode()).isEqualTo(605);
```

brfrn169 (commented):
Left a few comments. PTAL!

feeblefakie (approved):
LGTM! Thank you!

core/src/integration-test/java/com/scalar/db/storage/jdbc/RdbEngineStrategyExceptionIntegrationTest.java
Torch3333: Thanks for spotting the inconsistency. 
I would be leaning more towards just having these methods return false when the `IF NOT EXISTS` syntax is supported. It is easier to maintain, we don't need to test these error codes are correct, and also I don't think we should cater for the corner case of having the user create the same table at the same time, since the Admin API is not made to support concurrent usage in the first place.
What do you think?

core/src/integration-test/java/com/scalar/db/storage/jdbc/RdbEngineStrategyExceptionIntegrationTest.java
Torch3333: Thanks
https://github.com/scalar-labs/scalardb/pull/3611/commits/2b80b6f26a0ac011f5c7e3e56261762e6dbad5e6

core/src/integration-test/java/com/scalar/db/storage/jdbc/RdbEngineStrategyExceptionIntegrationTest.java
brfrn169: Agreed, that makes sense. Let's proceed with that. Are you planning to do that in a separate PR?

Torch3333: Yes, I will do that in another PR.

brfrn169 (approved):
LGTM! Thank you!

PR #3614, opened 2026-06-09
Fix read-only handling in coordinator write omission commit paths

brfrn169:
## Description

During the review of #3564, two pre-existing issues were found in how Consensus Commit handles read-only transactions together with the `coordinator.write_omission_on_read_only` setting. Both can produce incorrect coordinator behavior: one writes an unintended coordinator record, and the other crashes a read-only commit and leaks a group commit slot. This PR fixes both so that read-only transactions are handled consistently across all commit paths regardless of the write-omission and group-commit settings.

## Related issues and/or PRs

Related to #3564 (both issues were surfaced during its review).

## Changes made

- **`CommitHandler` (before-preparation hook failure paths):** The hook-failure paths aborted the coordinator state and rolled back records without the write-omission guard that the validation-failure path already applies. As a result, a read-only transaction with coordinator write omission enabled wrote a spurious `ABORTED` coordinator record when a before-preparation hook threw. The shared guard logic is now extracted into a single `abortStateAndRollbackRecordsIfNeeded` helper and applied to all three failure paths (hook invoke, hook future wait, and validation).
- **`ConsensusCommitManager#begin`:** Group commit slot reservation was skipped for every read-only transaction. With coordinator write omission disabled, a read-only transaction therefore kept a bare (unreserved) transaction ID and crashed with `AssertionError("Group commit unexpectedly failed")` when committing its state via group commit. The reservation is now performed whenever a transaction will write a coordinator state row — every non-read-only transaction, plus read-only transactions when write omission is disabled — so such read-only transactions are group-committed like any other state-writing transaction.
- **`ConsensusCommit` (rollback/close):** Releases the reserved group commit slot for read-only transactions that actually hold one (identified by a full-key transaction ID), so the slot is not leaked when such a read-only transaction is rolled back or closed.
- **`CommitHandlerWithGroupCommit#cancelGroupCommitIfNeeded`:** Skips `remove()` for bare (non-full-key) transaction IDs that never reserved a slot, avoiding key-parsing failures, and logs the exception when removal does fail.
- **Tests:** Added unit tests covering the hook-failure write-omission behavior (read-only and non-read-only, omission enabled and disabled, both the hook invoke and hook future paths), the group-commit slot lifecycle on those paths, the `begin` reservation behavior per write-omission setting, and the rollback slot-release behavior for reserved vs. non-reserved read-only transactions.

## Checklist

- No user-facing documentation change is needed; both fixes are internal corrections to existing behavior with no public API or configuration changes.
- Both issues exist on `master`, `3`, `3.18`, `3.17`, and `3.16` (the coordinator write-omission-on-read-only feature, the before-preparation hook, and the unconditional read-only reserve skip are all present there). They do **not** exist on `3.15` (the prerequisite features are not present). Backports to `3`/`3.18`/`3.17`/`3.16` are appropriate; `3.16`–`3.18` will need a minor adaptation for the hook-failure path because they predate #3564 and still use the `abortState(String)` signature.
- The crash (read-only + write omission disabled + group commit enabled) is the more severe of the two; the spurious `ABORTED` record (write omission enabled + before-preparation hook failure) causes no data corruption but is observable noise for recovery/backup tooling.

## Release notes

- Fixed an issue in Consensus Commit where a read-only transaction could write an unintended `ABORTED` record to the Coordinator table when a before-preparation hook failed and coordinator write omission on read-only was enabled.
- Fixed an issue in Consensus Commit where a read-only transaction could fail when coordinator write omission on read-only was disabled and coordinator group commit was enabled.


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -226,7 +229,12 @@ public void rollback() {
...
 
-    if (groupCommitter != null && !context.readOnly) {
+    // Release the reserved group commit slot if one was reserved. ConsensusCommitManager.begin()
+    // reserves a slot for any transaction that writes a coordinator state row — every non-read-only
+    // transaction, plus read-only transactions when coordinator write omission is disabled. In the
+    // latter case the reserved transaction ID is a group commit full key, so isFullKey() reliably
+    // distinguishes a read-only transaction that holds a slot from one that does not.
+    if (groupCommitter != null && (!context.readOnly || KEY_MANIPULATOR.isFullKey(getId()))) {
       groupCommitter.remove(getId());
     }
```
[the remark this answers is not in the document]
brfrn169: @gemini-code-assist Good catch — `rollback()` is a cleanup path and should not propagate or mask failures here. Wrapped `groupCommitter.remove(getId())` in a try-catch with a warning log, consistent with `CommitHandlerWithGroupCommit.cancelGroupCommitIfNeeded`. Fixed in 0465fba55.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -278,7 +285,11 @@ DistributedTransaction begin(
       String txId, Isolation isolation, boolean readOnly, boolean oneOperation) {
     checkArgument(!Strings.isNullOrEmpty(txId));
     checkNotNull(isolation);
-    if (!readOnly && isGroupCommitEnabled()) {
+    // Reserve a group commit slot whenever the transaction will write a coordinator state row. A
+    // read-only transaction writes one only when coordinator write omission is disabled, so it must
+    // also reserve a slot in that case; otherwise its state would be committed via the
+    // non-group-commit path with a bare (unreserved) transaction ID.
```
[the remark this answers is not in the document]
brfrn169: Good point — the comment overstated the reservation rule. begin() reserves a slot for every non-read-only transaction, and for a read-only transaction only when coordinator write omission is disabled; a write-less non-read-only transaction still reserves here and cancels the slot later on the commit path. Reworded the comment accordingly in 0465fba55.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommit.java
```diff
@@ -226,7 +229,12 @@ public void rollback() {
       logger.warn("Failed to close the scanner. Transaction ID: {}", getId(), e);
     }
 
-    if (groupCommitter != null && !context.readOnly) {
+    // Release the reserved group commit slot if one was reserved. ConsensusCommitManager.begin()
+    // reserves a slot for any transaction that writes a coordinator state row — every non-read-only
+    // transaction, plus read-only transactions when coordinator write omission is disabled. In the
+    // latter case the reserved transaction ID is a group commit full key, so isFullKey() reliably
+    // distinguishes a read-only transaction that holds a slot from one that does not.
```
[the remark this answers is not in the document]
brfrn169: Agreed — reworded this comment too so it no longer implies a 1:1 relationship between reservation and writing a coordinator state row. It now states that begin() reserves for every non-read-only transaction and for a read-only transaction only when write omission is disabled. Fixed in 0465fba55.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -278,7 +285,13 @@ DistributedTransaction begin(
...
-    if (!readOnly && isGroupCommitEnabled()) {
+    // Reserve a group commit slot for every non-read-only transaction, and for a read-only
+    // transaction only when coordinator write omission is disabled. Such a read-only transaction
+    // writes a coordinator state row, so without a slot it would reach the group commit path with a
+    // bare (unreserved) transaction ID and fail. A non-read-only transaction that turns out to be
+    // write-less under write omission still reserves a slot here but cancels it later on the commit
+    // path.
+    if (isGroupCommitEnabled() && (!readOnly || !coordinatorWriteOmissionOnReadOnlyEnabled)) {
       assert groupCommitter != null;
       txId = groupCommitter.reserve(txId);
```
komamitsu, line 305: How about adding `context.groupCommitSlotReserved` or something and setting it to true here? Then `ConsensusCommit.rollback()` and `cancelGroupCommitIfNeeded()` can check the flag directly instead of calling `isFullKey(getId())`.

brfrn169: Good idea — done in 03f7e432e3c3f5f998f980874e6d3247869394e8. I added `TransactionContext.groupCommitSlotReserved`, set it in `begin()` when a slot is reserved, and `ConsensusCommit.rollback()` / `cancelGroupCommitIfNeeded()` now check the flag instead of `isFullKey(getId())`.

While doing this I also removed the `abortStateWithoutWriteSet(String)` override in `CommitHandlerWithGroupCommit`. It was the only `cancelGroupCommitIfNeeded()` caller that had just a bare transaction ID and no context, which is why `cancelGroupCommitIfNeeded()` had to take a `String` and rely on `isFullKey()` in the first place. With the flag in place, every remaining caller has a `TransactionContext`, so the override is no longer needed and `cancelGroupCommitIfNeeded()` now takes the context directly.

One thing I noticed while making this change: the slot release in that override was only ever helping the same-instance case. `manager.rollback(txId)` / `abort(txId)` goes through `abortStateWithoutWriteSet(txId)`, which writes the ABORTED coordinator state under the full ID only. For a normal group that never conflicts with the parent-ID state the group commit writes, so a `manager.abort(txId)` issued from a *different* instance against an in-flight (normal-group) transaction can be silently lost — the group still commits and `getState()` returns COMMITTED. The old override only masked this locally by removing the slot on the same instance; it does nothing cross-instance. This is pre-existing (it has been there since coordinator group commit was introduced), not something this PR changes. The proper fix is to make `rollback(txId)` use the same two-step abort as `putStateForLazyRecoveryRollback` (write the parent-ID conflict marker first, then the full-ID ABORTED). I'll do that in a separate PR to keep this one focused.

brfrn169: Follow-up: removing that override also surfaced the same-instance side of this as two failing group-commit ITs — `abort_forOngoingTransaction_ShouldAbortCorrectly` and `rollback_forOngoingTransaction_ShouldRollbackCorrectly`. Same root cause as above: the ongoing transaction's own `commit()` still group-commits under the parent ID, so the by-ID abort written under the full ID doesn't take.

I've disabled these two tests for coordinator group commit via `@DisabledIf` in 9ed14ef9b, with a TODO to re-enable them in the same follow-up PR that adds the two-step abort. They remain enabled for non-group-commit runs.

feeblefakie:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
komamitsu: Thanks for the detailed explanation.

> While doing this I also removed the `abortStateWithoutWriteSet(String)` override in `CommitHandlerWithGroupCommit`. It was the only `cancelGroupCommitIfNeeded()` caller that had just a bare transaction ID and no context, which is why `cancelGroupCommitIfNeeded()` had to take a `String` and rely on `isFullKey()` in the first place.

Ah, sorry for missing the case where `manager.abort(txId)` is called. There's no `TransactionContext` available at that call site since the API takes only a transaction ID. But I think we still need to release the group commit slot here. Otherwise the in-flight slot stays in `NormalGroup` and the group emit later writes COMMITTED at the parent ID, which is exactly what makes `abort_forOngoingTransaction_ShouldAbortCorrectly()` fail.

> so a `manager.abort(txId)` issued from a *different* instance against an in-flight (normal-group) transaction can be silently lost

I might be missing something, but all operations on the same transaction ID are supposed to be executed on the same instance, right? It would be great if you could clarify cases where you're actually concerned.

komamitsu: If the scope of "different instances" includes different threads, yes, I think there is a potential bug with the following code:

```java
    CountDownLatch latch = new CountDownLatch(1);
    ExecutorService executorService = Executors.newSingleThreadExecutor();
    AtomicReference<String> txId = new AtomicReference<>();
    executorService.submit(() -> {
      DistributedTransaction transaction = manager.begin();
      txId.set(transaction.getId());
      latch.countDown();
      transaction.get(prepareGet(0, 0));
      transaction.put(Put.newBuilder(preparePut(0, 0)).intValue(BALANCE, 1).build());
      transaction.commit();
    });

    latch.await();
    manager.abort(txId.get());
```

I'm not sure if this usage is reasonable, though.

brfrn169: > But I think we still need to release the group commit slot here.

`DistributedTransactionManager.abort(String)` / `rollback(String)` takes only a transaction ID precisely so that a different instance (an operator tool, a cleanup process, or another client that only knows the ID) can terminate the transaction. In that intended case there's no local slot to release, so the old same-instance `groupCommitter.remove(id)` was just a band-aid — it didn't fix the real cross-instance behavior, where the full-ID ABORTED never conflicts with the parent-ID COMMITTED and the abort is silently lost.

> If the scope of "different instances" includes different threads, yes, I think there is a potential bug

Right, and it has the same root cause. The correct fix is to make the by-ID abort win at the coordinator level via the same two-step protocol as lazy recovery (parent-ID conflict marker → full-ID ABORTED), which works regardless of instance or thread.

I've done this in #3619, which also re-enables the two tests disabled here.

What do you think?

komamitsu: I see. Sounds reasonable.

Thanks for taking care of the issue!

komamitsu (approved):
LGTM, thank you!

KodaiD (approved):
LGTM. Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #3617, opened 2026-06-11
Add DistributedTransactionManager#recoverRecord

brfrn169:
## Description

When a Consensus Commit transaction crashes, it can leave a record in an uncommitted physical state (`PREPARED` or `DELETED`). Today such records are cleaned up either by lazy recovery (on the next read, and only after the writer is considered expired for the no-coordinator-state case) or by `finishTransaction(txId)`. However, `finishTransaction` is transaction-ID-scoped and write-set-driven: it only works for transactions that reached `DistributedTransaction#commit()` and therefore persisted a write set. Records left behind by a writer that has **no write set** (e.g. a transaction that crashed before `commit()`, or one from a binary that pre-dates the write-set column) cannot be repaired on demand and must wait for lazy recovery to happen to read them.

This PR adds `recoverRecord`, a key-scoped operational API that lets an operator deterministically repair a single such record on demand.

## Related issues and/or PRs

N/A

## Changes made

- Added `boolean recoverRecord(String namespace, String table, Key partitionKey, @Nullable Key clusteringKey)` to `DistributedTransactionManager`, documented as a low-level operational API (mirroring `finishTransaction`/`getState`/`rollback`).
- Implemented it in `ConsensusCommitManager`: reads the record, derives the writer transaction ID from it, and recovers via the existing `Coordinator`/`RecoveryHandler` machinery. It rolls the record forward or back based on the writer's coordinator state, keeps the expiration guard for the no-coordinator-state case (so a healthy in-flight writer is not aborted), and does **not** delete the Coordinator state row (other records of the same writer may still need it). Key-shape and table validation are delegated to the storage-layer `OperationChecker` invoked by `storage.get`.
- **Return value is a `boolean`, not a `TransactionState`.** It reports *whether* the record was recovered: `true` means the record was resolved (rolled forward/back, or already resolved), and `false` signals the writer may still be in flight, so the call can be retried later. It deliberately does not return the resolved terminal state because that cannot always be determined accurately in races (e.g. a concurrent `finishTransaction` that already cleaned up the Coordinator state row). Callers that need the actual outcome should read the record after a `true` return.
- Refactored `RecoveryHandler` to separate best-effort from guaranteed recovery:
  - Renamed the existing best-effort `recover` to `tryRecover` (still `void`; used by lazy recovery on the read path — a conflicting record is left for the winning actor or a later lazy read).
  - Added a new synchronous `recover` that physically resolves the record, overloaded to take either an `Optional<Coordinator.State>` (returning whether the record was recovered) or a concrete `Coordinator.State` (`void`).
  - Centralized the terminal-state dispatch (roll forward on `COMMITTED`, back on `ABORTED`) in a single helper.
- Generalized `RecoveryExecutor.executeSynchronously` with an `Optional<Coordinator.State>` overload returning whether the record was recovered, giving `finishTransaction` and `recoverRecord` a single guaranteed synchronous recovery entry point.
- Wired the cross-cutting surface: `DecoratedDistributedTransactionManager` and `TransactionService` delegate; `JdbcTransactionManager` and `SingleCrudOperationTransactionManager` throw `UnsupportedOperationException` with dedicated `CoreError` messages.
- Added unit tests (`ConsensusCommitManagerTest`, `RecoveryHandlerTest`, `RecoveryExecutorTest`, and the JDBC/single-CRUD manager tests) and Consensus Commit integration tests (`ConsensusCommitSpecificIntegrationTestBase`), including group-commit cases and the not-expired `false`-return case. Added Javadoc to `TransactionResult#getState`/`#isCommitted`/`#isDeemedAsCommitted`.

## Checklist

This PR is stacked on #3621 ("Return the before-image only after the lazy-recovery rollback is written on a read") and targets its branch `fix-stale-read-on-lazy-recovery-abort-conflict`. It has been rebased onto that branch, so the diff against the base contains only the `recoverRecord` changes plus the supporting `RecoveryHandler`/`RecoveryExecutor` refactor; the best-effort lazy-recovery path (renamed `tryRecover`) coexists with the base branch's synchronous lazy-recovery-rollback fix. Once #3621 is merged, this PR will be retargeted to `master`.

## Release notes

Added the `DistributedTransactionManager#recoverRecord` API for recovering a single record left in an uncommitted state by a crashed transaction. Note that this is a low-level operational API specific to the Consensus Commit transaction manager. Most applications should not call it directly; it is intended for advanced use cases. Callers are expected to understand the underlying transaction lifecycle and the implications of invoking this method directly.

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
@@ -2,10 +2,12 @@ package com.scalar.db.api;
 
 import com.scalar.db.exception.transaction.TransactionException;
 import com.scalar.db.exception.transaction.TransactionNotFoundException;
+import com.scalar.db.io.Key;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
+import javax.annotation.Nullable;
 
 public interface DistributedTransactionManager
     extends TransactionManagerCrudOperable, AutoCloseable {
@@ -479,12 +481,12 @@ public interface DistributedTransactionManager
   }
 
   /**
-   * Finishes a given terminated transaction by performing any remaining post-termination work and
-   * cleaning up the Coordinator state row. The transaction must already be in a terminal state
+   * Finishes a given terminated transaction by completing any remaining post-termination work and
+   * performing the Coordinator state cleanup. The transaction must already be in a terminal state
    * ({@code COMMITTED} or {@code ABORTED}); this method completes the per-record work that was
    * otherwise deferred to lazy recovery — rolling forward {@code PREPARED} or {@code DELETED}
    * records of a committed transaction, or rolling back {@code PREPARED} or {@code DELETED} records
-   * of an aborted one — and then removes the Coordinator state row.
+   * of an aborted one — and then performs the cleanup.
    *
    * <p>This is a best-effort, retryable cleanup API intended to be called after a transaction
    * terminates so that ScalarDB can complete per-record post-termination work eagerly and reclaim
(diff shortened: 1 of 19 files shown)


core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java

feeblefakie (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM! 👍

jnmt (approved):
LGTM, thank you!

KodaiD (approved):
LGTM. Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #3619, opened 2026-06-13
Abort in-flight group-committed transactions correctly on the manager rollback/abort-by-ID path

brfrn169:
## Description

When coordinator group commit is enabled, `DistributedTransactionManager.rollback(String)` / `abort(String)` could fail to abort an in-flight, group-committed transaction. The by-ID abort path wrote the `ABORTED` coordinator state under the full transaction ID only, which never conflicts with the parent-ID `COMMITTED` state that a normal group commit writes. As a result the abort was silently lost: the transaction still committed and `getState()` returned `COMMITTED`.

This issue was surfaced by #3614, where removing a same-instance-only slot-cancel band-aid made it visible as failing integration tests. #3614 temporarily disabled those tests for group commit; this PR fixes the behavior and re-enables them.

## Related issues and/or PRs

- #3614

## Changes made

- Add `CommitHandler.abortStateForRollback(String)` for the manager-level rollback/abort-by-ID path. For a group commit full key it uses the same two-step protocol as lazy recovery (`Coordinator.putStateForLazyRecoveryRollback`): it writes the parent-ID conflict marker before the full-ID `ABORTED` record, so the abort wins against an in-flight normal group commit. For a non-group-commit ID it just writes the `ABORTED` record.
- Route `ConsensusCommitManager.rollback(String)` through `abortStateForRollback`. `abortStateWithoutWriteSet` keeps its single-`putState` behavior for the two-phase commit rollback path. Conflict resolution is extracted into `resolveAbortStateConflict` and shared.
- Re-enable `abort_forOngoingTransaction_ShouldAbortCorrectly` and `rollback_forOngoingTransaction_ShouldRollbackCorrectly` for coordinator group commit.
- Add unit tests for `abortStateForRollback`.

## Checklist

For a same-instance abort of an in-flight transaction that is batched with other transactions under the same parent, the parent-ID conflict marker also makes those co-batched transactions fail their commit with a (retryable) conflict. This is acceptable because aborting an in-flight transaction by ID is an out-of-band operation rather than the normal `transaction.rollback()` path.

## Release notes

Fixed a bug where aborting an in-flight, group-committed transaction by ID via `DistributedTransactionManager.rollback(String)` / `abort(String)` could be lost when the Coordinator group commit feature was enabled.

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -528,11 +528,11 @@ public class CommitHandler {
    * DistributedTransactionManager.rollback(String)} / {@code abort(String)} path), where only the
    * transaction ID is known.
    *
-   * <p>For a group commit full key, it uses the same two-step protocol as lazy recovery (see {@link
-   * Coordinator#putStateForLazyRecoveryRollback(String)}): it writes the parent-ID conflict marker
-   * before the full-ID ABORTED record, so the abort wins against an in-flight normal group commit
-   * (which writes the COMMITTED state under the parent ID). For a non-group-commit ID it just
-   * writes the ABORTED record.
+   * <p>This delegates to {@link Coordinator#putStateForLazyRecoveryRollback(String)}, which
+   * branches on the given ID: for a group commit full key it uses the same two-step protocol as
+   * lazy recovery, writing the parent-ID conflict marker before the full-ID ABORTED record so the
+   * abort wins against an in-flight normal group commit (which writes the COMMITTED state under the
+   * parent ID); for a non-group-commit ID it just writes the ABORTED record.
    *
    * @param id the transaction ID
    * @return the resulting transaction state — either {@link TransactionState#ABORTED} or, if a
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -449,9 +449,8 @@ public class CommitHandler {
   }
 
   /**
-   * Writes the ABORTED state without persisting a {@code tx_write_set}. Intended for callers that
-   * don't have the originating transaction context (lazy recovery, manager-level rollback) or that
-   * don't want per-transaction write-set logging.
+   * Writes the ABORTED state without persisting a {@code tx_write_set} using a single {@code
+   * putState}. Intended for callers that don't want per-transaction write-set logging.
    *
    * @param id the transaction ID
    * @return the resulting transaction state — either {@link TransactionState#ABORTED} or, if a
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
brfrn169, line 1: This is a comment to block this PR from being merged.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -497,6 +479,31 @@ private TransactionState abortStateInternal(String id, @Nullable WriteSet writeS
...
+  // concurrent writer beat us, so follow whatever state is already persisted.
+  private TransactionState resolveAbortStateConflict(String id, CoordinatorConflictException e)
+      throws UnknownTransactionStatusException {
+    try {
+      Optional<Coordinator.State> state = coordinator.getState(id);
+      if (state.isPresent()) {
+        // successfully COMMITTED or ABORTED
+        return state.get().getState();
+      }
+      throw new UnknownTransactionStatusException(
```
komamitsu, line 492: The other transactions in the same group fail with `UnknownTransactionStatusException`?

For example, two transactions in one group under parent `P`, where `P:a` is aborted and `P:b` is the other one.

1. Abort `P:a`. `putStateForLazyRecoveryRollback` writes `P` (`tx_child_ids` empty, `ABORTED`) and `P:a` (`ABORTED`).
2. Group commit emits `putStateForGroupCommit(P, [a, b], COMMITTED)`, which conflicts with `P`, so both fail.
3. `P:a` commit: `getState` finds `P:a` = `ABORTED`, so `CommitConflictException` (retryable).
4. `P:b` commit: `getState` finds `P` (which doesn't list `b`) and no `P:b` record, so it returns empty and throws `UnknownTransactionStatusException`.

I think this issue has existed from https://github.com/scalar-labs/scalardb/pull/2135. Conflicts with delayed commit and lazy-recovery might be rare, though.

brfrn169: Good catch, and thanks for the detailed trace — the scenario is valid and your steps are exactly right. `P:a`'s abort writes the parent `P` marker (empty `tx_child_ids`, `ABORTED`), which poisons the whole group commit, but only `P:a` gets a full-ID `ABORTED` record. `P:b` then finds neither `b` in the parent nor a full-ID `P:b` row, so `getState` returns empty and the commit throws `UnknownTransactionStatusException`.

A couple of points:

1. **The collateral abort of `P:b` is inherent to group commit and can't be avoided here.** A group commit is atomic over the single coordinator row keyed by the parent, so reliably aborting an *uncommitted* group member necessarily prevents the whole group from committing. So `P:b` being aborted is the unavoidable part — what we *can* fix is the exception type: `P:b` should get a retryable `CommitConflictException`, not `UnknownTransactionStatusException`.

2. **You're right that this predates this PR (#2135).** The lazy-recovery two-step protocol could already trigger it via background recovery; this PR just adds another trigger by routing the manager rollback/abort-by-ID path through the same protocol (previously it wrote a single full-ID record that didn't conflict with the group commit at all — which was the bug this PR fixes).

3. **This is resolved in the follow-up branch [`fix-recovery-on-coordinator-state-deletion`](https://github.com/scalar-labs/scalardb/tree/fix-recovery-on-coordinator-state-deletion).** There, `handleCommitConflict` resolves the coordinator-state-absent case by rolling the records back and throwing a retryable `CommitConflictException` instead of `UnknownTransactionStatusException`. `P:b` hits exactly this path. The reasoning holds for a group-commit child: `getState` checks both the parent and full-ID rows, so an absent state genuinely means the transaction never committed, and the only other writer of its state is a lazy recovery (which only ever writes `ABORTED`) — so it's definitively aborted, hence a retryable conflict.

So the final outcome (`P:b` is aborted and retried) stays the same, but the surfaced exception becomes the proper retryable conflict.

I’ll create a PR for  [`fix-recovery-on-coordinator-state-deletion`](https://github.com/scalar-labs/scalardb/tree/fix-recovery-on-coordinator-state-deletion) after the dependent PRs are merged.

Thanks!

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTestBase.java
```diff
@@ -16,7 +16,6 @@
 import java.util.Optional;
 import java.util.Properties;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.DisabledIf;
 
 public abstract class ConsensusCommitIntegrationTestBase
```
feeblefakie, line 20: I might miss something, but it looks like no test deterministically reproduces the issue.
Can we do it somehow by using InOrder assertion or something?

komamitsu (approved):
LGTM, thank you!

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitIntegrationTestBase.java
brfrn169: Good catch. You're right that the re-enabled `abort/rollback_forOngoingTransaction` tests don't deterministically reproduce the issue — whether the in-flight transaction is committed via a normal group (parent ID) or an isolated/delayed commit (full ID) depends on group commit timing, so they don't reliably exercise the buggy path.

I addressed this in two ways (46d18cfb5):

1. **Unit (`CoordinatorTest`)** — `InOrder` assertions on `putStateForLazyRecoveryRollback`, verifying the parent-ID conflict marker is written *before* the full-ID `ABORTED` record. That ordering is what makes the abort win against an in-flight normal group commit, and it's now checked deterministically.

2. **Integration (`ConsensusCommitSpecificIntegrationTestBase`)** — a deterministic test that intercepts the group commit's `COMMITTED` write under the parent ID and rolls the transaction back by ID just before that write runs, so the two writes genuinely conflict. I confirmed it fails without the fix (commit succeeds and the state stays `COMMITTED`) and passes with it.

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #3650, opened 2026-06-22
Handle coordinator state row deletion in Consensus Commit recovery and read paths

brfrn169:
## Description

Consensus Commit has historically assumed that a transaction's Coordinator state row, once written, is never removed — so on the recovery and read paths an absent state row was taken to mean "this transaction never committed." `finishTransaction` (#3567) and an upcoming cleanup now **delete** Coordinator state rows after a transaction is finished. With deletion, an absent state row is ambiguous: the writer may have committed (or aborted) and then had its row cleaned up.

Under the old assumptions this leads to several correctness problems once cleanup is in play:

- a read returns a **stale before-image** for a record that actually committed and was cleaned up;
- lazy recovery / `recoverRecord` writes a **spurious `ABORTED`** state for a transaction that actually committed;
- the group-commit lazy-recovery rollback hits an **`AssertionError`** when the parent Coordinator row was cleaned up;
- `getState` / `rollback` / `abort`-by-ID give overly conservative or misleading results.

This PR makes the read path, lazy recovery, and commit/abort conflict resolution correct (or honestly `UNKNOWN`) when a Coordinator state cleanup process can delete state rows, and documents the resulting `getState` / `rollback` contracts.

The changes are split into self-contained commits so the PR can be reviewed commit by commit (hashes are current as of the latest push and will change if the branch is rebased):

- 25e7eea Re-read the record to resolve a coordinator-state-absent read under cleanup deletion (read path + `abortIfExpired` re-read guard)
- 8245c16 Handle a cleaned-up parent row in group-commit lazy-recovery rollback
- b51d989 Resolve a coordinator-state-absent commit/abort conflict by calling path
- d7a11d8 Document getState and rollback/abort semantics for cleaned-up transactions
- 83c81c1 Add integration tests for reads racing coordinator-state cleanup

## Related issues and/or PRs

- #3567 (`finishTransaction`, which introduced Coordinator state deletion)
- #3621 (read before-image fix, merged)
- #3619 (abort-by-ID, merged).

## Changes made

- **Read path (`RETURN_LATEST_RESULT_AND_RECOVER`, get/scan/streaming-scanner):** before resolving an expired writer that has no Coordinator state, physically re-read the record and resolve from its real state — return the committed value, return empty, or re-resolve against a different writer — instead of blindly returning the before-image. After writing the lazy-recovery rollback, re-read once more to close the cleanup TOCTOU. Bounded by `MAX_RESOLUTION_ATTEMPTS`.
- **Lazy recovery / `recoverRecord` (`abortIfExpired`):** apply the same pre-abort re-read guard so the background and guaranteed-recovery paths never write a spurious `ABORTED` for a committed-and-cleaned-up writer.
- **Group-commit lazy-recovery rollback:** treat a cleaned-up parent Coordinator row as normal flow (fall through to the full-ID write) instead of throwing `AssertionError`.
- **Commit/abort conflict resolution:** resolve a now-absent Coordinator state by what the calling path can prove — `handleCommitConflict` rolls back and throws `CommitConflictException`; the self-abort path and all Two-phase Commit aborts resolve to `ABORTED`; the One-phase Commit abort-by-ID path stays an honest `UNKNOWN` (it can target a transaction that actually committed and whose row `finishTransaction` can remove).
- **Documentation:** `getState` returns `UNKNOWN` for a finished, cleaned-up transaction (indistinguishable from never-existed / read-error); `rollback`/`abort` document the finished-transaction limitation.
- **Tests:** unit tests for all of the above, plus integration tests covering reads racing Coordinator-state cleanup (five scenarios across isolation levels, read-only modes, and commit types), verified on PostgreSQL with and without coordinator group commit enabled.

## Checklist

- The integration-test bases rename the `prepared_at` test locals to `preparedAt` for Java naming consistency; this touches a few pre-existing test variables in the touched files.

## Release notes

Fixed the Consensus Commit recovery and read paths to stay correct when Coordinator state records are removed by cleanup (for example, by `finishTransaction`), preventing stale reads, spurious aborts, and a crash during group-commit lazy recovery.

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
@@ -438,8 +438,21 @@ public interface DistributedTransactionManager
    * advanced use cases. Callers are expected to understand the underlying transaction lifecycle and
    * the implications of invoking this method directly.
    *
+   * <p><b>Finished or unknown transactions:</b> the state is read from the transaction's
+   * Coordinator state row. Once that row has been removed — by {@link #finishTransaction(String)}
+   * or another Coordinator state cleanup — the transaction's terminal outcome is no longer
+   * persisted and cannot be read, so this method returns {@link TransactionState#UNKNOWN}. A
+   * committed-and-cleaned-up transaction, a transaction ID that never existed, and a transaction
+   * whose state row could not be read are therefore indistinguishable through this method: all
+   * return {@link TransactionState#UNKNOWN}.
+   *
+   * <p>Relatedly, calling {@link #rollback(String)} or {@link #abort(String)} on a finished,
+   * cleaned-up transaction ID records a fresh ABORTED state, after which this method returns {@link
+   * TransactionState#ABORTED} for that ID even if the transaction had committed.
+   *
    * @param txId a transaction ID
-   * @return {@link TransactionState}
+   * @return the transaction's {@link TransactionState}, or {@link TransactionState#UNKNOWN} if its
+   *     state can no longer be determined (see above)
    * @throws TransactionException if getting the state of a given transaction fails
    * @throws UnsupportedOperationException if the underlying transaction manager does not support
    *     getting a transaction state
@@ -453,6 +466,15 @@ public interface DistributedTransactionManager
    * transaction manager. Most applications should not call it directly — it is intended for
    * advanced use cases. Callers are expected to understand the underlying transaction lifecycle and
    * the implications of invoking this method directly.
+   *
+   * <p><b>Limitation for finished transactions:</b> if the given transaction ID has already
+   * finished and been cleaned up (its Coordinator state row removed by {@link
+   * #finishTransaction(String)} or another Coordinator state cleanup), this method writes a fresh
+   * ABORTED Coordinator state and returns {@link TransactionState#ABORTED} — even if that
+   * transaction had actually committed. (If a Coordinator state row is still present, that existing
+   * state is returned and no new row is written.) A subsequent {@link #getState(String)} for the
+   * same ID then returns {@link TransactionState#ABORTED}, masking the true (possibly committed)
+   * outcome. Do not roll back a transaction ID that may already have finished.
    *
    * @param txId a transaction ID
    * @return {@link TransactionState}
@@ -469,6 +491,8 @@ public interface DistributedTransactionManager
    * transaction manager. Most applications should not call it directly — it is intended for
    * advanced use cases. Callers are expected to understand the underlying transaction lifecycle and
    * the implications of invoking this method directly.
+   *
+   * <p>The finished-transaction limitation described on {@link #rollback(String)} applies here too.
    *
    * @param txId a transaction ID
    * @return {@link TransactionState}
@@ -512,6 +536,10 @@ public interface DistributedTransactionManager
    * <p><b>Idempotency:</b> calling this method on a transaction ID whose state row is absent
    * (already finished, never started, or already cleaned up by a concurrent caller) returns {@code
    * true}. Callers may safely re-invoke this method on the same transaction ID.
+   *
+   * <p><b>Effect on {@link #getState(String)}:</b> once this method finishes a transaction and
+   * reclaims its Coordinator state row, {@link #getState(String)} for that transaction ID returns
+   * {@link TransactionState#UNKNOWN}, since the terminal outcome is no longer persisted.
    *
    * <p><b>Group commit:</b> when the transaction ID belongs to a child of a group commit, the call
    * processes the write sets of all sibling children in a single pass and then deletes the shared
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
@@ -187,7 +187,12 @@ public class RecoveryExecutor implements AutoCloseable {
       throws CrudException {
     TransactionResult current = result;
     for (int attempt = 0; ; attempt++) {
-      Optional<Coordinator.State> state = getCoordinatorState(current.getId());
+      // current is always an uncommitted record, so its tx_id is non-null here; getId() is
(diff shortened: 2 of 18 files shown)


core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
```diff
@@ -293,6 +293,23 @@ private boolean abortIfExpired(
...
+    // process can finalize the record and remove the writer's state row, so an expired writer with
+    // no coordinator state no longer implies the record is still uncommitted: the writer may have
+    // committed and been cleaned up. Only abort when the record is still uncommitted by this
+    // writer; otherwise writing the ABORTED state would leave a spurious tombstone for an
+    // already-resolved (possibly committed) transaction.
+    Optional<TransactionResult> latest =
+        ConsensusCommitUtils.rereadRecord(storage, tableMetadataManager, selection, result);
+    if (!latest.isPresent()
+        || latest.get().isCommitted()
+        || !result.getId().equals(latest.get().getId())) {
```
komamitsu: `result.getId()` has `@Nullable`, so `Objects.equals()` is safer just in case?

```suggestion
        || !Objects.equals(result.getId(), latest.get().getId())) {
```

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -200,23 +143,184 @@ public Result execute(
...
-        future = Futures.immediateFuture(null);
+      if (!fresh.isPresent()) {
+        // The record is gone (a committed delete, or a rolled-back insert). Return empty.
+        return new Result(key, Optional.empty(), Futures.immediateFuture(null), true);
+      }
+      if (fresh.get().isCommitted()) {
+        // The transaction committed and was cleaned up; the record carries the committed value.
+        return new Result(key, fresh, Futures.immediateFuture(null), false);
+      }
+      if (!current.getId().equals(fresh.get().getId())) {
```
komamitsu: `result.getId()` has `@Nullable`, so `Objects.equals()` is safer just in case?

```suggestion
      if (!Objects.equals(current.getId(), fresh.get().getId())) {
```

brfrn169: Thanks! This call site is on a recovery path where the `TransactionResult` is always an uncommitted record, so `getId()` is non-null here — it's `@Nullable` only for deemed-committed (imported) rows, which never reach this code. The surrounding calls (`getCoordinatorState`, `tryAbortExpiredTransaction`) share the same assumption, so rather than only switching `equals()` to `Objects.equals()`, I made the invariant explicit in 9ae1674de2c57e8218f0880d083e420627a4981d: capture the id once as a non-null local with an `assert` and route those calls through it. I also documented the nullability contract in `TransactionResult.getId()`'s Javadoc.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
brfrn169: Same as the reply on the `RecoveryExecutor` thread in 9ae1674de2c57e8218f0880d083e420627a4981d — `result` here is always an uncommitted record, so `getId()` is non-null. Made the invariant explicit with a non-null local + `assert` (and covered the same pattern at `RecoveryExecutor.java:280`).

komamitsu (approved):
LGTM! 👍

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
```diff
@@ -200,23 +143,189 @@ public Result execute(
...
-        future = Futures.immediateFuture(null);
+      if (!fresh.isPresent()) {
+        // The record is gone (a committed delete, or a rolled-back insert). Return empty.
+        return new Result(key, Optional.empty(), Futures.immediateFuture(null), true);
+      }
+      if (fresh.get().isCommitted()) {
+        // The transaction committed and was cleaned up; the record carries the committed value.
+        return new Result(key, fresh, Futures.immediateFuture(null), false);
+      }
+      if (!txId.equals(fresh.get().getId())) {
```
feeblefakie, line 238: I'm a bit concerned with the redundant predicates; `!present || isCommitted() || !txId.equals(getId())`.
I'm not sure about the right way to fix it, but my AI suggests the following. WDYT?

```
enum RereadOutcome {
    GONE,                   // record no longer exists
    COMMITTED,              // record finalized to a committed value
    REPREPARED,             // record re-prepared by a different transaction
    STILL_PREPARED_BY_SAME  // still uncommitted by the same expired writer (txId)
  }

  static RereadOutcome classifyReread(String txId, Optional<TransactionResult> fresh) {
    if (!fresh.isPresent()) return RereadOutcome.GONE;
    if (fresh.get().isCommitted()) return RereadOutcome.COMMITTED;
    if (!txId.equals(fresh.get().getId())) return RereadOutcome.REPREPARED;
    return RereadOutcome.STILL_PREPARED_BY_SAME;
  }
```

feeblefakie (approved):
LGTM! Thank you!
Left one comment for potential improvement, but it's not a blocker.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryExecutor.java
brfrn169: Thanks! I gave the `classifyReread` extraction a try across the re-read sites. It read nicely in `RecoveryHandler`, but in `RecoveryExecutor` these sites need the unwrapped record afterward, so folding the `isPresent()` check in just traded the named predicate for some `orElseThrow(...)` ceremony — about a wash. So I've left it as is for now. Agree the repeated predicate is a bit noisy though; happy to revisit if it grows. Thanks again!

PR #3653, opened 2026-06-23
Split CommitHandler into an orchestrator and specialized handlers

brfrn169:
## Description

This PR splits the monolithic `CommitHandler` into an orchestrator plus specialized handlers so that participant-side record operations and Coordinator-side state writes are cleanly separated. It also removes the `TransactionTableMetadataManager` argument from `BeforePreparationHook.handle`, since the metadata manager is a long-lived singleton that is better injected once via the constructor. This is a preparatory refactoring with no behavioral change for callers.

## Related issues and/or PRs

N/A

## Changes made

- **Split `CommitHandler`.**
  - `ParticipantCommitHandler`: participant-side data-record operations (`prepareRecords`, `validateRecords`, `commitRecords`, `rollbackRecords`, `onePhaseCommitRecords`).
  - `CoordinatorCommitHandler`: Coordinator-table state writes (`commitState`, `commitStateWithoutWriteSet`, `abortState`, `abortStateWithoutWriteSet`, `abortStateForRollback`) plus the commit-conflict resolution helper.
  - `CoordinatorCommitHandlerWithGroupCommit`: extends the Coordinator-side handler with the group-commit emitter / slot-cancellation behavior.
  - `CommitHandler` becomes the orchestrator owning `commit(context)`, `canOnePhaseCommit`, `invokeBeforePreparationHook`, and the prepare -> validate -> commitState -> commitRecords sequencing. Existing public methods remain as thin pass-throughs so direct callers continue to work unchanged.
- **Drop the `TransactionTableMetadataManager` argument from `BeforePreparationHook.handle`;** implementations now hold it via constructor injection.
- Reorganize and add unit tests for the new handlers (`ParticipantCommitHandlerTest`, `CoordinatorCommitHandlerTest`, `CoordinatorCommitHandlerWithGroupCommitTest`).

## Checklist

Pure refactoring; existing callers (`TwoPhaseConsensusCommit` / `ConsensusCommit` / `ConsensusCommitManager` / `RecoveryHandler`) are unaffected thanks to the pass-through methods. A paired change in `LogWriterBeforePreparationHook` is in a separate repository.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/BeforePreparationHook.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/BeforePreparationHook.java
@@ -2,7 +2,12 @@ package com.scalar.db.transaction.consensuscommit;
 
 import java.util.concurrent.Future;
 
+/**
+ * A hook invoked by {@link CommitHandler#commit} before the prepare phase begins.
+ *
+ * <p>This hook is invoked only on the two-phase commit path (prepare, validate, commit-state). It
+ * is not invoked when the one-phase commit optimization path is taken.
+ */
 public interface BeforePreparationHook {
-  Future<Void> handle(
-      TransactionTableMetadataManager transactionTableMetadataManager, TransactionContext context);
+  Future<Void> handle(TransactionContext context);
 }
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
@@ -2,16 +2,10 @@ package com.scalar.db.transaction.consensuscommit;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.errorprone.annotations.concurrent.LazyInit;
-import com.scalar.db.api.Delete;
 import com.scalar.db.api.DistributedStorage;
-import com.scalar.db.api.Mutation;
 import com.scalar.db.api.TransactionState;
 import com.scalar.db.common.CoreError;
-import com.scalar.db.exception.storage.ExecutionException;
-import com.scalar.db.exception.storage.NoMutationException;
-import com.scalar.db.exception.storage.RetriableExecutionException;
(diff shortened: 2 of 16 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/BeforePreparationHook.java
brfrn169, line 1: This is a comment to block this PR from being merged.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -211,438 +215,88 @@ public void commit(TransactionContext context)
...
-            context.transactionId);
-      }
-    } catch (CoordinatorException ex) {
-      throw new UnknownTransactionStatusException(
-          CoreError.CONSENSUS_COMMIT_CANNOT_GET_COORDINATOR_STATUS.buildMessage(ex.getMessage()),
-          ex,
-          context.transactionId);
-    }
+  public void prepareRecords(TransactionContext context) throws PreparationException {
+    participantCommitHandler.prepareRecords(context);
```
komamitsu, line 255: I might be missing something, but `prepareRecords()` is also called on the normal commit path. So, using `participantCommitHandler` seems a bit confusing since it sounds 2PC interface specific. How about `commitRecordHandler` (and `commitStateHandler`)?

brfrn169: Thanks, that's a good point.

The names here follow the consensus-commit protocol roles rather than the 2PC interface. Personally, I think `CoordinatorCommitHandler` is a good name as it's tied to the existing `Coordinator` class and the Coordinator table, so I don't see a need to rename that one. And given that, `ParticipantCommitHandler` works well as its counterpart in the usual 2PC coordinator/participant naming. Renaming just one of them would also break that pairing.

So I'd lean toward keeping them as-is. What do you think?

komamitsu (approved):
LGTM! 👍

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #3654, opened 2026-06-23
Unify commit-phase timestamps via orchestrator-provided timestamps

brfrn169:
## Description

In Consensus Commit, every record/row written during a single transaction previously generated its own `System.currentTimeMillis()` at the call site, so `preparedAt` / `committedAt` could differ across rows of the same transaction and between the data records and the COMMITTED Coordinator state row. This PR aligns those timestamps so that all writes that must logically share one value actually do — **including the group-commit path**, where a batch of transactions shares one COMMITTED Coordinator row. This is a preparatory refactoring with no externally observable behavioral change other than the now-consistent timestamps.

## Related issues and/or PRs

- Based on #3655 (Add an emit result type to the group commit framework). This PR uses #3655's emit result to hand the group-commit emit-time `committedAt` back to each batched transaction. (#3655 is itself stacked on #3653.)

## Changes made

- **`preparedAt`: generated once and passed to `prepareRecords`**, so every prepared row of the transaction shares it (and multiple participants on the Two-phase Commit interface can be given the same value).
- **`committedAt`: generated where the COMMITTED Coordinator state row is written, and returned to the caller** so the data records are stamped with the same value:
  - Non-group: `CoordinatorCommitHandler.commitState` / `commitStateWithoutWriteSet` generate `committedAt` at the `putState` point and return it. In the already-committed `putState`-conflict case, `handleCommitConflict` returns the persisted row's `committedAt`.
  - Group commit: the `Emitter` generates a single `committedAt` for the batched COMMITTED row and returns it; `GroupCommitter#ready` (via #3655) hands that one value back to every transaction in the batch, and `CoordinatorCommitHandlerWithGroupCommit.commitState` returns it.
  - The orchestrator (`CommitHandler#commit` and `TwoPhaseConsensusCommit#commit`) takes `commitState`'s returned `committedAt` and passes it to `commitRecords`, so the Coordinator row and every committed data row share one value — for both group and non-group commit.
- **Recovery roll-forward shares the Coordinator row's `committedAt` too.** When lazy recovery rolls a PREPARED record forward because the writer's Coordinator state is `COMMITTED`, the record is now stamped with that Coordinator state row's `committedAt` (`Coordinator.State#getCreatedAt`) rather than a fresh recovery-time `System.currentTimeMillis()` — so a record committed by recovery shares the same `committedAt` as the row. `RecoveryHandler` threads the state's `committedAt` from `rollRecordToState` → `rollforwardRecord` → `createCommitMutationComposer`.
- **Remove the no-timestamp constructors** of `PrepareMutationComposer` and `CommitMutationComposer`; the caller always supplies the value. `AbstractMutationComposer` keeps only the explicit-timestamp constructor and no longer calls `System.currentTimeMillis()` itself.
- **Deliberately left untouched** (no cross-operation alignment to enforce): one-phase commit (`onePhaseCommitRecords`), abort (`abortState` / `abortStateWithoutWriteSet`), rollback (`RollbackMutationComposer`), and `Coordinator.forceAbort`. See the commit message for the rationale per case.
- Update unit tests to assert the COMMITTED Coordinator row and the committed data records share `commitState`'s returned `committedAt` (group and non-group), and add end-to-end integration tests that verify the shared `committedAt` against real storage, including a group-commit batch case where multiple transactions committed concurrently land in one batched row.

## Checklist

This is part of a stack: `refactor-commit-handler` (#3653) → `add-emit-result-to-group-commit` (#3655) → this PR. It depends on #3655 (and #3653) being merged first. Pure refactoring; the only externally visible effect is that timestamps written within a single transaction are now consistent, including across a group-commit batch.

## Release notes

N/A


core/src/main/java/com/scalar/db/transaction/consensuscommit/AbstractMutationComposer.java
brfrn169, line 1: This is a comment to block this PR from being merged.

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
```diff
@@ -233,7 +233,8 @@ void rollforwardRecord(Selection selection, TransactionResult result) throws Exe
   CommitMutationComposer createCommitMutationComposer(Selection selection, TransactionResult result)
       throws ExecutionException {
     CommitMutationComposer composer =
-        new CommitMutationComposer(result.getId(), tableMetadataManager);
+        new CommitMutationComposer(
+            result.getId(), System.currentTimeMillis(), tableMetadataManager);
```
komamitsu: How about using `Coordinator.State.getCreatedAt()` somehow instead of `System.currentTimeMillis()`?

brfrn169: Good point — done (I mistakenly amend committed and force-pushed it 🙇 ). The roll-forward now stamps the record with the Coordinator state row's `committedAt` (`Coordinator.State#getCreatedAt()`) instead of a fresh `System.currentTimeMillis()`, so a record committed by recovery shares the same `committedAt` as the Coordinator row (the same invariant this PR establishes for the normal commit path).

`RecoveryHandler` now threads the state's `committedAt` from `rollRecordToState` (which takes the `Coordinator.State`) → `rollforwardRecord(..., committedAt)` → `createCommitMutationComposer(..., committedAt)`. Added a `RecoveryHandlerTest` case asserting that a `COMMITTED` state's `createdAt` is the value threaded into the commit composer.

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

PR #3660, opened 2026-06-27
Introduce the TwoPhaseCommit interface

brfrn169:
## Description

This PR introduces a new internal `TwoPhaseCommit` API for Consensus Commit. The API splits a single stateful transaction object into two ID-based roles — `Coordinator` (transaction lifecycle) and `Participant` (CRUD and record-level primitives) — so that the two roles can eventually live in separate services for true distributed two-phase commit. This PR adds only the interface; the rename refactor and the implementation follow in later PRs in the series.

## Related issues and/or PRs

- The implementation follows in #3662.

## Changes made

- Add `com.scalar.db.api.TwoPhaseCommit`, defining:
  - the `Coordinator` role (`begin` / `registerParticipant` / `commit` / `rollback` / `close`). `commit` reports prepare/validate failures as commit-level exceptions (`CommitConflictException` / `CommitException`) rather than surfacing `PreparationException` / `ValidationException`, since those phases are internal to it. `registerParticipant` is documented as idempotent per participant ID (registering an already-registered `getId()` is a no-op);
  - the `Participant` role (`getId` / `join` / CRUD / `prepareRecords` / `validateRecords` / `commitRecords` / `rollbackRecords` / `close`);
  - the `WriteSetEntry` type carrying a participant's per-record write set, and the `PreparationResult` envelope returned by `prepareRecords` (a dedicated return type so the prepare phase can report additional information in the future without changing the method signature).
- The Javadoc is written at the contract level — it documents required behavior and the transaction lifecycle (out-of-order operations throw `IllegalStateException`), not a specific implementation's exact steps.

## Checklist

- `TwoPhaseCommit` is an internal interface: it is not invoked by application code directly, breaking changes are expected, and users should not depend on it. This PR adds the interface only; there is no behavior to unit-test on its own (the implementation and its tests land in #3662).

## Release notes

N/A


core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
```diff
@@ -0,0 +1,627 @@
...
+ *       single participant may front one or multiple data stores (for example, when using
+ *       multi-storage).
+ * </ul>
+ *
+ * <p>This interface is identifier-based: every operation takes the transaction ID as an argument
+ * rather than relying on implicit per-connection or per-session state, so callers may distribute
+ * work across processes or resume after a restart. The implementations are not stateless, however —
+ * both roles hold per-transaction state internally, keyed by the transaction ID.
+ *
+ * <p>One-phase commit (an optimization that skips the PREPARE phase) is intentionally out of scope
```
feeblefakie, line 52: Can you elaborate on the `intentionally out of scope for this interface` part?
For using the one-phase commit optimization, do we expect users to use a different interface?

feeblefakie (approved):
LGTM! But I left one question. PTAL!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
brfrn169: Sure. This `TwoPhaseCommit` interface assumes a transaction that spans **multiple participants** — the `Coordinator` orchestrates prepare/validate/commit across participants that may eventually live in separate services.

One-phase commit is an optimization for a transaction **confined to a single participant**: all of its mutations target one participant and can be grouped into a single atomic storage mutation, so there is nothing to coordinate across participants and the whole PREPARE/validate/commit-state machinery can be skipped. Since it is by definition a single-participant shortcut, it falls outside this multi-participant orchestration contract — that's what "out of scope for this interface" means.

So no different interface for users. Single-participant transactions keep going through the existing local Consensus Commit execution path as before; this interface just doesn't cover that case.

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
brfrn169, line 1: This is a comment to block this PR from being merged.

PR #3662, opened 2026-06-27
Implement the TwoPhaseCommit Coordinator and Participant for Consensus Commit

brfrn169:
## Description

This PR implements the consensus-commit-backed `TwoPhaseCommit.Coordinator` and `TwoPhaseCommit.Participant`, wires them into the transaction providers, and adds a single-phase facade so existing `DistributedTransaction(Manager)` callers keep working unchanged.

## Related issues and/or PRs

- `TwoPhaseCommit.Coordinator` and `TwoPhaseCommit.Participant` are introduced in #3660.
- Depends on #3663, which widens `DistributedTransactionManager.join` to throw `TransactionException` (the facade's `join` propagates registration failures through it).

## Changes made

- Add `participant_id` to the `WriteSet` persistence proto and to `ConsensusCommitConfig`.
- Add a `WriteSetEncoder` entry-point for encoding aggregated `WriteSetEntry`s.
- Add `ConsensusCommitCoordinator` implementing the lifecycle role (`begin` / `registerParticipant` / `commit` / `rollback`). `commit` collapses prepare/validate failures into commit-level exceptions (`CommitConflictException` / `CommitException`), mirroring `CommitHandler`. `registerParticipant` is idempotent per participant ID: registering a participant whose `getId()` is already registered is a no-op (it is not joined again), since `commit` keys each participant's write set by its ID.
- Add `ConsensusCommitParticipant` implementing the CRUD role (`join` + CRUD + `prepareRecords` / `validateRecords` / `commitRecords` / `rollbackRecords`), with a strict per-transaction lifecycle state machine. `getScanner` returns a scanner whose `one()` / `all()` / `iterator()` / `close()` are serialized on the per-transaction lock — the same monitor a context release (commit / rollback) holds when it closes the scanners — so iteration of the not-thread-safe underlying scanner cannot race a concurrent release; an iteration that loses that race surfaces a retriable `CrudConflictException`.
- Honor a per-transaction isolation level supplied via the `cc-transaction-isolation` attribute at `join` (falling back to the configured default), matching `ConsensusCommitManager`.
- Wire the `TwoPhaseCommit.Coordinator` / `Participant` factories into the transaction providers and expose them through `ProviderManager` and `TransactionFactory` (`getTwoPhaseCommitCoordinator` / `getTwoPhaseCommitParticipant`). `DistributedTransactionProvider#createTwoPhaseCommitCoordinator` / `#createTwoPhaseCommitParticipant` throw `UnsupportedOperationException` when the transaction manager does not support the two-phase commit interface (JDBC and single-CRUD-operation transactions, using `CoreError.JDBC_TRANSACTION_TWO_PHASE_COMMIT_NOT_SUPPORTED` / `SINGLE_CRUD_OPERATION_TRANSACTION_TWO_PHASE_COMMIT_NOT_SUPPORTED`); the `ProviderManager` / `TransactionFactory` accessors propagate that exception.
- Add `TwoPhaseCommitBackedDistributedTransaction` and `TwoPhaseCommitBackedDistributedTransactionManager` as a single-phase facade over the new roles. The by-ID Coordinator-state operations that the `TwoPhaseCommit.Coordinator` interface does not expose — `getState`, `rollback(String)` / `abort(String)`, `finishTransaction`, `recoverRecord`, and `resume` — throw `UnsupportedOperationException`.
- Move `TRANSACTION_NOT_FOUND` to the concurrency error category and reuse it for the Coordinator/Participant transaction-not-found paths (CoreError).

## Checklist

- Unit tests are added/updated for the new components: `ConsensusCommitCoordinatorTest`, `ConsensusCommitParticipantTest`, `WriteSetEncoderTest`, `TwoPhaseCommitBackedDistributedTransactionManagerTest`, `ConsensusCommitConfigTest`, and `SynchronizedScannerTest`.
- The Coordinator's per-participant skip optimization (skipping `validateRecords` / `commitRecords` where unnecessary) is intentionally **not** in this PR; it lands separately in #3664.
- Group commit and one-phase commit are out of scope for the new Coordinator.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedTransactionProvider.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedTransactionProvider.java
@@ -46,4 +46,26 @@ public interface DistributedTransactionProvider {
    */
   @Nullable
   TwoPhaseCommitTransactionManager createTwoPhaseCommitTransactionManager(DatabaseConfig config);
+
+  /**
+   * Creates an instance of {@link TwoPhaseCommit.Coordinator} for the new multi-participant
+   * two-phase commit interface.
+   *
+   * @param config a database config
+   * @return an instance of {@link TwoPhaseCommit.Coordinator}
+   * @throws UnsupportedOperationException if the transaction manager does not support the two-phase
+   *     commit interface
+   */
+  TwoPhaseCommit.Coordinator createTwoPhaseCommitCoordinator(DatabaseConfig config);
+
+  /**
+   * Creates an instance of {@link TwoPhaseCommit.Participant} for the new multi-participant
+   * two-phase commit interface.
+   *
+   * @param config a database config
+   * @return an instance of {@link TwoPhaseCommit.Participant}
+   * @throws UnsupportedOperationException if the transaction manager does not support the two-phase
+   *     commit interface
+   */
+  TwoPhaseCommit.Participant createTwoPhaseCommitParticipant(DatabaseConfig config);
 }
--- a/core/src/main/java/com/scalar/db/common/CoreError.java
+++ b/core/src/main/java/com/scalar/db/common/CoreError.java
@@ -237,13 +237,6 @@ public enum CoreError implements ScalarDbError {
       "",
       ""),
   TRANSACTION_ALREADY_EXISTS(Category.USER_ERROR, "0047", "The transaction already exists", "", ""),
-  TRANSACTION_NOT_FOUND(
-      Category.USER_ERROR,
-      "0048",
-      "A transaction associated with the specified transaction ID is not found. "
-          + "The transaction might have expired",
-      "",
-      ""),
   SYSTEM_NAMESPACE_SPECIFIED(
       Category.USER_ERROR, "0049", "%s is the system namespace name", "", ""),
   NAMESPACE_ALREADY_EXISTS(
@@ -1101,6 +1094,25 @@ public enum CoreError implements ScalarDbError {
       "Recovering a record is not supported in single CRUD operation transactions",
       "",
       ""),
+  CONSENSUS_COMMIT_PARTICIPANT_ID_IS_REQUIRED(
+      Category.USER_ERROR,
+      "0293",
+      "A participant ID is required for the new TwoPhaseCommit.Participant. "
+          + "Set the property: scalar.db.consensus_commit.participant_id",
+      "",
+      ""),
+  JDBC_TRANSACTION_TWO_PHASE_COMMIT_NOT_SUPPORTED(
+      Category.USER_ERROR,
+      "0294",
+      "Two-phase commit is not supported in JDBC transactions",
+      "",
+      ""),
+  SINGLE_CRUD_OPERATION_TRANSACTION_TWO_PHASE_COMMIT_NOT_SUPPORTED(
+      Category.USER_ERROR,
+      "0295",
+      "Two-phase commit is not supported in single CRUD operation transactions",
+      "",
+      ""),
 
   //
   // Errors for the concurrency error category
@@ -1250,6 +1262,13 @@ public enum CoreError implements ScalarDbError {
       Category.CONCURRENCY_ERROR,
       "0030",
       "Resolving an uncommitted record exceeded the retry limit during recovery. Transaction ID: %s",
+      "",
+      ""),
+  TRANSACTION_NOT_FOUND(
+      Category.CONCURRENCY_ERROR,
+      "0031",
+      "A transaction associated with the specified transaction ID is not found. "
+          + "The transaction might have expired",
       "",
       ""),
 
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
@@ -1,0 +1,387 @@
+package com.scalar.db.transaction.consensuscommit;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.scalar.db.api.DistributedStorage;
+import com.scalar.db.api.TwoPhaseCommit;
+import com.scalar.db.api.TwoPhaseCommit.Participant;
+import com.scalar.db.api.TwoPhaseCommit.WriteSetEntry;
+import com.scalar.db.common.CoreError;
+import com.scalar.db.config.DatabaseConfig;
+import com.scalar.db.exception.transaction.CommitConflictException;
+import com.scalar.db.exception.transaction.CommitException;
+import com.scalar.db.exception.transaction.PreparationConflictException;
+import com.scalar.db.exception.transaction.PreparationException;
+import com.scalar.db.exception.transaction.TransactionException;
+import com.scalar.db.exception.transaction.TransactionNotFoundException;
+import com.scalar.db.exception.transaction.UnknownTransactionStatusException;
+import com.scalar.db.exception.transaction.ValidationConflictException;
+import com.scalar.db.exception.transaction.ValidationException;
+import com.scalar.db.service.StorageFactory;
+import com.scalar.db.transaction.consensuscommit.proto.v1.WriteSet;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Consensus-commit-backed {@link TwoPhaseCommit.Coordinator} implementation.
+ *
+ * <p>Drives the two-phase commit protocol across the participants registered for a transaction
+ * (prepare -&gt; validate -&gt; write COMMITTED state -&gt; commit records, with the abort/rollback
+ * path on failure) and holds the per-transaction state in an in-memory map keyed by canonical
+ * transaction ID. Each entry records the registered participants and the read-only flag and
+ * attributes passed at {@code begin}. The map is cleaned up after every {@code commit} or {@code
+ * rollback}.
+ *
+ * <p>TODO: support group commit on the new Coordinator (tracked separately from this change). This
+ * is a planned extension, not a design limitation: the {@code WriteSet} persistence proto's {@code
+ * child_id} (group-commit dimension) and {@code participant_id} (two-phase-commit dimension) are
+ * orthogonal and already compose, and {@link CoordinatorStateAccessor#getState(String)} already
+ * resolves full (parent + child) transaction IDs. The remaining work is reserving a group-commit
+ * slot in {@code begin} (so the transaction ID becomes a full key) and routing the COMMITTED-state
+ * write through the group-commit variant of the Coordinator-side handler.
+ */
+@ThreadSafe
+public class ConsensusCommitCoordinator implements TwoPhaseCommit.Coordinator {
+  private static final Logger logger = LoggerFactory.getLogger(ConsensusCommitCoordinator.class);
+
+  private final DistributedStorage storage;
+  private final CoordinatorCommitHandler coordinatorCommitHandler;
+  private final boolean coordinatorWriteOmissionOnReadOnlyEnabled;
+
+  private final ConcurrentMap<String, CoordinatorContext> contexts = new ConcurrentHashMap<>();
+
+  public ConsensusCommitCoordinator(DatabaseConfig databaseConfig) {
+    ConsensusCommitConfig config = new ConsensusCommitConfig(databaseConfig);
+    throwIfGroupCommitIsEnabled(config);
+    StorageFactory storageFactory = StorageFactory.create(databaseConfig.getProperties());
+    this.storage = storageFactory.getStorage();
+    this.coordinatorCommitHandler =
+        new CoordinatorCommitHandler(new CoordinatorStateAccessor(storage, config));
+    this.coordinatorWriteOmissionOnReadOnlyEnabled =
+        config.isCoordinatorWriteOmissionOnReadOnlyEnabled();
+  }
+
+  @VisibleForTesting
+  @SuppressFBWarnings("EI_EXPOSE_REP2")
+  ConsensusCommitCoordinator(
+      CoordinatorCommitHandler coordinatorCommitHandler, ConsensusCommitConfig config) {
+    throwIfGroupCommitIsEnabled(config);
+    this.storage = null;
+    this.coordinatorCommitHandler = checkNotNull(coordinatorCommitHandler);
+    this.coordinatorWriteOmissionOnReadOnlyEnabled =
+        config.isCoordinatorWriteOmissionOnReadOnlyEnabled();
+  }
+
+  // TODO: Remove this guard once group commit is supported on the new Coordinator (see the class
+  // Javadoc). It is a temporary not-yet-implemented gate, not a permanent restriction.
+  private static void throwIfGroupCommitIsEnabled(ConsensusCommitConfig config) {
+    if (CoordinatorGroupCommitter.isEnabled(config)) {
+      throw new IllegalArgumentException(
+          "Group commit is not yet supported on the new Coordinator implementation");
+    }
+  }
+
+  @Override
+  public String begin(
+      @Nullable String transactionId,
+      boolean readOnly,
+      Map<String, String> attributes,
+      @Nullable Participant participant)
+      throws TransactionException {
+    // Use the caller-supplied ID when present; otherwise generate one. Group commit is not yet
+    // supported on this path, so no parent-prefix logic is needed.
+    String canonical = transactionId != null ? transactionId : UUID.randomUUID().toString();
+    CoordinatorContext existing =
+        contexts.putIfAbsent(canonical, new CoordinatorContext(canonical, readOnly, attributes));
+    if (existing != null) {
+      throw new TransactionException(
+          CoreError.TRANSACTION_ALREADY_EXISTS.buildMessage(), canonical);
+    }
+    if (participant != null) {
+      // Register the optional participant exactly as registerParticipant would. If join fails, drop
+      // the just-created context so a failed begin leaves no orphaned state.
+      try {
+        registerParticipant(canonical, participant);
+      } catch (TransactionException e) {
+        releaseResources(canonical);
+        throw e;
+      }
+    }
+    return canonical;
+  }
+
+  @Override
+  public void registerParticipant(String transactionId, Participant participant)
+      throws TransactionException {
+    CoordinatorContext context = getContext(transactionId);
+    synchronized (context) {
+      context.checkActive();
+      // Registration is idempotent per participant ID: if a participant with the same getId() is
+      // already registered, this is a no-op (the participant is not joined again). commit() keys
+      // each participant's write set by getId(), so registering the same ID twice would otherwise
+      // merge into one EntryGroup and misattribute the persisted records.
+      for (Participant registered : context.participants) {
+        if (registered.getId().equals(participant.getId())) {
+          return;
+        }
+      }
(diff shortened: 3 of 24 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
```diff
@@ -0,0 +1,268 @@
...
+ * write through the group-commit variant of the Coordinator-side handler.
+ */
+@ThreadSafe
+public class ConsensusCommitCoordinator implements TwoPhaseCommit.Coordinator {
+  private static final Logger logger = LoggerFactory.getLogger(ConsensusCommitCoordinator.class);
+
+  private final DistributedStorage storage;
+  private final CoordinatorCommitHandler coordinatorCommitHandler;
+
+  private final ConcurrentMap<String, CoordinatorContext> contexts = new ConcurrentHashMap<>();
```
brfrn169, line 64: Follow-up (not in this PR): There's no timeout-based eviction here yet. Per the `TwoPhaseCommit` contract, a transaction's state may also terminate via a timeout when `Coordinator#commit/#rollback` is never reached, but this implementation removes entries only on explicit `commit`/`rollback`. A caller crash or a `Coordinator` restart therefore leaks the entry for the lifetime of the process. I plan to add a periodic expiry sweep — writing ABORTED for already-prepared transactions before eviction — in a follow-up PR.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitParticipant.java
```diff
@@ -0,0 +1,538 @@
...
+  private final DistributedStorage storage;
+  private final DistributedStorageAdmin admin;
+  private final TransactionTableMetadataManager tableMetadataManager;
+  private final ParallelExecutor parallelExecutor;
+  private final RecoveryExecutor recoveryExecutor;
+  private final CrudHandler crud;
+  private final ParticipantCommitHandler commit;
+  private final ConsensusCommitOperationChecker operationChecker;
+
+  private final ConcurrentMap<String, TransactionContext> contexts = new ConcurrentHashMap<>();
```
brfrn169: Follow-up (not in this PR): No timeout-based eviction yet. The `TwoPhaseCommit.Participant` contract (see the join Javadoc) allows this context to terminate via a timeout when the `Coordinator` never drives `commitRecords`/`rollbackRecords`, but here entries are removed only on those explicit terminal calls. A `Coordinator` crash, a participant restart, or a missed terminal call therefore leaks the `TransactionContext` — including its `Snapshot` (read/write sets) and any open scanners — for the lifetime of the process. A periodic expiry sweep (closing scanners and rolling back records for already-prepared transactions before eviction) is planned as a follow-up.

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/DistributedTransactionProvider.java
brfrn169, line 1: This is a comment to block this PR from being merged.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
```diff
@@ -0,0 +1,387 @@
...
+              hasWrites = true;
+            }
+            writeSetsByParticipant
+                .computeIfAbsent(participant.getId(), k -> new ArrayList<>())
+                .addAll(entries);
+          }
+          for (Participant participant : participants) {
+            participant.validateRecords(transactionId);
+          }
+        } catch (PreparationException | ValidationException e) {
```
feeblefakie, line 184: In ConsensusCommitCoordinator.commit(), the prepare/validate block (lines 169–201) only catches the checked PreparationException | ValidationException and TransactionNotFoundException. But Participant.prepareRecords/validateRecords can throw unchecked exceptions that escape this block; most reachably IllegalStateException(CONSENSUS_COMMIT_SCANNER_NOT_CLOSED) when the app calls commit() with an open scanner, and IllegalStateException(TRANSACTION_NOT_ACTIVE) from checkActive(), plus any RuntimeException from validateRecords/toSerializable.

When that happens, the exception skips abortAndRollbackRecords(...) and hits the finally (223–226), which runs releaseResources(),  and that only removes the Coordinator's context (contexts.remove(transactionId), line 335). 

Consequences:
  1. Participant context leak (permanent). A ParticipantContext is removed from ConsensusCommitParticipant.contexts only in commitRecords (440) / rollbackRecords (464). Neither runs here, and there's no idle-expiry sweep on that map, so the context
  (holding a Snapshot) leaks for the lifetime of the manager. Repeated occurrences grow it unboundedly.
  2. Peer participants' PREPARED records are orphaned. In a multi-participant commit, if participant A already prepared and participant B throws, A's PREPARED records are never actively rolled back — they stay locked until the expiration + lazy-recovery
  path aborts them, even though the transaction provably never committed (commitState was never reached).
  3. Wrong exception type to the caller — a raw IllegalStateException surfaces instead of CommitException.

How about broadening the catch around the prepare/validate block to also handle RuntimeException (or Exception), route it through abortAndRollbackRecords(transactionId, participants, hasWrites), and rethrow wrapped as CommitException. Aborting is always correct there since the transaction cannot have committed.

There's also a test gap: ConsensusCommitCoordinatorTest covers the checked prepare/validate failures but has no case for a participant throwing an unchecked exception (e.g. an unclosed-scanner IllegalStateException) from prepare/validate.

feeblefakie (commented):
Overall, looking good! 
Left one comment. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: Thanks—the analysis is accurate.

These unchecked exceptions (unclosed scanner, inactive transaction, etc.) are application programming errors. They should be rare and are the caller’s responsibility, so I’d rather surface the raw `IllegalStateException` and fail fast than wrap it as a `CommitException`, which could mask the bug. I see this as a robustness/eager-cleanup trade-off, not a correctness issue.

So I’m leaning toward keeping it as is. What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitParticipant.java
```diff
@@ -0,0 +1,786 @@
...
+        }
+      }
+      entries.add(
+          new WriteSetEntryImpl(
+              TwoPhaseCommit.WriteSetEntry.Type.WRITE,
+              put.forNamespace().get(),
+              put.forTable().get(),
+              put.getPartitionKey(),
+              put.getClusteringKey(),
+              Collections.unmodifiableList(filteredColumns)));
```
feeblefakie, line 521: I might miss something, but there is a case where `filteredColumns` are not required, so always creating filteredColumns is inefficient?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
```diff
@@ -0,0 +1,387 @@
...
+        // Only the Coordinator state writes (commitState/abortState) are gated on this; the
+        // record-level steps (prepareRecords / commitRecords / rollbackRecords) are still driven on
+        // every participant so their local contexts are released.
+        boolean hasWrites = false;
+
+        long preparedAt = System.currentTimeMillis();
+        try {
+          for (Participant participant : participants) {
+            List<WriteSetEntry> entries =
+                participant.prepareRecords(transactionId, preparedAt).getWriteSet();
```
feeblefakie, line 173: I might miss something, but `hasWrites` is only set from a returned write set, so if the first writer's prepareRecords throws after a partial prepare, hasWrites stays false and abortAndRollbackRecords(..., false) skips abortState() (omission on by default).
CommitHandler derives this up-front from the snapshot and always writes ABORTED on prepare failure; so this diverges, and the behavior is even participant-order-dependent (an earlier successful writer would flip hasWrites and mask it).
 
It's not a correctness bug (an absent row is treated as ABORTED by lazy recovery), but if the best-effort rollbackRecords also fails, the partially-prepared record is stranded PREPARED with no coordinator row and holds its lock until TRANSACTION_LIFETIME_MILLIS (~15s) instead of being aborted on the next read.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitParticipant.java
```diff
@@ -0,0 +1,786 @@
...
+  private final DistributedStorage storage;
+  private final DistributedStorageAdmin admin;
+  private final TransactionTableMetadataManager tableMetadataManager;
+  private final ParallelExecutor parallelExecutor;
+  private final RecoveryExecutor recoveryExecutor;
+  private final CrudHandler crud;
+  private final ParticipantCommitHandler commit;
+  private final ConsensusCommitOperationChecker operationChecker;
+
+  private final ConcurrentMap<String, ParticipantContext> contexts = new ConcurrentHashMap<>();
```
feeblefakie, line 87: It maps add an entry at join/begin and remove it only at a terminal step (commitRecords/rollbackRecords, commit/rollback). There's no timeout, sweeper, or eviction anywhere, yet the TwoPhaseCommit Javadoc repeatedly promises one — "reclaimed by the participant's own timeout", "discarded after … a timeout"  — and ~15 methods document TransactionNotFoundException … (e.g., it expired), a state nothing can actually produce (the Status enum has no EXPIRED). So a transaction that's begun but never committed/rolled back (orchestrator crash/abandon) leaks its context — and its Snapshot — permanently, and these raw roles aren't wrapped by the active-transaction decorators that evict the manager path.

The timeout feature is coming soon?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
```diff
@@ -0,0 +1,387 @@
...
+        } catch (TransactionNotFoundException e) {
+          // A participant no longer knows this transaction (its local context is gone, e.g. it
+          // expired). The transaction cannot commit: abort it and roll back the records already
+          // PREPARED on the other participants, then surface the TransactionNotFoundException as-is
+          // (the facade maps it to a retriable conflict).
+          abortAndRollbackRecords(transactionId, participants, hasWrites);
+          throw e;
+        }
+        long committedAt;
+        if (hasWrites || !coordinatorWriteOmissionOnReadOnlyEnabled) {
```
feeblefakie, line 204: This predicate occurs two times in this file, and two times in CommitHandler, so it might be better to unify them as a single helper method.

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -1252,6 +1264,13 @@ public enum CoreError implements ScalarDbError {
       "Resolving an uncommitted record exceeded the retry limit during recovery. Transaction ID: %s",
       "",
       ""),
+  TRANSACTION_NOT_FOUND(
+      Category.CONCURRENCY_ERROR,
```
feeblefakie, line 1268: Changing the category and number is as expected?
Is it a backward incompatible change?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
```diff
@@ -0,0 +1,387 @@
...
+      context.checkActive();
+      // Registration is idempotent per participant ID: if a participant with the same getId() is
+      // already registered, this is a no-op (the participant is not joined again). commit() keys
+      // each participant's write set by getId(), so registering the same ID twice would otherwise
+      // merge into one EntryGroup and misattribute the persisted records.
+      for (Participant registered : context.participants) {
+        if (registered.getId().equals(participant.getId())) {
+          return;
+        }
+      }
```
feeblefakie, line 140: It assumed every participant has a unique ID, but is there a way to validate or force it?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
feeblefakie: Agreed on surfacing the raw IllegalStateException; failing fast on a programming error is the right call.
But that's orthogonal to cleanup. 
Right now the unchecked path seems to skip abortAndRollbackRecords, and because the finally already drops the coordinator context, a later rollback() no-ops. So, the participant context (and its Snapshot) leaks permanently, unrecoverably, and in multi-participant, any peer's PREPARED records stay locked until expiry. 

WDYT?

core/src/main/java/com/scalar/db/common/CoreError.java
brfrn169: I don't think it's a backward-incompatible change.

In ScalarDB, error codes are identifiers for diagnosing and looking up an error — they're not meant to be used for programmatic error handling (callers branch on the exception type/category, never on the code string). So changing the code/category doesn't break any working caller code. Moving `TRANSACTION_NOT_FOUND` to `CONCURRENCY_ERROR` just makes the identifier reflect the error's nature more accurately: a not-found transaction is a transient condition that the API surfaces as a retriable conflict.

What do you think?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: Thanks — sounds like we're aligned that surfacing the raw `IllegalStateException` (fail-fast, no wrapping) is the right call, so let's consider that part settled. On cleanup: the eviction is intentionally deferred to the follow-up #3669, which adds the `ActiveTransactionManaged{Coordinator,Participant}` decorators. Their idle-expiry reaper calls `releaseContext` (the registry's disposal handler is `tracked -> participant.releaseContext(tracked.transactionId)`), so an abandoned/idle context — including a participant context orphaned after an unchecked-exception `commit()` — is reclaimed once it goes idle, bounded by `expirationTimeMillis` rather than leaking permanently. In production the raw roles are always wrapped by these decorators (`AbstractDistributedTransactionProvider#createTwoPhaseCommitCoordinator/Participant`); the raw roles in this PR alone have no eviction, which is why it looks unbounded here. So no data loss (lazy recovery reconciles PREPARED records) and no permanent context leak once #3669 lands. Does that address it?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: Good catch. You're right that computing `hasWrites` incrementally makes the abort path order-dependent, and that a participant throwing part-way through `prepareRecords` can leave records PREPARED that `hasWrites` never observed. I've made the abort path always write ABORTED (dropped the write-omission there), which removes the order dependence in a072007aa74076988c5082a24f9ef94f67db6354. This is actually a bit more conservative than `CommitHandler`: it can safely gate its abort on `hasWritesOrDeletesInSnapshot` because that's derived accurately up-front from the snapshot, whereas the Coordinator only learns each participant's write set incrementally during prepare, so it can't compute an accurate up-front flag and instead always writes ABORTED on abort. The write-omission still applies on the commit-*success* path, where `hasWrites` accurately reflects every participant's returned write set, so read-only transactions still skip the COMMITTED row.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: After the abort-path fix in a072007aa74076988c5082a24f9ef94f67db6354, this predicate is now used in only one place in the Coordinator (the commit-success gate) — the abort path no longer uses it — so there's no duplication left to extract here.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: We could make registering a duplicate ID throw an error, but we've intentionally not done that. The design assumes every participant has a unique ID, and registering the same ID again is treated as an idempotent no-op (the participant isn't joined twice). We prefer keeping it a no-op because it's more convenient for future implementation — an idempotent registration is more forgiving for callers that may end up re-registering the same participant, so we'd rather not lock ourselves into treating that as an error.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitParticipant.java
brfrn169: You're right — and in fact it's always wasted in this PR, not just in some cases: the Coordinator currently encodes the commit state with `includeColumns=false`, and the entry columns are only read when that flag is true, which happens once write-set logging (a later, opt-in change) is wired in. The participant doesn't see that flag today (it's decided Coordinator-side), so the clean optimization is to thread a "write-set-logging enabled" flag down to the participant and skip building the filtered columns when it's off — I'd rather do that together with the write-set-logging change that actually needs the columns, and keep this PR as-is. I'll leave a note so we pick it up there.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitParticipant.java
brfrn169: Yes — the timeout/eviction is the follow-up #3669. The `ActiveTransactionManaged{Coordinator,Participant}` decorators track each transaction in an `ActiveTransactionRegistry` and, on idle expiry, call `releaseContext` to discard the context (closing scanners / dropping the Snapshot). That idle-reap is exactly what "expired" refers to in the Javadoc: the context is reclaimed on timeout, after which a later call finds no context and throws `TransactionNotFoundException`. It's a lifecycle event, not a persistent status — so there's intentionally no `EXPIRED` in the Status enum. The interface Javadoc describes that contract; the raw role in this PR just doesn't fulfill the timeout part on its own (the #3669 decorator does), which is why it looks unbacked when viewed here alone. So I don't think the wording needs changing — but I'm happy to add a note that the timeout is provided by the active-transaction-management layer if you think it'd help.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
feeblefakie: Yes, I'm not concerned about data loss.
I was wondering whether the prepared records should be aborted immediately rather than lazily recovered.
I might miss something, though.

core/src/main/java/com/scalar/db/common/CoreError.java
feeblefakie: OK, got it. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
feeblefakie: I might miss something, but isn't it error-prone?
Assume there are two participants, A (id=p1) and B (id=p1), and A is registered. Then, if B is registered, is B silently not registered?

brfrn169: After discussing this, we concluded that the duplicate check should be done by the caller of this API rather than by the Coordinator. Thanks for the discussion!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: We deliberately leave the prepared records to lazy recovery here rather than aborting them eagerly, because this path is only reached on an unchecked exception — i.e. a programming error, not an expected failure. This is actually consistent with `CommitHandler`: its `catch (Exception e)` around the prepare/validate steps also just rethrows without writing ABORTED or rolling back, and only the checked `PreparationException` / `ValidationException` trigger an eager abort. As an overall design stance we don't try to handle application bugs defensively — the records are still reconciled (lazy recovery / expiry), so there's no data loss, just deferred cleanup on a should-never-happen path. If we ever did want eager abort here, it'd be a holistic change across both this Coordinator and `CommitHandler`. Thanks!

feeblefakie (approved):
LGTM! Thank you! Great work!

PR #3664, opened 2026-06-28
Skip unnecessary validate and commit steps per participant in the TwoPhaseCommit Coordinator

brfrn169:
## Description

The Consensus Commit `TwoPhaseCommit.Coordinator` introduced in #3662 drove `validateRecords` and `commitRecords` on every registered participant unconditionally. This PR makes the Coordinator drive only the record-level steps each participant actually needs — mirroring the single-phase `CommitHandler` (minus the one-phase fast path) — so it skips work that would have nothing to do. For remote deployments, where each participant call is an RPC, this avoids unnecessary round trips.

## Related issues and/or PRs

N/A

## Changes made

- Added `TwoPhaseCommit.PreparationResult#isValidationRequired()` and `#isCommitRequired()` so the prepare response tells the Coordinator which record-level steps a participant still needs. The participant derives both from a single source — `snapshot.hasWritesOrDeletes()` for the commit requirement — so the Coordinator's skip decision and the participant's own early context release stay in lockstep.
- The Coordinator now drives `validateRecords` only where the prepare result reports it required, and `commitRecords` only where the prepare result reports it required. (It still reads the write set itself only to decide whether to write the COMMITTED Coordinator state row.)
- A participant whose later steps are all skipped releases its own context at its last driven step: `validateRecords` for a write-less participant that requires validation, or `prepareRecords` when validation is not required either. `prepareRecords` itself can never be skipped because its result must be returned.
- `Participant#rollbackRecords` is now a lenient no-op for an unknown or already-released transaction, so the abort path can roll back every participant uniformly without tracking which ones released early.
- Added and updated unit tests for the per-participant skip behavior and the early context release.

## Checklist

- `TwoPhaseCommit` is an internal interface (breaking changes are expected; users should not depend on it). This PR is an internal optimization with no user-facing API change.
- One-phase commit and group commit remain out of scope.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
+++ b/core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
@@ -223,9 +223,13 @@ public interface TwoPhaseCommit {
    * record-level two-phase commit steps.
    *
    * <p>Each participant maintains a local context (snapshot, read set, write set) keyed by
-   * transaction ID. The context is created by {@link #join} and discarded after the transaction
-   * terminates—via {@link #commitRecords}, {@link #rollbackRecords}, or a timeout if {@link
-   * Coordinator#commit} or {@link Coordinator#rollback} is never reached.
+   * transaction ID. The context is created by {@link #join} and discarded once the participant has
+   * run the last step the Coordinator drives for it. For a participant with writes that is the
+   * commit (via {@link #commitRecords}) or the abort (via {@link #rollbackRecords}); for a
+   * write-less participant the Coordinator may skip those steps, so the context is discarded
+   * earlier — at {@link #validateRecords}, or at {@link #prepareRecords} when validation is not
+   * required either. It is also discarded on a timeout if {@link Coordinator#commit} or {@link
+   * Coordinator#rollback} is never reached.
    *
    * <p>The methods on this interface have two different intended callers:
    *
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitParticipant.java
```diff
@@ -410,7 +413,23 @@ public TwoPhaseCommit.PreparationResult prepareRecords(String transactionId, lon
...
         pc.toPrepared();
       }
-      return new PreparationResultImpl(buildWriteSetEntries(context));
+      List<TwoPhaseCommit.WriteSetEntry> writeSet = buildWriteSetEntries(context);
+      boolean validationRequired = context.isValidationRequired();
+      // commitRecords is required iff this participant has writes to commit. Derive it from the
+      // same snapshot predicate validateRecords uses for its self-release, so the commit-step skip
+      // and the validate-step self-release stay in lockstep on a single source. This drives the
+      // Coordinator's commit-step skip via PreparationResult#isCommitRequired.
+      boolean commitRequired = context.snapshot.hasWritesOrDeletes();
```
komamitsu: How about adding `Context.isCommitRequired()`?

brfrn169: Fixed in 33c04370b57a41c48b532582357818647bda110d. Thanks!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
brfrn169, line 1: This is a comment to block this PR from being merged.

feeblefakie (approved):
LGTM! Thank you!

PR #3668, opened 2026-06-30
Add releaseContext to TwoPhaseCommit for no-storage context release

brfrn169:
## Description

The forthcoming per-role context-reaper decorators need a way to reclaim an abandoned or idle transaction's role-local state **without** mutating storage — a plain rollback would be both unnecessary (the durable Coordinator state is the source of truth) and, for a PREPARED participant, wrong. This PR adds that primitive: a reap-only terminal `releaseContext(transactionId)` on both `TwoPhaseCommit` roles.

## Related issues and/or PRs

- `TwoPhaseCommit.Coordinator` / `TwoPhaseCommit.Participant` are introduced in #3660 and implemented in #3662.
- Depends on #3667 (move `AbstractDistributedTransactionProvider` to `common`).

## Changes made

- Add `releaseContext(String transactionId)` to `TwoPhaseCommit.Coordinator` and `TwoPhaseCommit.Participant`, implemented natively in `ConsensusCommitCoordinator` / `ConsensusCommitParticipant`.
- It releases only the role's in-memory state: the Coordinator drops its own context (it does **not** drive the participants), and the Participant closes its scanners and discards its snapshot — with no storage mutation. A PREPARED participant is **not** rolled back; its records are reconciled by lazy recovery via the Coordinator state row.
- An unknown or already-released transaction is a lenient no-op, mirroring `rollbackRecords` / `rollback`.
- `commit()` is unchanged; no caller is added in this PR.

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

- `TwoPhaseCommit` is an internal interface (breaking changes are expected; users should not depend on it). Adding `releaseContext` is an internal API addition with no user-facing surface.
- Unit tests for the native `releaseContext` are added to `ConsensusCommitCoordinatorTest` / `ConsensusCommitParticipantTest`.

## Release notes

N/A

PR #3669, opened 2026-06-30
Add core TwoPhaseCommit role decorators (attribute propagation and active transaction management)

brfrn169:
## Description

The legacy manager-shaped transaction API gets cross-cutting behavior — transaction-scoped attribute propagation and active-transaction (idle) management — from core-side decorators. This PR adds the equivalent decorators for the new id-based `TwoPhaseCommit` Coordinator/Participant roles, their forwarding bases, and the provider wiring that applies them, so deployments built on these roles get the same behavior without re-implementing it per role.

## Related issues and/or PRs

- `TwoPhaseCommit.Coordinator` / `TwoPhaseCommit.Participant` are introduced in #3660 and implemented in #3662.
- Builds on the reap-only `releaseContext` primitive added in #3668.

## Changes made

- **Forwarding bases** `DecoratedTwoPhaseCommitParticipant` / `DecoratedTwoPhaseCommitCoordinator`: role decorators override only the methods they need (id-based counterparts of the manager-API `DecoratedDistributedTransactionManager`).
- **`OperationAttributeMerger`**: the per-operation-type attribute-merge logic extracted from `AttributePropagatingDistributedTransaction` (operation value wins on key clash), plus a `mergeEach` helper for list merging that allocates a new list only when some element actually needs merging. Both the legacy manager-shaped decorator and the new participant decorator route through it (single implementation; legacy behavior unchanged, guarded by the existing tests).
- **`AttributePropagatingTwoPhaseCommitParticipant`**: captures the begin-attributes at `join` (keyed by transaction id) and merges them into every CRUD operation before delegating, so it sits outside any decorator that reads operation attributes (e.g. ABAC). Captured attributes are dropped on the transaction's terminal step (`commitRecords` / `rollbackRecords` / `releaseContext`).
- **`ActiveTransactionManagedTwoPhaseCommitCoordinator` / `...Participant`** (with `ActiveTransactionRegistry#touch`): track each transaction from `begin`/`join` until its terminal step and, on idle expiry, call `releaseContext` on the wrapped role (no storage rollback) instead of rollback, so records are recovered lazily. The participant reaps on true idle time (every CRUD/record step refreshes the timer); the coordinator measures idle from the last coordinator-observed call (`begin`/`registerParticipant`), since it does not observe CRUD against participants. Both are intended to be the outermost decorator so the reap traverses the inner decorators via `releaseContext`.
- Generalized `ActiveTransactionRegistry`'s expiration callback (`TransactionRollback#rollback` → `ExpirationHandler#onExpired`) so it serves any disposal action, not just rollback; the existing `ActiveTransactionManaged{DistributedTransaction,TwoPhaseCommitTransaction}Manager` are updated to the renamed handler with no behavior change.
- **`AbstractDistributedTransactionProvider` wiring**: `createTwoPhaseCommitCoordinator` wraps with the active-transaction-management decorator when enabled; `createTwoPhaseCommitParticipant` wraps with attribute propagation (inner) then active transaction management (outermost) when enabled, mirroring `createDistributedTransactionManager`. Subclasses provide the raw role instances via the new `createRawTwoPhaseCommit{Coordinator,Participant}` abstracts. When a raw factory returns `null` (a transaction manager that does not support the two-phase commit interface, e.g. JDBC or single-CRUD-operation), the wrapper short-circuits and returns `null` before any wrapping, mirroring `createTwoPhaseCommitTransactionManager`. The coordinator gets no attribute-propagation wrapping (no CRUD), and neither role gets state-management wrapping (handled natively).

## Checklist

- These decorators are generic over the `TwoPhaseCommit` interface; their `@ThreadSafe` guarantee relies on the wrapped role serializing its own per-transaction work (as `ConsensusCommit{Coordinator,Participant}` do on a per-context monitor), since the reaper may call `releaseContext` concurrently with an in-flight call for the same transaction id.
- Unit tests are added/updated for the forwarding bases, the merger (incl. `mergeEach`), both attribute-propagation decorators, both active-transaction-management reapers, the provider wiring (including the unsupported-manager null short-circuit), and the `ActiveExpiringMap` refresh-vs-eviction contract.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
+++ b/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
@@ -70,10 +70,6 @@ public abstract class AbstractDistributedTransactionProvider
   public TwoPhaseCommit.Coordinator createTwoPhaseCommitCoordinator(DatabaseConfig config) {
     TwoPhaseCommit.Coordinator coordinator = createRawTwoPhaseCommitCoordinator(config);
 
-    if (coordinator == null) {
-      return null;
-    }
-
     if (config.isActiveTransactionManagementEnabled()) {
       // Wrap the coordinator for active transaction management. This must be the outermost wrapping
       // so that the idle-expiry reap traverses every inner decorator via releaseContext.
@@ -91,10 +87,6 @@ public abstract class AbstractDistributedTransactionProvider
   @Override
   public TwoPhaseCommit.Participant createTwoPhaseCommitParticipant(DatabaseConfig config) {
     TwoPhaseCommit.Participant participant = createRawTwoPhaseCommitParticipant(config);
-
-    if (participant == null) {
-      return null;
-    }
 
     if (config.isAttributePropagationEnabled()) {
       // Wrap the participant for transaction-scoped attribute propagation
--- a/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
+++ b/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
@@ -11,7 +11,8 @@ public abstract class AbstractDistributedTransactionProvider
     implements DistributedTransactionProvider {
 
   @Override
-  public DistributedTransactionManager createDistributedTransactionManager(DatabaseConfig config) {
+  public final DistributedTransactionManager createDistributedTransactionManager(
+      DatabaseConfig config) {
     DistributedTransactionManager transactionManager =
         createRawDistributedTransactionManager(config);
 
@@ -41,7 +42,7 @@ public abstract class AbstractDistributedTransactionProvider
 
   @Nullable
   @Override
-  public TwoPhaseCommitTransactionManager createTwoPhaseCommitTransactionManager(
+  public final TwoPhaseCommitTransactionManager createTwoPhaseCommitTransactionManager(
       DatabaseConfig config) {
     TwoPhaseCommitTransactionManager transactionManager =
         createRawTwoPhaseCommitTransactionManager(config);
@@ -63,11 +64,12 @@ public abstract class AbstractDistributedTransactionProvider
     return transactionManager;
   }
 
+  @Nullable
   protected abstract TwoPhaseCommitTransactionManager createRawTwoPhaseCommitTransactionManager(
       DatabaseConfig config);
 
   @Override
-  public TwoPhaseCommit.Coordinator createTwoPhaseCommitCoordinator(DatabaseConfig config) {
+  public final TwoPhaseCommit.Coordinator createTwoPhaseCommitCoordinator(DatabaseConfig config) {
     TwoPhaseCommit.Coordinator coordinator = createRawTwoPhaseCommitCoordinator(config);
 
     if (config.isActiveTransactionManagementEnabled()) {
@@ -85,7 +87,7 @@ public abstract class AbstractDistributedTransactionProvider
       DatabaseConfig config);
 
   @Override
-  public TwoPhaseCommit.Participant createTwoPhaseCommitParticipant(DatabaseConfig config) {
+  public final TwoPhaseCommit.Participant createTwoPhaseCommitParticipant(DatabaseConfig config) {
     TwoPhaseCommit.Participant participant = createRawTwoPhaseCommitParticipant(config);
 
     if (config.isAttributePropagationEnabled()) {
--- a/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
+++ b/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
@@ -2,6 +2,7 @@ package com.scalar.db.common;
 
 import com.scalar.db.api.DistributedTransactionManager;
 import com.scalar.db.api.DistributedTransactionProvider;
+import com.scalar.db.api.TwoPhaseCommit;
 import com.scalar.db.api.TwoPhaseCommitTransactionManager;
 import com.scalar.db.config.DatabaseConfig;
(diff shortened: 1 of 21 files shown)


komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/common/AttributePropagatingTwoPhaseCommitParticipant.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  @Override
+  public void join(String transactionId, boolean readOnly, Map<String, String> attributes)
+      throws TransactionException {
+    super.join(transactionId, readOnly, attributes);
+    // Capture only after a successful join so a failed join leaves no stale entry. Skip empty
+    // attributes so there is nothing to merge later.
+    if (!attributes.isEmpty()) {
+      transactionAttributes.put(transactionId, ImmutableMap.copyOf(attributes));
```
komamitsu, line 84: Is an entry added here released even if the transaction doesn't have any write operation (== empty write-set)? https://github.com/scalar-labs/scalardb/pull/3664/changes seems to skip `commitRecords()`. Also,
`preparedRecords()` isn't overridden in this class.

brfrn169: Good catch, thank you. You're right — with the write-less skip from #3664, the Coordinator skips `commitRecords` for a write-less participant (including every read-only transaction), so the captured begin-attributes were never released on the normal commit path and leaked until idle reaping or JVM exit. (I assume `preparedRecords` means `prepareRecords`.)

Rather than mirror `ActiveTransactionManagedTwoPhaseCommitParticipant`'s terminal-step tracking, I drop the attributes at `prepareRecords`: they are only needed while CRUD operations are still issued, and `prepareRecords` is that boundary (no CRUD is accepted once prepared), so dropping them there is safe for every transaction — write-less or not — and covers all commit paths without detecting where the terminal step lands. `commitRecords` is therefore no longer a terminal step here (it always follows `prepareRecords`), so I removed its override; `rollbackRecords` / `releaseContext` keep dropping to cover transactions rolled back or reaped before ever preparing.

Fixed in c2e432c3c.

core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
brfrn169, line 1: This is a comment to block this PR from being merged.

feeblefakie:
**`AttributePropagatingTwoPhaseCommitParticipant`: the documented leak mitigation is incomplete and unenforced (minor / doc)**

The class Javadoc says the unbounded `transactionAttributes` map is reaped by enabling active transaction management: *"Enable active transaction management whenever this decorator is used."* Two gaps make that not quite hold:

1. **Independent config flags, no coupling check.** In `AbstractDistributedTransactionProvider`, the `AttributePropagating` wrap is gated on `isAttributePropagationEnabled()` and the `ActiveTransactionManaged` wrap on `isActiveTransactionManagementEnabled()` independently. Both default `true`, but `attribute.propagation.enabled=true` + `active.transaction.management.enabled=false` is a valid combination that nothing warns about — and it leaves the map with no reaper. (`DatabaseConfig.load()` validates cross-partition-scan combinations but not this one.)
2. **Enabling active management isn't sufficient by itself.** `ActiveExpiringMap` only starts its reaper thread when `valueLifetimeMillis > 0` (`ActiveExpiringMap.java:40`), and `active.transaction.management.expiration.time.millis` defaults to `-1`. So in the *default* config the reaper never runs and abandoned entries are never reaped — the mitigation additionally requires a **positive expiration time**.

To be clear on severity: this is **not a new leak beyond existing behavior** — the map leaks in lockstep with the wrapped `ConsensusCommitParticipant`'s own per-transaction context, which also isn't reaped when reaping is off. So this is a documentation-accuracy + config-footgun issue, not a correctness regression.

Suggestion: either (a) amend the Javadoc to state the real precondition (active management **and** a positive expiration time), and/or (b) add a startup WARN in `DatabaseConfig.load()` when attribute propagation is on while reaping is effectively off (disabled or expiration ≤ 0), mirroring the existing cross-partition-scan validation.

feeblefakie:
**`AbstractDistributedTransactionProvider`: `createTwoPhaseCommit{Coordinator,Participant}` are now non-`final` template methods — a retained override silently skips the decorators (minor / SPI hardening)**

This branch converts `createTwoPhaseCommitCoordinator` / `createTwoPhaseCommitParticipant` from subclass-implemented into concrete template methods that wrap the raw role with `ActiveTransactionManaged*` / `AttributePropagating*`, moving the extension point to the new abstract `createRawTwoPhaseCommit{Coordinator,Participant}`.

Because the two template methods are **not `final`**, a subclass that adds `createRaw*` (forced by the new abstract methods — good) but *also* keeps an old `createTwoPhaseCommit*` override will compile cleanly with no `@Override`/runtime signal, and the retained override silently bypasses all decorator wrapping (no active-transaction reaping, no attribute propagation).

Exposure is limited: this SPI is new/unreleased, so there's no external-migration burden yet, and the only in-lockstep consumer is the Cluster edition. Noting it so the choice is deliberate.

Options: (a) make the two template methods `final` so a retained override becomes a compile error — cleanest, but note the manager-side `createDistributedTransactionManager` / `createTwoPhaseCommitTransactionManager` are non-`final`, so this would be an intentional inconsistency; or (b) add Javadoc on the interface/base stating that `createTwoPhaseCommit*` are template methods and `createRaw*` is the sole override point, plus a release note.

feeblefakie (approved):
LGTM! Thank you!
My AI suggested two relatively-minor things, so PTAL just in case.

PR #3673, opened 2026-06-30
Cap active transactions and refresh the idle timer on each operation

brfrn169:
## Description

Active transaction management (the `ActiveTransactionManaged*` decorators) keeps a registry of in-flight transactions and reaps idle ones. This PR closes two gaps in that mechanism:

1. **The registry was unbounded.** A flood of abandoned or leaked transactions — clients that `begin()` but never `commit()`/`rollback()`, which is especially easy to hit over the network — could grow the registry without limit until the idle reaper eventually caught up, risking memory pressure under load.
2. **The idle timer was only refreshed when a transaction was resumed through the manager.** A client that holds a transaction reference directly and performs a long sequence of operations, or slowly consumes a large `Scanner`, could be reaped mid-work even though it is actively making progress, because operations on the held reference did not refresh the timer.

This PR bounds the registry with a size cap and refreshes the idle timer on every operation.

## Related issues and/or PRs

- Stacked on #3669 (`add-core-twophasecommit-decorators`); this PR is based on that branch.

## Changes made

- Add a configurable cap on the number of active transactions: `scalar.db.active_transaction_management.max_active_transactions` (default `10000`). A value of `0` or less disables the cap (unbounded).
- Reimplement `ActiveExpiringMap` as a thin adapter over [Caffeine](https://github.com/ben-manes/caffeine):
  - **Idle expiration** via `expireAfterAccess`, made proactive with a scheduler so idle entries are reaped promptly even without a subsequent access. (`Scheduler.systemScheduler()` requires Java 9+, so a small custom daemon-backed scheduler is used to stay Java 8 compatible.)
  - **Size-based eviction** via `maximumSize`, using Caffeine's recency- and frequency-aware **Window-TinyLFU** policy, so a frequently-touched (active) transaction is preferentially retained over a cold/abandoned one. This replaces the earlier custom eviction and removes its hash-bucket sampling bias.
  - Removal causes are routed to the right handler (EXPIRED → expiration handler, SIZE → eviction handler); explicit `remove()` and replacements invoke neither.
  - Reclamation handlers run on a **dedicated bounded daemon executor** (not `ForkJoinPool.commonPool()`, Caffeine's default) so a disposal that blocks on I/O (rollback / release-context) does not starve unrelated common-pool work.
- `ActiveTransactionRegistry`: use a single `DisposalHandler` (renamed from `ExpirationHandler`) for both idle-expiration and cap-eviction disposal, log which cause triggered disposal, and swallow handler exceptions so the admitting caller is unaffected.
- Refresh the idle timer on every operation performed on a held transaction — CRUD (`get`/`scan`/`getScanner`/`put`/`delete`/`insert`/`upsert`/`update`/`mutate`/`batch`) plus 2PC `prepare()`/`validate()` — in both the legacy `ActiveTransactionManaged{DistributedTransaction,TwoPhaseCommitTransaction}Manager` and the new TwoPhaseCommit `Coordinator`/`Participant` decorators. Terminal operations (`commit`/`rollback`/`abort`) deregister the transaction instead.
- Refresh the idle timer while consuming a `Scanner` via a new `ActiveTransactionRefreshingScanner` (`one()`/`all()`/`iterator()`), so a slowly-consumed scan is not reaped mid-iteration. Iteration is driven through the shared `ScannerIterator`.
- `DatabaseConfig`: add the new setting, a static accessor, and a getter.
- Wire the cap value through `AbstractDistributedTransactionProvider` to all `ActiveTransactionManaged*` constructors.
- Tests: deterministic idle-expiration/eviction assertions via an injectable `Ticker` and a same-thread executor; per-operation idle-timer refresh in both managers; scanner refresh; disposal-handler-throws still evicts; and the new config default/override.

## Checklist

- This PR is stacked on #3669 (`add-core-twophasecommit-decorators`); the base branch is set accordingly. Please review/merge that first.
- The legacy `ActiveTransactionManaged*` manager constructors gained a `maxActiveTransactions` parameter, and an external component constructs these managers directly, so a **paired external change** is required before/with merge.
- Caffeine (2.9.3, the Java-8-compatible line) is already a `core` dependency; no new dependency is introduced.
- Behavioral note: because size-based eviction is decided asynchronously, a successful `add` is not a retention guarantee — under sustained pressure at the cap a just-added transaction can be evicted moments later. This is documented on the relevant APIs.
- User-facing documentation on the ScalarDB docs site for the new `scalar.db.active_transaction_management.max_active_transactions` property is a follow-up.

## Release notes

Added a configurable limit on the number of active transactions tracked by active transaction management (`scalar.db.active_transaction_management.max_active_transactions`, default 10000), backed by a Caffeine cache with Window-TinyLFU eviction, and refreshed the idle timer on every transaction operation so long-running transactions are not reaped while still in use.

### DIFF
--- a/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
+++ b/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
@@ -31,7 +31,9 @@ public abstract class AbstractDistributedTransactionProvider
       // active transaction registry) carry the behavior of every inner decorator.
       transactionManager =
           new ActiveTransactionManagedDistributedTransactionManager(
-              transactionManager, config.getActiveTransactionManagementExpirationTimeMillis());
+              transactionManager,
+              config.getActiveTransactionManagementExpirationTimeMillis(),
+              config.getActiveTransactionManagementMaxActiveTransactions());
     }
 
     return transactionManager;
@@ -58,7 +60,9 @@ public abstract class AbstractDistributedTransactionProvider
       // Wrap the transaction manager for active transaction management
       transactionManager =
           new ActiveTransactionManagedTwoPhaseCommitTransactionManager(
-              transactionManager, config.getActiveTransactionManagementExpirationTimeMillis());
+              transactionManager,
+              config.getActiveTransactionManagementExpirationTimeMillis(),
+              config.getActiveTransactionManagementMaxActiveTransactions());
     }
 
     return transactionManager;
@@ -77,7 +81,9 @@ public abstract class AbstractDistributedTransactionProvider
       // so that the idle-expiry reap traverses every inner decorator via releaseContext.
       coordinator =
           new ActiveTransactionManagedTwoPhaseCommitCoordinator(
-              coordinator, config.getActiveTransactionManagementExpirationTimeMillis());
+              coordinator,
+              config.getActiveTransactionManagementExpirationTimeMillis(),
+              config.getActiveTransactionManagementMaxActiveTransactions());
     }
 
     return coordinator;
@@ -100,7 +106,9 @@ public abstract class AbstractDistributedTransactionProvider
       // so that the idle-expiry reap traverses every inner decorator via releaseContext.
       participant =
           new ActiveTransactionManagedTwoPhaseCommitParticipant(
-              participant, config.getActiveTransactionManagementExpirationTimeMillis());
+              participant,
+              config.getActiveTransactionManagementExpirationTimeMillis(),
+              config.getActiveTransactionManagementMaxActiveTransactions());
     }
 
     return participant;
--- a/core/src/main/java/com/scalar/db/common/ActiveTransactionRegistry.java
+++ b/core/src/main/java/com/scalar/db/common/ActiveTransactionRegistry.java
@@ -14,8 +14,6 @@ import org.slf4j.LoggerFactory;
 @ThreadSafe
 public class ActiveTransactionRegistry<T> {
 
-  private static final long TRANSACTION_EXPIRATION_INTERVAL_MILLIS = 1000;
-
   private static final Logger logger = LoggerFactory.getLogger(ActiveTransactionRegistry.class);
 
   private final ActiveExpiringMap<String, T> activeTransactions;
@@ -24,26 +22,46 @@ public class ActiveTransactionRegistry<T> {
    * Constructs an ActiveTransactionRegistry.
    *
    * @param expirationTimeMillis the expiration time in milliseconds
-   * @param expirationHandler the handler invoked when a transaction expires
+   * @param maxActiveTransactions the maximum number of active transactions to keep; when exceeded,
+   *     a transaction is evicted (disposed via {@code disposalHandler}) per the cache's recency-
+   *     and frequency-aware policy to make room. Non-positive means unbounded.
+   * @param disposalHandler the handler invoked to dispose a transaction when it expires or is
+   *     evicted
    */
   public ActiveTransactionRegistry(
-      long expirationTimeMillis, ExpirationHandler<T> expirationHandler) {
+      long expirationTimeMillis, int maxActiveTransactions, DisposalHandler<T> disposalHandler) {
     this.activeTransactions =
         new ActiveExpiringMap<>(
             expirationTimeMillis,
-            TRANSACTION_EXPIRATION_INTERVAL_MILLIS,
+            maxActiveTransactions,
             (id, t) -> {
               logger.warn("The transaction is expired. Transaction ID: {}", id);
-              try {
-                expirationHandler.onExpired(t);
-              } catch (Exception e) {
-                logger.warn("Failed to handle the expired transaction. Transaction ID: {}", id, e);
-              }
+              dispose(id, t, disposalHandler, "expired");
+            },
+            (id, t) -> {
+              logger.warn(
+                  "The transaction is evicted because the maximum number of active transactions "
+                      + "({}) was exceeded. Transaction ID: {}",
+                  maxActiveTransactions,
+                  id);
+              dispose(id, t, disposalHandler, "evicted");
             });
+  }
+
+  private void dispose(String id, T transaction, DisposalHandler<T> disposalHandler, String cause) {
+    try {
+      disposalHandler.onDisposed(transaction);
+    } catch (Exception e) {
+      logger.warn("Failed to handle the {} transaction. Transaction ID: {}", cause, id, e);
+    }
   }
 
   /**
    * Adds a transaction to the registry.
+   *
+   * <p>A {@code true} return means the transaction was inserted, not that it will remain
+   * registered: at the cap, a just-added transaction can be evicted (and disposed) moments later.
+   * Callers must not assume a successful add keeps the transaction alive.
    *
    * @param id the transaction ID
    * @param transaction the transaction to add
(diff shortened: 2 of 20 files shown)


komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
brfrn169, line 1: This is a comment to block this PR from being merged.

core/src/main/java/com/scalar/db/common/ActiveTransactionRegistry.java
```diff
@@ -24,27 +22,47 @@ public class ActiveTransactionRegistry<T> {
...
-              }
+              dispose(id, t, disposalHandler, "expired");
+            },
+            (id, t) -> {
+              logger.warn(
+                  "The transaction is evicted because the maximum number of active transactions "
+                      + "({}) was exceeded. Transaction ID: {}",
+                  maxActiveTransactions,
+                  id);
+              dispose(id, t, disposalHandler, "evicted");
```
feeblefakie, line 47: **🔴 Size-based eviction can silently dispose a live, healthy transaction — and the victim is never told.**

The cap picks a victim by Caffeine's recency/frequency policy (TinyLFU), not by transaction liveness. A live transaction that's momentarily "cold" (client doing app-side compute or a slow external call between ops) can become the LRU victim once we're over `maxActiveTransactions`, and disposal runs `ActiveTransaction.rollback()` on it.

I traced the outcome end-to-end (ConsensusCommit, no group commit):
- `ConsensusCommit.rollback()` before prepare only closes scanners + frees the group-commit slot — the snapshot/context stay intact, and no aborted flag is set.
- `commit()`'s only precondition is `areAllScannersClosed()`, which the disposal rollback *satisfies*. There is no "was rolled back" guard.
- So the victim's next `get()` reads its intact snapshot and **succeeds**, and it can go on to **commit successfully**. The eviction meant to reclaim the transaction is silently voided, and for ConsensusCommit it frees almost nothing (the client still holds the object + snapshot). For JDBC it's the opposite failure — `rollback()` closes the connection, so the next op throws an opaque `SQLException` with no "evicted" signal.

To be clear, this is **not** corruption or a race — the `synchronized` `ActiveTransaction` serializes disposal against client ops correctly. The problem is the *policy*: aborting live transactional state by cache-popularity with no clean client-visible outcome.

**Suggested direction:** reject *new* transactions at the cap (fail `begin()`/`start()`/`join()` fast with a dedicated "too many active transactions" error) rather than evicting an incumbent — this also fixes the admission-window case where a freshly-begun transaction can be evicted before the caller uses it. If eviction must stay, surface a specific "evicted due to active-transaction cap" exception on the victim's next call so no backend silently continues on a zombie. Best resolved together with the default-on concern on the config default below.

core/src/main/java/com/scalar/db/config/DatabaseConfig.java
```diff
@@ -66,6 +69,7 @@ public class DatabaseConfig {
   public static final String SCAN_FETCH_SIZE = PREFIX + "scan_fetch_size";
 
   public static final int DEFAULT_METADATA_CACHE_EXPIRATION_TIME_SECS = 60;
+  public static final int DEFAULT_ACTIVE_TRANSACTION_MANAGEMENT_MAX_ACTIVE_TRANSACTIONS = 10000;
```
feeblefakie, line 72: **🟡 The cap is ON by default (10000), which is a backward-incompatible behavior change.**

Previously, under default config, active transactions were never auto-disposed: `expiration_time_millis` defaults to `-1` and the old reaper only ran when lifetime `> 0`. This size cap disposes regardless of the expiration setting. After upgrading, any workload legitimately holding >10,000 concurrent active transactions (many paused/long-running 2PC transactions, connection-pool-per-tenant designs, or a pre-existing leak that was previously just a memory concern) will start having transactions rolled back underneath it — with the silent/opaque semantics noted on the eviction handler, and no flag or release note gating it.

**Suggested direction:** default the cap to unbounded (`0`) and make it opt-in, so upgrades are behavior-preserving. If it must ship on, gate it behind the eviction-semantics fix and call it out explicitly as a breaking change in the release notes with the failure mode + tuning guidance. Either way it needs docs — the property currently has zero references under `docs/`.

core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
```diff
@@ -1,234 +1,201 @@
...
+
+  // Runs the reclamation handlers (idle-expiry and size-eviction). Disposal work can block on
+  // network/storage I/O (e.g. rolling a transaction back or releasing its context), so it must not
+  // run on ForkJoinPool.commonPool() — Caffeine's default removal executor — where a burst of
+  // blocking disposals could starve unrelated common-pool work. Threads are daemon. The fixed
+  // thread
+  // count bounds how many disposals run concurrently, throttling a mass reclamation to at most that
+  // many concurrent backend calls rather than flooding the backend; any excess queues (the queue is
+  // unbounded, so no disposal is dropped) and drains as threads free up.
+  private static final Executor DISPOSAL_EXECUTOR = newDaemonDisposalExecutor();
```
feeblefakie, line 69: **🟡 The disposal pool has an unbounded queue and is shared JVM-wide — risky under mass reclamation.**

`newFixedThreadPool(availableProcessors())` uses an unbounded `LinkedBlockingQueue`, and the handler does blocking I/O (`rollback()` / `releaseContext()`). Under an idle-timeout wave or an eviction storm at the cap, thousands of transactions can be reclaimed near-simultaneously while only ~NCPU threads drain them — the queue holds thousands of `Runnable`s, each capturing a transaction + its snapshot (transient heap spike proportional to backlog). If a backend hangs without a socket timeout, all threads get pinned and the unbounded queue grows without bound → disposal stalls and heap climbs toward OOM.

It's also a `static` singleton shared by every `ActiveExpiringMap`/registry in the JVM, so a storm in one transaction manager can starve disposals (and cap-relief evictions) in every other manager.

**Suggested direction:** use a bounded queue with a defined saturation policy (e.g. `CallerRunsPolicy`, which backpressures the reaper/scheduler thread instead of accumulating unbounded work), and run disposal `rollback`/`releaseContext` with a finite timeout so a hung backend can't wedge the pool. Consider a per-manager (or per-config) pool to remove the cross-manager coupling.

feeblefakie (approved):
LGTM! Thank you!
My AI pointed out some P1/P2 issues, which don't look that critical.
But, PTAL just in case.

core/src/main/java/com/scalar/db/common/ActiveTransactionRegistry.java
brfrn169: Thanks for the detailed trace — it's accurate. We're treating this size cap purely as a last-resort safety net against OOM, not as a general transaction-lifecycle mechanism. Holding more than the default 10,000 concurrent active transactions is already a pathological state (typically a leak or severe misconfiguration), so in normal operation the cap never fires and no live transaction is evicted. In that safety-net role we're comfortable with the current behavior: it's best-effort (records left behind are reconciled by lazy recovery), and hitting the cap already signals something is wrong. So we'll keep it as-is for now.

core/src/main/java/com/scalar/db/config/DatabaseConfig.java
brfrn169: Agreed it's a behavior change on paper, but we don't think it breaks realistic workloads: holding >10,000 concurrent active transactions is already an irregular state (typically a leak), and the cap exists precisely to stop that from growing into an OOM. A workload that legitimately needs more can raise or disable the cap. So we'd like to keep it on by default as a safety net and keep the current default.

core/src/main/java/com/scalar/db/util/ActiveExpiringMap.java
brfrn169: Good points. We considered both bounding the queue and dropping the static shared pool, but we'll keep the current design. The backlog / cross-manager-starvation concern only materializes under the same >10,000 storm scenario, which is already pathological; the disposed transactions' state is already resident in memory, so the queued tasks add little incremental heap, and a stalled disposal is best-effort (reconciled by lazy recovery). Making the pool per-instance would multiply daemon threads (NCPU per registry, several per config) and would need a shutdown lifecycle these maps don't currently have — risking a thread leak — for a benefit that only applies in that rare storm. So we're keeping the shared daemon pool as-is.

PR #3678, opened 2026-07-04
Disable secondary-index Get tests for Object Storage

brfrn169:
## Description

The full CI run ([28697501820](https://github.com/scalar-labs/scalardb/actions/runs/28697501820)) failed with 12 test failures in the Blob Storage (Object Storage) integration tests, all throwing `IllegalArgumentException` from `OperationChecker` / `ConsensusCommitOperationChecker`.

The cause is that PR #3607 added four secondary-index `Get` tests to `ConsensusCommitSpecificIntegrationTestBase`, but did not add the corresponding `@Disabled` overrides in the Object Storage subclass. Object Storage does not support index-related operations, so these tests fail as soon as an index `Get` is issued. The subclass already disables all other index-related tests the same way; these four were simply missed.

This failure only surfaces in the manually triggered full CI (`workflow_dispatch`), which is why it was not caught when #3607 was merged.

## Related issues and/or PRs

- Follow-up to #3607, which introduced the four tests.

## Changes made

- Added four `@Disabled("Object Storage does not support index-related operations")` overrides to `ConsensusCommitSpecificIntegrationTestWithObjectStorage` for:
  - `get_GetWithIndexForTransientDeletedAndPreparedRecords_ShouldResolveToSingleRecord`
  - `get_GetWithIndexForTransientDeletedAndPreparedRecordsWhenWriterAborted_ShouldResolveToNoRecord`
  - `get_GetWithIndexForExistingCommittedRecordAndTransientPreparedRecord_ShouldReturnCommittedRecordInAllIsolations`
  - `get_GetWithIndexMatchingMultipleCommittedRecords_ShouldThrowIllegalArgumentException`

## Checklist

This PR only addresses the Object Storage failures. The same full CI run also shows unrelated `Lock wait timeout exceeded` failures in the MySQL/MariaDB `InHighestIsolation` (SERIALIZABLE) scan-recovery tests, which are being investigated separately.

## Release notes

N/A

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/storage/objectstorage/ConsensusCommitSpecificIntegrationTestWithObjectStorage.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/objectstorage/ConsensusCommitSpecificIntegrationTestWithObjectStorage.java
@@ -317,4 +317,26 @@ public class ConsensusCommitSpecificIntegrationTestWithObjectStorage
   @Disabled("Object Storage does not support index-related operations")
   public void
       commit_GetScannerWithScanAllIndexConditionInSerializable_WhenBeforeIndexHasPreparedRecordFromOtherTransaction_ShouldThrowCommitConflictException() {}
+
+  @Override
+  @Disabled("Object Storage does not support index-related operations")
+  public void get_GetWithIndexForTransientDeletedAndPreparedRecords_ShouldResolveToSingleRecord(
+      Isolation isolation) {}
+
+  @Override
+  @Disabled("Object Storage does not support index-related operations")
+  public void
+      get_GetWithIndexForTransientDeletedAndPreparedRecordsWhenWriterAborted_ShouldResolveToNoRecord(
+          Isolation isolation) {}
+
+  @Override
+  @Disabled("Object Storage does not support index-related operations")
+  public void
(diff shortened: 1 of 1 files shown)


komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

hiSandog:
This PR fixes the immediate missed overrides, but a backend capability flag might prevent the same class of issue from returning. If Object Storage explicitly advertises “no secondary index operations,” the shared integration suite could skip or expect failures centrally instead of requiring every new index test to be disabled in the subclass.

brfrn169:
@hiSandog Thanks, that's a good point — I agree that centralizing this would be more robust than relying on every new index test being disabled in the subclass by hand.

There's actually a precedent for exactly this in the suite: group-commit-specific tests are gated centrally via `@EnabledIf("isGroupCommitEnabled")` against an overridable hook, rather than being disabled per-subclass. We could introduce a similar `isSecondaryIndexSupported()` (defaulting to true, overridden to false for Object Storage) and annotate the index tests with `@DisabledIf`/`@EnabledIf`. That would let Object Storage opt out of the whole class of index tests in one place and remove the ~two dozen per-test `@Disabled` overrides currently in the subclass.

That refactor touches the shared base class and every index test, though, so I'd prefer to keep this PR as a minimal hotfix to unblock the full CI. I'll note it down as a follow-up. Thanks!

PR #3679, opened 2026-07-05
Fix ConsensusCommit scan-recovery integration tests on JDBC engines under the highest isolation level

brfrn169:
## Description

The manually triggered full CI run ([28697501820](https://github.com/scalar-labs/scalardb/actions/runs/28697501820)) failed or hung on **MySQL, MariaDB, SQL Server, Db2 and SQLite** in the `scan`/`getScanner` ConsensusCommit recovery-scenario integration tests.

Root cause: several of these scenarios simulate a finalize/cleanup race by **synchronously writing the scanned record from within lazy recovery while the storage scanner that read it is still open**. On engines that keep a shared/range lock for an open scan, that write blocks on the scan's own lock and the simulation self-deadlocks. Which engines block depends on their concurrency control:

- **MySQL / MariaDB** are MVCC below SERIALIZABLE and take shared read locks only under SERIALIZABLE, so only the **highest-isolation** test class is affected (surfaces as a lock-wait timeout).
- **SQL Server / Db2** are lock-based at **every** isolation level (including their default READ COMMITTED), so **every** JDBC test class that runs these scenarios is affected. Their lock wait is effectively unbounded, so the write blocks forever and the job **hangs** to the CI timeout.
- **PostgreSQL, Oracle and TiDB** use MVCC/SSI (no shared read locks) and are unaffected. **SQLite** uses whole-file locking and is handled via WAL mode.

**This is a test-only issue — production is not affected.** Production never synchronously writes a scanned row during recovery: the real synchronous recovery write during a scan targets the coordinator-state table (a different table, no lock conflict with the scan), and the record roll-forward/rollback is asynchronous and awaited only at commit, which first requires all scanners to be closed (`areAllScannersClosed()` guard, so the locks are already released). Empirically, the plain (non-simulated) scan-recovery tests, which exercise this real path, did **not** hang on SQL Server/Db2 — only the mock-write scenarios did.

## Related issues and/or PRs

- The scan-recovery scenarios were introduced by #3607 and #3650.

## Changes made

- Add an overridable `isConcurrentWriteToRowUnderOpenScanSupported()` predicate (default `true` in `ConsensusCommitSpecificIntegrationTestBase`), reached through a shared `assumeConcurrentWriteToRowUnderOpenScanSupported(Selection)` helper, and skip the scan-path variants of the finalize/cleanup-race simulations (including the group-commit-only `GroupCommitParentRow` scenario) where the engine blocks:
  - **highest-isolation JDBC class**: MySQL, MariaDB, SQL Server, Db2 (all take locks under SERIALIZABLE);
  - **default and metadata-decoupling JDBC classes**: SQL Server and Db2 only (MySQL/MariaDB are MVCC there and keep running).
  The resolved `RdbEngine` is used (`JdbcTestUtils.isMysql`/`isMariaDB`/`isSqlServer`/`isDb2`), so **TiDB** (which uses the `jdbc:mysql:` URL but resolves to a distinct `RdbEngine` whose highest isolation is `REPEATABLE_READ`) is correctly excluded. The `Get` variants keep covering the recovery decision logic on every engine — a point `Get` releases its connection immediately, so it never deadlocks.
- **Run the SQLite integration tests in WAL mode** (`journal_mode=WAL` in `ci/tests-config.yaml`), which lets a reader and a writer proceed concurrently, resolving SQLite **without skipping any tests**.
- Make `tearDown()` null-safe for `recoveryExecutor` (a test can now be skipped before `createConsensusCommitManager()` initializes it), matching the existing `groupCommitter` null-guard.
- Reflow a few test comments that had awkward mid-sentence line breaks.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/core/src/integration-test/java/com/scalar/db/storage/jdbc/ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/jdbc/ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation.java
@@ -72,8 +72,7 @@ public class ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsol
     // the scan still holds blocks until the scan finishes. The scan-path finalize/cleanup-race
     // recovery tests simulate exactly such a write from within lazy recovery while the scanner is
     // open, so they self-deadlock on these engines.
-    return !(JdbcTestUtils.isMysql(rdbEngine)
-        || JdbcTestUtils.isMariaDB(rdbEngine)
+    return !((JdbcTestUtils.isMysql(rdbEngine) && !JdbcTestUtils.isTidb(rdbEngine))
         || JdbcTestUtils.isSqlServer(rdbEngine)
         || JdbcTestUtils.isDb2(rdbEngine));
   }
--- a/core/src/integration-test/java/com/scalar/db/storage/jdbc/ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation.java
+++ b/core/src/integration-test/java/com/scalar/db/storage/jdbc/ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation.java
@@ -18,6 +18,7 @@ public class ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsol
     extends ConsensusCommitSpecificIntegrationTestBase {
 
   private JdbcAdminTestUtils jdbcAdminTestUtils;
+  private RdbEngineStrategy rdbEngine;
 
   @Override
   protected Properties getProperties(String testName) {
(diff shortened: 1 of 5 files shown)


core/src/integration-test/java/com/scalar/db/storage/jdbc/ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation.java
```diff
@@ -63,6 +65,19 @@ protected void truncateCoordinatorTables() throws ExecutionException {
     super.truncateCoordinatorTables();
   }
 
+  @Override
+  protected boolean isConcurrentWriteToRowUnderOpenScanSupported() {
```
komamitsu, line 69: This method name is the same as other helper methods that return false in case of `SqlServer` and `Db2`. How about using a different name (e.g., `...WithSerializable`)?

brfrn169: Since all three JDBC subclasses `@Override` the same base `isConcurrentWriteToRowUnderOpenScanSupported()` (the base `assumeConcurrentWriteToRowUnderOpenScanSupported(Selection)` helper dispatches through it), we can't rename just this one to `...WithSerializable`. For that reason I'd like to keep the shared name — the isolation-specific difference is documented in each override's comment instead. Let me know if you have concerns. Thanks!

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

PR #3689, opened 2026-07-09
Add a write-set detail option to TwoPhaseCommit.Participant#prepareRecords

brfrn169:
## Description

With the new `TwoPhaseCommit` interface, `Participant#prepareRecords` always returns `WriteSetEntry`s carrying the full non-key column values, while the Coordinator persists only the primary keys in the commit state (it encodes the write set with `includeColumns=false`). The column values are therefore always built, shipped, and then discarded. This is pure overhead today, and it becomes real serialization cost once the Coordinator and Participants are deployed across a transport boundary.

Since only the Coordinator knows how much write-set detail it needs, this PR makes the Coordinator request the detail level as a parameter of `prepareRecords`, rather than introducing a Participant-side configuration that could disagree with the Coordinator's persistence settings.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/3660
- https://github.com/scalar-labs/scalardb/pull/3662

## Changes made

- Added the `TwoPhaseCommit.WriteSetDetailLevel` enum (`KEYS_ONLY` / `FULL`) and added it as a parameter of `TwoPhaseCommit.Participant#prepareRecords`:
  - Under `KEYS_ONLY`, every returned `WriteSetEntry` must carry no non-key columns (`getColumns()` is empty).
  - Under `FULL`, every `WRITE` entry carries the non-key user columns it writes — the full user-visible record content of the write. Implementation-internal columns are not part of the record content at this interface, and this is the write's own content, not a full row image (a partial update carries exactly the columns it sets).
  - The enum is documented as a single ordered axis of record-content detail with `FULL` as its permanent ceiling: additions beyond the record content of a write (transaction metadata, before images, and the like) belong on a separate axis, not on higher levels.
  - The detail level never changes which entries are returned — the entries and their keys are identical at either level — so an empty write set still means a write-less participant regardless of the requested detail.
- `ConsensusCommitCoordinator` now requests `KEYS_ONLY`, matching the keys-only commit state it persists. This becomes configuration-driven once a consumer of column values (e.g., backup/changelog capture) ships.
- `ConsensusCommitParticipant#buildWriteSetEntries` skips the column pass — including the table-metadata lookup, which is needed only to filter transaction-meta columns — entirely under `KEYS_ONLY`.
- Updated the `TwoPhaseCommit` role decorators (`DecoratedTwoPhaseCommitParticipant`, `AttributePropagatingTwoPhaseCommitParticipant`, `ActiveTransactionManagedTwoPhaseCommitParticipant`) to forward the new parameter.
- Added unit tests verifying that `FULL` carries user columns only (transaction-meta columns such as `tx_id` are excluded) and that `KEYS_ONLY` returns the same entries with empty columns without touching the table metadata.

## Checklist

The `KEYS_ONLY` contract is intentionally specified as "must not carry columns" (rather than "may omit them") so that implementations are testable and the response size is guaranteed once the prepare call crosses a transport boundary.

## Release notes

N/A


core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitParticipant.java
```diff
@@ -564,26 +565,41 @@ private void checkMutation(String transactionId, Mutation mutation) throws CrudE
...
+        case KEYS_ONLY:
+          // KEYS_ONLY entries must carry no non-key columns. Skipping the column pass also skips
+          // the table-metadata lookup, which is needed only to filter transaction-meta columns.
+          columns = Collections.emptyList();
+          break;
+        case WITH_COLUMNS:
+          TableMetadata metadata = getTableMetadataForMutation(put);
+          List<Column<?>> filteredColumns = new ArrayList<>();
+          for (Column<?> column : put.getColumns().values()) {
+            if (!ConsensusCommitUtils.isTransactionMetaColumn(column.getName(), metadata)) {
```
komamitsu, line 595: Is it okay to filter out `tx_version`? Also, the previous `tx_id` might be useful information to traverse tx_id chains.

brfrn169: Thanks for pointing that out. Yes, I agree that including the transaction metadata would be useful. I’ll revisit it later in a separate PR. Thanks!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
```diff
@@ -693,12 +703,37 @@ enum Type {
...
+     * set is recorded for the active-recovery use case), and keeps the prepare result minimal.
+     */
+    KEYS_ONLY,
+
+    /**
+     * Every {@link WriteSetEntry.Type#WRITE} entry additionally carries the non-key columns it
+     * writes, excluding any implementation-internal columns. Intended for use cases that persist
+     * the full record content (for example, backup/changelog capture).
+     */
+    WITH_COLUMNS
```
feeblefakie: I think `FULL` or something similar might be more proper. WDYT?

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
```diff
@@ -693,12 +703,37 @@ enum Type {
...
+  /**
+   * How much detail the {@link WriteSetEntry}s returned by {@link Participant#prepareRecords} must
+   * carry, as requested by the Coordinator.
+   *
+   * <p>The detail level never affects which entries a write set contains — the entries and their
+   * primary keys are identical at every level; it only controls whether {@link
+   * WriteSetEntry#getColumns()} is populated. {@link WriteSetEntry.Type#DELETE} entries never carry
+   * columns at any level.
+   */
+  enum WriteSetDetail {
```
feeblefakie: As the comment says, it might be more proper to name it to something like `WriteSetDetailLevel` or `WriteSetLevel`. WDYT?

feeblefakie (commented):
Overall, looking good. Left some comment/suggestions. PTAL!

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
brfrn169: Thanks! Renamed it to `FULL`.

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
brfrn169: Thanks! Renamed it to `WriteSetDetailLevel` as well.

feeblefakie (approved):
LGTM! Thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
```diff
@@ -176,7 +177,9 @@ public void commit(String transactionId)
         long preparedAt = System.currentTimeMillis();
         try {
           for (Participant participant : participants) {
-            PreparationResult result = participant.prepareRecords(transactionId, preparedAt);
+            PreparationResult result =
+                participant.prepareRecords(
+                    transactionId, preparedAt, WriteSetDetailLevel.KEYS_ONLY);
```
feeblefakie, line 182: Late follow-up to my approval, and only a suggestion.

This `KEYS_ONLY` and the `includeColumns = false` passed to `WriteSetEncoder.encodeFromWriteSetEntries` further down in `commitState` have to agree, and nothing enforces it. They're ~200 lines apart, the coordinator is the only thing that picks both, and a mismatch would fail *silently* rather than loudly: requesting `KEYS_ONLY` while encoding with `includeColumns = true` would just encode zero columns.

Deriving the encoder flag from the requested `WriteSetDetailLevel` — rather than hardcoding both independently — would make them agree by construction and make the detail level the single source of truth.

Not urgent, since the two are consistent today. It'd matter more once a consumer actually requests `FULL` (per the `tx_version` / metadata discussion on `ConsensusCommitParticipant` that you're deferring to a separate PR) — that's exactly the change where the two could drift apart.

feeblefakie (commented):
Sorry to circle back after approving. One small suggestion inline about the coupling between the requested detail level and the encoder's `includeColumns` flag — non-blocking, and arguably better handled in the follow-up PR where `FULL` gets a real consumer.

For what it's worth, I traced the `KEYS_ONLY` path while reviewing the stack, and the optimisation is real and slightly better than the description suggests: the participant was previously doing a table-metadata lookup, a per-write `isTransactionMetaColumn` filter, and an `ArrayList` allocation per write to build a payload the coordinator then discarded unread. I also confirmed the empty-iff-write-less guarantee genuinely survives — the `hasWritesOrDeletes()` early return sits ahead of the `detailLevel` switch, so the entry set is identical at both levels — which matters, because the ABORTED-write omission in #3690 depends on it.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: Thanks — agreed this is a real sharp edge, and thanks for spelling out the silent-failure mode.

I'm going to defer it to the follow-up PR rather than address it here, for a couple of reasons:

- Today `KEYS_ONLY` and `includeColumns = false` are consistent, so there's no actual defect to fix right now — only the latent drift risk you point out.
- As you noted, that risk only becomes real once a consumer actually requests `FULL` (the `tx_version` / metadata work on `ConsensusCommitParticipant` that I'm deferring). That PR is also the natural place to get the design right: rather than deriving one flag from the other, I'd like the detail level and the encoder flag to be decided *together* from a single upstream intent (e.g. a future backup/changelog mode would set `FULL` + `includeColumns = true` in one place). Doing that half-way now, before that upstream switch exists, would be less coherent than doing it alongside the change that introduces it.

So I'll carry this over and handle it in that PR. Thanks again for the careful review!

Torch3333 (approved):
LGTM, thank you!

PR #3690, opened 2026-07-09
Omit the ABORTED Coordinator state write for known write-less transactions

brfrn169:
## Description

`ConsensusCommitCoordinator`'s abort path always wrote the ABORTED Coordinator state row, even when coordinator-write omission on read-only is enabled. This was conservative because `hasWrites` is not trustworthy on a prepare-phase failure: a participant that throws part-way through `prepareRecords` never returns its write set, so records may be left PREPARED that the Coordinator did not observe.

However, the omission can be applied safely when the transaction is *provably* write-less. For such a transaction, no PREPARED record exists anywhere, so the ABORTED row would never be consulted by lazy recovery. Skipping it saves a pointless storage write and, more importantly, avoids surfacing a spurious `UnknownTransactionStatusException` from a failed `abortState` for a transaction whose outcome is in fact fully known.

The key insight is that the strongest available signal differs by phase: during the prepare phase only the static `readOnly` flag can prove the transaction write-less, whereas once every `prepareRecords` has returned, `hasWrites` authoritatively reflects every participant's write set — matching `CommitHandler`'s abort path, which gates on the snapshot it owns.

## Related issues and/or PRs

- https://github.com/scalar-labs/scalardb/pull/3660
- https://github.com/scalar-labs/scalardb/pull/3662

## Changes made

- Split `ConsensusCommitCoordinator#commit`'s prepare-phase and validate-phase exception handling into separate try/catch blocks so each phase applies the omission with the strongest signal available to it:
  - On a prepare-phase failure (`PreparationException` / `TransactionNotFoundException`), the transaction is known write-less only if it is read-only (participants reject mutations on a read-only transaction, so no PREPARED record can exist even after a part-way failure).
  - On a validate-phase failure (`ValidationException` / `TransactionNotFoundException`), every `prepareRecords` has returned, so any observed write-less transaction (`!hasWrites`) qualifies, read-only or not.
- `abortAndRollbackRecords` now takes a `knownWriteLess` flag and skips the ABORTED state write when the transaction is known write-less and the omission is enabled. The participants are still rolled back to release their local resources. When the omission is disabled, the row is always written, preserving the legacy behavior that every terminated transaction leaves a Coordinator state row.
- Extracted a `shouldWriteCoordinatorState` helper in both `ConsensusCommitCoordinator` and `CommitHandler` so the COMMITTED gate (commit-success path) and the ABORTED gate (abort path) share a single policy in each class.
- Added unit tests covering the omission decision for prepare-phase and validate-phase failures (both exception types), for read-only and non-read-only transactions, with and without writes, and with the omission enabled and disabled.

## Checklist

The omission is intentionally NOT applied on a prepare-phase failure of a non-read-only transaction even when no writes were observed: the throwing participant may have left records PREPARED without reporting them, and the ABORTED row is what lets lazy recovery abort those records on the next read instead of leaving them locked until the transaction lifetime expires.

## Release notes

N/A


komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/transaction/consensuscommit/CommitHandler.java
```diff
@@ -244,14 +244,22 @@ public void commit(TransactionContext context)
...
     if (hasWritesOrDeletesInSnapshot) {
       rollbackRecords(context);
     }
   }
 
+  // The single policy behind both Coordinator state gates: the row (COMMITTED on the commit path,
+  // ABORTED on the abort path) is written unless the transaction has no writes or deletes and
+  // coordinator write omission on read-only is enabled. Mirrored by ConsensusCommitCoordinator's
+  // gate of the same name.
+  private boolean shouldWriteCoordinatorState(boolean hasWritesOrDeletes) {
```
feeblefakie: Should means it's nice to have, so isn't it better to be `must`?
```suggestion
  private boolean mustWriteCoordinatorState(boolean hasWritesOrDeletes) {
```
As for the naming, my Copilot recommends the following in order.

- isCoordinatorStateWriteRequired(...)
- mustWriteCoordinatorState(...)
- requiresCoordinatorStateWrite(...)

brfrn169: Thanks! Agreed that `should` reads as optional. I went with `isCoordinatorStateWriteRequired` since the `is...Required` form is the dominant convention in this project, and renamed the mirrored gate in both `CommitHandler` and `ConsensusCommitCoordinator` in f5acec9f5. PTAL!

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #3708, opened 2026-07-13
Probe participants before reaping an expired coordinator context

brfrn169:
## Description

The coordinator role observes only `begin` and `registerParticipant` — the CRUD a transaction issues goes directly to its participants — so its active-transaction expiration time acted as a **total transaction lifetime bound** rather than an idle timeout: a healthy long-running transaction was silently reaped and its commit then failed with `TransactionNotFoundException`. This affects the in-process `TwoPhaseCommitBackedDistributedTransactionManager` facade and, more generally, any coordinator built on this decorator — including deployments whose participants are remote.

A coordinator context is useful exactly as long as some participant still holds the transaction, so this PR reaps on that condition instead: expired transactions are probed against their registered participants before any reap, and the context is released only when every participant reports the transaction gone — at which point it can no longer commit anyway. The expiration time thereby becomes the *probing period*, and a context lives exactly as long as some participant holds the transaction or its absence cannot be confirmed.

## Related issues and/or PRs

- #3669 (the active-transaction-management decorators this PR extends)

## Changes made

- Add a liveness probe to `TwoPhaseCommit.Participant`: `boolean hasTransactionContext(transactionId)`. The probe is **quiet** by contract (it must not refresh any idle timer or eviction recency, or probing would keep abandoned transactions alive forever) and **time-bounded**. There is no default implementation: a not-yet-updated implementation fails to compile rather than silently degrading. A thrown `TransactionNotFoundException` is a definitive "no context" — an alternative carrier of the `false` answer — so a remote implementation's conventional not-found error mapping (including a probe rerouted to a successor node after a crash) works as-is and leads to a prompt reap. Any other exception is fail-open (possibly present).
- `ActiveTransactionManagedTwoPhaseCommitCoordinator` now tracks a per-transaction expiration time (pushed out by `begin` / `registerParticipant`) and sweeps its tracked transactions on a dedicated daemon thread every second (or every expiration period, if shorter), probing every expired transaction **in place, while it stays tracked**. If any participant still holds it (or cannot be probed), the expiration time is pushed another period out; only when every participant reports it gone is the context released. A sweep pass is one volatile read per unexpired entry, and probes fire at most once per period per transaction, so the short sweep interval buys prompt expiry detection without inflating the probe rate.
- Fail-open retention is deliberately unbounded: while a participant is entirely unreachable no new transactions can join it (so the affected entries are bounded to the in-flight snapshot), every cycle re-probes (so the state self-heals the moment probes answer again), the registry cap stays as the hard memory bound, and each fail-open retention logs a WARN.
- Probing in place means an entry never has a not-registered moment, which shrinks the concurrency surface to a single handshake: a registration pushes the expiration time out under the entry monitor before delegating, and the sweep re-checks it under the same monitor (releasing before removing) before reaping — a reap never overlaps a registration that is about to succeed. The decorator assumes transaction IDs are not reused (documented in the class Javadoc).
- Idle expiration on the underlying registry is disabled (the registry contributes only the size cap, whose eviction remains an unconditional release); `ActiveExpiringMap` / `ActiveTransactionRegistry` gain only an iteration view (`forEach`, pinned by a test to not count as access).
- The participant-side decorator deliberately does **not** override the probe, so it passes through to the wrapped participant's context map and the quiet contract holds by construction.
- Route the test-only facade IT provider through the role-level decorator factory methods (it previously bypassed them), and add an integration test that keeps issuing CRUD past the expiration and then commits — it failed with `CommitConflictException` before this change and passes after it.
- The decorator's unit tests are deterministic: they force entries' expiration times and invoke the sweep directly, with one wall-clock smoke test pinning that the sweep is actually scheduled.

## Checklist

- For a remote `Participant` implementation of the probe (out of scope for this PR), "no context held" may be reported either as a `false` return or through the implementation's conventional not-found error mapping — both are safe by the probe's contract. Such an implementation must keep the probe quiet end-to-end (no activity tracking updated on the probe path) and must bound it with a deadline.

## Release notes

N/A


feeblefakie (approved):
LGTM! Thank you!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
```diff
@@ -78,23 +218,42 @@ public String begin(
...
+    TrackedTransaction tracked = current.get();
+    // Push the expiration time out under the entry monitor BEFORE delegating: the sweep re-checks
+    // it under the same monitor before reaping, so a reap never overlaps a registration that is
+    // about to succeed (see the class Javadoc).
+    tracked.updateExpirationTime(nextExpirationTimeMillis());
     super.registerParticipant(transactionId, participant);
+    // First-wins per participant ID, mirroring the wrapped coordinator's idempotent registration.
+    // Instances registered under the same participant ID front the same stores, so they are
+    // interchangeable for probing.
+    tracked.addParticipant(participant);
```
feeblefakie, line 267: Sorry to come back after approving — I found one thing here that I think is a real bug, and I was able to reproduce it.

**The sweep can reap a transaction whose `registerParticipant` has already succeeded.**

The ordering in this method is:

```java
tracked.updateExpirationTime(nextExpirationTimeMillis());  // push deadline to T2, under the entry monitor
super.registerParticipant(transactionId, participant);      // delegate — join() happens here, and succeeds
tracked.addParticipant(participant);                        // <-- publish, only now
```

So between the delegation and this line there is a window where the participant has genuinely joined — the client legitimately owns the transaction on it — but `tracked.participants` is still empty. If the first `join()` outlasts the expiration period, `probeAndDecide` takes the `participants.isEmpty()` fast path and reaps on the wall clock **without probing at all**, releasing the context of a live transaction. The client's next coordinator call then fails with `TransactionNotFoundException`.

**Why the handshake doesn't catch this.** The handshake defends against a registration that moves the deadline *after* the sweep observed it. Here the push happened *before* the observation, so the sweep sees a deadline that is expired, consistent, and not moving, while the registration is still in flight — and `reapUnlessExtended`'s `expirationTimeMillis != observedExpirationTimeMillis` check passes. It makes the *deadline* atomic with the reap, but the *participant set* is published at a different point and isn't covered.

**Note `begin` gets this right** — it calls `addParticipant` *before* `registry.add`, so a begin-supplied participant is never invisible to a sweep. This method inverts that order.

**Reproduction** (fails on this branch with `NeverWantedButInvoked` — note the participant is stubbed `hasTransactionContext == true`, and the probe is never called):

```java
coordinator.begin(null, false, Collections.emptyMap(), /* participant= */ null);

doAnswer(inv -> { joinStarted.countDown(); releaseJoin.await(5, SECONDS); return null; })
    .when(delegate).registerParticipant(eq(TX), any());

new Thread(() -> coordinator.registerParticipant(TX, participant)).start();
assertThat(joinStarted.await(5, SECONDS)).isTrue();   // inside join: after the push, before addParticipant

registry.get(TX).get().updateExpirationTime(0);       // the join outlives the expiration period
coordinator.sweep();
releaseJoin.countDown();                              // join completes SUCCESSFULLY

verify(delegate, never()).releaseTransactionContext(any());  // FAILS
```

**Scope.** Liveness only, no integrity impact — `releaseTransactionContext` touches no storage and nothing is PREPARED yet, so no records are stranded. And it's **not reachable at the default config**, since `expiration_time_millis` defaults to `-1` and disables the sweeper. It needs the sweep enabled, `begin(..., participant = null)` (a documented path), and a first `join()` slower than the period — which needs no clock anomaly, since `ConsensusCommitCoordinator.registerParticipant` holds `synchronized (context)` across `participant.join(...)`.

**Suggestion.** Move `addParticipant` above the delegation. A failed `join()` then leaves the participant in `tracked`, but that looks benign: a participant that never joined answers `false`, so the reap proceeds as before; only an *unreachable* one pins the entry, and fail-open pinning is already the designed behaviour elsewhere (bounded by the cap, re-probed each cycle).

Related: the class Javadoc frames this window as begin-to-first-registration and attributes it to a forward clock jump. It's neither — it extends *through* the first registration, past the point where `join()` succeeded, and a slow join reaches it on a well-behaved clock. Worth correcting alongside the fix.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
```diff
@@ -129,4 +288,219 @@ public void releaseContext(String transactionId) {
...
+    }
+  }
+
+  /**
+   * One probing pass: probes, in place, every expired transaction, and reaps the ones whose absence
+   * is confirmed. Invoked on the sweeper thread; tests invoke it directly.
+   */
+  @VisibleForTesting
+  void sweep() {
+    long now = System.currentTimeMillis();
```
feeblefakie: Minor: `now` is sampled once per pass and every entry is compared against that one sample. With a slow pass (remote probes), entries visited late look not-yet-expired and get skipped to the next pass.

This is fail-safe — it only ever under-reaps, and every real reap decision rests on the probe plus the monitor handshake, not on `now` — so it's cleanliness rather than a bug. But sampling per entry costs a single `currentTimeMillis()` call, which is nothing next to the volatile read already happening, let alone a probe.

Tiny related thing: the `closed` check just below returns from the `BiConsumer`, which skips one entry rather than the traversal, so the walk continues over all remaining entries after `close()`. Harmless, just not the early exit the shape implies.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
```diff
@@ -129,4 +288,219 @@ public void releaseContext(String transactionId) {
...
+        new ConcurrentHashMap<>();
+
+    // The absolute wall-clock time at which the transaction becomes a probe candidate. Pushed out
+    // by begin/registerParticipant and by a sweep that found (or failed to rule out) a
+    // participant still holding the transaction. Written under the entry monitor (the
+    // registration-vs-reap handshake, see the class Javadoc); volatile so the sweep's cheap
+    // pre-check can read it without the monitor.
+    private volatile long expirationTimeMillis;
+
+    @VisibleForTesting
```
feeblefakie: Three of the four `@VisibleForTesting` annotations on `TrackedTransaction` have production callers, so the annotation is asserting something that isn't true:

- this constructor — called from `begin`
- `updateExpirationTime` — called from `registerParticipant` and from `probeAndDecide`'s keep-alive path
- `addParticipant` — called from `begin` and `registerParticipant`

Only `hasParticipant` is genuinely test-only. Suggest dropping the other three and keeping that one.

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
```diff
@@ -583,6 +584,45 @@ void commitRecords(String transactionId, long committedAt)
...
+     *     callers act on terminally (by reaping). An implementation must therefore throw it only
+     *     when the absence is authoritative for the participant as a whole - never for a transient
+     *     miss, such as a probe answered by a node that does not hold the context while another
+     *     node still might. A condition like that must surface as {@link TransactionException}
+     *     (fail-open) instead
+     * @throws TransactionException if the probe fails due to transient or nontransient faults (for
+     *     example, a remote participant that cannot be reached, or one that does not implement the
+     *     probe). Callers must treat a failed probe conservatively, as possibly present
+     */
+    boolean hasTransactionContext(String transactionId)
```
feeblefakie, line 638: While the probe is being added: could the per-transaction serialization requirement be stated on `TwoPhaseCommit.Coordinator` / `Participant` directly?

Both decorators ground their `@ThreadSafe` claim in the wrapped role serializing its own per-transaction work, and cite `ConsensusCommitCoordinator` by name as the evidence — while this interface says nothing about concurrency anywhere. The participant-side decorator is explicit that the dependency is unchecked ("A wrapped participant that does not serialize per-transaction calls would break this guarantee with no signal at this layer").

That's a `common`-layer decorator depending on an implementation detail of one `transaction.consensuscommit` class, and this PR makes it newly load-bearing: the sweep introduces a second thread calling `releaseTransactionContext` concurrently with application threads, where previously the coordinator decorator was single-threaded with respect to its wrapped role. A short paragraph here — implementations must serialize calls for the same transaction ID, and `releaseTransactionContext` may be invoked concurrently with any other method for that ID — would let both decorators cite the contract instead.

feeblefakie (commented):
Following up on my approval — I took another pass over the concurrency here, and one of the findings I think is a real (if narrow) bug, so I wanted to flag it before this merges. Reproduced with a unit test; details inline on `registerParticipant`. It isn't reachable at the default config (`expiration_time_millis` defaults to `-1`), and it's liveness-only with no integrity impact, so I'm commenting rather than blocking — your call on whether it's worth fixing here or as a follow-up.

The rest are minor.

One more that has no in-diff line to hang off: **`expiration_time_millis` now means two different things across the two roles** — an idle timeout on the participant, and a probing period that explicitly bounds nothing on the coordinator — while `AbstractDistributedTransactionProvider` feeds both from the same value. The reasoning is sound and well-argued in the class Javadoc, but an operator setting one number gets two behaviours, and the knob can't be tuned for one role without moving the other. Worth a note in the operator-facing docs, along with the fact that the sweep is off by default.

I also looked hard at the fail-open retention, the entry-monitor lock ordering, `ActiveExpiringMap.forEach`'s quietness, and the double-release-vs-cap-eviction interaction, and couldn't break any of them — the reasoning in the class Javadoc holds up. Nice work.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
brfrn169: Great catch, and thanks for the reproduction. You're right that the window extends through the join, not just begin-to-first-registration.

Fixed in d9bf2079d by publishing the participant *before* delegating the join (mirroring `begin`), so a sweep firing during a slow join probes the already-published participant instead of taking the no-participants fast path and reaping on the wall clock alone. As you noted, a failed join now leaves the participant tracked, which is benign: a participant that never joined answers `false` to a probe, so the reap still proceeds; only an unreachable one pins the entry, which is the fail-open retention used elsewhere.

Also added a regression test that reproduces the slow-join window, and corrected the class Javadoc framing (the window extends through the first registration and a slow join reaches it on a well-behaved clock).

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
brfrn169: Fixed the per-pass `now` sampling in a90203ed0 — it's now sampled per entry, so a slow pass (remote probes) no longer compares late-visited entries against a stale reading and skips ones that expired mid-pass. The extra `currentTimeMillis()` per entry is negligible next to the volatile read and probe already on that path.

On the `closed` check: agreed it's harmless and only skips a single entry rather than terminating the walk. Since `registry.forEach` takes a `BiConsumer` and can't break early, I left it as-is.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
brfrn169: You're right — dropped `@VisibleForTesting` from the constructor, `updateExpirationTime`, and `addParticipant` (all have production callers) and kept it only on `hasParticipant`, which is genuinely test-only. Fixed in 6c37ee392.

core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
brfrn169: Good point — done in 037761af9. Added a **Concurrency** paragraph to both `TwoPhaseCommit.Coordinator` and `TwoPhaseCommit.Participant`: implementations must serialize per-transaction work, and `releaseContext` (plus the participant probe) may be invoked concurrently with any other method for the same transaction ID, while calls for different transaction IDs may proceed concurrently.

Both decorators now cite that interface contract rather than naming `ConsensusCommitCoordinator` / `ConsensusCommitParticipant` as the only evidence, so the dependency is stated where it belongs.

Separately, on the review-summary note that `expiration_time_millis` now means two things across the two roles (an idle timeout on the participant, a probing period on the coordinator) and that the sweep is off by default — agreed; I'll capture that in the operator-facing docs (they live in the separate docs repo, so it's a follow-up there rather than in this PR).

Torch3333 (approved):
LGTM, thank you!

PR #3709, opened 2026-07-13
Release participant contexts when the commit outcome is unknown

brfrn169:
## Description

When `ConsensusCommitCoordinator#commit` surfaces an `UnknownTransactionStatusException` — the COMMITTED-state write failed with an unknown outcome (or its conflict-resolution read-back did), or the ABORTED-state write did on an internal abort path — the records must stay PREPARED for lazy recovery to resolve against the Coordinator state row, so neither `commitRecords` nor `rollbackRecords` may run. The participants' in-memory contexts, however, are already unreachable at that point: commit is terminal on every outcome, so nothing can ever drive those participants again. They previously just sat there until each participant's own idle reaping collected them. This PR releases them promptly instead, and completes the exception vocabulary of the context-releasing paths (including the `releaseContext` → `releaseTransactionContext` rename deferred from #3708).

## Related issues and/or PRs

- #3664 (where `rollbackRecords`' `TransactionNotFoundException` declaration was removed; this PR restores it under redefined semantics — see below)

## Changes made

- Both unknown-status paths in `ConsensusCommitCoordinator#commit` now drive a best-effort `Participant#releaseTransactionContext` — the reap-only terminal that touches no storage — before rethrowing: the commit-state path releases the participants that still hold a context (`toCommit`; write-less participants already self-released), and the abort path releases every registered participant (a participant that never prepared still holds an ACTIVE context there). The durable story is unchanged: PREPARED records are still reconciled by lazy recovery, and a failed release just falls back to the participant's own idle reaping (logged at DEBUG — unlike the record-level best-effort steps, nothing is left locked).
- Declare `throws TransactionNotFoundException, TransactionException` on both `TwoPhaseCommit` `releaseTransactionContext` methods, in line with the rest of the interface: a remote implementation surfaces transport faults through the declared exception and may surface the unknown-transaction no-op through its conventional not-found error mapping — `TransactionNotFoundException` as an alternative carrier of "there was no context to release", which callers treat exactly like a normal return. A time-bounded requirement is documented on the participant side (the release can now run while a client awaits commit's terminal outcome). The in-process implementations remain lenient no-ops.
- Restore `throws TransactionNotFoundException` on `Participant#rollbackRecords`. It was removed in #3664 because the abort path rolls back every participant uniformly, and a self-released participant reporting not-found *as an error* would have produced a spurious records-left-PREPARED WARN on every such abort. That argument was against not-found as an error; redefining it as the documented no-op's alternative carrier — with the best-effort caller explicitly quiet about it — restores the declaration without the spurious logs, and lets a remote participant's conventional error mapping work unmodified instead of requiring a swallow-not-found special case.
- Every reap-style caller treats `TransactionNotFoundException` as the benign outcome it wanted: the coordinator-side sweep's release, both `ActiveTransactionRegistry` disposal paths (participant idle-reap and coordinator cap-eviction), and the new unknown-status release.
- Rename `releaseContext` → `releaseTransactionContext` on both roles (interface, implementations, decorators, callers, tests, Javadoc), aligning the context-releasing terminal with `hasTransactionContext`. Kept as a separate, purely mechanical commit for reviewability.
- Tests: the commit-state and abort-path unknown-status scenarios (release scoping per path, no commit/rollback of records, retried commit rejected as unknown, a failing release neither masking the unknown status nor stopping other releases), not-found tolerance on every reap path, and `rollbackRecords` treating not-found as a benign no-op while still rolling back the other participants.

## Checklist

- This PR is two commits by design: the behavior/contract change, then the mechanical rename. Reviewing them separately is recommended.
- The interface-wide rule after this PR: "definitely no context" may always arrive as `TransactionNotFoundException`, and each caller interprets it per path — the probe reads it as *absent*, `releaseTransactionContext` / `rollbackRecords` read it as a *benign no-op*, prepare/validate read it as an *abort trigger*, and `commitRecords` absorbs it best-effort. A remote implementation can therefore use one conventional not-found mapping for every method.
- A remote implementation of these interfaces (out of scope for this PR) can adopt the renamed methods and signatures as-is, should apply the same not-found tolerance in its own reap-style callers, and must bound the remote release with a deadline per the new Javadoc contract.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
+++ b/core/src/main/java/com/scalar/db/api/TwoPhaseCommit.java
@@ -78,10 +78,11 @@ public interface TwoPhaseCommit {
    *
    * <p><b>Concurrency.</b> Implementations must serialize per-transaction work: for one transaction
    * ID the methods here must be mutually exclusive, so concurrent calls cannot corrupt that
-   * transaction's state. This explicitly covers {@link #releaseContext}, which a context-reaper
-   * drives from its own thread and may therefore invoke concurrently with any other method for the
-   * same transaction ID. Calls for different transaction IDs may proceed concurrently. Decorators
-   * that add a background reaper rely on this contract for their own thread-safety.
+   * transaction's state. This explicitly covers {@link #releaseTransactionContext}, which a
+   * context-reaper drives from its own thread and may therefore invoke concurrently with any other
+   * method for the same transaction ID. Calls for different transaction IDs may proceed
+   * concurrently. Decorators that add a background reaper rely on this contract for their own
+   * thread-safety.
    */
   interface Coordinator extends AutoCloseable {
 
@@ -234,7 +235,7 @@ public interface TwoPhaseCommit {
      *     faults. The context may still be held; reap-style callers treat the failure as
      *     best-effort, since the context is reclaimed by other means (e.g. a later reap or close)
      */
-    void releaseContext(String transactionId)
+    void releaseTransactionContext(String transactionId)
         throws TransactionNotFoundException, TransactionException;
 
     /**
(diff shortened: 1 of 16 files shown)


komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

PR #3712, opened 2026-07-14
Add a disposal-handler constructor to the two-phase commit participant reaper

brfrn169:
## Description

`ActiveTransactionManagedTwoPhaseCommitParticipant`'s idle-expiry reaper releases an abandoned transaction's context by calling `releaseTransactionContext` down the wrapped participant chain, on an internal timer thread that carries no caller credentials. When an embedder interposes a cross-cutting decorator that checks credentials on every call — for example, an authorization decorator — between this decorator and the wrapped participant, that check has no token to see on the reaper thread and rejects the release, so the abandoned context (and its locks) is never reclaimed. The node-local safety-net reaper is silently disabled exactly when it is needed.

The 1PC `ActiveTransactionManagedDistributedTransactionManager` already solves the identical problem by exposing a disposal-handler constructor overload, so an embedder can substitute the reap-driven action and run it in a privileged execution context that bypasses the per-call check. The 2PC participant decorator had no equivalent seam; this PR adds the mirrored overload.

## Related issues and/or PRs

- Mirrors the disposal-handler seam already present on the 1PC `ActiveTransactionManagedDistributedTransactionManager`

## Changes made

- Added a public constructor to `ActiveTransactionManagedTwoPhaseCommitParticipant` that takes an `ActiveTransactionRegistry.DisposalHandler<String>` keyed by transaction ID. When supplied, the reaper invokes it with the transaction ID — for both idle expiry and cap eviction — instead of releasing the context on the wrapped participant itself, so an embedder can wrap the release in its own execution context.
- Kept the existing 3-arg constructor behaviorally unchanged; it is now expressed in terms of the new one, with the default action releasing the wrapped participant's context directly and treating a `TransactionNotFoundException` from it as the already-released no-op it denotes. The handler-supplied path fully replaces that default action, so the embedder owns the release and any not-found tolerance (documented on the constructor).
- The registry entry type stays package-private, so the handler payload is the transaction ID (a `String`) rather than the entry — all the reaper has, and all an embedder needs to close over its own wrapped participant for the release.
- Added a unit test that a supplied disposal handler is invoked with the correct transaction ID on idle expiry and that the default release on the wrapped participant is not called. Backward compatibility of the 3-arg constructor (default release on idle expiry) is covered by the existing reaper test.

## Checklist

- This mirrors the 1PC seam exactly: the supplied handler replaces the default reap action (as `DistributedTransaction::rollback` is replaced on the 1PC side), so an embedder gets full control over how the reap-driven release runs.
- The default (no-handler) path is unchanged, so deployments that do not interpose a credential-checking decorator between this decorator and the wrapped participant are unaffected.
- The symmetric `Coordinator` decorator (`ActiveTransactionManagedTwoPhaseCommitCoordinator`) is intentionally not given the same overload: its reaper releases the coordinator's own context and, in the deployments this addresses, does not sit behind a credential-checking decorator, so the seam has no consumer there. It can be added later if that changes.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitParticipant.java
+++ b/core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitParticipant.java
@@ -43,6 +43,18 @@ import javax.annotation.concurrent.ThreadSafe;
  * traverses the inner decorators via {@code releaseTransactionContext}. It is the participant-side
  * counterpart of {@link ActiveTransactionManagedTwoPhaseCommitCoordinator}.
  *
+ * <p>By default the reaper releases the context by calling {@link
+ * TwoPhaseCommit.Participant#releaseTransactionContext} directly on the wrapped participant. An
+ * embedder that interposes a cross-cutting decorator between this decorator and the wrapped
+ * participant — for example, an authorization decorator that credential-checks every call — may
+ * need that reap-driven release to run in a different execution context than a normal caller-driven
+ * one, because the reaper runs on an internal timer thread that carries no caller credentials. The
+ * {@linkplain #ActiveTransactionManagedTwoPhaseCommitParticipant(TwoPhaseCommit.Participant, long,
+ * int, ActiveTransactionRegistry.DisposalHandler) disposal-handler constructor} lets such an
+ * embedder substitute the reap-driven release action (e.g. wrap it in a privileged mode) while
+ * leaving every other path untouched. This mirrors the seam {@link
+ * ActiveTransactionManagedDistributedTransactionManager} already exposes for its 1PC reaper.
+ *
  * <p>A write-less participant does not always reach {@link #commitRecords}: the Coordinator skips
  * the steps a participant no longer needs, so for such a participant the last driven step is {@link
  * #prepareRecords} (when no validation is required) or {@link #validateRecords} (when it is). To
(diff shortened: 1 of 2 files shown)


komamitsu (approved):
LGTM! 👍

Torch3333 (approved):
LGTM, thank you!

core/src/test/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitParticipantTest.java
```diff
@@ -97,6 +100,33 @@ void join_WhenReleaseThrowsNotFound_ShouldTreatItAsAlreadyGoneAndKeepReaping() t
     verify(delegate, timeout(PAST_SWEEP_MILLIS * 4)).releaseTransactionContext("tx-2");
   }
 
+  @Test
+  void disposalHandler_OnIdleExpiry_ShouldInvokeHandlerWithTransactionId_NotDefaultRelease()
```
feeblefakie, line 104: Test only exercises idle expiry, but the new constructor's Javadoc promises the handler fires for both idle expiry **and** cap eviction. A future change to `ActiveTransactionRegistry` that routes cap eviction through a different path would silently break the embedder use case this seam exists for. Worth adding a companion `_OnCapEviction_` test.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitParticipant.java
```diff
@@ -107,6 +157,16 @@ public ActiveTransactionManagedTwoPhaseCommitParticipant(
     this.registry = registry;
   }
 
+  private static void releaseTransactionContextQuietly(
```
feeblefakie: This helper is `private static`, but the constructor Javadoc explicitly asks the embedder to "treat [TransactionNotFoundException] as the benign no-op it denotes". A conforming embedder writing ``id -> runPrivileged(() -> p.releaseTransactionContext(id))`` will forget the swallow and log `"Failed to handle the expired transaction"` on every already-released context — the seam ends up noisier than the default. Consider exposing this as `public static`, or providing a `defaultReleaseHandler(participant)` factory so embedders can compose it inside their privileged wrapper.

feeblefakie (commented):
Left some comments. PTAL!

core/src/test/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitParticipantTest.java
brfrn169: Fixed in df4c26049b07d6068db232efa8b9c9cb6fa4c3cb. Thanks!

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitParticipant.java
brfrn169: Fixed in c59283785f1ad6d550ce43fe0ae465be6ca3219f. Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #3713, opened 2026-07-15
Persist the write set on validate-phase aborts in the coordinator

brfrn169:
## Description

`ConsensusCommitCoordinator` persists the aggregated write set (keys only) on the COMMITTED Coordinator state row, but always wrote the ABORTED state row with no write set — unlike the 1PC `CommitHandler`, which persists the keys-only write set on its abort path too. The write set records which records a transaction touched, so omitting it on abort leaves an active-recovery / write-set-logging consumer without that information for aborted transactions, and leaves the commit and abort paths asymmetric.

The coordinator can supply a *complete* write set only on the validate-phase abort paths: by then every `prepareRecords` has returned, so the aggregate is the same map the commit path persists. On the prepare-phase abort paths a participant may have thrown before returning its entries, leaving the aggregate incomplete — a partial write set would be worse than none — so those keep writing no write set. This PR persists the write set on the validate-phase aborts only.

## Related issues and/or PRs

- Related to #3689, which introduced `WriteSetDetailLevel` and the keys-only write-set persistence on the commit path that this abort path now mirrors

## Changes made

- `abortAndRollbackRecords` now takes the aggregated `writeSetsByParticipant` map (or `null`), mirroring `commitState`, and `abortState` encodes it here just as `commitState` does. The validate-phase callers pass the complete map; the prepare-phase callers pass `null`.
- Extracted the shared keys-only encoding (`encodeKeysOnlyWriteSet`) used by both `commitState` and `abortState`, so the COMMITTED and ABORTED rows are encoded identically.
- A write-less validate-phase abort persists an empty write set when the row is written at all (coordinator-write omission disabled), symmetric with the write-less commit which likewise persists an empty COMMITTED write set.
- Added a test that a two-writer validate-phase abort persists the same participant-grouped, keys-only write set the commit path encodes; updated the validate-phase abort tests to expect the write set — the write-less ones asserting an empty (but non-null) write set. The prepare-phase abort tests still assert no write set is written.
- Tidied a few pre-existing oddly-wrapped comments in the same test file (words stranded on their own line).

## Checklist

- Prepare-phase aborts intentionally persist no write set: the aggregate is incomplete there (a participant may have thrown mid-`prepareRecords`), the same reason the existing `hasWrites` value is documented as not trustworthy on those paths.
- The write-less validate-phase abort writing an empty write set is a deliberate symmetry with the write-less commit path (which also encodes an empty write set when the row is written) and with the 1PC `CommitHandler`. If persisting nothing (rather than an empty write set) for write-less aborts is preferred, that is a one-line change (`hasWrites ? map : null`).

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
@@ -211,7 +211,8 @@ public class ConsensusCommitCoordinator implements TwoPhaseCommit.Coordinator {
           // transaction write-less. The prepare phase is internal to commit(), so the failure is
           // reported as a commit-level exception: a conflict as CommitConflictException so the
           // caller's retry logic still sees it as retriable, anything else as CommitException.
-          abortAndRollbackRecords(transactionId, context.readOnly, participants);
+          abortAndRollbackRecords(
+              transactionId, context.readOnly, /* writeSetsByParticipant= */ null, participants);
           if (e instanceof PreparationConflictException) {
             throw new CommitConflictException(e.getMessage(), e, e.getTransactionId().orElse(null));
           }
@@ -222,7 +223,8 @@ public class ConsensusCommitCoordinator implements TwoPhaseCommit.Coordinator {
           // records already PREPARED on the other participants, then surface the
           // TransactionNotFoundException as-is (the facade maps it to a retriable conflict). As
           // above, hasWrites is not trustworthy mid-prepare, so only readOnly counts.
-          abortAndRollbackRecords(transactionId, context.readOnly, participants);
+          abortAndRollbackRecords(
+              transactionId, context.readOnly, /* writeSetsByParticipant= */ null, participants);
           throw e;
         }
 
@@ -237,7 +239,7 @@ public class ConsensusCommitCoordinator implements TwoPhaseCommit.Coordinator {
           // abort on the snapshot it owns). The failure is reported as a commit-level exception,
           // mirroring the prepare phase: a conflict as CommitConflictException, anything else as
           // CommitException.
-          abortAndRollbackRecords(transactionId, !hasWrites, participants);
+          abortAndRollbackRecords(transactionId, !hasWrites, writeSetsByParticipant, participants);
           if (e instanceof ValidationConflictException) {
             throw new CommitConflictException(e.getMessage(), e, e.getTransactionId().orElse(null));
           }
@@ -245,7 +247,7 @@ public class ConsensusCommitCoordinator implements TwoPhaseCommit.Coordinator {
         } catch (TransactionNotFoundException e) {
           // A participant no longer knows this transaction while validating. As above, hasWrites
           // is authoritative at this point.
-          abortAndRollbackRecords(transactionId, !hasWrites, participants);
+          abortAndRollbackRecords(transactionId, !hasWrites, writeSetsByParticipant, participants);
           throw e;
         }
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinatorTest.java
```diff
@@ -322,11 +322,12 @@ void commit_WithWrites_WithWriteOmission_WhenValidateFails_ShouldWriteAbortedAnd
...
 
     // Act Assert — PREPARED records exist (hasWrites is true), so the ABORTED state row is written
-    // for lazy recovery even though the omission is enabled, and the records are rolled back.
+    // for lazy recovery even though the omission is enabled, now carrying the keys-only write set
+    // (validate-phase aborts have the complete aggregate), and the records are rolled back.
     assertThatThrownBy(() -> consensusCommitCoordinator.commit("tx-1"))
         .isInstanceOf(CommitConflictException.class)
         .hasCauseInstanceOf(ValidationConflictException.class);
-    verify(coordinatorCommitHandler).abortState("tx-1", null);
+    verify(coordinatorCommitHandler).abortState(eq("tx-1"), notNull());
```
komamitsu: Don't we need to check if the given write-set is the expected one?

core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinatorTest.java
```diff
@@ -373,10 +374,11 @@ void commit_WithWrites_WithWriteOmission_WhenValidateFails_ShouldWriteAbortedAnd
 
     // Act Assert — the exception is rethrown as-is, and PREPARED records exist (hasWrites is
     // true), so the ABORTED state row is written for lazy recovery even though the omission is
-    // enabled; the records are rolled back.
+    // enabled, now carrying the keys-only write set (validate-phase aborts have the complete
+    // aggregate); the records are rolled back.
     assertThatThrownBy(() -> consensusCommitCoordinator.commit("tx-1"))
         .isInstanceOf(TransactionNotFoundException.class);
-    verify(coordinatorCommitHandler).abortState("tx-1", null);
+    verify(coordinatorCommitHandler).abortState(eq("tx-1"), notNull());
```
komamitsu: Same as above

core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinatorTest.java
brfrn169: Fixed in c95d0e0d0a8a2bc0bf91b689973bfcc2d54c1d18. Thanks!

core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinatorTest.java
brfrn169: Fixed in c95d0e0d0a8a2bc0bf91b689973bfcc2d54c1d18. Thanks!

komamitsu (approved):
LGTM! 👍

Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
```diff
@@ -211,7 +211,8 @@ public void commit(String transactionId)
           // transaction write-less. The prepare phase is internal to commit(), so the failure is
           // reported as a commit-level exception: a conflict as CommitConflictException so the
           // caller's retry logic still sees it as retriable, anything else as CommitException.
-          abortAndRollbackRecords(transactionId, context.readOnly, participants);
+          abortAndRollbackRecords(
```
feeblefakie: Two of the four `abortAndRollbackRecords` callers pass `/* writeSetsByParticipant= */ null` — the named-arg comment is a hint that the shape is awkward. Two overloads (a 3-arg form for prepare-phase, the 4-arg form for validate-phase) would make the intent explicit at each call site and drop the null marker.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
```diff
@@ -461,13 +474,27 @@ private CoordinatorContext getContext(String transactionId) throws TransactionNo
...
+
+  // Writes the ABORTED state row via the Coordinator-side handler, persisting the write set encoded
+  // from writeSetsByParticipant (keys only), or none when null. Mirrors commitState, which likewise
+  // takes the aggregated map and encodes it here. See abortAndRollbackRecords for when a
+  // write set is supplied.
+  private void abortState(
+      String transactionId, @Nullable Map<String, List<WriteSetEntry>> writeSetsByParticipant)
+      throws UnknownTransactionStatusException {
+    WriteSet writeSet =
+        writeSetsByParticipant == null ? null : encodeKeysOnlyWriteSet(writeSetsByParticipant);
```
feeblefakie, line 497: `abortState` null-guards the map (`map == null ? null : encodeKeysOnlyWriteSet(map)`), but `commitState` (line 477) calls `encodeKeysOnlyWriteSet(map)` unconditionally. `CoordinatorCommitHandler.commitState` already accepts `@Nullable WriteSet`, so if some future refactor decides to pass `null` from `commitState`, it NPEs. Making both wrappers null-tolerant removes the trap without changing today's behavior.

feeblefakie (commented):
Left some comments. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: Fixed in ac7a675371e56f2a4be8072bba4febe06c71fa75. Thanks!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinator.java
brfrn169: Fixed in 26684fbfdd21f45837fea9977c95c53f53603dda. Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #3714, opened 2026-07-15
Split TwoPhaseCommit into TwoPhaseCommitCoordinator and TwoPhaseCommitParticipant

brfrn169:
## Description

The internal two-phase commit primitives lived as nested interfaces under a single `TwoPhaseCommit` umbrella interface (`TwoPhaseCommit.Coordinator`, `TwoPhaseCommit.Participant`, plus the participant-facing `PreparationResult`, `WriteSetEntry`, and `WriteSetDetailLevel`). Nesting the two roles under one umbrella made the coordinator and participant contracts read as one type; promoting them to independent top-level interfaces lets each stand on its own and keeps the participant-only types with the role they belong to.

## Related issues and/or PRs

- Refactors the `TwoPhaseCommit` interface family introduced across the earlier PRs in this stack (#3660 and onward)

## Changes made

- `TwoPhaseCommit.Coordinator` becomes the top-level `TwoPhaseCommitCoordinator`.
- `TwoPhaseCommit.Participant` becomes the top-level `TwoPhaseCommitParticipant`, and `PreparationResult`, `WriteSetEntry` (with its `Type`), and `WriteSetDetailLevel` are nested inside it, since they belong to the participant's contract (they describe what `prepareRecords` returns).
- The `TwoPhaseCommit` umbrella interface is removed. Its overall-design Javadoc (the two-role model, the identifier-based contract, the typical flow, the lazy-recovery note) is carried onto `TwoPhaseCommitCoordinator`, which drives the protocol; `TwoPhaseCommitParticipant` points to it for that context.
- Everything else is a mechanical update of references, imports, and Javadoc links across `core` and `integration-test`. No behavior change.

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

[...]

PR #3735, opened 2026-07-21
Refine the TwoPhaseCommit coordinator and participant interfaces

brfrn169:
## Description

After the split into the top-level `TwoPhaseCommitCoordinator` / `TwoPhaseCommitParticipant` interfaces (#3714), the contracts still carried shapes inherited from the earlier design: `begin` on the coordinator took a participant argument even though a transaction conceptually begins with no participants, the coordinator's `registerParticipant` did not match the "join" verb used by the participant side (`TwoPhaseCommitParticipant.join`) and `DistributedTransactionManager.join`, and the coordinator's `releaseTransactionContext` declared checked exceptions although it is a pure in-memory reap invoked only in-process. This PR refines the interfaces to their intended final shape.

The interface changes at a glance (all on `TwoPhaseCommitCoordinator`; `TwoPhaseCommitParticipant` has no signature changes):

| Member | Before | After |
|---|---|---|
| `begin` | `begin(transactionId, readOnly, attributes, participant)` | `begin(transactionId, readOnly, attributes)` — a transaction begins with no participants; each participant joins afterward via `joinParticipant` |
| `registerParticipant` | `registerParticipant(transactionId, participant)` | Renamed to `joinParticipant(transactionId, participant)` (exceptions unchanged) — the coordinator begins the transaction, and participants join it |
| `releaseTransactionContext` | `releaseTransactionContext(transactionId) throws TransactionNotFoundException, TransactionException` | `releaseTransactionContext(transactionId)` with no checked exception — a pure in-memory reap, mirroring `close()` |

## Related issues and/or PRs

N/A

## Changes made

- Dropped the participant argument from `TwoPhaseCommitCoordinator.begin`: a transaction is begun without any participant, and each participant joins it afterward via `joinParticipant`.
- Renamed `registerParticipant` to `joinParticipant` on the coordinator, so the primitive-layer verb matches the participant-side `TwoPhaseCommitParticipant.join` and `DistributedTransactionManager.join`. The relationship now reads as: the coordinator begins the transaction, and participants join it.
- Made `releaseTransactionContext` on the coordinator a pure in-memory reap that declares no checked exception (mirroring `close()`), since it is invoked only in-process; the participant-side counterpart keeps its exceptions for remote implementations that genuinely produce them. Simplified the reaper accordingly and dropped the tests that stubbed release to throw.
- Trimmed accumulated history and narrative from the two-phase-commit interfaces, implementations, and decorators.

## Checklist

- `TwoPhaseCommitBackedDistributedTransactionManagerTest` is removed in this PR: it stubbed the old `begin(participant)` / release-throws shapes wholesale. The manager's unit tests are reintroduced in a later PR in this stack, reshaped for the refined interfaces.
- Stacked PR: the "dependent changes merged and published" item is checked on the understanding that this PR merges in stack order after its base (#3714).

## Release notes

N/A


Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM! 👍

feeblefakie (commented):
Left some comments. PTAL!

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedDistributedTransactionManager.java
```diff
@@ -91,7 +91,18 @@ public DistributedTransaction beginReadOnly(String txId, Map<String, String> att
...
-    String canonicalId = coordinator.begin(txId, readOnly, attributes, participant);
+    String canonicalId = coordinator.begin(txId, readOnly, attributes);
+    if (participant != null) {
+      try {
+        coordinator.joinParticipant(canonicalId, participant);
+      } catch (TransactionException e) {
+        // A failed join would leave the just-begun coordinator context orphaned; release it so
+        // begin-then-join is atomic from the caller's perspective. The release is a pure
+        // in-memory reap that no-ops on an unknown transaction and does not throw.
+        coordinator.releaseTransactionContext(canonicalId);
```
feeblefakie, line 102: This new try/catch — release the just-begun coordinator context on failed `joinParticipant` — has no unit test coverage. `TwoPhaseCommitBackedDistributedTransactionManagerTest.java` was deleted (-386 lines) in this PR with no replacement, so the manager class has zero unit tests after this merges. A regression that reorders the throw, passes the wrong id, or forgets the release ships silently. The PR body defers coverage to "a later PR in this stack," but if that PR slips it ships with a hole in newly-added code. Also, ~25 of the 31 deleted tests exercised behavior invariant to the signature change (`rollbackByTransactionId`, `getState`, `resume`, `finishTransaction`, `close`, coordinator-only-mode CRUD) — worth keeping with a couple of edits rather than deleting wholesale.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
```diff
@@ -394,21 +361,23 @@ private void reap(String transactionId, String reason) {
...
-    // wrapped coordinator rejects that registration instead of accepting it onto a context this
-    // reap is destroying.
-    releaseOnWrapped(transactionId);
+    // wrapped coordinator rejects that join instead of accepting it onto a context this reap
+    // is destroying. releaseTransactionContext is a pure in-memory reap that no-ops on an
+    // unknown transaction and does not throw — trusted like the removal below — so it is not
+    // guarded per entry; any unexpected failure is caught by sweepSafely and retried on the next
+    // pass.
+    super.releaseTransactionContext(transactionId);
     registry.remove(transactionId);
```
feeblefakie, line 373: With `releaseOnWrapped` deleted, `super.releaseTransactionContext(transactionId)` and `registry.remove(transactionId)` run without a try/finally. The new "cannot fail" contract on `Coordinator.releaseTransactionContext` justifies the simplification, but if any embedder's decorator (or a future in-tree impl) violates the contract with an unchecked throw, `registry.remove` is skipped, the entry re-expires on the next pass, and the `logger.warn("The transaction is expired and {}; releasing the context...")` re-fires every sweep interval indefinitely. A `try/finally` around the two lines would keep the pass moving.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedDistributedTransactionManager.java
brfrn169: This manager itself is removed later in this stack: #3737 replaces it with a test-only adapter in `integration-test` (`GlobalTransactionBackedDistributedTransactionManager`), which lands together with its own unit tests — including the begin-then-join failure cleanup in its new form (rolling back the just-begun global transaction). So tests restored here would be deleted again two PRs up, which is why I deferred the coverage rather than reshaping the old file. Given the stack merges in order, I'd accept the bounded window instead of re-introducing tests that are about to disappear. WDYT?

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
brfrn169: `releaseTransactionContext` is specified in this PR's refined contract as a pure in-memory reap that does not throw, and this decorator trusts that contract the same way it trusts the wrapped coordinator's serialization contract (see the thread-safety note) rather than guarding individual calls against violations. If an implementation does violate it with an unchecked throw, `sweepSafely` keeps the schedule alive and the entry is retried on the next pass, so the broken implementation stays loudly visible in the WARNs while nothing is silently lost. So I'd like to keep these two lines unguarded, per the contract. WDYT?

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
```diff
@@ -85,61 +79,50 @@ public interface TwoPhaseCommitCoordinator extends AutoCloseable {
...
-   *     faults. You can retry the transaction from the beginning
-   * @throws TransactionException if registering the participant fails due to transient or
-   *     nontransient faults
+   * @param participant the participant to join
+   * @throws TransactionNotFoundException if joining the participant fails due to transient faults.
+   *     You can retry the transaction from the beginning
+   * @throws TransactionException if joining the participant fails due to transient or nontransient
+   *     faults
    */
-  void registerParticipant(String transactionId, TwoPhaseCommitParticipant participant)
```
feeblefakie, line 138: Can you elaborate on the reason for the rename?
I feel `registerParticipant` is more natural than `joinParticipant` from the coordinator's perspective.

feeblefakie (commented):
Sorry, left one more question. PTAL!

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
brfrn169: `joinParticipant` internally invokes `TwoPhaseCommitParticipant.join()`, so I renamed it to match the verb on the participant side. With `registerParticipant`, the same operation would have two different names across the two roles. WDYT?

feeblefakie: The subjects are different (coordinator and participant), so using different verbs is more proper?
Currently, it reads like `a coordinator joins participant`, so it sounds pretty misleading.

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
```diff
@@ -85,61 +79,50 @@ public interface TwoPhaseCommitCoordinator extends AutoCloseable {
...
-   * @param participant the participant to register
-   * @throws TransactionNotFoundException if registering the participant fails due to transient
-   *     faults. You can retry the transaction from the beginning
-   * @throws TransactionException if registering the participant fails due to transient or
-   *     nontransient faults
+   * @param participant the participant to join
+   * @throws TransactionNotFoundException if joining the participant fails due to transient faults.
+   *     You can retry the transaction from the beginning
+   * @throws TransactionException if joining the participant fails due to transient or nontransient
+   *     faults
```
feeblefakie, line 126: Now that `begin` doesn't take a participant, a failed `joinParticipant` leaves the coordinator context alive and the caller has to release it. `TwoPhaseCommitBackedDistributedTransactionManager.beginInternal` does exactly that by hand, so the obligation is real, but it isn't documented anywhere: the `@throws` tags only cover retriability, and the usage example at the top of the interface shows the plain sequence with no cleanup.

This might be a problem for embedders following the docs — they'd orphan a context per failed join. How about stating the obligation here and adding the release to the example?

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
```diff
@@ -211,17 +194,14 @@ void commit(String transactionId)
    * <p>An unknown transaction ID (never begun, or already finished) is a no-op. A decorator may
    * treat this as a terminal step, releasing any per-transaction resources it holds.
    *
+   * <p>Because it only discards in-memory state — with no I/O and no transport hop (unlike {@link
+   * TwoPhaseCommitParticipant#releaseTransactionContext}, which a remote participant may fail or
+   * answer on its not-found channel) — this operation cannot fail and declares no checked
```
feeblefakie: This says the operation cannot fail because there is no I/O and no transport hop, but the intro above (~line 37) says the primitives are identifier-based so callers "may distribute work across processes". A remote coordinator would then have no legal way to report a release failure.

This might be a problem. How about scoping the claim to in-process implementations, or just leaving it out of the contract?

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
```diff
@@ -172,19 +155,19 @@ void registerParticipant(String transactionId, TwoPhaseCommitParticipant partici
    *     retry or roll back; determine the outcome (for example, by checking the coordinator state,
    *     which lazy recovery also relies on) before deciding how to proceed
    * @throws TransactionNotFoundException if no transaction with this ID is registered with this
-   *     Coordinator (never begun, or already finished by a prior commit/rollback), or if a
-   *     registered participant no longer knows the transaction while preparing or validating (its
-   *     local context is gone, e.g. it expired). You can retry the transaction from the beginning
+   *     Coordinator (never begun, or already finished by a prior commit/rollback), or if a joined
```
feeblefakie, line 165: Minor: "registered with this Coordinator" on the line above looks like the last "registered" left after the rename to joined/begun. How about "known to this Coordinator"?

feeblefakie (commented):
Left some comments, PTAL!

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
feeblefakie: One more consequence worth adding to this thread: the throw also escapes `ActiveExpiringMap.forEach`, which is just `cache.asMap().forEach(action)`, so it aborts the whole pass — every other expired transaction later in iteration order goes unreaped too.

So it is not only this entry re-warning every interval; the rest of the sweep is starved behind it. The inline comment saying "any unexpected failure is caught by sweepSafely and retried on the next pass" is true for this entry but not for the others.

This might be a problem. How about a try/finally here after all, or at least a per-entry try/catch in `sweep()` so one bad entry does not truncate the pass?

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
brfrn169: Good catch — the obligation was real and undocumented. Documented in 1e4bbf6b5, at the class level rather than here, since the rule spans begin/commit/rollback:

> Every transaction begun with `begin` must be terminated with either `commit` or `rollback`. This holds on every path out of the flow above, including a failed `joinParticipant`: the failure does not terminate the transaction — the Coordinator's context and any already-joined participants stay alive — so a caller that gives up must still roll it back. An unterminated transaction leaves the Coordinator's per-transaction state in place until the implementation's own reaping reclaims it, if it has any.

One deliberate difference from your suggestion: the prescribed terminal is `rollback`, not `releaseTransactionContext`. The example joins two participants, so if the second join fails, a release would discard only the Coordinator's state and leave the first participant's context to its own timeout, whereas `rollback` drives `rollbackRecords` on the joined participants, which discards their contexts too. And with nothing PREPARED it writes no Coordinator state row and mutates no record, so it costs nothing. That keeps the rule simple: begin, then commit or rollback.

I left the example itself as the happy path and put the rule directly under it. `TwoPhaseCommitBackedDistributedTransactionManager.beginInternal` still calls `releaseTransactionContext`, but that caller moves out of `core` in #3737 and already rolls back there, so I didn't touch it.

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
brfrn169: Agreed. The justification was stated as a physical property ("no I/O and no transport hop"), which is a fact about the in-tree implementation, not something this interface can assert on behalf of every implementation. Rewritten in 1e4bbf6b5 as a contract plus the constraint that makes the contract satisfiable:

> Not part of the normal caller lifecycle, which is `begin` then `commit` or `rollback`. This terminal is driven in-process by whoever owns this Coordinator's per-transaction state — typically a context reaper co-located with it — and is not invoked across a transport boundary.
>
> Implementations must therefore make it a pure in-memory discard that does not fail; that is why it declares no checked exception. An implementation fronting a remote Coordinator should reclaim the remote side's state with that side's own expiration-based reaping rather than turning this call into a wire operation. This is the one asymmetry with `TwoPhaseCommitParticipant#releaseTransactionContext`, which the Coordinator drives on participants that may be remote and which may therefore fail or answer on its not-found channel.

The asymmetry with the participant side is now explained by who drives it rather than by physics, so it no longer conflicts with the intro. I left the intro itself alone.

core/src/main/java/com/scalar/db/common/ActiveTransactionManagedTwoPhaseCommitCoordinator.java
brfrn169: You're right about the mechanics: `ActiveExpiringMap.forEach` is `cache.asMap().forEach(action)`, so an escaping throw aborts the pass and every entry after it waits for the next one. And the inline comment did read as though the pass survives. Corrected in 1e4bbf6b5:

> ... so it is not guarded per entry. If an implementation broke that contract, the throw would abort the whole pass, leaving the entries after this one for the next one; `sweepSafely` keeps the schedule alive and logs the failure.

On the guard itself, I'd still rather not add one. Reaching it takes an unchecked throw out of `releaseTransactionContext`, which the refined contract now states as an obligation ("Implementations must therefore make it a pure in-memory discard that does not fail"). This interface is internal — "Breaking changes can and will be introduced; users should not depend on it" — so the implementations behind it are ours rather than an embedder extension point, and in-tree the wrapped call is a `ConcurrentHashMap` get, a flag write and a remove, with no throw path. Guarding it would mean writing defensive code against a contract we own and have just made explicit; I'd rather let a violation surface as a failed sweep in the log. WDYT?

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
brfrn169: Done in 1e4bbf6b5. Not a rename leftover, as it turns out — the subject here is the transaction rather than a participant, and it's the only "registered" in the file. But "known to" is the better word anyway, for consistency: `rollback` and `releaseTransactionContext` both describe the same condition as "an unknown transaction ID (never begun, or already finished)", and the second half of this very sentence says a joined participant "no longer knows the transaction". Now the Coordinator knowing and the participant knowing line up.

core/src/main/java/com/scalar/db/api/TwoPhaseCommitCoordinator.java
brfrn169: Sorry — my previous reply was too thin. "It matches the verb on the participant side" doesn't actually answer your point about the subjects, so let me lay out the whole reasoning.

- **`join` is transitive here.** `joinParticipant(transactionId, participant)` reads as "join this participant to the transaction identified by `transactionId`": the participant is the object, and the ID argument carries the destination. The Javadoc's first line already puts it that way — "Joins a participant to the transaction."
- **The bare verb and the verb-plus-object are different constructions.** `TwoPhaseCommitParticipant.join(transactionId, ...)` names no object, so the subject joins. `TwoPhaseCommitCoordinator.joinParticipant(transactionId, participant)` names its object, so the subject joins that object to the transaction. The two subjects do get different readings, and the name is what distinguishes them — this is the part I failed to explain last time.
- **The operation is idempotent per participant ID.** A repeat call is a no-op. That fits "join" — a membership state that converges — better than "register", where repeating a registration is normally an error or a duplicate entry.

WDYT?

brfrn169: As discussed offline, we will rename the method from `joinParticipant` to `enlist` in a separate PR. Thank you for the discussion!

feeblefakie: Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #3736, opened 2026-07-21
Extract the two-phase-commit primitive getters into TwoPhaseCommitFactory

brfrn169:
## Description

`TwoPhaseCommitCoordinator` and `TwoPhaseCommitParticipant` are internal primitives — their Javadoc says users should not depend on them — yet their getters sat on the application-facing `TransactionFactory`. This PR moves them to a separate factory so `TransactionFactory` stays limited to the application-facing types.

## Related issues and/or PRs

N/A

## Changes made

- Removed the `TwoPhaseCommitCoordinator` / `TwoPhaseCommitParticipant` getters from `TransactionFactory`.
- Added `TwoPhaseCommitFactory`, which exposes those getters for the infrastructure code that wires the primitives.

## Checklist

- Pure relocation of the getters; the wiring they return is unchanged, so the existing tests cover it.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/service/TransactionFactory.java
+++ b/core/src/main/java/com/scalar/db/service/TransactionFactory.java
@@ -2,8 +2,6 @@ package com.scalar.db.service;
 
 import com.scalar.db.api.DistributedTransactionAdmin;
 import com.scalar.db.api.DistributedTransactionManager;
-import com.scalar.db.api.TwoPhaseCommitCoordinator;
-import com.scalar.db.api.TwoPhaseCommitParticipant;
 import com.scalar.db.api.TwoPhaseCommitTransactionManager;
 import com.scalar.db.config.DatabaseConfig;
 import java.io.File;
@@ -18,7 +16,7 @@ import javax.annotation.Nullable;
  * A factory class to instantiate {@link DistributedTransactionManager} and {@link
  * DistributedTransactionAdmin} and {@link TwoPhaseCommitTransactionManager}
  */
-public class TransactionFactory {
+public final class TransactionFactory {
   private final DatabaseConfig config;
 
   /**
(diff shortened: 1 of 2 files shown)


Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/service/TransactionFactory.java
```diff
@@ -18,7 +16,7 @@
  * A factory class to instantiate {@link DistributedTransactionManager} and {@link
  * DistributedTransactionAdmin} and {@link TwoPhaseCommitTransactionManager}
  */
-public class TransactionFactory {
+public final class TransactionFactory {
```
feeblefakie, line 19: The `public class` → `public final class` change looks like unrelated scope creep — a source and binary compat break for any downstream subclass. If no one currently subclasses in-repo, that doesn't cover Enterprise/cluster/user code, and it's not called out in the PR body. Suggest dropping the `final` unless the compat break is deliberate.

Also in this file: `getTwoPhaseCommitCoordinator()` and `getTwoPhaseCommitParticipant()` were `public` methods on a `public` class. Their return-type Javadoc says the returned types are internal primitives, but the methods themselves were part of the ABI. Callers written against them get `NoSuchMethodError` at runtime after upgrade. Consider keeping them as `@Deprecated` bridges that delegate to `new TwoPhaseCommitFactory(config).getXxx()`, or splitting the removal into a major-version PR.

core/src/main/java/com/scalar/db/service/TwoPhaseCommitFactory.java
```diff
@@ -0,0 +1,92 @@
...
+    return ProviderManager.createTwoPhaseCommitParticipant(config);
+  }
+
+  /**
+   * Returns a {@link TwoPhaseCommitFactory} instance.
+   *
+   * @param properties properties to use for configuration
+   * @return a {@link TwoPhaseCommitFactory} instance
+   */
+  public static TwoPhaseCommitFactory create(Properties properties) {
```
feeblefakie, line 56: All four static `create(...)` overloads (`Properties`/`Path`/`File`/`String`), the `DatabaseConfig` field, and the private constructor are literal copies of `TransactionFactory`'s. A change in one (validation, new overload, exception mapping) will silently drift. An accessor on the existing factory — `TransactionFactory#getTwoPhaseCommitFactory()` (constructed with the shared `DatabaseConfig`) — plus a package-private `TwoPhaseCommitFactory(DatabaseConfig)` constructor would drop ~30 LOC and give both factories one source of properties parsing. Users who need both factories today parse and instantiate `DatabaseConfig` twice.

feeblefakie (commented):
Left some comments. PTAL!

core/src/main/java/com/scalar/db/service/TransactionFactory.java
brfrn169: On `final`: I actually think `TransactionFactory` should have been `final` from the start — it's a static-factory-style class whose only public constructor has long been deprecated, and I don't expect anyone subclasses it externally, so I took this chance to formalize that.

On the removed getters: they have never been part of a release — they were added by #3662 after v3.18.0 was cut — so no released caller can hit `NoSuchMethodError`. The only consumer is unreleased work in a separate repository, which switches to `TwoPhaseCommitFactory` in a paired change. So I think we can remove them cleanly without a deprecation bridge.

core/src/main/java/com/scalar/db/service/TwoPhaseCommitFactory.java
brfrn169: I'd like to keep the two factories independent: the overloads are one-line delegations to the `DatabaseConfig` constructors, so the drift surface is small in practice, and routing through `TransactionFactory#getTwoPhaseCommitFactory()` would put a reference to the internal-primitives factory back onto the application-facing factory — the very coupling this PR removes. If the duplication still feels heavy, I'm happy to trim `TwoPhaseCommitFactory` down to just the `create(Properties)` overload instead. WDYT?

feeblefakie: I lean toward having just the create(Properties), but it's internal APIs, so I think we can refactor them later as necessary.
Thanks!

feeblefakie (approved):
LGTM! Thank you!

PR #3737, opened 2026-07-21
Add GlobalTransactionManager, GlobalTransaction, and BranchTransaction

brfrn169:
## Description

Applications need an application-facing way to run one transaction across multiple processes without depending on the internal two-phase-commit primitives (`TwoPhaseCommitCoordinator` / `TwoPhaseCommitParticipant`, whose Javadoc forbids user code from touching them). This PR adds that layer: a global transaction that is begun once, joined by any number of branches (each branch doing CRUD in its own process against its own participant), and committed or rolled back as a whole through the global handle.

### Sample code

Service 1 (the initiator) begins the global transaction, works on its own branch, and drives the final commit:

```java
GlobalTransactionManager manager =
    TransactionFactory.create(properties).getGlobalTransactionManager();

GlobalTransaction global = manager.begin();
try {
  // Work on this service's own branch, then end it.
  BranchTransaction branch = manager.beginBranch(global.getId());
  Optional<Result> result = branch.get(get);
  branch.put(put);
  branch.end();

  // Pass global.getId() to the other services (e.g., via an RPC call) and wait until their
  // branches have done their work and ended.
  callOtherService(global.getId());

  // Commit the global transaction as a whole.
  global.commit();
} catch (Exception e) {
  global.rollback();
}
```

Service 2 (another process) joins the same global transaction by its ID and only does CRUD — commit/rollback stays with the initiator's global handle:

```java
GlobalTransactionManager manager =
    TransactionFactory.create(properties).getGlobalTransactionManager();

BranchTransaction branch = manager.beginBranch(transactionId);
Optional<Result> result = branch.get(get);
branch.insert(insert);
branch.end();
// Return to the caller; the initiator commits or rolls back the global transaction.
```

## Related issues and/or PRs

N/A

## Changes made

- Added the API interfaces in `com.scalar.db.api`:
  - `GlobalTransactionManager` — `begin` / `beginReadOnly` (with optional transaction attributes) and `beginBranch(transactionId)` (with optional per-branch attributes).
  - `GlobalTransaction` — the overall handle: `getId` / `commit` / `rollback`. It holds no data-store resources itself.
  - `BranchTransaction` — a CRUD-only handle (`extends TransactionCrudOperable`) for one branch of the global transaction.
- Added `TwoPhaseCommitBackedGlobalTransactionManager`: backs the API with the two-phase-commit primitives — `begin` begins on the coordinator, `beginBranch` joins a participant, and `commit` drives the full 2PC across the joined participants.
- Added `DistributedTransactionManager-backed` `GlobalTransactionManager` for the fully-shared deployment, symmetric to the above: `begin` begins a distributed transaction, `beginBranch` joins it by ID, and all branches share the single underlying transaction.
- Added `DecoratedGlobalTransaction` / `DecoratedGlobalTransactionManager` as the decorator bases.
- Per-branch attributes passed to `beginBranch` are propagated to the underlying participant/manager.
- Tests:
  - `GlobalTransactionTestBase` — integration tests exercising the API directly (multi-branch commit/rollback, conflict at commit rolling back all branches, etc.), with per-storage test classes for both backings.
  - `GlobalTransactionBackedDistributedTransactionManager(Test)` in `integration-test` — a test-only adapter that runs the existing `DistributedTransaction` suite on top of the Global layer (replacing the previous core-resident `TwoPhaseCommitBackedDistributedTransaction(Manager)`), plus its unit tests.

## Checklist

- The Global layer is deliberately thin: it adds no recovery, retry, or coordinator-state semantics of its own — those stay in the underlying primitives.

## Release notes

N/A


Torch3333 (approved):
LGTM, thank you!

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseCommitBackedConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -1,260 +0,0 @@
...
-      // Assert: with omission disabled, a COMMITTED Coordinator state row is written even for the
-      // read-only transaction.
-      Optional<CoordinatorStateAccessor.State> state = coordinator.getState(readOnlyId);
-      assertThat(state).isPresent();
-      assertThat(state.get().getState()).isEqualTo(TransactionState.COMMITTED);
-    }
-  }
-
-  @Test
-  public void commit_TransactionExceedingExpirationWithContinuousCrud_ShouldCommitSuccessfully()
```
komamitsu, line 170: Is this scenario covered elsewhere in this PR?

core/src/main/java/com/scalar/db/api/BranchTransaction.java
```diff
@@ -0,0 +1,36 @@
...
+   *
+   * <p>Every branch must call this exactly once, after its last CRUD operation and before the
+   * owning {@link GlobalTransaction} is committed or rolled back. Issuing CRUD on this branch once
+   * {@code end()} has been called is not allowed.
+   *
+   * @throws CrudConflictException if ending the branch fails due to transient faults (e.g., a
+   *     conflict). You can retry the transaction from the beginning
+   * @throws CrudException if ending the branch fails due to transient or nontransient faults
+   */
+  void end() throws CrudConflictException, CrudException;
```
komamitsu, line 39: > `Issuing CRUD on this branch once {@code end()} has been called is not allowed.`

This Javadoc sounds like CRUD operations following `end()` should fail, but it's not actually validated. Is the inconsistency acceptable?

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedGlobalTransactionManager.java
```diff
@@ -0,0 +1,94 @@
...
+    BranchTransaction branch =
+        new TwoPhaseCommitBackedBranchTransaction(participant, transactionId);
+    return attributes.isEmpty()
+        ? branch
+        : new AttributePropagatingBranchTransaction(branch, attributes);
+  }
+
+  @Override
+  public void close() {
+    coordinator.close();
```
komamitsu: Don't we need to wrap `coordinator.close()` with a `try` block and close the participant in the `finally` block?

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/TwoPhaseCommitBackedConsensusCommitSpecificIntegrationTestBase.java
brfrn169: It wasn't — the probe keep-alive was covered only at the unit level, and this end-to-end scenario was lost with this file. Restored it as a GlobalTransaction API test in d1cf7eca952986dd62700a80fd9b8b4e040bf452: `commit_GlobalTransactionExceedingExpirationWithContinuousCrud_ShouldCommitSuccessfully` in `TwoPhaseCommitBackedConsensusCommitGlobalTransactionTestBase` keeps issuing CRUD for three expiration periods and asserts the global transaction still commits. (It lives in the two-phase-commit backing's test base rather than the shared corpus because the scenario is specific to that backing's coordinator reaper.)

core/src/main/java/com/scalar/db/api/BranchTransaction.java
brfrn169: Good point — it wasn't enforced. Fixed in 5fc6041abc830e71098e2727b80ad0be5f38e528: both backings now track the ended state and reject CRUD after `end()` — and a second `end()` — with `IllegalStateException`, and the Javadoc now states that explicitly.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedGlobalTransactionManager.java
brfrn169: Good catch — fixed in 19fbd3024354f4e9188a78ac9deb33716cc51f79: the participant is now closed in a `finally` block, with a unit test covering a coordinator whose `close()` throws.

core/src/main/java/com/scalar/db/api/BranchTransaction.java
komamitsu: Just in case, is `end()` potentially called from a different thread? If so, `volatile` or something should be used.

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/BranchTransaction.java
brfrn169: A branch handle is meant to be driven by one thread at a time, like the other transaction handles (e.g., `ConsensusCommit` / `TwoPhaseConsensusCommit`, which are `@NotThreadSafe`); under that contract, the caller's handoff provides the happens-before, so the plain field is intentional. To make the contract explicit, I annotated the global/branch transaction handles with `@NotThreadSafe` — and the manager backings with `@ThreadSafe` — in 19a63fb968f7d451e8f9ff96c4d71732199fa851.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedGlobalTransaction.java
```diff
@@ -0,0 +1,61 @@
...
+  public void commit() throws CommitException, UnknownTransactionStatusException {
+    // Drive the full 2PC; the coordinator generates the prepared/committed timestamps internally.
+    // The coordinator already reports prepare/validate failures as CommitConflictException /
+    // CommitException, so those propagate as-is; only an unknown transaction needs translating.
+    try {
+      coordinator.commit(transactionId);
+    } catch (TransactionNotFoundException e) {
+      // The coordinator no longer knows this transaction (e.g., it expired). Surface it as a
+      // retriable commit conflict so the caller can restart the transaction from the beginning.
+      throw new CommitConflictException(e.getMessage(), e, transactionId);
```
feeblefakie, line 51: This translates every `TransactionNotFoundException` from `coordinator.commit` into `CommitConflictException` (retriable). But `TwoPhaseCommitCoordinator.commit`'s Javadoc explicitly says `TransactionNotFoundException` covers "never begun, or already finished by a prior commit/rollback". So a duplicate `global.commit()` (retry wrapper, resume-after-crash, application bug) surfaces as retriable — the caller reruns their business logic on top of an already-committed state. `UnknownTransactionStatusException` is the correct signal for this indeterminate case.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedBranchTransaction.java
```diff
@@ -0,0 +1,169 @@
...
+  @Override
+  public List<CrudOperable.BatchResult> batch(List<? extends Operation> operations)
+      throws CrudException {
+    return crud(() -> participant.batch(transactionId, operations));
+  }
+
+  @Override
+  public void end() throws CrudException {
+    checkNotEnded();
+    ended = true;
```
feeblefakie, line 176: `end()` just flips a local `ended` boolean — no coordinator/participant notification. The initiator can call `global.commit()` while a straggler branch is mid-CRUD (RPC race, retry, timeout); the coordinator then issues `prepareRecords` to a participant with in-flight CRUD and blows up with `IllegalStateException` from deep in the primitive, or corrupts the per-transaction context. The API sells "one transaction across multiple processes" but ships no primitive to enforce the precondition.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedBranchTransaction.java
```diff
@@ -0,0 +1,169 @@
...
+ * default target. See {@link TwoPhaseCommitBackedGlobalTransactionManager} for how the participant
+ * is wired and the branch is begun.
+ */
+@NotThreadSafe
+public class TwoPhaseCommitBackedBranchTransaction implements BranchTransaction {
+
+  private final TwoPhaseCommitParticipant participant;
+  private final String transactionId;
+
+  private boolean ended;
```
feeblefakie, line 50: `ended` is a plain (non-volatile) field. The documented flow has one thread (initiator) drive `commit()` after another thread (branch) sets `ended=true`. No happens-before edge exists, so on weaker memory models the initiator may see stale `ended` and drive commit against still-in-flight CRUD. Same issue in `DistributedTransactionBackedBranchTransaction`.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedBranchTransaction.java
```diff
@@ -0,0 +1,169 @@
...
+    return crud(() -> participant.get(transactionId, get));
+  }
+
+  @Override
+  public List<Result> scan(Scan scan) throws CrudException {
+    return crud(() -> participant.scan(transactionId, scan));
+  }
+
+  @Override
+  public TransactionCrudOperable.Scanner getScanner(Scan scan) throws CrudException {
```
feeblefakie, line 80: Scanner leak past `end()`. A caller obtains a scanner before `end()`; subsequent `one()` / `all()` / iteration calls delegate directly to the participant and bypass `checkNotEnded()`. Wrap the returned scanner so each access checks branch state. Same issue in `DistributedTransactionBackedBranchTransaction.getScanner`.

core/src/main/java/com/scalar/db/common/DistributedTransactionBackedGlobalTransactionManager.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @Override
+  public BranchTransaction beginBranch(String transactionId, Map<String, String> attributes)
+      throws TransactionException {
+    // Join the already-begun global transaction by its ID. All branches share the single underlying
+    // distributed transaction. The per-branch attributes are propagated client-side into each CRUD
+    // operation by AttributePropagatingBranchTransaction; they are not sent to join (which carries
+    // no attributes) and differ from the transaction-scoped attributes supplied at beginGlobal.
+    BranchTransaction branch =
+        new DistributedTransactionBackedBranchTransaction(manager.join(transactionId));
```
feeblefakie, line 90: `manager.join(transactionId)` returns the same underlying `DistributedTransaction`/`ConsensusCommit` instance for every branch, and `ConsensusCommit` is `@NotThreadSafe`. The Javadoc invites multi-branch usage (in-process orchestration or misuse from separate threads); concurrent CRUD across two branches on different threads will silently corrupt the shared `Snapshot`/`CrudHandler` or NPE on `HashMap` mutation. The `@NotThreadSafe` annotation is on the branch class, but the API surface doesn't communicate that two branches share the same non-thread-safe object.

core/src/main/java/com/scalar/db/common/DistributedTransactionBackedGlobalTransactionManager.java
```diff
@@ -0,0 +1,93 @@
...
+ * therefore assumes the branches of a global transaction operate on disjoint data: under that
+ * assumption, sharing one context has no functional impact, since there is no read/write
+ * interference and the combined validation is equivalent to per-branch validation. The assumption
+ * is a precondition on the caller rather than something this class enforces. Branches that do touch
+ * the same records observe each other's uncommitted writes, unlike the per-participant snapshots
+ * that {@link TwoPhaseCommitBackedGlobalTransactionManager} gives each branch.
+ *
+ * <p>Because branches join by transaction ID, this backing requires a {@link
+ * DistributedTransactionManager} whose {@link DistributedTransactionManager#join(String)} resolves
+ * a transaction begun elsewhere. Running branches in separate processes therefore requires an
```
feeblefakie: This claim isn't achievable with any shipped provider. `DistributedTransactionManager.join(id)` defaults to `resume` (in-process only), and Consensus Commit / JDBC / SingleCrudOperation all use it. A second-service process calling `beginBranch` immediately throws `TransactionNotFoundException`. Either the doc is wrong, or the DT backing genuinely only supports in-process use.

feeblefakie (commented):
Left some comments. PTAL!

core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
```diff
@@ -117,4 +118,10 @@ public final TwoPhaseCommitParticipant createTwoPhaseCommitParticipant(DatabaseC
 
   protected abstract TwoPhaseCommitParticipant createRawTwoPhaseCommitParticipant(
       DatabaseConfig config);
+
+  @Override
+  public GlobalTransactionManager createGlobalTransactionManager(DatabaseConfig config) {
+    return new TwoPhaseCommitBackedGlobalTransactionManager(
```
feeblefakie, line 124: `DistributedTransactionBackedGlobalTransactionManager` (93 LOC + branch/global classes ~186 LOC) is never wired here — this method unconditionally returns the 2PC-backed manager. The DT backing is only referenced by an integration-test class. Either wire the config toggle in this PR or move the whole DT backing into `integration-test/src/main` (matching the pattern of `GlobalTransactionBackedDistributedTransactionManager`).

core/src/main/java/com/scalar/db/api/DistributedTransactionProvider.java
```diff
@@ -68,4 +68,14 @@ public interface DistributedTransactionProvider {
...
+
+  /**
+   * Creates an instance of {@link GlobalTransactionManager}.
+   *
+   * @param config a database config
+   * @return an instance of {@link GlobalTransactionManager}
+   * @throws UnsupportedOperationException if the transaction manager does not support global
+   *     transactions
+   */
+  GlobalTransactionManager createGlobalTransactionManager(DatabaseConfig config);
```
feeblefakie, line 80: Adding this as an abstract method breaks source compat for every external `DistributedTransactionProvider` implementation, even ones that don't support global transactions. Make it a `default` method that throws `UnsupportedOperationException`; providers that support it can override.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedGlobalTransactionManager.java
```diff
@@ -0,0 +1,99 @@
...
+    if (participant == null) {
+      throw new UnsupportedOperationException(
+          CoreError.COORDINATOR_ONLY_GLOBAL_TRANSACTION_MANAGER_BRANCH_NOT_SUPPORTED
+              .buildMessage());
+    }
+    // Join the in-process participant to the global transaction. joinParticipant establishes the
+    // participant's local context, forwarding the readOnly flag and the transaction-scoped
+    // attributes supplied at beginGlobal. The per-branch attributes passed here are propagated
+    // client-side into each CRUD operation by AttributePropagatingBranchTransaction.
+    coordinator.joinParticipant(transactionId, participant);
```
feeblefakie, line 84: Comment above says "joining is idempotent per participant ID," but `beginBranch` will happily hand out a second `BranchTransaction` sharing the same underlying participant context. `end()` on one doesn't affect the other's `ended` flag, so CRUD after "the" `end()` still succeeds via the other handle — the `BranchTransaction.end()` contract ("exactly once … rejected with `IllegalStateException`") isn't enforced across handles.

core/src/main/java/com/scalar/db/api/GlobalTransactionManager.java
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * Begins a new global transaction. This method is an alias of {@link #beginGlobal()}.
+   *
+   * @return the {@link GlobalTransaction} handle used to drive the transaction's outcome
+   * @throws TransactionNotFoundException if the transaction fails to begin due to transient faults.
+   *     You can retry the transaction from the beginning
+   * @throws TransactionException if the transaction fails to begin due to transient or nontransient
+   *     faults
+   */
+  default GlobalTransaction startGlobal()
```
feeblefakie: The six `start*` aliases (`startGlobal`, `startGlobalReadOnly`, `startBranch`, with/without attributes) are pure forwarders to `beginXxx`. This is a brand-new interface — no legacy shim to keep. Interface bloat forever for zero v1 benefit; every implementer/decorator has to document and maintain three synonyms per operation.

core/src/main/java/com/scalar/db/api/BranchTransaction.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * The handle for one branch of a global transaction, obtained via {@link
+ * GlobalTransactionManager#beginBranch(String)}.
+ *
+ * <p>A branch performs CRUD (inherited from {@link TransactionCrudOperable}) against its own
+ * portion of the data. The transaction's overall outcome — commit or rollback — is driven by the
+ * owning {@link GlobalTransaction}, not by the branch.
+ */
+public interface BranchTransaction extends TransactionCrudOperable {
```
feeblefakie, line 14: `extends TransactionCrudOperable` drags in `put(Put)`, `put(List)`, `delete(List)` — all `@Deprecated As of release 3.13.0. Will be removed in release 4.0.0`. So this brand-new API ships already-deprecated methods, and every implementer/decorator (4 in this PR alone) has to carry them.

core/src/main/java/com/scalar/db/api/GlobalTransactionManager.java
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * Begins a new global transaction.
+   *
+   * @return the {@link GlobalTransaction} handle used to drive the transaction's outcome
+   * @throws TransactionNotFoundException if the transaction fails to begin due to transient faults.
+   *     You can retry the transaction from the beginning
+   * @throws TransactionException if the transaction fails to begin due to transient or nontransient
+   *     faults
+   */
+  default GlobalTransaction beginGlobal()
```
feeblefakie: Since it's a `GlobalTransactionManager`, `beginGlobal()` seems redundant. How about just making it `begin()`?

core/src/main/java/com/scalar/db/api/DistributedTransactionProvider.java
brfrn169: The implementers of this SPI are assumed to be internal components only — I'm not aware of any external `DistributedTransactionProvider` implementations — and the internal ones will be updated in a paired change. Keeping it abstract also makes every implementation state its decision explicitly (the three in-tree providers each throw with their own error code today) rather than silently inheriting an unsupported default. So I'd like to keep it as is. WDYT?

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedBranchTransaction.java
brfrn169: Good catch. I addressed it from the other end: instead of guarding every scanner access after `end()`, `end()` itself now fails with `IllegalStateException` while a scanner obtained from this branch is still open — mirroring how consensus commit rejects `commit()` with an unclosed scanner. A scanner can then never legitimately outlive `end()`, which closes the bypass at its source. Fixed in bcaf29d530306a0044740cb1d751cb0144424da0.

core/src/main/java/com/scalar/db/common/DistributedTransactionBackedGlobalTransactionManager.java
brfrn169: Documented in d29087bfb3ba1746e40382d45dd669dfe4cfd1b5: the sharing paragraph now states explicitly that every branch handle drives the same non-thread-safe `DistributedTransaction`, so CRUD on different branches of one global transaction must never run concurrently — branches are driven one at a time, with the caller providing the happens-before when they hop threads. The branch class also points there now.

core/src/main/java/com/scalar/db/common/DistributedTransactionBackedGlobalTransactionManager.java
brfrn169: You're right that the claim overreached. Stepping back, `beginBranch`'s contract is just "begin a branch of the identified global transaction" — resolving it via `DistributedTransactionManager#join` is an implementation detail of this backing, and the Javadoc wrongly promoted it to a contract-level requirement. Reworded in d29087bfb3ba1746e40382d45dd669dfe4cfd1b5: the backing is now documented as in-process orchestration by design (the shared transaction lives in the wrapped manager instance; the separated, multi-process arrangement is what the two-phase-commit backing is for), the join mention is demoted to an implementation note, and the interface-level wording (`GlobalTransactionManager` / `GlobalTransaction`) is mechanism-neutral now.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedGlobalTransaction.java
brfrn169: The already-finished case can't actually arise through this API: there is deliberately no way to obtain a `GlobalTransaction` handle for an existing transaction ID (no global-side resume/join), so the only committer is the handle returned by `begin()`. A duplicate `global.commit()` is therefore only possible by calling `commit()` twice on that same handle — a contract violation, not a recovery path — and the resume-after-crash scenario is not expressible. The case this mapping is actually about is the coordinator context reaped after idle expiry, where nothing has committed and retrying from the beginning is exactly right; `UnknownTransactionStatusException` there would force an outcome check whose answer is already known. So I'd like to keep the mapping as is. WDYT?

core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionProvider.java
brfrn169: This is intentional: the provider wires the backing suited to the library's own in-process composition, while the single-phase backing is meant to be constructed directly by the infrastructure that assembles the fully-shared deployment — the same way the two-phase-commit primitives are obtained through `TwoPhaseCommitFactory` rather than wired into the application-facing path. A config toggle here would add a knob whose right answer is decided by the deployment, not by application configuration, so I'd rather not introduce one. The integration-test suite is the in-tree consumer that keeps the backing exercised until that external wiring lands. So I'd like to keep it in core as is. WDYT?

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedGlobalTransactionManager.java
brfrn169: Right — the exactly-once bookkeeping is per handle. I didn't want to put manager-side tracking behind it: the manager never observes commit/rollback, so an id-to-handle map would have no cleanup hook, and the enforcement machinery would be disproportionate for a misuse guard. Documented on both backings in 05324c8e1f907277d4da29820244c883bd71d94b instead: a repeated `beginBranch` for the same transaction returns a new handle fronting the same underlying context, so each branch is to be begun once and driven through that one handle. WDYT?

core/src/main/java/com/scalar/db/api/GlobalTransactionManager.java
brfrn169: Implementers and decorators don't actually carry these: the `start*` methods are `default` aliases forwarding to the corresponding `beginXxx`, so an implementation implements only the three abstract methods (`begin(attributes)` / `beginReadOnly(attributes)` / `beginBranch(id, attributes)`) either way, and `DecoratedGlobalTransactionManager` deliberately forwards only the abstract methods and inherits the defaults — dropping the aliases wouldn't change what anyone implements or maintains. I kept them for symmetry with `DistributedTransactionManager`'s begin/start convention, so the cost is interface surface only. That said, they're trivial to drop now and impossible later, so if you'd rather ship the new API lean, I'm happy to remove them. WDYT?

core/src/main/java/com/scalar/db/api/GlobalTransactionManager.java
brfrn169: Agreed — renamed in 5a1149cb2c8fc0aae3b450aa096fc1eadb9e5dff: `begin()` / `beginReadOnly()` (the `start` aliases follow the same shape), alongside the unchanged `beginBranch()`.

core/src/main/java/com/scalar/db/api/BranchTransaction.java
brfrn169: That's transitional: master is 4.0.0-SNAPSHOT, and those methods are scheduled to be removed from `TransactionCrudOperable` itself for 4.0.0 — `BranchTransaction` sheds them then automatically, along with every other CRUD surface. Defining a separate deprecated-free CRUD interface just for branches would fork the CRUD surface permanently to fix a pre-release cosmetic, so I'd rather keep inheriting the shared one. WDYT?

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedBranchTransaction.java
brfrn169: You're right that this can surface as a partial commit: if the initiator commits while a branch is still working, the commit proceeds with the writes staged so far, and the straggler's rejected operation doesn't stop a commit already in flight. Driving commit only after every branch has finished has always been the application's responsibility with the two-phase commit interface family, and this API keeps that contract rather than adding a coordination primitive — `end()` is the lifecycle hook where a backing that can enforce more (e.g., a buffering remote one) gets to do so. I documented the precondition explicitly on `GlobalTransaction#commit()` in c114bf1ae4c0bd2ab27c70adc0a5b063722f665e.

core/src/main/java/com/scalar/db/common/TwoPhaseCommitBackedBranchTransaction.java
brfrn169: `global.commit()` never reads the `ended` flag — it lives on the branch handle purely to reject calls made through that handle, so there is no path where a stale read lets commit proceed against in-flight CRUD. As for the flag itself, a branch handle follows the same contract as the other transaction handles (`ConsensusCommit` etc., all `@NotThreadSafe` with plain fields): it is driven by one thread at a time, and a caller that hands it across threads provides the happens-before — that contract is stated by the `@NotThreadSafe` annotations. The commit-side precondition (commit only after every branch has ended) is now documented on `GlobalTransaction#commit()` as well (see the other thread).

feeblefakie (approved):
LGTM! Thank you!

PR #3738, opened 2026-07-21
Revert "Widen DistributedTransactionManager.join to throw TransactionException (#3663)"

brfrn169:
## Description

#3663 widened `DistributedTransactionManager.join(String)` from `throws TransactionNotFoundException` to `throws TransactionException` for its sole intended consumer: the Coordinator-backed `DistributedTransactionManager`, which registered a participant into an existing transaction during `join` and needed to surface registration failures faithfully. With the GlobalTransaction API introduced earlier in this stack, that consumer is gone from core — joining a transaction across processes is now done via `GlobalTransactionManager.beginBranch`, which declares `TransactionException` on its own contract. Nothing needs the widened `join` anymore, so this PR reverts the widening and restores the original (narrower) application-facing contract, undoing the breaking change before it ships.

## Related issues and/or PRs

- Reverts #3663

## Changes made

- Restored `DistributedTransactionManager.join(String)` to `throws TransactionNotFoundException` only, including its Javadoc.
- Restored the matching signatures on `AbstractDistributedTransactionManager` and `DecoratedDistributedTransactionManager`.

## Checklist

- Clean `git revert` of #3663's squash commit (`df344f9ba`); `core` and `integration-test` compile with no remaining dependency on the widened signature.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
@@ -415,12 +415,8 @@ public interface DistributedTransactionManager
    * @return {@link DistributedTransaction}
    * @throws TransactionNotFoundException if the transaction associated with the specified
    *     transaction ID is not found. You can retry the transaction from the beginning
-   * @throws TransactionException if the transaction fails to join due to transient or nontransient
-   *     faults. You can try retrying the transaction from the beginning, but you may not be able to
-   *     join the transaction due to nontransient faults
    */
-  default DistributedTransaction join(String txId)
-      throws TransactionNotFoundException, TransactionException {
+  default DistributedTransaction join(String txId) throws TransactionNotFoundException {
     return resume(txId);
   }
 
--- a/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionManager.java
+++ b/core/src/main/java/com/scalar/db/common/AbstractDistributedTransactionManager.java
@@ -11,7 +11,6 @@ import com.scalar.db.api.Scan;
 import com.scalar.db.api.Update;
 import com.scalar.db.api.Upsert;
 import com.scalar.db.config.DatabaseConfig;
-import com.scalar.db.exception.transaction.TransactionException;
 import com.scalar.db.exception.transaction.TransactionNotFoundException;
 import com.scalar.db.util.ScalarDbUtils;
 import java.util.List;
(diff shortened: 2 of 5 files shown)


Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
```diff
@@ -415,12 +415,8 @@ DistributedTransaction start(String txId, Isolation isolation, SerializableStrat
    * @return {@link DistributedTransaction}
    * @throws TransactionNotFoundException if the transaction associated with the specified
    *     transaction ID is not found. You can retry the transaction from the beginning
-   * @throws TransactionException if the transaction fails to join due to transient or nontransient
-   *     faults. You can try retrying the transaction from the beginning, but you may not be able to
-   *     join the transaction due to nontransient faults
    */
-  default DistributedTransaction join(String txId)
-      throws TransactionNotFoundException, TransactionException {
+  default DistributedTransaction join(String txId) throws TransactionNotFoundException {
```
feeblefakie, line 419: Revert is clean here — both /ce:review and /codex:review confirm no regressions and no callers rely on the widened throws.

One loose end that's outside this diff but tied to this contract: `integration-test/src/test/java/com/scalar/db/common/GlobalTransactionBackedDistributedTransactionManagerTest#join_WhenRegisterParticipantFails_ShouldPropagateTransactionException` (around line 279–289) is `@Disabled` and still asserts that `manager.join(...)` throws a `TransactionException` that is not a `TransactionNotFoundException` — the exact contract this PR reverts. If someone drops `@Disabled` later, it either fails to compile or re-asserts a repudiated contract. Worth deleting now, or rewriting the comment to point at `GlobalTransactionManager#beginBranch` as the surface where non-not-found registration failures propagate.

feeblefakie (commented):
Left some comments. PTAL!

core/src/main/java/com/scalar/db/api/DistributedTransactionManager.java
brfrn169: Good catch — and the sibling disabled tests had the same problem, so I removed all five relics in ced15e586ab3a84cb88db0e088c7daa61fa0f7d9 (they described the adapter's previous by-ID registration shape; the live `join_ShouldThrowUnsupportedOperationException` covers the actual behavior). The intent of the flagged one is preserved at the surface where the contract lives now: `TwoPhaseCommitBackedGlobalTransactionManagerTest#beginBranch_WhenJoinParticipantFails_ShouldPropagateTransactionException` asserts that a non-not-found registration failure propagates from `beginBranch` as-is.

feeblefakie (approved):
LGTM! Thank you!

PR #3739, opened 2026-07-22
Share ConsensusCommitManager's storage and table metadata manager with subclasses and narrow the commit-handler surface

brfrn169:
## Description

`ConsensusCommitManager` keeps its `storage`, `admin`, and `tableMetadataManager` fields `private` with no accessor, so a subclass that needs the same resources has no way to reuse them and is forced to build its own duplicate set from the same configuration. Separately, the `commit` field is `protected` solely so a subclass can call `commit.setBeforePreparationHook(...)`, which exposes the entire `CommitHandler` just to reach one setter.

This PR reshapes what `ConsensusCommitManager` offers to subclasses in both directions: it exposes the storage and table metadata manager for reuse, and it narrows the commit-handler surface to the single operation a subclass actually needs. This lets a subclass stop duplicating a `DistributedStorage`, a `DistributedStorageAdmin`, and a table-metadata cache built from the same configuration.

## Related issues and/or PRs

- A paired downstream change consumes this new API and will be opened/merged after this PR.

## Changes made

- Added `protected DistributedStorage getStorage()` and `protected TransactionTableMetadataManager getTableMetadataManager()` so subclasses can reuse the manager's instances instead of constructing duplicates. Their Javadoc records the ownership contract: both instances are owned by the manager and closed by `close()`, so a borrower must not close them. No `getAdmin()` accessor was added because the only use of the admin was to build a `TransactionTableMetadataManager`, which is now borrowed directly.
- Changed the `commit` field from `protected final` to `private final` and added `protected void setBeforePreparationHook(BeforePreparationHook)` that delegates to it. This is a breaking change to `protected` API and thus targets 4.0.0. `createCommitHandler` is `private`, so the field was the only commit-handler-related subclass surface; `CommitHandler.setBeforePreparationHook` remains `public` and unchanged.
- Added a unit test verifying `setBeforePreparationHook` delegates to the `CommitHandler`.

## Checklist

The `commit` field visibility change is a breaking change to `protected` API. Its only known subclass is updated in the paired change, so no in-tree or known out-of-tree consumer is left stranded.

The new accessors return interface-typed fields, which SpotBugs does not flag (`EI_EXPOSE_REP` was verified not to fire on them), so no suppression annotation is needed.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
@@ -71,7 +71,7 @@ public class ConsensusCommitManager extends AbstractDistributedTransactionManage
   private final ParallelExecutor parallelExecutor;
   private final RecoveryExecutor recoveryExecutor;
   private final CrudHandler crud;
-  protected final CommitHandler commit;
+  private final CommitHandler commit;
   private final Isolation isolation;
   private final ConsensusCommitOperationChecker operationChecker;
   @Nullable private final CoordinatorGroupCommitter groupCommitter;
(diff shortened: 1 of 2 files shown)


Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM! 👍

PR #3740, opened 2026-07-23
Deprecate the external API for Consensus Commit implicit pre-read and insert mode

brfrn169:
## Description

"Implicit pre-read" and "insert mode" are Consensus Commit implementation details that leaked onto the public `Put` API — via typed `PutBuilder` methods, the public `ConsensusCommitOperationAttributes` utility (methods and attribute-key constants), and the `OperationBuilder.ImplicitPreReadEnabled` / `InsertModeEnabled` interfaces. The modern `insert()` / `upsert()` / `update()` operations already cover the user-facing use cases, and the only operation that carried these knobs, `put(Put)`, is already deprecated. This PR deprecates the external-facing surface so users are guided to `insert` / `upsert` / `update`, while the mechanism itself stays fully functional for Consensus Commit's internal use. The change is annotation- and Javadoc-only — no runtime behavior, attribute mechanism, or wire-format changes.

## Related issues and/or PRs

N/A

## Changes made

- Deprecated the implicit-pre-read / insert-mode methods (`enable*` / `disable*` / `is*Enabled`) and the public attribute-key constants (`IMPLICIT_PRE_READ_ENABLED`, `INSERT_MODE_ENABLED`) in `ConsensusCommitOperationAttributes`. The `transaction-isolation` attribute API in the same class is untouched.
- Deprecated the `OperationBuilder.ImplicitPreReadEnabled<T>` / `InsertModeEnabled<T>` interfaces and their methods, and the corresponding overrides in `PutBuilder` (`Buildable` / `BuildableFromExisting`).
- Refreshed the existing `@deprecated` Javadoc on `Put.isImplicitPreReadEnabled()` / `isInsertModeEnabled()` so it no longer points at the now-also-deprecated `ConsensusCommitOperationAttributes` readers.
- Enriched the `@deprecated` Javadoc of the already-deprecated `put(Put)` / `put(List<Put>)` methods in `CrudOperable`, `TransactionCrudOperable`, and `TransactionManagerCrudOperable` to recommend `insert` / `upsert` / `update`.
- All deprecation messages state "As of release 3.19.0. Will be removed in release 4.0.0."

No logic changes; the operation-attribute mechanism and internal Consensus Commit usage are unchanged.

## Checklist

The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- Annotation/Javadoc-only; there is no behavior change, so no new tests were added — the existing unit and integration suites already exercise these methods and the underlying implicit-pre-read / insert-mode behavior, and the full `:core:test` suite passes.
- No `@SuppressWarnings` were added: the build has no `-Werror`/`-Xlint`, and the codebase already tolerates internal deprecation warnings unsuppressed (e.g. `AbstractDistributedTransaction.mutate` calls the deprecated `put(Put)`) and deprecated-type imports (e.g. `Put.java` imports the deprecated `Value`). The internal callers (`CrudHandler`, `Snapshot`, the mutation composers, `ConsensusCommitUtils`) and the tests are intentionally left untouched. Because this deprecates public API that internal/test code still uses, `-Xlint:deprecation` reports warnings by design; the build itself introduces no new failures or Error Prone errors.
- Out of scope and verified untouched: `TwoPhaseCommitParticipant` and its implementers already deprecate `put(String, Put)` as of 3.19.0; the `parallel_implicit_pre_read.enabled` config is a separate performance setting (not the per-operation attribute); `CrudHandler.readIfImplicitPreReadEnabled` is internal orchestration, not a user knob.
- Follow-up (separate repo): confirm `scalardb-cluster` does not escalate deprecation warnings to errors where it references these surfaces; a paired change would only be needed if it does.

## Release notes

Deprecated the external API for Consensus Commit implicit pre-read and insert mode — the `Put` builder methods, the `ConsensusCommitOperationAttributes` methods and attribute-key constants, and the `OperationBuilder.ImplicitPreReadEnabled` / `InsertModeEnabled` interfaces. Use the `insert`, `upsert`, or `update` operations instead.

### DIFF
--- a/core/src/main/java/com/scalar/db/api/PutBuilder.java
+++ b/core/src/main/java/com/scalar/db/api/PutBuilder.java
@@ -264,18 +264,24 @@ public class PutBuilder {
       return this;
     }
 
+    /** @deprecated As of release 3.19.0. Will be removed in release 4.0.0. */
+    @Deprecated
     @Override
     public Buildable disableImplicitPreRead() {
       ConsensusCommitOperationAttributes.disableImplicitPreRead(attributes);
       return this;
     }
 
+    /** @deprecated As of release 3.19.0. Will be removed in release 4.0.0. */
+    @Deprecated
     @Override
     public Buildable enableImplicitPreRead() {
       ConsensusCommitOperationAttributes.enableImplicitPreRead(attributes);
       return this;
     }
 
+    /** @deprecated As of release 3.19.0. Will be removed in release 4.0.0. */
+    @Deprecated
     @Override
     public Buildable implicitPreReadEnabled(boolean implicitPreReadEnabled) {
       if (implicitPreReadEnabled) {
@@ -286,18 +292,24 @@ public class PutBuilder {
       return this;
     }
 
+    /** @deprecated As of release 3.19.0. Will be removed in release 4.0.0. */
+    @Deprecated
     @Override
     public Buildable disableInsertMode() {
       ConsensusCommitOperationAttributes.disableInsertMode(attributes);
       return this;
     }
 
+    /** @deprecated As of release 3.19.0. Will be removed in release 4.0.0. */
+    @Deprecated
     @Override
     public Buildable enableInsertMode() {
       ConsensusCommitOperationAttributes.enableInsertMode(attributes);
       return this;
     }
 
+    /** @deprecated As of release 3.19.0. Will be removed in release 4.0.0. */
+    @Deprecated
     @Override
     public Buildable insertModeEnabled(boolean insertModeEnabled) {
       if (insertModeEnabled) {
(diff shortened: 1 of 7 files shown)


Torch3333 (approved):
LGTM, thank you!

komamitsu (approved):
LGTM! 👍

core/src/main/java/com/scalar/db/api/PutBuilder.java
```diff
@@ -264,18 +264,24 @@ public Buildable value(Column<?> column) {
       return this;
     }
 
+    /** @deprecated As of release 3.19.0. Will be removed in release 4.0.0. */
```
feeblefakie: Two things about the `PutBuilder` deprecations:

**(1) Terse Javadoc.** All 12 `@Deprecated` overrides in `Buildable` and `BuildableFromExisting` use this bare form, but the parent interfaces in `OperationBuilder` and every method in `ConsensusCommitOperationAttributes` use the fuller form ending `"...Implicit pre-read and insert mode are Consensus Commit internal details; use the insert, upsert, or update operations instead."` `Put.newBuilder()...enableImplicitPreRead()` is the primary user-facing entry point, and IDE hover on the concrete override usually wins over the interface's Javadoc — so migrating users see the version string with no hint of what to switch to.

**(2) `Buildable` class implements deprecated interfaces without `@SuppressWarnings("deprecation")`.** The enclosing `Buildable` (and `BuildableFromExisting`) is a public, non-deprecated class now `implements ImplicitPreReadEnabled<Buildable>, InsertModeEnabled<Buildable>` — both `@Deprecated`. Any consumer compiling with `-Werror -Xlint:deprecation` (a normal library-hardening choice) fails on core's own class declaration, not just on their own call sites. The PR body flags a `scalardb-cluster` follow-up for this — a class-level `@SuppressWarnings("deprecation")` on `Buildable` and `BuildableFromExisting` scopes it to core and avoids the follow-up.

feeblefakie (commented):
Left some comments. PTAL!

core/src/main/java/com/scalar/db/api/PutBuilder.java
brfrn169: **(1)** Good point about IDE hover resolving to the concrete override — expanded all twelve to the full wording (with the migration path) in 90746158110b6a1152894ba18361055a681c17da.

**(2)** I'd rather not add the suppression, for three reasons. First, a consumer's compilation doesn't emit warnings for core's class declaration: javac reports deprecation in the compilation unit that references the deprecated element, so a consumer merely using the non-deprecated `Buildable` sees nothing — and warnings for a consumer's own calls to the deprecated methods can't be silenced from core anyway. Second, a class-level `@SuppressWarnings("deprecation")` on these large builder classes would also mask future accidental uses of deprecated API inside them; the build intentionally has no `-Werror`/`-Xlint`, and the codebase already tolerates internal deprecation warnings unsuppressed (see the Additional notes). Third, the separate-repo follow-up is about that repository's own references to these surfaces, which a suppression on core's declaration wouldn't affect. WDYT?

feeblefakie (approved):
LGTM! Thank you!

PR #3765, opened 2026-07-30
Record only record identities in the Coordinator write set

brfrn169:
## Description

The `tx_write_set` payload persisted on the Coordinator table carries three fields that no production code reads: `Entry.entry_type`, `Entry.participant_id`, and `Entry.columns` — the last of which is never written at all. They were placed ahead of a backup/changelog consumer that has not been designed yet.

Two things follow from that. First, `WriteSetDetailLevel.FULL` and `WriteSetEntry.getColumns()` commit the participant API to a shape chosen before the requirements are known, and the enum's javadoc declares `FULL` a "permanent ceiling" — a promise we may not want once a real consumer needs a previous transaction ID or a before image. Second, and more urgently, the payload cannot grow into a compressed encoding later: turning `repeated EntryGroup` into a message is wire-incompatible, so if we want compression without spending `schema_version` on it, the wrapper has to exist now.

The format is not released on any branch or tag, so this is the last point at which the shape can be changed freely.

## Related issues and/or PRs

N/A

## Changes made

**Proto (`core/src/main/proto/consensus_commit.proto`)**

- Drop `Entry.entry_type` and `Entry.columns`, remove the nested `EntryType` enum, and renumber the remaining fields to 1–5. No `reserved` markers: nothing is persisted anywhere, so the numbers need no protection.
- Keep `Entry.participant_id`. Unlike the other two it cannot be recovered from the record itself — a record does not say which participant wrote it, whereas put-vs-delete is recoverable from its own `tx_state` (which is already how the recovery loop decides).
- Introduce an `EntryGroups` wrapper message carried by a `oneof payload`. A `oneof` cannot hold a repeated field directly, and a future compressed encoding needs a named message to serialize. Compression can now be added as another `oneof` case rather than by bumping `schema_version`.
- Rewrite the `schema_version` comment to state **when it is bumped** rather than what a version-1 row contains. Fields that carry their own presence (`optional` scalars, enums with an `UNSPECIFIED` zero value, message fields, new `oneof` cases) can be added without a bump, so the field set of a version-1 row may grow and readers must not assume otherwise. Only a field without presence — practically, a `repeated` field — forces a bump or a companion marker, because "written empty" and "never written" are identical on the wire. Recording column values is the case that will hit this.

**Encoder / recovery path**

- `WriteSetEncoder` loses the `includeColumns` and entry-type parameters and its `TransactionTableMetadataManager` dependency, becoming a static utility. `CommitHandler` and `CommitHandlerWithGroupCommit` drop the corresponding constructor parameters.
- A read-only commit now encodes an empty `EntryGroups` instead of leaving the payload unset, so the `oneof` case is always set.
- `ConsensusCommitManager#finishTransaction` and `WriteSetEncoder#mergeChildWriteSets` assert the payload case before reading `getEntryGroups()`. Both invariants hold by construction today — `ENTRY_GROUPS` is the oneof's only case, and the encoder always sets it — so they are assertions rather than coded errors. The comments record what would make them reachable and what it would cost, since `getEntryGroups()` returns an empty default instance for a case it does not know: `finishTransaction` would delete the Coordinator state row without recovering anything, leaving the still-`PREPARED` records of a committed transaction to be rolled back by lazy recovery, and `mergeChildWriteSets` would drop a child's entries from the parent row in silence. Adding a payload case does not bump `schema_version`, so a version check would not catch either one; whoever adds a case must replace the assertions with recoverable errors, because reading a row written by a newer binary is an operational situation rather than a bug.

**Participant API**

- Drop `WriteSetDetailLevel.FULL` and `WriteSetEntry.getColumns()`, which have no destination now that the payload records no column values. `ConsensusCommitParticipant#buildWriteSetEntries` loses its detail-level switch, and `getTableMetadataForMutation` goes with it (its only caller was the `FULL` branch).
- The `WriteSetDetailLevel` enum and the `prepareRecords` signature stay, so a level conveying more record content can return as a new constant without touching the signature or the transports that carry it. The javadoc no longer claims `FULL` is a permanent ceiling; what a richer level should look like is left to the consumer that first needs it.
- `WriteSetEntry.Type` stays. It is an in-memory API that is never persisted, so the "empty vs. absent" ambiguity that motivated the other removals does not apply to it.

**Tests**

- Updated across `core` and `integration-test` for the new shape. Removed the tests that covered column recording and transaction-meta-column filtering in the write set, since neither capability exists on this path any more; the filtering logic itself (`ConsensusCommitUtils#isTransactionMetaColumn`) is unchanged and still covered through its other callers.
- The read-only recovery test asserts the empty-`EntryGroups` encoding. The unset-payload assertion is covered by `ConsensusCommitManagerTest#finishTransaction_WriteSetWithUnsetPayload_ShouldFailAssertionWithoutDeletingState`, which also verifies that recovery does not run and the Coordinator state row survives.
- The multi-participant encoding tests (`ConsensusCommitCoordinatorTest`, `WriteSetEncoderTest`) assert each entry's partition key. With `entry_type` gone, asserting only the participant id left one participant's entries indistinguishable from each other, so the assertions would have held even if the encoder had emitted one of them twice or swapped them.

## Checklist

**Documentation.** `tx_write_set` is an internal persistence format with no user-facing documentation; the proto comments are the documentation, and they are rewritten here.

**Two traps recorded for whoever adds column recording later**, both of which fail silently rather than loudly:

- Do not derive insert-vs-update from the presence of a previous transaction ID. `TransactionResult#getId()` is null for a deemed-committed record — a row imported via `schema-loader --import` — so the first write to such a row is an update with no previous transaction ID. Reading that as an insert would apply the write as "these columns, everything else null".
- Do not add compression by wrapping the payload in an outer envelope. Existing rows are bare `WriteSet` bytes, and parsing one as an envelope would read field 1 (`schema_version`, varint) as a codec enum and field 2 as an opaque payload — a plausible-looking misparse with no exception.

## Release notes

N/A

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
@@ -633,7 +633,23 @@ public class ConsensusCommitManager extends AbstractDistributedTransactionManage
     // composer-level NoMutationException absorption still acts as the safety net for races against
     // concurrent recovery.
     WriteSet writeSet = state.getWriteSet().get();
-    for (EntryGroup entryGroup : writeSet.getEntryGroupsList()) {
+
+    // The writer always sets the payload, so an unset case — or one this binary does not know —
+    // means the row was written by a newer schema version. getEntryGroups() would return an empty
+    // default instance for such a row, which would silently skip recovery and report success, so
+    // fail loudly instead.
+    if (writeSet.getPayloadCase() != WriteSet.PayloadCase.ENTRY_GROUPS) {
+      throw new TransactionException(
+          CoreError.CONSENSUS_COMMIT_FINISHING_TRANSACTION_FAILED.buildMessage(
+              txId,
+              "The write set uses an unsupported payload encoding. Schema version: "
+                  + writeSet.getSchemaVersion()
+                  + ", Payload case: "
+                  + writeSet.getPayloadCase()),
+          txId);
+    }
+
+    for (EntryGroup entryGroup : writeSet.getEntryGroups().getEntryGroupsList()) {
       String expectedTxId =
           entryGroup.hasChildId()
               ? KEY_MANIPULATOR.fullKey(state.getId(), entryGroup.getChildId())
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
@@ -634,20 +634,17 @@ public class ConsensusCommitManager extends AbstractDistributedTransactionManage
     // concurrent recovery.
     WriteSet writeSet = state.getWriteSet().get();
 
-    // The writer always sets the payload, so an unset case — or one this binary does not know —
-    // means the row was written by a newer schema version. getEntryGroups() would return an empty
-    // default instance for such a row, which would silently skip recovery and report success, so
-    // fail loudly instead.
-    if (writeSet.getPayloadCase() != WriteSet.PayloadCase.ENTRY_GROUPS) {
-      throw new TransactionException(
-          CoreError.CONSENSUS_COMMIT_FINISHING_TRANSACTION_FAILED.buildMessage(
-              txId,
-              "The write set uses an unsupported payload encoding. Schema version: "
-                  + writeSet.getSchemaVersion()
-                  + ", Payload case: "
-                  + writeSet.getPayloadCase()),
-          txId);
-    }
+    // Holds by construction today: ENTRY_GROUPS is the payload oneof's only case, and the encoder
+    // always sets it — a read-only commit is encoded as an empty EntryGroups, never as an unset
+    // payload. It stops holding the moment a second case (e.g., the compressed encoding the oneof
+    // was introduced for) ships, because such a row parses here with the case unset: getEntryGroups
+    // then returns an empty default instance, the loop below does nothing, and the Coordinator
+    // state row is deleted anyway — leaving the still-PREPARED records of a committed transaction
+    // to be rolled back by lazy recovery. Note that adding a payload case does not bump
+    // schema_version (see the proto), so a version check would not catch this either. Whoever adds
+    // that case must replace this assertion with a recoverable error: reading a row written by a
+    // newer binary is an operational situation, not a bug.
+    assert writeSet.getPayloadCase() == WriteSet.PayloadCase.ENTRY_GROUPS;
 
     for (EntryGroup entryGroup : writeSet.getEntryGroups().getEntryGroupsList()) {
       String expectedTxId =
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/WriteSetEncoder.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/WriteSetEncoder.java
@@ -7,9 +7,7 @@ import com.google.protobuf.ByteString;
 import com.scalar.db.api.Delete;
 import com.scalar.db.api.Mutation;
 import com.scalar.db.api.Put;
-import com.scalar.db.api.TableMetadata;
 import com.scalar.db.api.TwoPhaseCommitParticipant;
-import com.scalar.db.exception.storage.ExecutionException;
 import com.scalar.db.io.BigIntColumn;
 import com.scalar.db.io.BlobColumn;
 import com.scalar.db.io.BooleanColumn;
@@ -37,6 +35,7 @@ import com.scalar.db.transaction.consensuscommit.proto.v1.Column.TimestampTZValu
 import com.scalar.db.transaction.consensuscommit.proto.v1.Column.TimestampValue;
 import com.scalar.db.transaction.consensuscommit.proto.v1.Entry;
 import com.scalar.db.transaction.consensuscommit.proto.v1.EntryGroup;
+import com.scalar.db.transaction.consensuscommit.proto.v1.EntryGroups;
 import com.scalar.db.transaction.consensuscommit.proto.v1.WriteSet;
 import com.scalar.db.util.TimeRelatedColumnEncodingUtils;
 import java.nio.ByteBuffer;
@@ -48,13 +47,11 @@ import javax.annotation.Nullable;
  * Encodes the proto {@link WriteSet} / {@link EntryGroup} payload persisted in the Coordinator
  * table's {@code tx_write_set} column.
  *
- * <p>Primary keys (namespace, table, partition key, optional clustering key) are always recorded.
- * Non-key column values are only included when {@code includeColumns} is true — this is intended
- * for backup/changelog use cases that need the full record content; the default mode keeps the
- * persisted BLOB compact. When {@code includeColumns} is true, transaction-meta columns (e.g.,
- * {@code tx_state}, {@code tx_version}, {@code before_*}) that the snapshot may have injected for
- * ConsensusCommit's own bookkeeping are filtered out so the persisted payload contains only user
- * data.
+ * <p>An {@link Entry} currently records only a record's identity — namespace, table, partition key,
+ * optional clustering key — plus the owning participant on the multi-participant TwoPhaseCommit
+ * path. The written column values are not recorded, and neither is whether the record was put or
+ * deleted (that is recoverable from the record's own {@code tx_state}). The proto's {@code
+ * schema_version} comment states which of those can be added without bumping the version.
  */
 final class WriteSetEncoder {
   /**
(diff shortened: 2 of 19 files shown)


Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
```diff
@@ -633,7 +633,23 @@ public boolean finishTransaction(String txId) throws TransactionException {
...
     WriteSet writeSet = state.getWriteSet().get();
-    for (EntryGroup entryGroup : writeSet.getEntryGroupsList()) {
+
+    // The writer always sets the payload, so an unset case — or one this binary does not know —
+    // means the row was written by a newer schema version. getEntryGroups() would return an empty
+    // default instance for such a row, which would silently skip recovery and report success, so
+    // fail loudly instead.
+    if (writeSet.getPayloadCase() != WriteSet.PayloadCase.ENTRY_GROUPS) {
+      throw new TransactionException(
+          CoreError.CONSENSUS_COMMIT_FINISHING_TRANSACTION_FAILED.buildMessage(
```
feeblefakie: This reuses `CONSENSUS_COMMIT_FINISHING_TRANSACTION_FAILED`, so an unsupported payload surfaces under code `0068` together with storage failures and `deleteState` failures. Since this one signals a persisted-format problem, would a dedicated constant be better so it can be told apart? `0070` looks free.

Also a nit: the template separates fields with `;`, but the inline text uses `,`.

Unrelated to the code, the PR description says the read-only recovery test is what exercises this guard. I think the actual coverage is `finishTransaction_WriteSetWithUnsetPayload_ShouldThrowWithoutDeletingState`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/WriteSetEncoder.java
```diff
@@ -369,16 +314,19 @@ private static Entry encodeWriteSetEntry(
...
   static WriteSet mergeChildWriteSets(List<ChildWriteSet> children) {
-    WriteSet.Builder builder = WriteSet.newBuilder().setSchemaVersion(SCHEMA_VERSION);
+    EntryGroups.Builder entryGroups = EntryGroups.newBuilder();
     for (ChildWriteSet child : children) {
       if (child.writeSet == null) {
         continue;
       }
-      for (EntryGroup entryGroup : child.writeSet.getEntryGroupsList()) {
-        builder.addEntryGroups(entryGroup.toBuilder().setChildId(child.childId).build());
+      for (EntryGroup entryGroup : child.writeSet.getEntryGroups().getEntryGroupsList()) {
```
feeblefakie, line 328: `mergeChildWriteSets` reads `getEntryGroups()` without checking the payload case, unlike the new guard in `finishTransaction`. Not reachable today since the children are built in-process by this same encoder, but it would become reachable if the compressed case the `oneof` is designed for is ever added — a compressed child would be dropped silently.

Could we add the same check here, or a short comment saying why it's safe?

core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinatorTest.java
```diff
@@ -1116,19 +1112,16 @@ void joinParticipant_WhenDuplicateParticipantId_ShouldBeNoOp() throws Exception
     ArgumentCaptor<WriteSet> captor = ArgumentCaptor.forClass(WriteSet.class);
     verify(coordinatorCommitHandler).commitState(eq("tx-1"), captor.capture());
     WriteSet writeSet = captor.getValue();
-    assertThat(writeSet.getEntryGroupsList()).hasSize(2);
+    assertThat(writeSet.getEntryGroups().getEntryGroupsList()).hasSize(2);
 
-    EntryGroup group1 = writeSet.getEntryGroups(0);
+    EntryGroup group1 = writeSet.getEntryGroups().getEntryGroups(0);
     assertThat(group1.getEntriesList()).hasSize(2);
     assertThat(group1.getEntries(0).getParticipantId()).isEqualTo("participant-1");
```
feeblefakie, line 1121: Now that `entry_type` is gone, both entries here are only asserted on `getParticipantId()`, so the test would still pass if the encoder emitted the same record twice or swapped the two. The setup uses distinct keys (`pk=1` for the write, `pk=2` for the delete), so could we assert the partition key per entry? Same shape at `WriteSetEncoderTest:408`.

A few stale comments nearby while you're here:
- line 1111 still says "and the right entry type"
- line 1316 still says "includeColumns=false, so getColumns is not read"
- `ConsensusCommitParticipantTest:642` still says "so the FULL meta-column filtering is observable"

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -10750,10 +10753,12 @@ void finishTransaction_AbortedWithSomeRecordsPartiallyPrepared_ShouldRollBackAnd
...
             .setSchemaVersion(1)
-            .addEntryGroups(
-                EntryGroup.newBuilder()
-                    .addEntries(intKeyWriteEntry(namespace1, TABLE_1, 0, 0))
-                    .addEntries(intKeyWriteEntry(namespace1, TABLE_1, 0, 1)))
+            .setEntryGroups(
+                EntryGroups.newBuilder()
+                    .addEntryGroups(
+                        EntryGroup.newBuilder()
+                            .addEntries(intKeyWriteEntry(namespace1, TABLE_1, 0, 0))
```
feeblefakie: Tiny naming nit: `intKeyWriteEntry` no longer sets an entry type, so the `Write` part is a bit misleading now. `intKeyEntry`?

feeblefakie (approved):
LGTM! Thank you! Left minor comments. PTAL!

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitManager.java
brfrn169: Good point on the shared code. We went the other way in c1940abab rather than adding `0070`: the condition turned out to be unreachable by construction — `ENTRY_GROUPS` is the payload oneof's only case, and the encoder always sets it, including for a read-only commit — so it is now an `assert` with a comment, matching how `WriteSetDecoder` already treats a proto shape its own encoder cannot produce. That settles the `;`/`,` nit too, since the `CoreError` message is gone.

The comment records what would make it reachable and what it would cost, since none of it fails loudly: a second payload case parses here with the case unset, `getEntryGroups()` returns an empty default instance, and the Coordinator state row would be deleted without recovering anything — leaving the still-`PREPARED` records of a committed transaction to be rolled back by lazy recovery. Adding a payload case does not bump `schema_version`, so a version check would not catch it either. Whoever adds one has to replace the assertion with a recoverable error, since reading a row written by a newer binary is an operational situation rather than a bug.

And you are right about the test, thanks — the coverage is `finishTransaction_WriteSetWithUnsetPayload_ShouldFailAssertionWithoutDeletingState` (renamed in the same commit). The PR description is corrected.

core/src/main/java/com/scalar/db/transaction/consensuscommit/WriteSetEncoder.java
brfrn169: Added the same assertion in b5924348a, so the two sites are no longer asymmetric.

The invariant is stronger here than in `finishTransaction`, and the comment says so: every child is encoded in-process by this same encoder, so this path can never meet a row written by a newer binary — it would only break if the encoder itself started emitting a second payload case. The cost is recorded either way, since `getEntryGroups()` returns an empty default instance for a case it does not know: the child's entries would be dropped from the parent row in silence.

core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitCoordinatorTest.java
brfrn169: Fixed in 0a55d4d9b. Both sites now assert the partition key per entry — `pk=1`/`pk=2`/`pk=3` here, and `pk="a"`/`pk="c"`/`pk="b"` at `WriteSetEncoderTest:408` — so emitting one entry twice or swapping the two no longer passes.

The three stale comments are updated in the same commit. The `ConsensusCommitParticipantTest` one still had something to say, so it now says it: the injected `tx_id` column is what shows that a write whose content would need meta-column filtering still builds its entry without consulting the table metadata.

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
brfrn169: Renamed to `intKeyEntry` in c42e1baac.

PR #3791, opened 2026-08-05
feat: add scalar.db.collation to align text comparison with storage

Torch3333:
## Description

ScalarDB evaluates many text comparisons on the JVM itself — object-storage scan ordering and filtering, in-memory cross-partition/conjunction filtering, the Consensus Commit snapshot's bookkeeping and scan-after-write checks, and conditional mutations. Historically those comparisons used Java's natural `String` order (UTF-16 code units), which matches none of the supported backends: MySQL 8's default collation is case- and accent-insensitive, Cassandra and DynamoDB compare UTF-8 bytes, and PostgreSQL follows its configured collation. Any divergence can make an in-memory decision disagree with the storage's own answer, yielding wrong scan results, stale reads of a transaction's own writes, or missed/spurious conflict detection.

This PR adds a `scalar.db.collation` setting that aligns ScalarDB's in-memory text **ordering**, **equality**, and — under `ICU`, in the Consensus Commit transaction layer — **key identity** with the collation configured on the underlying storage.

Two deliberate design choices:

- **The collation flavor _is_ the determinism (MySQL model).** A case-insensitive `ICU` collation makes ordering **and** `=`/`!=` **and** transaction-layer key identity case-insensitive together, exactly as a CI collation behaves on MySQL/MariaDB/SQL Server/Oracle. There is no separate deterministic/nondeterministic toggle (PostgreSQL's decoupling is intentionally not replicated; see `docs/collation-equality-backend-behavior.md`).
- **`BINARY` is the default.** When `scalar.db.collation` is absent, ScalarDB compares text by unsigned UTF-8 byte order. Equality and key identity are byte-exact, identical to pre-collation behavior; the only behavioral delta versus older releases is the in-memory *ordering* of supplementary-plane text (above U+FFFF), documented as an upgrade note.

## Related issues and/or PRs

None.

## Changes made

**Configuration and comparator**
- `scalar.db.collation` (`BINARY` — the default — or `ICU`) with `scalar.db.collation.icu.locale`, `.icu.strength`, and `.icu.rules` (custom tailoring rules compose on top of the locale's collation). Invalid values fail fast at startup with `CoreError` messages.
- `com.scalar.db.io.CollationComparator` — a frozen (thread-safe) ICU4J collator or unsigned UTF-8 byte order — always present and threaded through the storage adapters and Consensus Commit (no `Optional` plumbing). Under ICU it also exposes a canonical text form (collation key bytes) via a per-thread thawed collator (the frozen collator's `getCollationKey` serializes on a JVM-wide lock).

**Ordering and equality (all in-memory evaluation sites)**
- Ordering: object-storage scan sort and range filtering, in-memory conjunction range filtering, snapshot scan-after-write range membership, conditional-mutation range predicates.
- Equality: `=`/`!=` on `TEXT` at the in-memory sites (conjunction/scan filtering, conditional-mutation `EQ`/`NE`, snapshot equality-based overlap) follows the configured collation via a single `ScalarDbUtils.columnEquals` chokepoint. `IS_NULL`/`IS_NOT_NULL`/`LIKE`, non-text, and null-text comparisons stay byte-exact. Under `BINARY`, text equality is exact `String` equality (avoiding unpaired-surrogate conflation in UTF-8 encoding).

**Collation-canonical key identity (Consensus Commit, ICU only)**
- `Snapshot.Key` identity is collation-canonical: two collate-equal keys (the `'apple'` your application typed and the `'Apple'` the backend stores and returns from scans) are one logical key across the snapshot's read/write/delete bookkeeping, read-your-own-writes merging, the prepare-time before-image join, write-write and scan-after-write conflict detection, and `MutationsGrouper` grouping. Stored bytes are never rewritten; public `Key`/`Column` `equals()`/`hashCode()` are untouched.
- Without this, a byte-exact snapshot on a CI-collated backend silently returns stale reads of its own writes, aborts valid read-modify-writes at prepare, splits one row's writes in two, and misses scan-after-write conflicts — all four defects were reproduced against MySQL `utf8mb4_0900_ai_ci` before building, and are now covered by automated integration tests.
- The aligned-backend operator contract (CI uniqueness on keys + collation-aware point-read resolution for recovery), UCA version-skew bounds, and misalignment hazards are documented in `docs/collation.md`.

**Integration tests (new, capability-gated)**
- `DistributedStorageCollationIntegrationTestBase`: ICU scan ordering, conditional-mutation `EQ`/`NE` across collate-equal spellings, range operators across case boundaries, cross-partition filtering. Runs on MySQL 8.x (`utf8mb4_0900_ai_ci`), SQL Server (`Latin1_General_100_CI_AI`, basic-Latin data), and object storage (in-memory ICU, no backend collation involved).
- `ConsensusCommitCollationIntegrationTestBase`: the four key-identity scenarios end-to-end through real transactions, with raw-SQL physical-row-count assertions. Runs on MySQL 8.x and SQL Server.
- Tables are created through normal ScalarDB DDL, then altered to the target collation via a new test-only `AdminTestUtils.alterTableCollation` hook (MySQL `CONVERT TO`; SQL Server transactional PK drop → per-column `ALTER COLUMN ... COLLATE` preserving ASC/DESC → PK recreate).
- Gating is a cached, logged `JdbcEnv` capability probe (excludes TiDB/MariaDB behind `jdbc:mysql` URLs and MySQL 5.x); the MySQL CI job sets `scalardb.jdbc.collation_test=required` so a probe failure errors loudly instead of silently skipping the suites.

**CI default-collation normalization (`ci/tests-config.yaml`)**
- MySQL 8.x → `utf8mb4_0900_bin` (NO PAD, exact `BINARY` match; 5.7 keeps `utf8mb4_bin` via a version conditional — mysqld 5.7 exits on unknown collations), MariaDB → `utf8mb4_nopad_bin`, PostgreSQL → pinned `--locale=C --encoding=UTF8`. SQL Server (`Japanese_BIN2`) and AlloyDB (posix ICU locale) were already exact.

**Docs**
- `docs/collation.md` (setting reference, key-identity operator contract, automated coverage, upgrade notes), `docs/collation-storage-compatibility.md` (per-backend matrix; SQL Server softened to "practically compatible for basic-Latin data" with `_CI_AI`), plus empirical research docs.

**Dependency:** ICU4J added to core (~13 MB).

## New concepts

**Collation-driven comparison (ICU / Unicode Collation Algorithm).** A collation decides how text compares — sort order, what counts as *equal*, and (for a transaction layer sitting on a CI-collated backend) what counts as *the same row*. This PR makes the configured collation the single authority for all three, via [ICU4J](https://unicode-org.github.io/icu/userguide/collation/), the reference UCA implementation that MySQL 8's defaults and PostgreSQL's ICU provider are also built on.

| Mode | Ordering | Equality (`=`/`!=` on TEXT) | Consensus Commit key identity |
|---|---|---|---|
| `BINARY` (default) | unsigned UTF-8 byte order | byte-exact | byte-exact |
| `ICU` | UCA per locale/strength/rules | the collation's equality (e.g. `'Apple' = 'apple'` at `PRIMARY`) | collation-canonical (one logical key per collation class) |

Why key identity must follow the collation: a CI-collated backend enforces one physical row per collation class and resolves point reads collation-aware — MySQL returns the stored `'Apple'` row for `get('apple')`. A byte-exact snapshot treats those as two unrelated keys and mis-books its own reads and writes. Conversely, `ICU` on a byte-order backend (Cassandra, DynamoDB, PostgreSQL `C`) is a misalignment — use `BINARY` there; the docs spell out the hazards.

## Post-Deploy Monitoring & Validation

No additional operational monitoring required — this repository ships a library; the change carries no service deployment. Validation is the CI matrix: the collation suites must show as executed (not skipped) on the mysql_8.0/8.4 and sqlserver jobs (`scalardb.jdbc.collation_test=required` enforces this on MySQL), and all existing jobs must stay green under the normalized default collations.

## Checklist


Torch3333:
Superseded by a stack of four PRs based on the new `feature/collation` trunk, so each layer can be reviewed on its own:

1. #3842 the `scalar.db.collation` setting, the comparator, and ICU rejection on storages that cannot honor it
2. #3843 in-memory text comparison by UTF-8 bytes or the configured ICU collation
3. #3844 Consensus Commit
4. #3845 integration tests and CI

Same content as this branch, reconciled with current `master` (error codes renumbered to 0304 to 0311) and without the `docs/` files. The `collation` branch stays as is. Closing this PR in favor of the stack.

PR #2801, opened 2025-06-20
Optimize JDBC adaptor for MySQL

brfrn169:
## Description

When benchmarking the new release, I observed some performance degradation with the JDBC adapter for MySQL. After investigating the issue, I found the following:

- Using `Connection.setReadOnly(true)` causes performance degradation.
- Enabling cursor fetch mode in MySQL also results in performance degradation.

To address these issues, this PR introduces optimizations to the JDBC adapter for MySQL.

## Related issues and/or PRs

N/A

## Changes made

Added some inline comments. Please take a look for the details.

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

N/A

## Release notes

N/A

PR #2651, opened 2025-05-14
Optimize JDBC storage by using Connection.setReadOnly()

brfrn169:
## Description

This PR updates the JDBC storage to use `Connection.setReadOnly()` to optimize read operations.

For details about the `Connection.setReadOnly()` method, see:  
https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#setReadOnly-boolean-

## Related issues and/or PRs

N/A

## Changes made

- Updated the JDBC storage to use `Connection.setReadOnly()` for the Get and Scan operations.

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

N/A

## Release notes

N/A

PR #3799, opened 2026-08-09
Fix ReadOnly for MySQL family

yito88:
## Description
Before MariaDB Java client `3.5.10`, `setReadOnly()` on a single-host connection sent no statement at all. Now it propagates to the server by default. ScalarDB's `JdbcUtils` calls `setReadOnly`, TiDB rejects the statement, and every metadata check fails.

## Related issues and/or PRs
- The MariaDB Java client was bumped up: #3775

## Changes made

- Override `setConnectionToReadOnly()` in `RdbEngineMysql` to do nothing. `RdbEngineMariaDB` and `RdbEngineTidb` extend it, so the whole MySQL family is covered through inheritance. This follows the existing `RdbEngineSqlite.setConnectionToReadOnly()` precedent.
- Update the five `setReadOnly` assertions in `JdbcAdminTest` to expect no call for `MYSQL` and `MARIADB` as well as `SQLITE`, via a new `verifyConnectionSetToReadOnly()` helper.
- Add `RdbEngineMysqlTest.setConnectionToReadOnly_ShouldDoNothing()` and a new `RdbEngineTidbTest` as regression guards.

## Checklist

One trade-off worth recording: for multi-host `jdbc:mysql:replication://` configurations, the driver uses the client-side read-only flag to route reads to a replica, and a no-op override drops that behavior. This is acceptable, and arguably desirable, since routing ScalarDB reads to a lagging replica would break our consistency assumptions anyway.

Read-only enforcement itself is unaffected: it is done at the ScalarDB layer (`ReadOnlyDistributedTransaction`), and the JDBC read paths only issue `SELECT`s.

## Release notes

N/A


yito88:
The TiDB integration tests in the weekly CI failed on 8th and 15th August. Those of 1st August passed. It makes sense since the version upgrade of the MariaDB Java client was merged on 2nd August.

https://github.com/scalar-labs/scalardb/actions/runs/31259224644
https://github.com/scalar-labs/scalardb/actions/runs/31886522088

Every error looks the same:
```
[Test worker] WARN org.mariadb.jdbc.message.server.ErrorPacket - Error: 1235-42000:
function READ ONLY has only noop implementation in tidb now, use tidb_enable_noop_functions...
175 tests completed, 133 failed
```

Torch3333 (approved):
LGTM, thank you!

brfrn169:
Thanks for the quick fix — the TiDB failure clearly needs to be addressed. Before this gets merged, I'd like to discuss widening the scope a bit, because I believe the same change is warranted for the whole MySQL family, and that a different mechanism might be more robust. I checked the driver and pool behavior locally and here is what I found.

### 1. MySQL and MariaDB are affected too, just silently

- `StandardClient.setReadOnly()` in 3.5.10 issues `SET SESSION TRANSACTION READ ONLY` / `READ WRITE` whenever the read-only state changes and the server is 5.6.5 or later.
- HikariCP resets the read-only state back to the pool default on connection return (`PoolBase.resetConnectionState()`), and we set `hikariConfig.setReadOnly(false)` in `JdbcUtils`.

So on MySQL / MariaDB / Aurora MySQL, every `JdbcDatabase.get()`, every `scan()` and every `beginReadOnly()` now costs two extra server round trips:

```
borrow -> SET SESSION TRANSACTION READ ONLY
       -> SELECT ...
close  -> SET SESSION TRANSACTION READ WRITE
```

For a point read that triples the number of round trips. It doesn't fail loudly like TiDB does, but it is a real performance regression introduced by the driver bump in #3775.

### 2. On MySQL there is no upside to offset that cost

`setConnectionToReadOnly()` was introduced as an optimization (#2651), but that rationale does not hold for the MySQL family with this driver:

- `RdbEngineMysql.requiresExplicitCommit()` is `false`, so `JdbcDatabase.get()` runs as a single `SELECT` with autocommit enabled. InnoDB already applies its read-only transaction optimization automatically in that case, so the `SET SESSION` statement buys nothing.
- The optimization does pay off on PostgreSQL, because pgjdbc's default `readOnlyMode=transaction` piggybacks the read-only state onto `BEGIN READ ONLY` at zero extra round trips. That is a PostgreSQL-specific property, not a general one.

We also don't lose any safety property by turning it off: read-only enforcement is done at the ScalarDB layer (`ReadOnlyDistributedTransaction`, plus the fact that the read paths only issue SELECTs).

### 3. Suggestion: override `setConnectionToReadOnly()` in `RdbEngineMysql` instead

```java
// RdbEngineMysql
@Override
public void setConnectionToReadOnly(Connection connection, boolean readOnly) {
  // Do nothing. Setting a connection to read-only brings no benefit in MySQL. A read is a single
  // SELECT with autocommit enabled, for which InnoDB applies the read-only transaction
  // optimization automatically, while MariaDB Connector/J 3.5.10 and later issue
  // SET SESSION TRANSACTION READ ONLY / READ WRITE, which adds two round trips per read since
  // HikariCP restores the pool default on connection return. TiDB rejects the statement entirely.
}
```

Compared to the JDBC URL option, this has a few advantages:

- `RdbEngineMariaDB` overrides `adjustJdbcUrl()` to return the URL as is, so a URL-based fix would not reach MariaDB even if it were moved up to `RdbEngineMysql`. An override of `setConnectionToReadOnly()` covers MySQL, MariaDB and TiDB through inheritance.
- The driver silently ignores unknown options. I verified this with `Configuration.parse()`: an unknown URL parameter and an unknown `Properties` entry both parse without any exception. So if `readOnlyPropagatesToServer` is renamed or removed in a future driver release, the fix would stop working with no signal at all and TiDB would break again.
- The diff gets smaller: the `adjustJdbcUrl()` override and its URL-format tests all go away. There is precedent for this shape in `RdbEngineSqlite.setConnectionToReadOnly()`.
- It also makes the bootstrap concern Copilot raised moot. (For what it is worth, I don't think that concern is valid as written: the detection connection in `RdbEngineFactory.createMysqlOrTidbEngine()` only calls `getMetaData()`, and Hikari's `setupConnection()` skips `setReadOnly()` when the value already matches the pool default, so no statement is sent there. That is consistent with the CI logs, where the failures are in the metadata and read paths rather than at bootstrap.)

One trade-off worth naming: with the URL option the client-side read-only flag is still set, and multi-host setups (`jdbc:mysql:replication://`) use that flag to route reads to a replica. A no-op override drops that behavior. I think that is acceptable, and arguably desirable, since routing ScalarDB reads to a lagging replica would break our consistency assumptions anyway.

### Test impact if we go this way

- `JdbcAdminTest` has five `if (rdbEngine == RdbEngine.SQLITE) ... else verify(connection).setReadOnly(true)` branches (around lines 318, 620, 2776, 3014 and 5089) that would need to cover MYSQL and MARIADB as well.
- `JdbcDatabaseTest` and `JdbcTransactionManagerTest` use POSTGRESQL, so they are unaffected.
- The test-only `RdbEngine` enum has no TiDB entry, so a direct test in `RdbEngineTidbTest` asserting that `setConnectionToReadOnly()` does not call `Connection#setReadOnly()` would be the regression guard for this issue.

WDYT? If you prefer to keep the blast radius of this PR small, an alternative is to merge the TiDB fix as is and handle the MySQL and MariaDB side in a follow-up, but I'd rather not leave the extra round trips in place for a release if we can avoid it.

yito88:
@brfrn169 Thanks, I measured it locally and switched to your suggestion.

Reproduced one ScalarDB read (borrow connection ‚`setConnectionToReadOnly(conn, true)` -> `SELECT` -> return), using HikariCP 4.0.3 and the same pool default `setReadOnly(false)` as `JdbcUtils`. 3000 iterations, counting both wall time and statements the server actually received (Questions delta).

I've tried it with MySQL 8.4.11
- 3.5.9 (before the bump) + setReadOnly: 1 statements/read, 244-250 ms
- 3.5.10 (current version) + setReadOnly: 3 statements/read, 616-646 ms
- 3.5.10 + no-op (this fix): 1 statements/read, 244-258 ms

The general log confirms the pattern:
```sql
SET SESSION TRANSACTION READ ONLY
SELECT v FROM t WHERE id = 1
SET SESSION TRANSACTION READ WRITE
```

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -492,6 +492,15 @@ public String adjustJdbcUrl(String jdbcUrl) {
     return jdbcUrl + (jdbcUrl.contains("?") ? "&" : "?") + "permitMysqlScheme=true";
   }
 
+  @Override
+  public void setConnectionToReadOnly(Connection connection, boolean readOnly) {
+    // Do nothing. Setting a connection to read-only brings no benefit in MySQL. A read is a single
+    // SELECT with autocommit enabled, for which InnoDB applies the read-only transaction
+    // optimization automatically, while MariaDB Connector/J 3.5.10 and later issue
+    // SET SESSION TRANSACTION READ ONLY / READ WRITE, which adds two round trips per read since
+    // HikariCP restores the pool default on connection return. TiDB rejects the statement entirely.
```
brfrn169: A small refinement to the comment I suggested earlier. The "single SELECT with autocommit enabled" part is narrower than the set of call sites this override now covers:

- `JdbcDatabase.get()` and all six `JdbcAdmin` call sites gate `setAutoCommit(false)` on `requiresExplicitCommit()`, which is `false` for the MySQL family, so those really are single autocommit SELECTs.
- `JdbcDatabase.scan()` calls `connection.setAutoCommit(false)` unconditionally, and `JdbcTransactionManager.begin(txId, true)` runs an explicit transaction. Neither is an autocommit single SELECT.

There is no behavior regression here: before 3.5.10 the driver sent nothing on single-host connections, so those two paths never had the READ ONLY marker to begin with. It is only that the stated rationale does not hold for them.

Since this comment is effectively what keeps the override from being dropped again, I would rather it rest on a reason that cannot go stale. The strongest one is not in the comment yet: read-only is enforced at the ScalarDB layer, so the JDBC flag was never load-bearing. Suggestion:

```suggestion
    // Do nothing. Setting a connection to read-only brings no benefit in MySQL: read-only is
    // enforced at the ScalarDB layer (see ReadOnlyDistributedTransaction) and the read paths
    // only issue SELECTs, while MariaDB Connector/J 3.5.10 and later issue SET SESSION
    // TRANSACTION READ ONLY / READ WRITE, which adds two round trips per read since HikariCP
    // restores the pool default on connection return. TiDB rejects the statement entirely.
```

core/src/test/java/com/scalar/db/storage/jdbc/RdbEngineMysqlTest.java
```diff
@@ -33,4 +38,15 @@ void adjustJdbcUrl_WithPermitMysqlSchemeAlreadyPresent_ShouldReturnAsIs() {
     String result = rdbEngineMysql.adjustJdbcUrl(url);
     assertThat(result).isEqualTo(url);
   }
+
+  @Test
+  void setConnectionToReadOnly_ShouldDoNothing() throws SQLException {
+    // MariaDB Connector/J 3.5.10 and later issue SET SESSION TRANSACTION READ ONLY / READ WRITE
+    // from Connection#setReadOnly(), which adds two round trips per read and which TiDB rejects.
```
brfrn169, line 46: Worth recording how this override was lost, since it happened once already and the same path could repeat.

- #2801 (`f3969d9ce`) originally added this exact no-op to `RdbEngineMysql`, with the comment "Observed performance degradation when using read-only connections in MySQL".
- #3428 (`fc8a6092c`) dropped it during the driver swap and, in the same commit, relaxed five `JdbcAdminTest` assertions from `MYSQL || SQLITE` to `SQLITE` (one was collapsed entirely). Reasonable at the time, since the MariaDB driver did not propagate `setReadOnly()` on single-host connections, so the override had no effect.
- This PR restores exactly those five assertion sites.

So the guard existed and did fire; it was relaxed to match the removal. That is the path worth closing, and the person who hits it next will be reading this test, not the production comment. Suggestion:

```suggestion
    // MariaDB Connector/J 3.5.10 and later issue SET SESSION TRANSACTION READ ONLY / READ WRITE
    // from Connection#setReadOnly(), which adds two round trips per read and which TiDB rejects.
    // This override was first added in #2801 and dropped in #3428; if this test starts failing,
    // check whether the driver still propagates the read-only state before relaxing it.
```

core/src/test/java/com/scalar/db/storage/jdbc/RdbEngineMysqlTest.java
```diff
@@ -33,4 +38,15 @@ void adjustJdbcUrl_WithPermitMysqlSchemeAlreadyPresent_ShouldReturnAsIs() {
...
+    Connection connection = mock(Connection.class);
+
+    rdbEngineMysql.setConnectionToReadOnly(connection, true);
+
+    verify(connection, never()).setReadOnly(true);
```
brfrn169: Nit: `anyBoolean()` would be a slightly stronger assertion here, and it matches what `JdbcAdminTest.verifyConnectionSetToReadOnly()` uses in this same PR (`JdbcAdminTest.java:136`, which inherited it from the pre-existing `if (rdbEngine == RdbEngine.SQLITE)` branches).

`setReadOnly(true)` passes identically today since the override body is empty. The difference only shows up if someone later changes the override to call `connection.setReadOnly(false)` explicitly, which `never().setReadOnly(true)` would not catch.

```suggestion
    verify(connection, never()).setReadOnly(anyBoolean());
```

This also needs `import static org.mockito.ArgumentMatchers.anyBoolean;` (goes between the `assertThat` and `Mockito.mock` static imports).

core/src/test/java/com/scalar/db/storage/jdbc/RdbEngineTidbTest.java
```diff
@@ -0,0 +1,40 @@
...
+  @Test
+  void setConnectionToReadOnly_ShouldDoNothing() throws SQLException {
+    // TiDB rejects "SET SESSION TRANSACTION READ ONLY" with error 1235 unless
+    // tidb_enable_noop_functions is enabled, and MariaDB Connector/J 3.5.10 and later issue that
+    // statement from Connection#setReadOnly(). Inherited from RdbEngineMysql.
+    Connection connection = mock(Connection.class);
+
+    rdbEngineTidb.setConnectionToReadOnly(connection, true);
+
+    verify(connection, never()).setReadOnly(true);
```
brfrn169: Same as in `RdbEngineMysqlTest` — `anyBoolean()` to match `JdbcAdminTest.verifyConnectionSetToReadOnly()`.

```suggestion
    verify(connection, never()).setReadOnly(anyBoolean());
```

Also needs `import static org.mockito.ArgumentMatchers.anyBoolean;` as the first static import.

brfrn169 (approved):
Left several minor comments. Other than that, LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #3805, opened 2026-08-21
Support the AWS Advanced JDBC Wrapper for Aurora failover

brfrn169:
## Description

Running ScalarDB on Aurora means that a writer failover stalls writes while the client waits for DNS to propagate. Measured against real Aurora clusters, that stall was about 26 seconds.

The AWS Advanced JDBC Wrapper avoids the wait by discovering the cluster topology and reconnecting to the new writer directly, but ScalarDB could not use it: `RdbEngineFactory` dispatches on the JDBC URL prefix and did not recognize `jdbc:aws-wrapper:`, so startup failed outright.

The interesting part turned out not to be the URL handling. ScalarDB already reports failover outcomes correctly, but for two different reasons on two different code paths, and neither was covered by a test. Most of this PR is about making those reasons explicit and keeping them from being broken.

## Related issues and/or PRs

N/A

## Changes made

**Accept the wrapper's URLs**

- `RdbEngineFactory` recognizes `jdbc:aws-wrapper:` and picks the engine from the URL underneath the prefix. Only Aurora PostgreSQL and Aurora MySQL are accepted; the wrapper also supports MariaDB, but allowing an untested combination would let it fail in subtler ways than an unsupported-engine error at startup.
- `JdbcUtils` hands the connection pool the wrapper's driver. Each engine's `getDriverClassName()` is left untouched.
- The URL given to HikariCP keeps the `jdbc:aws-wrapper:` prefix. Stripping it is only how the RDB dialect is identified; the prefix is what routes the connection through the wrapper at all. The unsupported-engine error also reports the URL the user configured rather than the stripped one.
- MySQL gets `wrapperTargetDriverDialect=mariadb-connector-j-3` added automatically. ScalarDB reaches MySQL through MariaDB Connector/J and does not bundle MySQL Connector/J, but the wrapper defaults to the MySQL Connector/J dialect for a `jdbc:mysql://` URL. A dialect the user set explicitly is left alone.

**Bundle the wrapper**

- Added to `core` as an `implementation` dependency, following how the AlloyDB connector is handled. All of the wrapper's declared dependencies are optional, so nothing is pulled in transitively — in particular, MySQL Connector/J is not among them and no MySQL code ships.

**Pin the failover exception handling**

No new classification code was needed, but the existing behavior was undocumented and untested:

- `JdbcTransaction#commit()` infers "outcome unknown" from the rollback failing. That only works because HikariCP evicts a connection whose SQLState starts with `08`. Setting HikariCP's `exceptionOverrideClassName`, which AWS recommends, would keep the connection alive, let the rollback succeed as a no-op, and report an unknown outcome as a `CommitException` — whose contract tells the caller it is safe to retry.
- `JdbcDatabase` relies on `isConflict` not matching failover SQLStates. Eviction is irrelevant there; adding `08*` to `isConflict` would turn an unknown outcome into a `RetriableExecutionException`.

Added a driver that simulates the failover exceptions so both paths can be exercised against a real HikariCP pool without a database or AWS credentials, plus a test that pins the wrapper's own exception contract so a version bump cannot silently change it.

**Refuse to start when HikariCP has an exception override configured**

The inference above only holds while HikariCP actually discards the connection. `HikariConfig`'s constructor reads `hikaricp.configurationFile`, so `exceptionOverrideClassName` can arrive without ScalarDB configuration being touched — and AWS documentation recommends setting it. Startup now fails instead, because the alternative is invisible: it surfaces only during a failover, looks like an ordinary retry, and gets found by noticing duplicated data.

**Classify what a scanner throws**

A scanner reports a read failure as an `ExecutionException` wrapping the `SQLException`, so iterating one was the only CRUD path that could not recognize a conflict. It now unwraps and classifies like a single operation.

**Stop discarding the commit failure**

When the follow-up rollback failed, the thrown `UnknownTransactionStatusException` described only the failed rollback, so a caller hitting a failover saw "Rolling back the transaction failed" with no trace of the SQLState that made the outcome unknown. The original failure is now kept as a suppressed exception.

## Checklist

## Additional notes

**One change reaches beyond Aurora.** Refusing to start when `exceptionOverrideClassName` is set applies to every JDBC deployment, not only wrapper users, because the inference it protects is not Aurora-specific either — a plain connection drop mid-commit reports the same SQLState class on any backend. An existing deployment that sets the override will now fail to start. That is intended: such a deployment is already exposed. Worth a look during review.

An earlier revision of this PR also reported a CRUD-phase connection failure as a conflict. That has been dropped after review: SQLState class 08 describes the cause rather than whether a retry helps, and it covers both a writer failover and a database that has gone away for good. `CrudException` already tells callers they may retry, so nothing changes for applications, and Consensus Commit reports the same event the same way.

**Verified against real Aurora clusters** (PostgreSQL 17.10 and MySQL 8.0.44, two instances each, since a managed failover promotes a reader). The clusters were deleted afterwards.

| | Write failures / attempts | Error window |
|---|---|---|
| With the wrapper | 1 / 560 | ~0 ms |
| Without | 90 / 617 | 26,159 ms |

Also confirmed there: Aurora MySQL connects through the existing MariaDB Connector/J, it is not mistaken for TiDB, and every failover exception carries its `08` SQLState on the exception itself rather than on a cause — which matters because HikariCP walks `getNextException()` but never `getCause()`.

**A note on `08S02`.** AWS documents it as "failover succeeded outside of a transaction", but that does not hold for how ScalarDB drives transactions. The wrapper decides between `08S02` and `08007` using `pluginService.isInTransaction()`, which tracks transaction state by parsing SQL text (`BEGIN` / `COMMIT`). ScalarDB uses the JDBC API instead, so the wrapper does not see a transaction in progress, and a commit that failed with uncommitted changes still reports `08S02`. Observed on a real cluster. It is handled correctly either way — both states start with `08`, so HikariCP evicts and the outcome is reported as unknown — but it is the reason no failover SQLState is treated as "nothing was applied".

**User-facing documentation is not in this repo.** A hand-off covering the configuration reference, the exception semantics, and why `exceptionOverrideClassName` must not be set has been prepared separately for the docs repository.

**Aurora is not needed to test this.** The wrapper works against any MySQL or PostgreSQL, so pointing the existing Docker-based integration tests at a `jdbc:aws-wrapper:` URL would put this on CI without AWS credentials. Worth a follow-up issue.

## Release notes

Added support for the AWS Advanced JDBC Wrapper, which enables fast Aurora failover for Aurora PostgreSQL and Aurora MySQL. Configuring HikariCP with an exception override is now rejected at startup, since it would cause a transaction with an unknown outcome to be reported as a definite failure.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -102,6 +102,7 @@ subprojects {
         alloyDbJdbcConnectorVersion = '1.3.2'
         googleCloudStorageVersion = '2.71.0'
         spannerDriverVersion = '2.42.0'
+        awsAdvancedJdbcWrapperVersion = '4.3.0'
         picocliVersion = '4.7.7'
         commonsTextVersion = '1.15.0'
         caffeineVersion = '2.9.3'
--- a/build.gradle
+++ b/build.gradle
@@ -98,6 +98,9 @@ subprojects {
         sqliteDriverVersion = '3.53.2.1'
         yugabyteDriverVersion = '42.7.3-yb-4'
         db2DriverVersion = '12.1.5.0'
+        // Bumping the major version means updating the wrapperTargetDriverDialect value in
+        // RdbEngineMysql too -- the AWS wrapper names its MariaDB dialect after the driver's major
+        // version, and only defines one for v3.
         mariadDbDriverVersion = '3.5.10'
         alloyDbJdbcConnectorVersion = '1.3.2'
         googleCloudStorageVersion = '2.71.0'
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
@@ -2,6 +2,7 @@ package com.scalar.db.storage.jdbc;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.hash.Hashing;
+import com.scalar.db.common.CoreError;
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import java.nio.charset.StandardCharsets;
@@ -80,6 +81,18 @@ public final class JdbcUtils {
       @Nullable Long maxLifetime,
       @Nullable Long keepaliveTime) {
     HikariConfig hikariConfig = new HikariConfig();
+
+    // The constructor above also reads whatever the hikaricp.configurationFile system property
+    // points at, so this setting can arrive without ScalarDB configuration being involved at all --
+    // and AWS documentation recommends setting it when using their JDBC wrapper. Refuse to start
+    // rather than run with the inference in JdbcTransaction#commit() quietly broken. That failure
+    // would only surface during a failover, would look like an ordinary retry, and would be found
+    // by noticing duplicated data.
+    if (hikariConfig.getExceptionOverrideClassName() != null) {
+      throw new IllegalArgumentException(
+          CoreError.JDBC_HIKARICP_EXCEPTION_OVERRIDE_NOT_SUPPORTED.buildMessage(
+              hikariConfig.getExceptionOverrideClassName()));
+    }
 
     // Do not set exceptionOverrideClassName here. JdbcTransaction#commit() infers an unknown
     // outcome from HikariCP evicting a connection whose SQLState starts with "08"; see that method
--- a/core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
@@ -19,6 +19,8 @@ public final class JdbcUtils {
   // SQL Server: 128).
   @VisibleForTesting static final int MAX_INDEX_NAME_LENGTH = 63;
 
+  private static final String AWS_WRAPPER_URL_PREFIX = "jdbc:aws-wrapper:";
+
   private JdbcUtils() {}
 
   public static HikariDataSource initDataSource(JdbcConfig config, RdbEngineStrategy rdbEngine) {
@@ -79,12 +81,16 @@ public final class JdbcUtils {
       @Nullable Long keepaliveTime) {
     HikariConfig hikariConfig = new HikariConfig();
 
+    // Do not set exceptionOverrideClassName here. JdbcTransaction#commit() infers an unknown
+    // outcome from HikariCP evicting a connection whose SQLState starts with "08"; see that method
+    // for why AWS recommends the override and why setting it would break the inference.
+
     /*
      * We need to set the driver class of an underlying database to the dataSource in order
      * to avoid the "No suitable driver" error when ServiceLoader in java.sql.DriverManager doesn't
      * work (e.g., when we dynamically load a driver class from a fatJar).
      */
-    hikariConfig.setDriverClassName(rdbEngine.getDriverClassName());
+    hikariConfig.setDriverClassName(getDriverClassName(config, rdbEngine));
 
     hikariConfig.setJdbcUrl(rdbEngine.adjustJdbcUrl(config.getJdbcUrl()));
     rdbEngine.setConnectionCredentials(config, hikariConfig);
@@ -126,6 +132,32 @@ public final class JdbcUtils {
   @VisibleForTesting
   static HikariDataSource createDataSource(HikariConfig hikariConfig) {
     return new HikariDataSource(hikariConfig);
+  }
+
+  static boolean isAwsWrapperUrl(String jdbcUrl) {
+    return jdbcUrl != null && jdbcUrl.startsWith(AWS_WRAPPER_URL_PREFIX);
+  }
+
+  /**
+   * Exposes the URL of the underlying database. Use this only to decide which RDB engine to use:
+   * the URL passed to the connection pool must keep the prefix, because that is what makes the
+   * wrapper handle the connection.
+   */
+  static String removeAwsWrapperPrefix(String jdbcUrl) {
+    assert isAwsWrapperUrl(jdbcUrl);
+    return "jdbc:" + jdbcUrl.substring(AWS_WRAPPER_URL_PREFIX.length());
+  }
+
+  /**
+   * The AWS Advanced JDBC Wrapper supplies its own driver, so the underlying database's driver
+   * class must not be used when the URL routes through it. The engine's own {@code
+   * getDriverClassName()} is left untouched; the substitution happens only here.
+   */
+  private static String getDriverClassName(JdbcConfig config, RdbEngineStrategy rdbEngine) {
+    if (isAwsWrapperUrl(config.getJdbcUrl())) {
+      return software.amazon.jdbc.Driver.class.getName();
+    }
+    return rdbEngine.getDriverClassName();
   }
 
   private static String toHikariTransactionIsolation(Isolation isolation) {
--- a/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineFactory.java
+++ b/core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineFactory.java
@@ -14,6 +14,10 @@ public final class RdbEngineFactory {
   public static RdbEngineStrategy create(JdbcConfig config) {
     String jdbcUrl = config.getJdbcUrl();
 
+    if (JdbcUtils.isAwsWrapperUrl(jdbcUrl)) {
+      return createForAwsWrapper(config, jdbcUrl);
+    }
+
     if (jdbcUrl.startsWith("jdbc:mysql:")) {
       return createMysqlOrTidbEngine(config);
     } else if (jdbcUrl.startsWith("jdbc:postgresql:")) {
@@ -36,6 +40,36 @@ public final class RdbEngineFactory {
       throw new IllegalArgumentException(
           CoreError.JDBC_RDB_ENGINE_NOT_SUPPORTED.buildMessage(jdbcUrl));
     }
+  }
+
+  /**
+   * Selects the engine for a URL that routes through the AWS Advanced JDBC Wrapper.
+   *
+   * <p>Only Aurora PostgreSQL and Aurora MySQL are supported. The wrapper itself also accepts
+   * MariaDB, but allowing a combination that is never tested would let it fail in subtler ways than
+   * an unsupported-engine error at startup. TiDB is rejected for the same reason: it shares MySQL's
+   * connection string, so it is only recognized after the metadata probe.
+   *
+   * @param config the config
+   * @param jdbcUrl the original JDBC URL, including the wrapper prefix
+   * @return the engine for the underlying database
+   */
+  private static RdbEngineStrategy createForAwsWrapper(JdbcConfig config, String jdbcUrl) {
+    String underlyingUrl = JdbcUtils.removeAwsWrapperPrefix(jdbcUrl);
+
+    if (underlyingUrl.startsWith("jdbc:postgresql:")) {
+      return new RdbEnginePostgresql();
+    } else if (underlyingUrl.startsWith("jdbc:mysql:")) {
+      RdbEngineStrategy engine = createMysqlOrTidbEngine(config);
+      if (!(engine instanceof RdbEngineTidb)) {
+        return engine;
+      }
+    }
+
+    // Report the URL the user actually configured, not the one left after stripping the prefix.
+    throw new IllegalArgumentException(
+        CoreError.JDBC_RDB_ENGINE_NOT_SUPPORTED_WITH_AWS_ADVANCED_JDBC_WRAPPER.buildMessage(
+            jdbcUrl));
   }
 
   /**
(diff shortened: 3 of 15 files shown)


core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
```diff
@@ -79,12 +81,16 @@ private static HikariDataSource createDataSource(
       @Nullable Long keepaliveTime) {
     HikariConfig hikariConfig = new HikariConfig();
 
+    // Do not set exceptionOverrideClassName here. JdbcTransaction#commit() infers an unknown
```
feeblefakie, line 97: `new HikariConfig()` also loads properties from the `hikaricp.configurationFile` system property, so an operator can set `exceptionOverrideClassName` without touching ScalarDB config, and AWS docs recommend exactly that override. Should we fail fast here if `getExceptionOverrideClassName() != null` instead of relying on this comment?

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -357,14 +373,54 @@ public void rollback() throws RollbackException {
...
+   * committed: the pool opens transactional connections with autoCommit disabled, so nothing has
+   * been applied, and the server discards the open transaction when the connection dies. The
+   * transaction is therefore safe to retry from the beginning.
+   *
+   * <p>This reasoning holds only in the CRUD phase. Do not reuse it for {@link #commit()}, where a
+   * lost connection leaves the outcome genuinely unknown, nor for {@code
+   * RdbEngineStrategy#isConflict}, which the storage layer uses without knowing which phase it is
+   * serving.
+   */
+  private static boolean isConnectionException(SQLException e) {
```
feeblefakie: This matches every 08 state for every JDBC backend, and the retry contract depends on the application doing backoff. For example, the MariaDB driver uses 08000/08006 for plain socket timeouts, so an overloaded database can look retriable. Should we mention that here and in the release notes? Or narrow this to the failover states?

feeblefakie (approved):
LGTM! Thank you! I left minor comments. PTAL!

build.gradle
```diff
@@ -102,6 +102,7 @@ subprojects {
         alloyDbJdbcConnectorVersion = '1.3.2'
         googleCloudStorageVersion = '2.71.0'
         spannerDriverVersion = '2.42.0'
+        awsAdvancedJdbcWrapperVersion = '4.3.0'
```
Torch3333, line 108: The driver doc mentions this limitation for MariaDB but I assume this does not apply when using the MariaDB driver with MySQL ?
https://github.com/aws/aws-advanced-jdbc-wrapper#mariadb

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -482,14 +491,32 @@ public TimestampTZColumn parseTimestampTZColumn(ResultSet resultSet, String colu
...
+
+    // ScalarDB talks to MySQL through MariaDB Connector/J and does not bundle MySQL Connector/J.
+    // The AWS Advanced JDBC Wrapper, however, defaults to the MySQL Connector/J dialect for a
+    // "jdbc:mysql://" URL, so without this the wrapper asks for a driver that is not on the
+    // classpath. Users who set the dialect themselves keep their choice.
+    if (JdbcUtils.isAwsWrapperUrl(jdbcUrl)
+        && !adjustedJdbcUrl.contains(WRAPPER_TARGET_DRIVER_DIALECT_PARAMETER)) {
+      adjustedJdbcUrl =
+          appendParameter(
+              adjustedJdbcUrl, WRAPPER_TARGET_DRIVER_DIALECT_PARAMETER + "=mariadb-connector-j-3");
```
Torch3333, line 516: In the string `mariadb-connector-j-3`, is `j-3` related to the current MariaDB driver major version (currently version 3) used by ScalarDB ?
If so, when the major version of the MariaDB driver is updated to v4, do we need to udpate this string value to `j-4`?

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -357,14 +373,54 @@ public void rollback() throws RollbackException {
...
+    String sqlState = e.getSQLState();
+    return sqlState != null && sqlState.startsWith("08");
```
Torch3333: Even though this exception code seems common for all storages, should we move this error code assertion in the `RdbEngine` with other storages specific error codes ?

Torch3333 (commented):
I left some comments, PTAL.

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
```diff
@@ -357,14 +373,37 @@ public void rollback() throws RollbackException {
   }
 
   private CrudException createCrudException(SQLException e, String message) {
-    if (rdbEngine.isConflict(e)) {
+    if (rdbEngine.isConflict(e) || isConnectionException(e)) {
```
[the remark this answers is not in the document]
brfrn169: Good catch, fixed. `createCrudException` now has an overload that unwraps the `ExecutionException` a scanner reports and classifies the `SQLException` underneath, so `one()` and `all()` go through the same path as a single operation.

The classification itself has since changed: connection failures are no longer treated as retriable (see the discussion on the `isConnectionException` thread), so what this now recovers is conflict detection during scanner iteration. That was worth keeping on its own.

Fixed in 231be1cd7 (`JdbcTransaction#createCrudException(ExecutionException)`, with the scanner wrapper routed through it).

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineFactory.java
```diff
@@ -38,6 +42,31 @@ public static RdbEngineStrategy create(JdbcConfig config) {
...
+   * @param jdbcUrl the original JDBC URL, including the wrapper prefix
+   * @return the engine for the underlying database
+   */
+  private static RdbEngineStrategy createForAwsWrapper(JdbcConfig config, String jdbcUrl) {
+    String underlyingUrl = JdbcUtils.removeAwsWrapperPrefix(jdbcUrl);
+
+    if (underlyingUrl.startsWith("jdbc:postgresql:")) {
+      return new RdbEnginePostgresql();
+    } else if (underlyingUrl.startsWith("jdbc:mysql:")) {
+      return createMysqlOrTidbEngine(config);
```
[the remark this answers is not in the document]
brfrn169: Right, that contradicted what the method documented. Fixed: the wrapper path now rejects a TiDB result rather than returning it, and it reports a dedicated error rather than the generic unsupported-engine one, so the message says what actually went wrong.

Fixed in 231be1cd7 (`RdbEngineFactory#createForAwsWrapper`, plus a dedicated `CoreError` entry).

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
```diff
@@ -482,14 +491,32 @@ public TimestampTZColumn parseTimestampTZColumn(ResultSet resultSet, String colu
...
+    if (!adjustedJdbcUrl.contains(PERMIT_MYSQL_SCHEME_PARAMETER)) {
+      adjustedJdbcUrl = appendParameter(adjustedJdbcUrl, PERMIT_MYSQL_SCHEME_PARAMETER + "=true");
+    }
+
+    // ScalarDB talks to MySQL through MariaDB Connector/J and does not bundle MySQL Connector/J.
+    // The AWS Advanced JDBC Wrapper, however, defaults to the MySQL Connector/J dialect for a
+    // "jdbc:mysql://" URL, so without this the wrapper asks for a driver that is not on the
+    // classpath. Users who set the dialect themselves keep their choice.
+    if (JdbcUtils.isAwsWrapperUrl(jdbcUrl)
+        && !adjustedJdbcUrl.contains(WRAPPER_TARGET_DRIVER_DIALECT_PARAMETER)) {
```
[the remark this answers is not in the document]
brfrn169: Accurate, but not fixing it here. Triggering it needs `wrapperTargetDriverDialect` to appear in a database name or another parameter value, which nothing realistic produces, and the same substring check has guarded `permitMysqlScheme` on this line since well before this PR. Changing one and not the other would be worse than leaving both, and changing both is a separate cleanup rather than something this PR should carry.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcUtils.java
brfrn169: You are right, and I had not appreciated how reachable this is. Confirmed in the HikariCP bytecode: the no-arg `HikariConfig` constructor reads `hikaricp.configurationFile` and loads it, so the setting can arrive without ScalarDB configuration being touched at all -- and AWS documentation recommends exactly this override. A comment in our source reaches nobody on that path.

Added the check, and it refuses to start rather than warning. The failure it prevents is invisible: it only surfaces during a failover, looks like an ordinary retry when it does, and gets found by noticing duplicated data.

The check is not limited to wrapper URLs, because the inference it protects is not either -- a plain connection drop mid-commit reports the same SQLState class on any JDBC backend. That does mean an existing deployment setting the override will now fail to start, which is intended: such a deployment is already exposed.

Added in 811370f1f (`JdbcUtils#createDataSource`, with a test that injects the setting through `hikaricp.configurationFile` so the path you described is the one exercised).

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
brfrn169: Removed the classification entirely rather than narrowing it.

Your point stands on its own: SQLState class 08 describes the cause -- a connection problem -- and says nothing about whether retrying helps. A writer failover belongs in the retriable set; a database that has gone away for good does not, and both arrive as 08. `CrudConflictException` promises retry works with no hedge, so claiming it for the whole class was an overpromise.

What settled it was that nothing is lost by dropping it. `CrudException` already tells callers they may retry, just without the guarantee, so an application behaves identically either way. And Consensus Commit already reports the same event as a plain `CrudException`, so keeping this only in the JDBC transaction manager created an asymmetry with no basis behind it -- a connection lost during the CRUD phase applies nothing in either.

On backoff: agreed that it is the application's responsibility, and that is unchanged by this.

Removed in a09a292fb (`JdbcTransaction#createCrudException`; the `isConnectionException` helper is gone).

build.gradle
brfrn169: Looked into this, and the honest answer is that it is not as clear-cut as either of us assumed.

The limitation names `usePipelineAuth` and `useBatchMultiSend`. Neither exists in MariaDB Connector/J 3.5.10 -- they were v2 options, so the documented workaround cannot be applied as written. That part of the doc has not been updated for v3.

But the functionality did not go away. v3 has a single `disablePipeline` option covering both: the driver uses it in the `StandardClient` constructor (authentication pipelining, the old `usePipelineAuth`) and in `executePipeline` (the old `useBatchMultiSend`). The wrapper does not set it for us either -- nothing in its jar references any of the three names.

So the limitation is about the driver rather than the engine, and it would apply to us. What I could not find is any statement that v3 resolved the incompatibility.

Not setting `disablePipeline`, on the basis that pipelining is a performance optimization and there is no evidence it misbehaves here: the Aurora runs for this PR connected fine (which exercises authentication pipelining), got through the integration suite, and showed nothing unusual across four failovers. That is absence of evidence rather than evidence of absence, so if you would rather we set it explicitly, I have no strong argument against it.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineMysql.java
brfrn169: Yes, and yes. Confirmed in the wrapper jar: its dialect table defines `pgjdbc`, `mysql-connector-j` and `mariadb-connector-j-3`, and MariaDB is the only one carrying a version. There is one `MariadbTargetDriverDialect` class, for v3.

So moving to a v4 driver needs the wrapper to name a v4 dialect first, and then this string updated. ScalarDB cannot get ahead of it.

Added a note in both places, since the trigger and the fix are in different files: at `mariadDbDriverVersion` in build.gradle so whoever bumps it knows this string exists, and here so whoever finds the string knows why it carries a number.

Noted in 7d6952822 (`build.gradle` at `mariadDbDriverVersion`, and `RdbEngineMysql#adjustJdbcUrl`).

core/src/main/java/com/scalar/db/transaction/jdbc/JdbcTransaction.java
brfrn169: This one resolved itself -- the check is gone. See the thread on the same method: SQLState class 08 turned out to mix conditions that a retry fixes with ones it never will, so treating the whole class as retriable was claiming more than it could support, and it has been dropped.

Worth recording your instinct anyway, because it was right about placement. Had the check stayed, `RdbEngine` is where it belonged: SQLState interpretation is per-backend and already lives there in `isConflict`. Having a second interpreter in the transaction manager was a smell.

The removal is a09a292fb.

Torch3333 (approved):
LGTM, thank you!

PR #3806, opened 2026-08-21
Replace BranchTransaction.end() with end(Status)

brfrn169:
## Description

`BranchTransaction.end()`'s contract and its implementations disagreed, and a branch had no way to declare that its work failed.

The Javadoc required `end()` "after its last CRUD operation and before the owning `GlobalTransaction` is committed **or rolled back**". Nothing enforced the rollback half — `TwoPhaseCommitBackedGlobalTransaction.rollback()` ignores branch state entirely — and no test followed it. `GlobalTransactionTestBase` called `end()` zero times, including on its commit paths, so the obligation the contract exists to state had no coverage at all.

Separately, a process that runs a branch but holds no `GlobalTransaction` handle had no failure-path hook. `end()` was a success-only call, so when such a process failed mid-branch there was nothing to release what it held.

This replaces `end()` with `end(Status)`, so a branch declares the outcome of its work wherever it is ended.

## Related issues and/or PRs

N/A

## Changes made

**API.** `void end()` becomes `void end(BranchTransaction.Status)`, with a nested `Status` enum of `SUCCESS` and `FAILURE`. There is no no-arg overload and no alias — a no-arg success path would keep success free to declare and failure the thing you must remember, which is the bias this change exists to correct.

**Semantics.** `end(SUCCESS)` keeps the previous behavior of `end()`: it rejects a branch that has already been ended and requires every scanner obtained from the branch to be closed. `end(FAILURE)` is lenient so it can be called from a `catch` block without masking the original failure — an already-ended branch is a no-op, and an open scanner is accepted rather than refused. `end(null)` throws `NullPointerException` before any state change, leaving the branch usable.

**`end(FAILURE)` deliberately does not close open scanners.** A scanner's `close()` runs `putIntoScanSetInSnapshot` / `putIntoScannerSetInSnapshot` and `verifyNoOverlap`, and `Snapshot.toSerializable` re-validates the scanner set at commit. Closing from a branch's cleanup path could therefore abort the whole transaction; in the single-phase backing all branches share one snapshot, so one branch's cleanup would inject validation targets the others are checked against. Nothing durable is lost by leaving them: `ConsensusCommit.rollback()`, `ConsensusCommitParticipant.rollbackRecords`, and `releaseTransactionContext` all close scanners one layer down, log-and-swallowing as a cleanup path should.

**Contract.** The Javadoc now names one obligation per side rather than bundling both into one sentence: the process running a branch ends it exactly once with no exemption, and whoever drives the outcome does not commit until every branch ended with `SUCCESS`. Naming the actors matters, because the two usually fall to different processes — one holding no `GlobalTransaction` cannot carry the second. The second is explicitly marked as unenforced; restating an unchecked obligation without saying so is the defect this change removes. `GlobalTransaction` and `GlobalTransactionManager` state the same contract and move with it.

Also documented on `BranchTransaction`: that a branch handle is not thread-safe and `end()` is no exception, that scanners should be closed *before* ending (closing one afterwards is worse than leaving it open, for the reason above), and that ending grants a branch no lifetime guarantee.

**Call sites.** Every branch in `GlobalTransactionTestBase` now declares an outcome. The `abort()` terminal takes `FAILURE`: `abort()` is a `default` alias of `rollback()`, so treating it as a commit path would have put the shared test base on record demonstrating `SUCCESS` followed by a rollback.

`GlobalTransactionBackedDistributedTransaction` held both handles and ended the branch on neither path, which put the whole `DistributedTransaction` integration corpus on a contract-violating path. Its two terminals need different handling:

- `commit()` maps `end(Status)`'s checked exceptions onto its own, a conflict staying retriable. `CommitConflictException extends CommitException`, so the existing `throws` clause already admits both.
- `rollback()` cannot do the same, because `RollbackException` and `CrudException` are unrelated siblings under `TransactionException`. It logs and swallows, and rolls back in a `finally`, so a failure while ending can never strand an in-flight transaction.

The remaining sites are mock-based manager wiring tests that never run a transaction. They end their branches too rather than being exempted — the contract admits no exemption, and a rule with an asterisk is worth less than five lines.

## Checklist

**This is a breaking change to a public API.** `BranchTransaction` shipped in 3.19.0 and carries no `@Beta`, so removing `end()` breaks callers, not only implementors. That is accepted deliberately.

**Tests.** Every cell of the `end(Status)` behavior matrix is pinned per backing, including all four ordering pairs, `end(null)`, and that ended-ness is per handle rather than per transaction. `BranchTransactionIdiomTest` is new: it holds the three idioms published in `BranchTransaction`'s Javadoc verbatim and runs them against a real backing, because a Javadoc code block is never compiled and a published sample can otherwise drift from the API without anything noticing.

The cells that `end(Status)`'s Javadoc states as contract are pinned a second time in `GlobalTransactionTestBase`, which ships as `scalardb-integration-test` and is therefore the corpus an implementation of the API runs against — unit tests of the backings this repository ships cover no one else's, and the published idioms depend on these rules. One cell is deliberately left out: that ended-ness is per handle rather than per transaction. The Javadoc does not state it, and promoting it would bind every implementation to two live handles for one transaction on one manager, which an implementation whose `end` has a backing effect could not provide.

**Verified locally.** `GlobalTransactionTestBase` was run against PostgreSQL 17 on both backings — 17 and 18 tests, no failures — which covers the scanner test that pins the half of the `end(FAILURE)` decision unit tests cannot reach, and the lifecycle cells added alongside it. The new lifecycle tests were also checked for vacuity: making `end(FAILURE)` strict fails exactly the two leniency tests and nothing else. The other storages remain to CI.

## Release notes

Replaced `BranchTransaction.end()` with `BranchTransaction.end(BranchTransaction.Status)`, which requires a branch of a global transaction to declare the outcome of its work. Replace existing `end()` calls with `end(BranchTransaction.Status.SUCCESS)`, and call `end(BranchTransaction.Status.FAILURE)` on failure paths.


integration-test/src/main/java/com/scalar/db/api/GlobalTransactionTestBase.java
```diff
@@ -325,20 +343,53 @@ public void beginBranch_ForTwoBranches_ShouldReturnBranchesKeyedByGlobalTransact
...
     assertThat(branch1.getId()).isEqualTo(global.getId());
     assertThat(branch2.getId()).isEqualTo(global.getId());
+    branch1.end(BranchTransaction.Status.FAILURE);
+    branch2.end(BranchTransaction.Status.FAILURE);
     global.rollback();
   }
 
+  @Test
+  public void
+      rollback_WhenBranchEndedWithFailureLeavingScannerOpen_ShouldNotPersistAndShouldReleaseState()
```
feeblefakie, line 360: The `end(Status)` matrix (FAILURE after SUCCESS is a no-op, double FAILURE, SUCCESS after end is rejected) is only tested in the unit tests of the two built-in backings. Since the idioms in the Javadoc rely on this behavior, should `GlobalTransactionTestBase` also cover it so that other implementations are tested against it?

feeblefakie (approved):
LGTM! Thank you! I left a minor comment. PTAL!

Torch3333 (approved):
LGTM, thank you!

Torch3333 (approved):
LGTM, thank you!

integration-test/src/main/java/com/scalar/db/api/GlobalTransactionTestBase.java
brfrn169: Good call — done in 422959899.

`GlobalTransactionTestBase` now pins the documented cells of the matrix: `end(SUCCESS)` twice and `end(SUCCESS)` after `end(FAILURE)` are refused; `end(FAILURE)` after `SUCCESS` and `end(FAILURE)` twice are quiet; `end(null)` leaves the branch usable; `end(SUCCESS)` refuses an open scanner and accepts it once closed; and CRUD after `end` is refused whichever status ended it.

I picked those by one rule — promote a cell only where `end(Status)`'s Javadoc states it as contract. That deliberately leaves out one cell the unit tests cover: that ended-ness is per handle rather than per transaction. The Javadoc does not state it, and it is the cell I would least want to bind every implementation to, since it requires two live handles for the same transaction on one manager — which an implementation whose `end` has a backing effect could not provide. It stays a unit test.

Two smaller things that came out of it. The tests that leave their branch ended with `SUCCESS` commit rather than roll back, so the shared corpus never demonstrates `SUCCESS` followed by a rollback — the same reason `abort()` takes `FAILURE`. And the class Javadoc no longer claims every test drives two branches.

Verified on PostgreSQL 17 against both backings (17 and 18 tests, no failures). I also checked the new tests actually bite rather than passing vacuously: making `end(FAILURE)` strict fails exactly the two leniency tests and nothing else.

PR #3808, opened 2026-08-21
Make ConsensusCommitAdmin.tableExists() a single-table lookup

brfrn169:
## Description

`ConsensusCommitAdmin.tableExists()` is far more expensive than it looks.

It is not overridden, so it uses the default implementation in `Admin`:

```java
default boolean tableExists(String namespace, String table) throws ExecutionException {
  return getNamespaceTableNames(namespace).contains(table);
}
```

and `ConsensusCommitAdmin.getNamespaceTableNames()` loads the metadata of every table in the namespace in order to filter out non-transaction tables:

```java
for (String table : admin.getNamespaceTableNames(namespace)) {
  // Remove non transaction table
  if (getTableMetadata(namespace, table) != null) {
    tables.add(table);
  }
}
```

So a single existence check costs a number of storage queries proportional to the number of tables in the namespace, and nothing in the admin chain caches metadata. Measured against PostgreSQL with a namespace holding 18 tables, one `tableExists()` call issued **39 SQL statements** and took about **44 ms**. Callers that check existence before an operation pay this every time, and the cost grows as users add tables.

## Related issues and/or PRs

None.

## Changes made

Overrode `tableExists()` in `ConsensusCommitAdmin` so that it looks at one table rather than the whole namespace:

```java
return getTableMetadata(namespace, table) != null && admin.tableExists(namespace, table);
```

This is the same condition `getNamespaceTableNames()` filters on — the storage has the table, and it has transaction metadata — so the result is unchanged, including for tables that exist in the storage without transaction metadata.

The storage is asked first, in the same order as `getNamespaceTableNames()`, which only reaches for metadata once it knows the table is there. Reading metadata for a table the storage does not have is not always harmless: on DynamoDB, `getTableMetadata()` fails outright while the metadata table has yet to be created, so checking it first breaks creating the very first table. The coordinator namespace is short-circuited to false to match `getNamespaceTableNames()`, which returns an empty set for it.

Measured against the same PostgreSQL namespace, one `tableExists()` call now issues **5 SQL statements** and takes about **4.7 ms**.

## Checklist

### Why the default in `Admin` is left alone

Answering a single-table question by listing the whole namespace is an odd shape, and resolving the default from `getTableMetadata(namespace, table) != null` is tempting: that is what every caller in this repository actually means by `tableExists()`, and it would fix the storage admins too, including DynamoDB's account-wide `ListTables`. It was tried and it does not work.

The listing the default uses reads only table names, while `getTableMetadata()` parses the stored metadata. Repairing a table whose metadata is corrupt depends on that difference. With the default changed, `ConsensusCommitAdminRepairIntegrationTest` fails: `repairTable_ForCorruptedMetadataTable_ShouldRepairProperly` writes a metadata row with an unparseable data type, and the subsequent `dropTable(..., ifExists)` throws `No enum constant DataType.corrupted` from inside `tableExists()` instead of reporting whether the table is there. Repair can no longer clean up after itself, the coordinator tables are left behind, and every following test fails on `Coordinator tables already exist`.

Measured on a clean database, changing the default fails 4 of the 11 repair tests, while this PR passes all 11. Making the default a single-table lookup has to wait until the repair paths stop depending on `tableExists()` tolerating unreadable metadata, and that is a separate change.

### Testing

Added unit tests covering a transaction table, a storage table without transaction metadata, a table the storage does not have, and the coordinator namespace, each asserting that the namespace is no longer listed. Two of them pin the ordering that DynamoDB depends on: the metadata is not read when the storage does not have the table, and the storage is not reached at all for the coordinator namespace. The existing coordinator namespace test was extended to cover `tableExists()` as well.

`:core:test` and `:core:spotlessCheck` pass. `integrationTestJdbc` against PostgreSQL 17 passes (6847 tests). One run of that suite hit `ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation.get_GetWithIndexForTransientDeletedAndPreparedRecordsWhenWriterAborted_ShouldResolveToNoRecord`, which failed on a PostgreSQL serialization conflict during recovery (`could not serialize access due to read/write dependencies among transactions ... might succeed if retried`); a second run of the full suite passed, the class passes on its own, and `tableExists()` is not on that code path.

`:core:spotbugsMain` fails on `KeyBytesEncoder` both with and without this change, so that failure is pre-existing and unrelated.

`DistributedStorageAdmin` has the same default `tableExists()`, but its `getNamespaceTableNames()` implementations do not perform a per-table metadata lookup, so the storage-level check does not have this amplification. It was left alone.

## Release notes

Reduced the number of storage queries that `ConsensusCommitAdmin.tableExists()` issues, which previously scaled with the number of tables in the namespace.

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdmin.java
@@ -308,6 +308,23 @@ public class ConsensusCommitAdmin implements DistributedTransactionAdmin {
   }
 
   @Override
+  public boolean tableExists(String namespace, String table) throws ExecutionException {
+    // The default implementation checks the table against getNamespaceTableNames(), which loads the
+    // metadata of every table in the namespace. Answer the same question with a single table
+    // instead: the table is a transaction table when the storage has it and it has transaction
+    // metadata.
+    //
+    // The storage is asked first, in the same order as getNamespaceTableNames(), which only reaches
+    // for metadata once it knows the table is there. Reading metadata for a table the storage does
+    // not have is not always harmless: on DynamoDB it fails outright while the metadata table has
+    // yet to be created.
+    if (namespace.equals(coordinatorNamespace)) {
+      return false;
+    }
+    return admin.tableExists(namespace, table) && getTableMetadata(namespace, table) != null;
+  }
+
+  @Override
   public boolean namespaceExists(String namespace) throws ExecutionException {
     if (namespace.equals(coordinatorNamespace)) {
       return false;
--- a/core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdminTestBase.java
+++ b/core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitAdminTestBase.java
@@ -600,6 +600,93 @@ public abstract class ConsensusCommitAdminTestBase {
   }
 
   @Test
+  public void tableExists_ForTransactionTable_ShouldReturnTrueWithoutListingTheNamespace()
+      throws ExecutionException {
+    // Arrange
+    when(distributedStorageAdmin.tableExists(NAMESPACE, TABLE)).thenReturn(true);
+    when(distributedStorageAdmin.getTableMetadata(NAMESPACE, TABLE))
+        .thenReturn(transactionTableMetadata());
+
+    // Act
+    boolean actual = admin.tableExists(NAMESPACE, TABLE);
+
+    // Assert
+    assertThat(actual).isTrue();
+    verify(distributedStorageAdmin).tableExists(NAMESPACE, TABLE);
+    verify(distributedStorageAdmin).getTableMetadata(NAMESPACE, TABLE);
+    verify(distributedStorageAdmin, never()).getNamespaceTableNames(anyString());
+  }
+
+  @Test
+  public void tableExists_ForStorageTable_ShouldReturnFalse() throws ExecutionException {
+    // Arrange
+    TableMetadata storageTableMetadata =
+        TableMetadata.newBuilder()
+            .addPartitionKey("id")
+            .addColumn("id", DataType.TEXT)
+            .addColumn("created_at", DataType.BIGINT)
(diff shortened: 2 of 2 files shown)


brfrn169:
Closing this without merging.

The measured win is real — 39 SQL statements down to 5 per `tableExists()` call — but this override is a stopgap for a problem whose proper fix is in `Admin.tableExists()` itself, and that fix would delete it. `ConsensusCommitAdmin.getTableMetadata()` already returns null for the coordinator namespace and for non-transaction tables, which is exactly the filter `getNamespaceTableNames()` applies, so no override is needed once the default resolves from the table's own metadata (2 statements rather than 5).

Three reasons for not landing it in the meantime:

1. The latency that prompted the investigation has been addressed on the caller side, by caching the existence check there. The remaining gain from this PR is real but modest.
2. This code is explicitly temporary. Shipping a public-API behaviour and then deleting it spends reviewer attention twice.
3. It embeds a non-obvious ordering contract: the storage has to be asked before the metadata is read, because `DynamoAdmin.readMetadata()` throws rather than returning null while the metadata table has yet to be created. The first revision of this PR got that order wrong and failed both DynamoDB jobs. A temporary construct carrying a trap that has already fired once is a poor thing to leave in place, even with the comment and the `tableExists_WhenStorageDoesNotHaveTheTable_ShouldNotReadTheMetadata` test guarding it.

Worth revisiting if either of these happens:

- **Large-schema DDL is reported as slow.** `SchemaOperator.createTables()` calls `tableExists()` once per table, so with the current default, loading N transaction tables into one namespace costs O(N^2) metadata reads — on the order of 10,000 for a 100-table schema. `deleteTables()` and `repairTables()` have the same shape. This is standalone motivation that does not depend on the original caller.
- **The prerequisites for fixing the default turn out to be far off.** Those are (a) `DynamoAdmin.readMetadata()` returning null instead of throwing when the metadata table itself is absent, which would bring it in line with every other storage and is arguably worth doing on its own, and (b) the repair paths no longer depending on `tableExists()` tolerating unparseable metadata — see "Why the default in `Admin` is left alone" in the description above for the reproduction.

The analysis and measurements stay in the description.

PR #3816, opened 2026-08-25
Use CosmosException.getShortMessage() for Cosmos error messages

brfrn169:
## Description

When Cosmos DB returns 429 (`RequestRateTooLarge`), the exception message that ScalarDB builds grows to about 36 KB. This is because `CosmosException.getMessage()` embeds the whole `CosmosDiagnostics` (retry history, replica addresses, response headers) as JSON. The useful part, `innerErrorMessage`, is only about 160 bytes; the remaining 99.5% is diagnostics. Since the Cosmos DB SDK retries a throttled request several times before giving up (9 by default), the diagnostics grow with the number of retries.

Consensus Commit then wraps the message again as it propagates, so what finally surfaces is a single error message of tens of kilobytes. An observed message was 35,955 bytes.

An error message of that size is not usable. It buries the one line that matters, it bloats logs when requests fail at a high rate, and callers that carry error messages across a boundary generally impose size limits far below it, so the message can be truncated or dropped and the error information lost entirely. In that case the cause of the failure — that the provisioned RU was insufficient — is not visible at all.

## Related issues and/or PRs

N/A

## Changes made

**Use `getShortMessage()` instead of `getMessage()` for `CosmosException`**

The Azure SDK provides `CosmosException.getShortMessage()` for exactly this purpose. Its javadoc states that it "returns the error message without any diagnostics — using this method is only useful when also logging the `CosmosException#getDiagnostics()` separately". ScalarDB was using `getMessage()`.

Added `CosmosUtils.buildErrorDetails(CosmosException)` as the single entry point that builds the error details, and applied it to every place where a `CosmosException` is caught:

- `MutateStatementHandler.throwException()` — both the `RETRY_WITH` and the general path
- `BatchHandler.throwException()` — same two paths
- `SelectStatementHandler.handle()` — the `catch (CosmosException e)` path

The helper lives in `CosmosUtils` rather than in a base class because `BatchHandler` does not extend `StatementHandler`.

With this change, the message for the case above becomes about 160 bytes plus the status codes and the template, so the reason for the failure is immediately visible.

**Include the status code and the sub status code**

`getShortMessage()` returns `innerErrorMessage()` as is and does not carry the codes, so `statusCode` and `subStatusCode` are now included explicitly. They cost a few dozen bytes and help identify failures other than 429. The `CoreError` message templates are unchanged: `COSMOS_ERROR_OCCURRED_IN_MUTATION` is also used from the `catch (RuntimeException e)` paths where no `CosmosException` is available, so the details string is composed at the call site and passed as the existing `Details: %s` argument.

**Wrap the `CosmosException` thrown from `ScannerImpl`**

`SelectStatementHandler.executeQuery()` returns `iterableByPage(fetchSize).iterator()`, and the pages are fetched lazily. A `CosmosException` can therefore be thrown from `ScannerImpl.one()` / `all()`, after the handler has already returned the scanner. It was not caught there, so the raw `CosmosException` reached the callers with the whole diagnostics in its message. Since `CosmosException` is unchecked, this compiled fine even though `Scanner` declares `throws ExecutionException`.

`one()` and `all()` now wrap it in an `ExecutionException`, following the same pattern as `storage/jdbc/ScannerImpl`. The existing `COSMOS_ERROR_OCCURRED_IN_SELECTION` code is reused rather than adding a new one, matching what `storage/objectstorage/StreamingRecordIterator` does.

**Not changed**

- The `catch (RuntimeException e)` paths in the three handlers keep using `getMessage()`. `CosmosException` is caught before them, so what reaches these blocks has no short message.
- `CosmosAdmin` needs no change. It has no `getMessage()` call; it passes only a fixed format string and the cause.
- `SingleRecordScanner` needs no change. It holds an already-fetched record and does no lazy fetching.
- No logger was added. The `CosmosException` stays attached as the cause, so logging with a stack trace still surfaces the diagnostics. Adding an explicit `logger.warn` would emit megabytes of logs within seconds when requests fail at a high rate.

## Checklist

**Audit of the other storages**

Since the underlying problem is "an oversized error message", the other storages were checked as well:

| Storage | Verdict | Basis |
|---|---|---|
| Cosmos | Broken | 35,955 bytes observed |
| Cassandra | Safe | `NoHostAvailableException` in driver 3.12.1 caps the message at 3 hosts (`MAX_ERRORS_IN_DEFAULT_MESSAGE = 3`) and excludes stack traces |
| DynamoDB | Needs measurement | `BatchHandler` allows up to 100 items per `TransactWriteItems`, and `TransactionCanceledException` enumerates every cancellation reason. Estimated at 2-3 KB. Much smaller than the Cosmos case, but not measured |
| JDBC | Driver-dependent | No structural bound, but typical driver messages are under 1 KB |
| Object storage | Depends on the underlying SDK | `ObjectStorageWrapperException` passes the message through unchanged |

Two findings are left out of this PR and are worth handling separately:

1. `Put.toString()` / `Update.toString()` include `columns`, and `TextColumn.toString()` prints the whole string value, so places that embed an operation in a message have no size bound. Most of them are shape-dependent `IllegalArgumentException`s that surface during development, but two are data-dependent and can first appear in production: the empty-index-column check in `DynamoOperationChecker` (dumps the whole `put`), and `MULTI_{RECORD,PARTITION,TABLE,NAMESPACE,STORAGE}_MUTATION` in `OperationChecker` (dumps the whole mutation list, up to 100 items). The impact is a degraded error message rather than a broken transaction state, so it is lower priority than this fix.
2. Measuring the actual `TransactionCanceledException` message length for a 100-item batch.

**On adding a generic cap**

Capping the message length in `ScalarDbError.buildMessage()` was considered as defense in depth. It was left out because the design questions are non-trivial: character count vs. UTF-8 byte length, per-argument vs. whole-message cap, and the fact that `buildMessage()` is a default method on a shared interface, so a cap would change behavior for every error enum that implements it. Bundling that with this fix would make the change much harder to review.

**Testing**

All new assertions were verified to actually fail when `getShortMessage()` is reverted to `getMessage()`, so the regression guards are not vacuous.

Note that the `catch (CosmosException e)` block in `SelectStatementHandler` may not fire for query paths at all, since `iterator()` likely does not trigger a fetch. If so, 429s on query paths all surface in `ScannerImpl`, which makes that part of the change the main path for scans rather than a supplement.

## Release notes

Fixed an issue where the error message for a Cosmos DB error was excessively large because it included the Cosmos DB diagnostics, which could cause the error information to be lost.


core/src/main/java/com/scalar/db/storage/cosmos/ScannerImpl.java
jnmt, line 93: Nit: Should these two be in the `finally` block for cleanup?

jnmt (approved):
LGTM, thank you for the fix! I left a question, but it's not a blocker.

core/src/main/java/com/scalar/db/storage/cosmos/ScannerImpl.java
brfrn169: Thanks for the catch! I thought about it, but since a `Scanner` is (implicitly) not reusable once an error occurs, I think we don't really need to release them in that path, so I'd like to leave them as is. Happy to change it if you'd prefer, though!

jnmt: > Thanks for the catch! I thought about it, but since a Scanner is (implicitly) not reusable once an error occurs, I think we don't really need to release them in that path, so I'd like to leave them as is. Happy to change it if you'd prefer, though!

Ahh, got it. Let's keep it as is. Thank you!

Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #3820, opened 2026-08-28
Retry conflict errors on catalog statements in the JDBC admin path

brfrn169:
## Description

Schema operations can fail partway with a conflict error on the JDBC admin path, leaving the schema half-created.

This was reported on Oracle with `scalar.db.jdbc.isolation_level=SERIALIZABLE`: after enough tables have been created, schema loading aborts with `ORA-08177: can't serialize access for this transaction` on an `INSERT` into `scalardb.metadata`, and an orphan physical table is left behind.

**Transaction metadata decoupling is where this surfaces in practice.** On Oracle the feature requires SERIALIZABLE — `ConsensusCommitAdmin` rejects it unless `StorageInfo.isConsistentVirtualTableReadGuaranteed()` holds, and `RdbEngineOracle.getMinimumIsolationLevelForConsistentVirtualTableRead()` returns SERIALIZABLE — so those users do not get to choose the isolation level, and will hit this once `scalardb.metadata` grows past the threshold described below. Users not using decoupling reach the same wall only if they set `scalar.db.jdbc.isolation_level=SERIALIZABLE` themselves, which is why this went unnoticed until now.

The underlying mechanism, however, has nothing to do with decoupling, and the gap it exposes is broader than that report. The data plane already treats conflicts as retriable — `JdbcDatabase` and `JdbcTransaction` convert them into `RetriableExecutionException` / `CrudConflictException` so the application can retry the transaction — but the admin path never consulted `rdbEngine.isConflict()` at all. Every engine implements it, and none of them were consulted here:

| Engine | `isConflict()` | Consulted from the admin path |
|---|---|---|
| Oracle | ORA-08177 / ORA-00060 / ORA-08176 | ❌ |
| MySQL | 1213 deadlock / 1205 lock wait timeout | ❌ |
| PostgreSQL | SQLSTATE 40001 / 40P01 | ❌ |
| SQL Server | 1205 deadlock victim | ❌ |
| Db2 | -911 | ❌ |
| SQLite | 5 / 6 database locked | ❌ |
| Spanner | ABORTED | ❌ |

So a MySQL deadlock or a PostgreSQL serialization failure during a schema operation aborts it just as readily. The Oracle report is simply where this surfaced first.

## Related issues and/or PRs

None.

## Changes made

**`JdbcAdmin`: retry conflict errors on catalog statements.**

The retried region covers the statement execution **and** the commit, because a conflict can surface at either point. The explicit `commit()` runs only for engines that require one — today that means Oracle at SERIALIZABLE — and on an autocommit connection the commit happens inside the statement call, so a conflict raised at commit time falls in the same region either way. PostgreSQL, for instance, raises `40001` during the commit attempt rather than at the statement. Leaving the loop as soon as the commit succeeds guarantees a committed statement is never re-executed, which keeps the SQL warning handling (it runs after the commit and can throw) outside the retried region.

A rollback precedes each retry on connections that commit explicitly. Oracle has already rolled the transaction back by the time the conflict error is returned, so the call is defensive: it costs nothing there and keeps the helper correct for any engine that later requires explicit commit while leaving its transaction alive after a conflict.

Retrying a single statement is safe here because one statement is one transaction on the admin path, whether the connection commits explicitly or uses autocommit. The retry is therefore not gated on the engine.

A retry re-runs the whole action, so the `paramSetter` and `resultMapper` passed to these helpers must be free of side effects. That contract is now documented on the helpers, and the one mapper that violated it is fixed: `JdbcAdminTestUtils.dropAllIndexesForTable` appended to a list captured from the enclosing scope, which a retry would have filled twice.

No backoff is applied. These conflicts resolve on their own, so waiting does not improve the odds, and failing fast keeps admin operations responsive. The retry count is bounded by `MAX_CONFLICT_RETRY_COUNT`.

**Supporting refactor.** The statement helpers in `JdbcAdmin` are static and did not hold an `RdbEngineStrategy`, so they could not call `isConflict()`. They now take one as their second parameter, matching the existing argument order of `JdbcAdmin.internalTableExists()`. This is a mechanical change across 64 call sites with no behavior change, kept in its own commit so the retry diff stays reviewable.

## Checklist

### Root cause

`ORA-08177` here is raised when the **primary key index of `scalardb.metadata` splits a leaf block inside a SERIALIZABLE transaction**.

Under SERIALIZABLE, `RdbEngineOracle.requiresExplicitCommit()` returns true, so each catalog statement commits individually — one column of table metadata is one transaction. As rows accumulate, the index eventually has to split, and a split inside a serializable transaction raises `ORA-08177`.

The threshold follows from how many index entries fit in one block:

```
threshold rows ≈ block size / (index key length + 13)
```

Measured on Oracle Free 23 (8 KB blocks), **single session, no concurrency, no ScalarDB involved**:

| Index key length | Predicted | Actual first ORA-08177 |
|---|---|---|
| ~18 bytes | row 264 | row 277 |
| ~58 bytes | row 115 | row 117 |
| ~98 bytes | row 74 | row 74 |
| ~209 bytes | row 36 | row 37 |

This explains why the reported failure looked deterministic, and why it moved when the schema was split into smaller batches: it tracks the cumulative row count in `scalardb.metadata`, not how many tables are loaded in one run.

The mechanism itself is independent of transaction metadata decoupling: `requiresExplicitCommit` depends only on the isolation level. Decoupling does not meaningfully accelerate it either — it adds roughly one row per primary key column per table, moving the wall by about one table, not halving it. Its relevance is the one described in the Description: on Oracle it forces the isolation level that triggers this.

### Reproduction

No ScalarDB build required.

```bash
docker run -d --name ora23 -p 15210:1521 -e ORACLE_PASSWORD=Oracle1234 gvenzl/oracle-free:23-slim
# wait for "DATABASE IS READY TO USE" in the container logs
```

Create a table with the same shape as `scalardb.metadata`:

```sql
CREATE TABLE md(
  "full_table_name" VARCHAR2(128), "column_name" VARCHAR2(128),
  "data_type" VARCHAR2(20) NOT NULL, "key_type" VARCHAR2(20),
  "clustering_order" VARCHAR2(10), "indexed" NUMBER(1) NOT NULL,
  "ordinal_position" INTEGER NOT NULL,
  CONSTRAINT pk_md PRIMARY KEY ("full_table_name", "column_name"));
```

Then, from a single session with no concurrency, insert and commit one row at a time, varying the key:

```sql
ALTER SESSION SET ISOLATION_LEVEL = SERIALIZABLE;
INSERT INTO md VALUES ('ns.tbl1_data', 'col1', 'INT', NULL, NULL, 0, 1);
COMMIT;
-- repeat
```

`ORA-08177` arrives at the row the formula predicts (row 277 for keys of this length). Re-running the same statement after a `ROLLBACK` succeeds within one or two attempts.

Repeating the run with `ALTER SESSION SET ISOLATION_LEVEL = READ COMMITTED` completes 640 rows with no error, confirming the isolation level is the lever.

### Mitigations that were tried and do not work

All measured, all ineffective — the failure stays at the same row:

| Attempted | Why it does not apply |
|---|---|
| `ROWDEPENDENCIES` on the metadata table | Stores a per-row SCN in the heap; index leaf blocks have no equivalent |
| `INITRANS 3` on the table | Governs the interested transaction list, not leaf splits |
| `INITRANS 3` on the index | Same |
| Full scan between batches to force block cleanout | Not a delayed-cleanout problem |
| Batching a table's metadata rows into one transaction | The threshold tracks rows, not transactions |

### Verification

An integration test reproducing the leaf split was written, run against Oracle Free 23 with a CI-equivalent user, and then removed (see below):

```
block size 8192, index key length 131, threshold ~56 rows, writing 280 rows across 7 tables
Retrying a conflicting statement (attempt 1) ... error code: 8177
Retrying a conflicting statement (attempt 2) ... error code: 8177
BUILD SUCCESSFUL
```

- Predicted threshold 56 rows; the first `ORA-08177` arrived at row 57
- Three conflicts across 280 rows, each resolved within two retries
- Setting the retry limit to zero reproduces the original failure at the same row

### On testing

`JdbcAdminConflictRetryTest` adds 15 unit tests covering the retry region boundaries (commit inside, warning handling outside), the conditional rollback, autocommit connections, retry exhaustion preserving the original exception, non-conflict errors bypassing retry, `rollback()` failure being suppressed onto the original exception, per-statement retry within the array overload, and result mapper / param setter re-execution not duplicating results — including a conflict raised part-way through reading a result set, which is how `ORA-08176` surfaces. `RdbEngineOracleTest` already covers `isConflict()` for the three Oracle error codes.

The Oracle integration test was removed after being verified. Its only unique contribution was confirming that immediate retry resolves a real `ORA-08177` — a fact about Oracle's behavior rather than a guard against regressions in this repository. It also could not assert that a retry actually occurred without adding test-only mutable state to `JdbcAdmin`, so it would have silently stopped exercising anything if the threshold were ever missed. The reproduction above serves the same purpose without that risk.

### Not addressed here

- **Catalog writes remain non-atomic.** If a statement fails for a non-conflict reason partway through writing a table's metadata, partial rows and an orphan `_data` table remain. This is pre-existing — `ConsensusCommitAdmin.createTable` has no compensation, and `withConnection`'s rollback is a no-op once statements have committed individually. This change removes the most likely trigger, not the non-atomicity.
- **Recovering an environment that already hit this** requires dropping the orphan table and deleting its metadata rows manually. The public API cannot do it, because every decoupling branch keys off whether the virtual table is registered in `scalardb.virtual_tables`, and a partially created table is not.
- **Running catalog writes at `READ COMMITTED`** instead also resolves the issue (verified), and removes the cause rather than handling its symptom. It is a more structural change — it would need the isolation check in `JdbcAdmin.getStorageInfo()` reworked — and it would not help the other six engines, so it is not attempted here.

## Release notes

Fixed an issue where schema operations could fail with a conflict error, such as `ORA-08177` on Oracle when using the SERIALIZABLE isolation level. Conflict errors are now retried.


core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -1617,15 +1649,90 @@ static void execute(
...
+   * loop is left immediately, which guarantees that a committed statement is never re-executed.
+   *
+   * <p>Retrying a single statement is safe on the admin path because one statement is one
+   * transaction there: with explicit commit the failed transaction is rolled back below, and with
+   * autocommit the driver has already ended it.
+   *
+   * <p>No backoff is applied. The conflicts seen here resolve on their own (an index leaf block
+   * split has completed by the time the retry runs; a deadlock victim is chosen before the error is
+   * returned), so waiting does not improve the odds, and failing fast keeps admin operations
+   * responsive.
```
feeblefakie: isConflict() also matches errors that mean "I waited for a lock and gave up", not just "the server already rolled someone back" — MySQL 1205 (ER_LOCK_WAIT_TIMEOUT) and SQLite 5 (SQLITE_BUSY). For those the wait already happened inside the driver/server before the error was returned, so an immediate retry can wait the whole timeout again. There's no setQueryTimeout or socket timeout on this path, so 11 attempts means 11 x the lock timeout — with our own SQLite integration-test URL (busy_timeout=50000, ci/tests-config.yaml) that's up to ~550s for one statement.

Should we retry only the "already rolled back" codes here, or bound the loop by elapsed time in addition to the attempt count?

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -58,6 +59,14 @@ public class JdbcAdmin implements DistributedStorageAdmin {
 
   private static final String INDEX_NAME_PREFIX = "index";
 
+  /**
+   * The maximum number of retries for a conflict error on a catalog statement. Conflict errors on
+   * the admin path are transient: one statement is one transaction here, so re-executing the
+   * statement after a rollback is safe. The measured worst case is two consecutive retries, so this
```
feeblefakie: "The measured worst case is two consecutive retries" — was that measured on Oracle? The same bound now also governs MySQL and SQLite, where the failure mode is a lock timeout rather than an index leaf split. Worth scoping the sentence to the engine it was measured on?

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
```diff
@@ -1617,15 +1649,90 @@ static void execute(
...
+   *
+   * <p>The retried region covers the statement execution <em>and</em> the commit, because a
+   * conflict can surface at either point. The explicit {@code commit()} below runs only for engines
+   * that require one, which today means Oracle at SERIALIZABLE; on an autocommit connection the
+   * commit happens inside the statement call, so a conflict raised at commit time (PostgreSQL
+   * raises {@code 40001} there) falls in the same region. Conversely, once the commit succeeds the
+   * loop is left immediately, which guarantees that a committed statement is never re-executed.
+   *
+   * <p>Retrying a single statement is safe on the admin path because one statement is one
+   * transaction there: with explicit commit the failed transaction is rolled back below, and with
```
feeblefakie: These helpers also run on the user data path, not just the admin path: JdbcDatabase builds a JdbcAdmin over the table-metadata pool and passes it to TableMetadataManager, whose cache loader calls getTableMetadata on a miss (60s expireAfterWrite, so the reload blocks the calling thread). The safety argument still holds — one statement is one transaction there too — but should the wording be widened?

feeblefakie (commented):
Overall looking good. Left some comments. PTAL!

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
brfrn169: Neither, in the end -- but the question changed what I think the comment needed to say, so thank you for it.

**Narrowing to the "already rolled back" codes would be actively wrong for MySQL.** With `innodb_deadlock_detect` disabled -- a real configuration for high-concurrency deployments -- InnoDB does not report 1213 at all and relies on `innodb_lock_wait_timeout` to break the cycle, so a deadlock surfaces as **1205**. Excluding 1205 would stop retrying the exact case this helper exists for on those deployments. Db2 has the same shape from the other direction: -911 covers deadlock *or* timeout under a single code and cannot be split at all.

**On bounding by elapsed time:** for the errors this retry is designed for, the amplification does not arise, because they consume no wait before they are reported -- a deadlock victim is rolled back as soon as the server detects the cycle, and PostgreSQL's `deadlock_timeout` adds ~1s, not 50s. The repeated waits only add up under *sustained* contention on the catalog tables, and that cannot originate from this code: `initDataSourceForAdmin` and `initDataSourceForTableMetadata` both pass `transactional=false`, and `withConnection` flips autocommit off only when `requiresExplicitCommit`, where every statement commits individually. No lock is held across statements on either path. Sustained contention therefore means a leaked or external transaction -- a bug to surface rather than a condition to absorb, which is what the per-attempt warning does. A deadline would truncate that signal, and would need a constant with no principled value behind it.

You are right that the Javadoc was resting on all of this without saying it. Recorded the assumptions in a4b73f10e: the "conflicts resolve on their own" claim is now scoped to the errors it is actually true for, with separate paragraphs on why the lock-wait codes are deliberately not excluded and why there is no elapsed-time bound.

Behavior unchanged.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
brfrn169: Correct -- it was measured on Oracle only, while reproducing the `ORA-08177` leaf split, and the sentence presented it as a general figure.

Rewritten in 4050f3d75 to attribute the number to that reproduction, and to state the bound's justification for the other engines separately rather than folding them into the same measurement: a deadlock or a serialization failure is settled by the server before the error is returned, so a retry either succeeds or has met a genuinely contended table.

core/src/main/java/com/scalar/db/storage/jdbc/JdbcAdmin.java
brfrn169: Confirmed, and widened in 329c9b7d7. `JdbcDatabase` builds a `JdbcAdmin` over `initDataSourceForTableMetadata`'s pool and hands it to `TableMetadataManager` (`JdbcDatabase.java:82-85`), whose Caffeine loader calls `getTableMetadata()` on a miss and blocks the caller during the reload.

The safety argument holds unchanged, so I widened the wording rather than qualifying it: both pools are created with `transactional=false` and apply `requiresExplicitCommit` identically through `withConnection`, so one statement is one transaction on either path.

I also added a note that the data path is the less exposed of the two -- the statement it runs is a non-locking read of the catalog, so there are no row locks to contend over, and the lock-wait errors discussed further down are effectively out of reach there.

feeblefakie (approved):
LGTM! Thank you!

Torch3333 (approved):
LGTM, thank you!

yu2scalar:
It worked fine as you described:
```
yu2@yu2-Virtual-Machine:~/cluster/schema-loader$ java -jar sl-after.jar -c scalardb.properties.core.oracle -f oracle_test.json
[main] INFO com.scalar.db.schemaloader.command.SchemaLoaderCommand - Config path: scalardb.properties.core.oracle
[main] INFO com.scalar.db.schemaloader.command.SchemaLoaderCommand - Schema path: oracle_test.json
[main] INFO com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Starting...
[main] WARN com.zaxxer.hikari.util.DriverDataSource - Registered driver with driverClassName=oracle.jdbc.driver.OracleDriver was not found, trying direct instantiation.
[main] INFO com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Start completed.
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table EMPLOYEES in the namespace HR succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table DEPARTMENTS in the namespace HR succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table JOBS in the namespace HR succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table DEPARTMENTS in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table ORDER_ITEMS in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table PRODUCT_DESCRIPTIONS in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table PRODUCT_ASSETS in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table CONTRACT_SCHEDULE in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table CHANGE_AUDIT in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table CUSTOMER_ATTRIBUTES in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table ITEM_CATEGORY in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table customer_feedback in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table PRODUCT_CONFIG in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table PRODUCT_TAGS in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table IDENTIFIER_BOUNDARY_LIMIT_TEST in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table NLS_TEST_TABLE in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table PARTITIONED_ORDERS in the namespace SALES succeeded
[main] WARN com.scalar.db.storage.jdbc.JdbcAdmin - Retrying a conflicting statement (attempt 1). SQL: INSERT INTO "scalardb"."metadata" VALUES ('SALES.WIDE_TABLE_100_tx_metadata','before_DATA_COL_088','TEXT',NULL,NULL,0,62), error code: 8177
[main] WARN com.scalar.db.storage.jdbc.JdbcAdmin - Retrying a conflicting statement (attempt 2). SQL: INSERT INTO "scalardb"."metadata" VALUES ('SALES.WIDE_TABLE_100_tx_metadata','before_DATA_COL_088','TEXT',NULL,NULL,0,62), error code: 8177
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table WIDE_TABLE_100 in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table COMPRESSED_TABLE in the namespace SALES succeeded
[main] INFO com.scalar.db.schemaloader.SchemaOperator - Creating the table NOLOGGING_TABLE in the namespace SALES succeeded
[main] INFO com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Shutdown initiated...
[main] INFO com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Shutdown completed.
```

However it may be a little difficult to explain it to a customer when he/she asks for the root cause of the WARN.

PR #3842, opened 2026-09-02
Collation 1/4: introduce collation setting

Torch3333:
## Description

ScalarDB compares text on the JVM in several places, and those comparisons must agree with the text order of the backend that stores the data. A backend that orders text only by UTF-8 bytes or code points can never agree with an ICU collation, so such a configuration must fail at startup rather than produce results that differ from the storage.

This PR adds the `scalar.db.collation` setting, the `CollationComparator` that implements the two collations it names, and the startup checks that reject ICU on every storage that cannot honor it. `CollationComparator` has no production caller, so the comparator, its locale and rule validation, and the storage gates can be reviewed on their own. One comparison result does change: `TextColumn.compareTo` moves from UTF-16 code unit order to Unicode code point order, which shifts BINARY-mode ordering for supplementary characters.

## Related issues and/or PRs

This PR is the first (bottom) PR of a stack of four PRs based on `feature/collation` that together replace #3791.

## Changes made

- Added the `scalar.db.collation` property to `DatabaseConfig` (values `BINARY`, the default, or `ICU`, case-insensitive) together with `scalar.db.collation.icu.locale` and `scalar.db.collation.icu.rules`, exposed through getters.
- Added the `Collation` enum in `com.scalar.db.config` and `CollationComparator` in `com.scalar.db.common`. BINARY orders text by unsigned UTF-8 byte sequence and compares equality with `String.equals`. ICU builds a collator from the configured BCP 47 locale (root collation when unset), optionally extended by a custom tailoring-rule string, and exposes text, column, and key comparators, collation-aware text equality, and canonical collation-key bytes.
- ICU comparison and collation-key generation both run on a per-thread collator cloned from a frozen prototype, because a frozen ICU collator guards its single collation buffer with a lock that every call takes.
- Changed `TextColumn.compareTo` to order text by Unicode code point rather than UTF-16 code unit, so BINARY-mode ordering matches the byte order of the backends that store the data, and made `TextColumn.compareByCodePoint` public so `CollationComparator` can read it.
- `CollationComparator` fails construction with a coded error for a malformed rule string, an ill-formed or unrecognized locale, a `-u-co-` collation type the locale has no data for, and a locale that carries collation settings while a rule string is also configured. It logs the resolved locale at INFO and warns once per JVM when the ICU4J version loaded at runtime differs from the one the build declares.
- Cassandra, Cosmos DB, DynamoDB, and object storage throw `IllegalArgumentException` at construction when the collation is ICU.
- Added `RdbEngineStrategy.throwIfCollationNotSupported`, which rejects ICU by default so that an engine nobody has classified fails at startup. MySQL, PostgreSQL, Oracle, SQL Server, and Db2 override it to accept, so MariaDB and TiDB inherit the MySQL verdict and Yugabyte inherits the PostgreSQL one. SQLite and Cloud Spanner reject. `JdbcDatabase` and `JdbcTransactionManager` call it at construction.
- `MultiStorageConfig` rejects any per-storage `collation` or `collation.*` property.
- Added `CoreError` codes 0304 to 0311 for the collation errors.
- Added `com.ibm.icu:icu4j` 77.1 as an `implementation` dependency of `core`, pinned through `icu4jVersion` and excluded from Dependabot, allowed its Unicode-3.0 license, and added a `processResources` step that writes that version into `scalardb-collation.properties`.
- Added unit tests for the configuration, the comparator, the storage and JDBC engine gates, and the multi-storage rejection, plus tests asserting that `Key` and `TextColumn` equality stays byte-exact. `RdbEngineCollationTest` pins the ICU verdict of every RDB engine, inherited verdicts included. `CollationComparatorTest` pins the spellings the ICU root collation treats as equal, covering default-ignorable characters and NFC against NFD.

## Checklist

- `CoreError` 0308 (`COLLATION_ICU_LIKE_CONDITION_NOT_SUPPORTED`) is defined in this PR but has no caller in it.
- ICU4J is an `implementation` dependency of `core`, so it becomes a runtime dependency for every ScalarDB user, including those that keep the default BINARY collation.
- The ICU4J version is pinned and excluded from Dependabot, and the comment next to `icu4jVersion` in `build.gradle` gives the reason. `processResources` expands that version into `scalardb-collation.properties`, which `CollationComparator` reads to emit the version-mismatch warning.
- Under ICU, text equality is many-to-one. The root collation treats default-ignorable characters and canonically equivalent forms as equal, so distinct stored strings share one identity. This has no effect in this PR because nothing consumes the comparator yet.

## Release notes

- `TextColumn.compareTo` now orders text by Unicode code point instead of UTF-16 code unit. For text containing supplementary characters, ordering relative to the range U+E000 to U+FFFF changes, which affects object storage scan order and clustering key boundaries.

### DIFF
--- a/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
+++ b/core/src/main/java/com/scalar/db/config/DatabaseConfig.java
@@ -9,6 +9,7 @@ import static com.scalar.db.config.ConfigUtils.getStringArray;
 
 import com.google.common.collect.ImmutableList;
 import com.scalar.db.common.CoreError;
+import com.scalar.db.io.Collation;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.File;
 import java.io.FileInputStream;
@@ -16,6 +17,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.Properties;
 import javax.annotation.Nullable;
@@ -42,6 +44,9 @@ public class DatabaseConfig {
   private boolean crossPartitionScanOrderingEnabled;
   private String systemNamespaceName;
   private int scanFetchSize;
+  private Collation collation;
+  @Nullable private String collationIcuLocale;
+  @Nullable private String collationIcuRules;
 
   public static final String PREFIX = "scalar.db.";
   public static final String CONTACT_POINTS = PREFIX + "contact_points";
@@ -67,6 +72,9 @@ public class DatabaseConfig {
   public static final String CROSS_PARTITION_SCAN_ORDERING = SCAN_PREFIX + "ordering.enabled";
   public static final String SYSTEM_NAMESPACE_NAME = PREFIX + "system_namespace_name";
   public static final String SCAN_FETCH_SIZE = PREFIX + "scan_fetch_size";
+  public static final String COLLATION = PREFIX + "collation";
+  public static final String COLLATION_ICU_LOCALE = PREFIX + "collation.icu.locale";
+  public static final String COLLATION_ICU_RULES = PREFIX + "collation.icu.rules";
 
   public static final int DEFAULT_METADATA_CACHE_EXPIRATION_TIME_SECS = 60;
   public static final int DEFAULT_ACTIVE_TRANSACTION_MANAGEMENT_MAX_ACTIVE_TRANSACTIONS = 10000;
@@ -138,6 +146,11 @@ public class DatabaseConfig {
     systemNamespaceName = getSystemNamespaceName(getProperties());
 
     scanFetchSize = getInt(getProperties(), SCAN_FETCH_SIZE, DEFAULT_SCAN_FETCH_SIZE);
+
+    String collationValue = getString(getProperties(), COLLATION, Collation.BINARY.name());
+    collation = Collation.valueOf(collationValue.toUpperCase(Locale.ROOT));
+    collationIcuLocale = getString(getProperties(), COLLATION_ICU_LOCALE, null);
+    collationIcuRules = getString(getProperties(), COLLATION_ICU_RULES, null);
   }
 
   public List<String> getContactPoints() {
@@ -206,6 +219,18 @@ public class DatabaseConfig {
 
   public int getScanFetchSize() {
     return scanFetchSize;
+  }
+
+  public Collation getCollation() {
+    return collation;
+  }
+
+  public Optional<String> getCollationIcuLocale() {
+    return Optional.ofNullable(collationIcuLocale);
+  }
+
+  public Optional<String> getCollationIcuRules() {
+    return Optional.ofNullable(collationIcuRules);
   }
 
   public static String getTransactionManager(Properties properties) {
--- a/core/src/main/java/com/scalar/db/io/Collation.java
+++ b/core/src/main/java/com/scalar/db/io/Collation.java
@@ -1,0 +1,17 @@
+package com.scalar.db.io;
+
+/**
+ * Selects ScalarDB's in-memory text comparison mode: it governs ordering, equality, and — under
+ * {@link #ICU}, within the Consensus Commit transaction layer — key identity. Comparisons delegated
+ * to the backend are unaffected, and stored bytes are never rewritten.
+ */
+public enum Collation {
+  /** Orders text by unsigned UTF-8 byte sequence; equality is byte-exact. */
+  BINARY,
+
+  /**
+   * Orders text according to the Unicode Collation Algorithm, configured by locale and optional
+   * tailoring rules; equality follows the collation.
+   */
+  ICU
+}
--- a/core/src/main/java/com/scalar/db/io/CollationComparator.java
+++ b/core/src/main/java/com/scalar/db/io/CollationComparator.java
@@ -1,0 +1,394 @@
+package com.scalar.db.io;
+
+import com.google.common.collect.Ordering;
+import com.ibm.icu.text.Collator;
+import com.ibm.icu.text.RuleBasedCollator;
+import com.ibm.icu.util.IllformedLocaleException;
+import com.ibm.icu.util.ULocale;
+import com.ibm.icu.util.VersionInfo;
+import com.scalar.db.common.CoreError;
+import com.scalar.db.config.DatabaseConfig;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.concurrent.atomic.AtomicBoolean;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+import javax.annotation.concurrent.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A thread-safe, immutable comparator that orders text according to the configured {@link
+ * Collation}, together with the text equality and canonical-form operations that follow from that
+ * order.
+ */
+@Immutable
+@ThreadSafe
+public final class CollationComparator {
+  private static final Logger logger = LoggerFactory.getLogger(CollationComparator.class);
+
+  private static final String VERIFIED_ICU_VERSION_RESOURCE = "scalardb-collation.properties";
+
+  private static final AtomicBoolean ICU_VERSION_CHECKED = new AtomicBoolean();
+
+  private final Comparator<String> textComparator;
+  private final Comparator<Column<?>> columnComparator;
+  private final Comparator<Key> keyComparator;
+  private final boolean byteExactEquality;
+
+  /**
+   * Per-thread canonical-form producer, present only for {@link Collation#ICU}. Collation-key
+   * generation on a shared frozen collator serializes on a JVM-wide lock, so each thread gets its
+   * own thawed clone; a thawed collator is safe for single-threaded use and produces the same
+   * collation keys as the frozen original it was cloned from.
+   */
+  @Nullable private final ThreadLocal<Collator> canonicalizer;
+
+  private CollationComparator(
+      Comparator<String> textComparator,
+      boolean byteExactEquality,
+      @Nullable ThreadLocal<Collator> canonicalizer) {
+    this.textComparator = textComparator;
+    Comparator<String> nullsFirstText = Comparator.nullsFirst(textComparator);
+    this.columnComparator = buildColumnComparator(nullsFirstText);
+    this.keyComparator = buildKeyComparator(this.columnComparator);
+    this.byteExactEquality = byteExactEquality;
+    this.canonicalizer = canonicalizer;
+  }
+
+  /**
+   * Creates a {@code CollationComparator} from the given configuration.
+   *
+   * @param config the database configuration
+   * @return the comparator for the configured collation ({@link Collation#BINARY} when {@code
+   *     scalar.db.collation} is unset)
+   * @throws IllegalArgumentException if an ICU custom tailoring-rule string is malformed or the
+   *     configured ICU locale is not recognized
+   */
+  public static CollationComparator from(DatabaseConfig config) {
+    Collation collation = config.getCollation();
+    switch (collation) {
+      case BINARY:
+        return new CollationComparator(TextColumn::compareByCodePoint, true, null);
+      case ICU:
+        {
+          warnOnIcuVersionMismatch();
+          Collator frozen = buildFrozenIcuCollator(config);
+          return new CollationComparator(
+              frozen::compare, false, ThreadLocal.withInitial(frozen::cloneAsThawed));
+        }
+      default:
+        throw new AssertionError("Unknown collation: " + collation);
+    }
+  }
+
+  /**
+   * Warns once per JVM when the ICU4J library loaded at runtime differs from the version this build
+   * was verified against. ICU4J is a plain transitive dependency, so an embedding application's
+   * dependency resolution can substitute another version, and ICU collation behavior (root sort
+   * order, CLDR data) shifts between versions — ordering and equality results would silently differ
+   * from the verified behavior. The check never fails collation setup: it cannot tell a harmless
+   * substitution from a harmful one, only make it diagnosable.
+   */
+  private static void warnOnIcuVersionMismatch() {
+    if (!ICU_VERSION_CHECKED.compareAndSet(false, true)) {
+      return;
+    }
+    String verifiedVersion = null;
+    try (InputStream stream =
+        CollationComparator.class
+            .getClassLoader()
+            .getResourceAsStream(VERIFIED_ICU_VERSION_RESOURCE)) {
+      if (stream != null) {
+        Properties properties = new Properties();
+        properties.load(stream);
+        verifiedVersion = properties.getProperty("icu4j.version");
+      }
+    } catch (IOException e) {
+      logger.debug("Failed to read {}", VERIFIED_ICU_VERSION_RESOURCE, e);
+    }
+    if (verifiedVersion == null) {
+      logger.debug(
+          "The verified ICU4J version is unavailable from {}; skipping the ICU4J version check",
+          VERIFIED_ICU_VERSION_RESOURCE);
+      return;
+    }
+    try {
+      if (loadedIcuVersionDiffersFrom(verifiedVersion)) {
+        logger.warn(
+            "The ICU4J library loaded at runtime (version {}) differs from the version this "
+                + "ScalarDB build was verified against ({}). ICU collation behavior shifts "
+                + "between ICU4J versions, so ICU-mode text ordering and equality may differ "
+                + "from the verified behavior. Align the application's icu4j dependency with "
+                + "ScalarDB's",
+            VersionInfo.ICU_VERSION,
+            verifiedVersion);
+      }
+    } catch (IllegalArgumentException e) {
+      logger.debug("Failed to compare the ICU4J versions", e);
+    }
+  }
+
+  /**
+   * Returns whether the ICU4J library on the classpath differs from the given version. Package-
+   * private and pure so it can be unit-tested without manipulating the classpath.
+   */
+  static boolean loadedIcuVersionDiffersFrom(String version) {
+    return !VersionInfo.getInstance(version).equals(VersionInfo.ICU_VERSION);
+  }
+
+  private static Collator buildFrozenIcuCollator(DatabaseConfig config) {
+    Collator collator = buildIcuCollator(config);
+    logResolvedIcuCollator(config, collator);
+
+    // An unfrozen ICU Collator is mutable and not thread-safe; a frozen one is safe for
+    // concurrent compare.
+    return collator.freeze();
+  }
+
+  private static Collator buildIcuCollator(DatabaseConfig config) {
+    Optional<String> localeName = config.getCollationIcuLocale();
+    Collator base =
+        localeName.isPresent()
+            ? buildValidatedLocaleCollator(config, localeName.get())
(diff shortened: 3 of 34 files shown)


core/src/main/java/com/scalar/db/io/CollationComparator.java
```diff
@@ -0,0 +1,394 @@
...
+    Collation collation = config.getCollation();
+    switch (collation) {
+      case BINARY:
+        return new CollationComparator(TextColumn::compareByCodePoint, true, null);
+      case ICU:
+        {
+          warnOnIcuVersionMismatch();
+          Collator frozen = buildFrozenIcuCollator(config);
+          return new CollationComparator(
+              frozen::compare, false, ThreadLocal.withInitial(frozen::cloneAsThawed));
```
feeblefakie, line 85: `canonicalTextFormOf` gets a per-thread thawed clone because collation-key generation locks `frozenLock`, but `compare` stays on the shared frozen collator here. Doesn't `doCompare` take that same lock whenever the `CollationFastLatin` path bails out, i.e. for any text above U+017F?

A quick benchmark on this branch, Japanese values of 20 chars: 16.9 Mops/s single-threaded, 12.9 at 8 threads (throughput goes down as threads are added). Routing `compare` through the `ThreadLocal` that already exists in ICU mode gave 81.2 at 8 threads, costing roughly 10% on ASCII. Is there a reason to keep `compare` on the frozen instance?

Separately, the comment at line 47 describes `frozenLock` as a JVM-wide lock. Isn't it a per-collator-instance `Lock`? As written it suggests a second collator wouldn't help.

core/src/main/java/com/scalar/db/io/Collation.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public enum Collation {
+  /** Orders text by unsigned UTF-8 byte sequence; equality is byte-exact. */
+  BINARY,
+
+  /**
+   * Orders text according to the Unicode Collation Algorithm, configured by locale and optional
+   * tailoring rules; equality follows the collation.
+   */
+  ICU
```
feeblefakie, line 16: Should this say that ICU equality is many-to-one, rather than just "equality follows the collation"?

At `ULocale.ROOT` and default (tertiary) strength, with no locale and no rules configured, `compare` returns 0 for all of these against `admin`: ZWSP, ZWJ, ZWNJ, soft hyphen, ZWNBSP/BOM, RLO and VS-1 inserted into the string. Also `café` NFC vs NFD. Case and hyphen correctly stay non-zero, so this is the default configuration rather than a misconfigured collator.

Since PR 3 feeds `canonicalTextFormOf` into the snapshot key while the backend's primary key index stays byte-exact, doesn't that mean two distinct stored rows can collapse to one key? And that a `PutIf` on a TEXT column becomes satisfiable by a different spelling, or a not-exists check defeatable by a character you can't see? The ignorables compose freely, so the set of colliding strings is unbounded.

Would it be worth tests pinning the ignorable and NFC/NFD cases, so this is a chosen behaviour with a regression guard? And is it worth rejecting the configurations that widen it most (primary or secondary strength, `ka-shifted`) through the same path as `COLLATION_ICU_LOCALE_SETTING_WITH_RULES_NOT_SUPPORTED`?

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
```diff
@@ -74,6 +75,13 @@ default void throwIfInvalidNamespaceName(String namespaceName) {}
 
   default void throwIfInvalidTableName(String tableName) {}
 
+  /**
+   * Throws if the given collation configuration cannot match this engine's text order. The default
+   * accepts any collation; engines whose text order is fixed to binary UTF-8 byte order (no
+   * UCA-based collation exists for {@link Collation#ICU} to approximate) reject {@code ICU}.
+   */
+  default void throwIfCollationNotSupported(Collation collation) {}
```
feeblefakie, line 83: Two questions about accepting by default.

MySQL, MariaDB, TiDB, Oracle, SQL Server, Db2 and Yugabyte all take ICU with no check, but nothing reads the server, database or column collation, and the usual defaults don't match ICU root: MySQL 8 is `utf8mb4_0900_ai_ci` (accent and case insensitive), PostgreSQL is typically glibc, Oracle defaults to `NLS_SORT=BINARY`. `parseBcp47LanguageTag`'s javadoc argues that no ICU collation reproduces a libc one, which is why `en_US.UTF-8` is rejected. Doesn't that argument apply to a glibc-collated PostgreSQL database just as much as to the locale string?

Second, is fail-open the right default for this predicate specifically? `RdbEngineDb2` was added recently and silently inherited "ICU is fine". Db2 LUW fixes collation at database creation time and defaults to IDENTITY with no per-statement COLLATE, so is it really in a different category from Spanner? Same question for Yugabyte and TiDB, which inherit acceptance from their parent classes rather than by a decision.

I realise every other `throwIf*` here is a `default` no-op, so opt-out is the established direction. The reason to consider inverting just this one is that a wrong default produces silent disagreement with the storage instead of an unsupported-operation error. Would rejecting by default, with each engine opting in, be worth going against the pattern?

core/src/main/java/com/scalar/db/config/DatabaseConfig.java
```diff
@@ -9,13 +9,15 @@
 
 import com.google.common.collect.ImmutableList;
 import com.scalar.db.common.CoreError;
+import com.scalar.db.io.Collation;
```
feeblefakie, line 12: This makes `config` depend on `io`, while `CollationComparator` depends on `config`, so the two packages now import each other where they didn't before. Was that intended?

Both classes are new and public, so moving them is free right now and a breaking change later. Would `com.scalar.db.config.Collation` plus `com.scalar.db.common.CollationComparator` work? `common` already depends on both packages, and it's where the later consumers sit (`OperationChecker`, `FilterableScanner`, the consensus commit classes, `ClusteringKeyComparator`).

`Isolation` is the closest existing config-valued enum and it isn't in `io` either, which is what made me ask.

core/src/main/java/com/scalar/db/io/TextColumn.java
```diff
@@ -62,10 +65,32 @@ public int compareTo(Column<String> o) {
...
   }
 
+  /**
+   * Orders two strings by Unicode code point, which for well-formed text is the unsigned UTF-8 byte
+   * order that byte-order backends use. {@link String#compareTo} orders by UTF-16 code unit and so
+   * places supplementary characters before U+E000..U+FFFF; comparing the UTF-8 encodings instead
+   * would conflate distinct strings whose unpaired surrogates all encode to the replacement byte.
+   * This order returns 0 only for {@link String#equals equal} strings, ill-formed ones included.
+   */
+  static int compareByCodePoint(String left, String right) {
```
feeblefakie, line 79: The description says no comparison results change in this PR. Doesn't this change `TextColumn.compareTo` for supplementary characters relative to U+E000 to U+FFFF, in BINARY mode, i.e. the default? It reaches `Key.compareTo` and the object storage clustering order through `Ordering.natural()`, so object storage scan order and clustering key boundaries shift for those characters.

To be clear, the change itself looks right to me: code point order is what the byte-ordered backends use, and the loop is correct (`i` and `j` advance together, so the length tail is a proper prefix tiebreak). I only wondered whether it wants a release note instead of being described as no change.

Also, is `TextValue.compareTo` staying on UTF-16 order a deliberate call because the class is deprecated for 4.0.0?

core/src/main/java/com/scalar/db/common/CollationComparator.java
```diff
@@ -0,0 +1,394 @@
...
+      return;
+    }
+
+    throw new IllegalArgumentException(
+        CoreError.COLLATION_ICU_LOCALE_SETTING_WITH_RULES_NOT_SUPPORTED.buildMessage(
+            localeName, String.join(", ", settingKeywords)));
+  }
+
+  /**
+   * Returns the locale's collation-setting keywords. ICU maps every one to a {@code col}-prefixed
```
brfrn169: This sentence is not accurate against the pinned icu4j 77.1: `-u-vt-` maps to the legacy keyword `variabletop`, which is neither `col`-prefixed nor `kv`, so it is a third case the claim does not cover.

Since `build.gradle` pins icu4j precisely because ICU behavior can shift between releases, spelling out the keyword mapping here seems likely to go stale on the next bump. Stating what the predicate is for, without enumerating the mapping, would probably age better.

core/src/main/java/com/scalar/db/common/CollationComparator.java
```diff
@@ -0,0 +1,394 @@
...
+  /**
+   * Rejects a locale ICU has no collation data for. ICU silently falls back to the root collation
+   * for such a locale, which would order text differently from the intended locale with no error,
+   * so a misconfiguration fails at startup instead of producing wrong ordering. An empty language
+   * means an explicitly root-rooted tag (und), which asks for the root collation itself, the same
+   * collation used when no locale is configured at all.
+   */
+  private static void rejectLocaleWithoutCollationData(String localeName, ULocale locale) {
+    String language = locale.getLanguage();
+    if (!language.isEmpty()
```
brfrn169, line 209: `und-Hant` and `und-Cyrl` also have an empty `getLanguage()`, so they take this early exit as well, not only bare `und`. Against icu4j 77.1 both resolve to a collator whose `getRules()` is identical to `ULocale.ROOT`'s, so the script subtag is dropped with no error, which is the outcome the Javadoc above says this method exists to prevent. Is that intended? The comment only mentions `und`, so I wasn't sure these were considered.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
brfrn169: nit: this Javadoc has no `@param collation` or `@throws`, unlike the other documented `throwIf*` methods in this interface (e.g. `throwIfConjunctionsColumnNotSupported`), and `javadoc` reports `no @param for collation` here. Unrelated to the discussion about the default behaviour on this line.

brfrn169 (commented):
Left several comments. PTAL!

core/src/main/java/com/scalar/db/io/CollationComparator.java
```diff
@@ -0,0 +1,438 @@
...
+@ThreadSafe
+public final class CollationComparator {
+  private static final Logger logger = LoggerFactory.getLogger(CollationComparator.class);
+
+  private static final String VERIFIED_ICU_VERSION_RESOURCE = "scalardb-collation.properties";
+
+  private static final AtomicBoolean ICU_VERSION_CHECKED = new AtomicBoolean();
+
+  private static final Set<String> ISO_LANGUAGES =
+      Collections.unmodifiableSet(new HashSet<>(Arrays.asList(ULocale.getISOLanguages())));
```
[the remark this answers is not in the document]
Torch3333: This does not apply to the current code. The guard is `rejectLocaleWithoutCollationData`, which checks the parsed language against `Collator.getAvailableULocales()` rather than `ULocale.getISOLanguages()`, which is what this comment asks for. I verified that `chr` passes it on icu4j 77.1, and `lij` is covered by `from_WhenIcuWithLocaleOutsideTwoLetterIsoLanguages_ShouldBuildComparator`. A language ICU has no collation data for, such as `zz`, is still rejected.

core/src/main/java/com/scalar/db/io/CollationComparator.java
Torch3333: Confirmed and fixed in 52ad43bf0.

I traced it in the icu4j 77.1 source. `doCompare` calls `getCollationBuffer()` on every fast-Latin bail-out, and `getCollationBuffer()` takes `frozenLock` when the collator is frozen. So any comparison whose first differing character is above U+017F serialized against every other comparison, which is consistent with the throughput you measured.

`compare` now runs on the per-thread thawed clone that collation-key generation already used. The frozen collator is kept only as the clone prototype, because cloning an unfrozen collator concurrently is not safe.

You are right about the comment too. `frozenLock` is a per-instance `ReentrantLock` (field at line 1768, and `cloneAsThawed` sets it to null on the clone), not a JVM-wide lock. The javadoc now says the lock belongs to the collator and guards its single collation buffer.

core/src/main/java/com/scalar/db/config/Collation.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public enum Collation {
+  /** Orders text by unsigned UTF-8 byte sequence; equality is byte-exact. */
+  BINARY,
+
+  /**
+   * Orders text according to the Unicode Collation Algorithm, configured by locale and optional
+   * tailoring rules; equality follows the collation.
+   */
+  ICU
```
Torch3333: Confirmed and pinned in 76e1305c0.

The `Collation.ICU` javadoc now states that equality is many-to-one, and `CollationComparatorTest` pins what that covers. It asserts that `textEquals("admin", ...)` is true for zero-width space, zero-width joiner, zero-width non-joiner, soft hyphen, byte order mark, right-to-left override, variation selector 1, and combinations of those, plus `café` NFC against NFD. Case and hyphen differences are asserted non-equal in the same class so the behaviour reads as the configuration rather than a broken collator. Every case you listed reproduced exactly as you described.

On rejecting the configurations that widen it most, I would rather not, and I want to check that with you. Root tertiary already collapses every ignorable listed above, so gating `ks-level1`, `ks-level2` and `ka-shifted` narrows the colliding set without closing it, while removing case-insensitive collation, which is one of the main reasons to choose ICU. The identity consequence you describe is real. I think it belongs in PR 3, documented where `canonicalTextFormOf` gets its first consumer, rather than as a config rejection here. If you still want the gate, it is a small addition to the same validation path.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
Torch3333: Inverted in 61b4aca8c, as you suggested. The default now rejects ICU, and each engine opts in by naming the mechanism that makes a UCA-based collation reachable. MySQL, PostgreSQL, Oracle, SQL Server and Db2 accept. MariaDB and TiDB inherit the MySQL verdict, Yugabyte inherits the PostgreSQL one, and SQLite and Cloud Spanner keep rejecting. `RdbEngineCollationTest` pins every verdict, inherited ones included, and the default.

On Db2 specifically, it turns out not to be in Spanner's category. Db2 LUW has locale-sensitive UCA-based collations, selected at database creation with `CREATE DATABASE ... COLLATE USING CLDR181_LCS`, documented at https://www.ibm.com/docs/en/db2/12.1.x?topic=database-locale-sensitive-uca-based-collation . Creation time only, so there is no per-statement COLLATE as you said, but a UCA collation is reachable.

Two things I should be straight about. First, no engine's verdict changed, so runtime behaviour is identical to before. What the inversion buys is that a future engine fails closed instead of silently inheriting acceptance, which was your point about `RdbEngineDb2`. Second, your deeper point stands. The predicate asserts that an ICU-comparable collation is achievable on the engine, not that the server's actual collation matches ICU. Nothing reads `pg_collation` or `SHOW COLLATION`, so a glibc-collated PostgreSQL database still passes. I left that gap open rather than pretending the gate closes it.

The rejection reuses error 0306. One wrinkle there: 0306 says the storage supports only binary UTF-8 byte order, which is accurate for SQLite and Spanner but is an unverified claim for an engine nobody has classified. It is unreachable today because all ten engines have an explicit verdict.

core/src/main/java/com/scalar/db/config/DatabaseConfig.java
Torch3333: Moved in 0bb3349b6, as you suggested. `Collation` is now `com.scalar.db.config.Collation` and `CollationComparator` is `com.scalar.db.common.CollationComparator`. `config` already depended on `common` and `common` already depended on both packages, so neither move adds an edge, and `io` no longer imports `config` at all.

One consequence I would like your opinion on. `CollationComparator` reads `TextColumn.compareByCodePoint` for the BINARY text order, and that method was package-private in `io`, so it is public now. The alternative is moving only `Collation` to `config`, which breaks the cycle on its own and leaves the method package-private, at the cost of `io` depending on `DatabaseConfig` and the comparator sitting away from the consumers you listed. I took the public method as the smaller cost, but say the word and I will do it the other way.

core/src/main/java/com/scalar/db/io/TextColumn.java
Torch3333: You are right on both counts.

The ordering change is real, and the PR description said otherwise. `compareTo` moves from UTF-16 code unit order to Unicode code point order, which changes BINARY-mode ordering for supplementary characters relative to U+E000 to U+FFFF, and it reaches `Key.compareTo` and the object storage clustering order through `Ordering.natural()`. I have corrected the description, which no longer claims that no comparison result changes, and added a release note naming the object storage scan order and clustering key boundary effect.

On `TextValue.compareTo`, yes, deliberate. The class is deprecated for 4.0.0 and `Key` holds `Column` rather than `Value`, so `TextValue.compareTo` is not on the key ordering or scan ordering path. Changing it would move nothing.

core/src/main/java/com/scalar/db/common/CollationComparator.java
Torch3333: Confirmed against icu4j 77.1: `de-u-vt-0020` yields the keyword `variabletop`, which matches neither branch, so the sentence was wrong as written.

Took the suggestion and dropped the mapping from the Javadoc. It now states what the predicate is for: ICU applies a setting to the collator object and leaves that collator's rule string unchanged, so a collator rebuilt from those rules loses the setting. `de-u-kn-true` and plain `de` return byte-identical `getRules()`, and rebuilding from them flips `compare("file2", "file10")` from -1 to 1. A `-u-co-` type is excluded because it does change the rule string, so it survives the rebuild. I also noted that the returned names are ICU's legacy spellings (`colnumeric`), since `startsWith("col")` reads as arbitrary next to the BCP 47 subtag `kn`.

I did not add `variabletop` to the predicate. `Collator.getInstance` throws `UnsupportedOperationException: locale keyword vt/variableTop` for any locale carrying `vt`, with or without rules, so it never reaches the silent-drop failure mode this guard exists to prevent. Routing it here would also produce a misleading message, since `COLLATION_ICU_LOCALE_SETTING_WITH_RULES_NOT_SUPPORTED` implies that dropping the rules would make the locale work, which is false for `vt`.

Fixed in 2f174ac10.

core/src/main/java/com/scalar/db/common/CollationComparator.java
Torch3333: You are right that `und-Hant` and `und-Cyrl` take this early exit and resolve to root's rules, and the comment naming only `und` is why that was not visible.

It is not specific to the empty language, though. The guard checks the language and nothing else, so `de-Hant`, `de-Cyrl` and `de-Zzzz` also resolve to plain `de` with the script dropped and no error. `und-Hant` narrowing to root is the same narrowing as `de-Hant` narrowing to `de`, which `logResolvedIcuCollator` already documents as accepted and reports at INFO. Rejecting it only for the root language would be inconsistent, so I left the behavior alone and fixed the comment: it now says only the language is checked, and that a narrower subtag narrows away rather than failing. That Javadoc said "region or variant" where it should have said "script, region, or variant", so I corrected the cross-reference too.

Widened the root-locale test to `und`, `und-Cyrl`, `und-Hant` and `und-DE` to pin it.

Fixed in 44824d9ea.

core/src/main/java/com/scalar/db/storage/jdbc/RdbEngineStrategy.java
Torch3333: Added `@param collation` and `@throws IllegalArgumentException`, matching `throwIfConjunctionsColumnNotSupported`. `:core:javadoc` no longer reports the warning.

Fixed in 3365908da.

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #3843, opened 2026-09-02
Collation 2/4: compare in-memory text through the configured collation

Torch3333:
## Description

ScalarDB compares text in memory when it filters conjunctions a storage cannot evaluate, and when the object-storage adapter orders clustering keys, applies scan boundaries, and checks mutation conditions. Each of these called `Ordering.natural()` on the column, so none of them could see the configured collation. This PR passes the `CollationComparator` from the first PR into all of them, so a TEXT comparison follows the configured collation and every site takes its order from one comparator.

Under the default BINARY collation the resulting order is unchanged. The first PR ordered `TextColumn` by Unicode code point, which is the unsigned UTF-8 byte order the BINARY comparator produces, so these sites already compared text that way.

## Related issues and/or PRs

This PR is the second of a stack of four PRs based on `feature/collation` that together replace #3791. It is based on the first PR's branch, `collation-pr1-config-and-capability-gate`.

## Changes made

- `ScalarDbUtils.columnsMatchAnyOfConjunctions` takes a `CollationComparator`: GT, GTE, LT and LTE on TEXT columns compare through it, EQ and NE on two non-null TEXT values go through the new public `ScalarDbUtils.columnEquals`, and IS_NULL, IS_NOT_NULL, LIKE, NOT_LIKE, null text and non-TEXT columns keep `Column.equals` and `compareTo`. `columnEquals` compares TEXT values without their column names, so equality and the range operators cannot diverge.
- The two-argument overload stays and delegates with a BINARY comparator so the unchanged Consensus Commit call sites in `Snapshot` and `CrudHandler` still compile. The next PR in this stack removes the overload.
- `FilterableScanner` takes the comparator. Cassandra, Cosmos DB, DynamoDB and object storage build one from `DatabaseConfig` and pass it.
- In object storage, `ClusteringKeyComparator` and the scan start and end boundary checks in `SelectStatementHandler` order TEXT clustering keys with the comparator, and the conditional Put and Delete checks in `ObjectStoragePartition` use it for EQ, NE and the range operators, threaded through `MutateStatementHandler` and `ObjectStoragePartitionSnapshot`.
- `CollationConformanceTest` checks that the text, column and key comparators and conjunction filtering order a shared corpus identically under BINARY and ICU, and that an unset collation orders like BINARY. Unit tests cover the scanner and the utils under both collations, and the object-storage handlers under BINARY.

## Checklist

Every storage that reaches these sites rejects the ICU collation at construction since the first PR, and the two-argument overload is hard-wired to BINARY, so no runtime path on this branch uses the ICU collation. `CollationConformanceTest`, `ScalarDbUtilsTest` and `FilterableScannerTest` exercise the ICU order directly against the comparator.

## Release notes

N/A


core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
```diff
@@ -363,13 +367,48 @@ private static Set<String> getColumnNamesUsedIn(Set<Conjunction> conjunctions) {
...
+    return columnsMatchAnyOfConjunctions(columns, conjunctions, BINARY_COLLATION_COMPARATOR);
+  }
+
+  /**
+   * Returns whether the given columns match any of the given conjunctions.
+   *
+   * <p>Range and equality operators on TEXT columns compare through the given comparator. {@code
+   * LIKE} and {@code NOT_LIKE} always match on exact characters, whatever the collation.
+   */
+  public static boolean columnsMatchAnyOfConjunctions(
```
brfrn169, line 402: nit: this overload's Javadoc has no `@param`/`@return`, unlike the two-argument overload above and `columnEquals` below, and `javadoc` reports four warnings here. Since this is the overload that remains once the bridge is removed in the next PR, it seems worth fixing here.

brfrn169 (approved):
Left a non-blocking comment. Other than that, LGTM! Thank you!

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
```diff
@@ -363,13 +367,48 @@ private static Set<String> getColumnNamesUsedIn(Set<Conjunction> conjunctions) {
     return columns;
   }
 
+  private static final CollationComparator BINARY_COLLATION_COMPARATOR =
+      binaryCollationComparator();
+
+  private static CollationComparator binaryCollationComparator() {
+    Properties properties = new Properties();
+    properties.setProperty(DatabaseConfig.CONTACT_POINTS, "localhost");
+    properties.setProperty(DatabaseConfig.STORAGE, "jdbc");
```
feeblefakie, line 376: Are these two required? I think they can be deleted since they are not used, but if we need them for some reasons, should we make them like "dummy" to make it explicit?

Also, using DababaseConfig might make this tightly-coupled with it?
For example, if DatabaseConfig adds something as a required property, this needs to be changed.

Or, can it be simply `CollationComparator.binary()`?

Or, it might be better to define CollationConfig or something that only holds collation-related settings so that we don't need to create DatabaseConfig with unused values. WDYT?

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
```diff
@@ -383,23 +422,25 @@ public static boolean columnsMatchAnyOfConjunctions(
...
   private static <T> boolean columnMatchesCondition(
-      Column<T> column, ConditionalExpression condition) {
+      Column<T> column, ConditionalExpression condition, CollationComparator collationComparator) {
     assert column.getClass() == condition.getColumn().getClass();
     switch (condition.getOperator()) {
       case EQ:
+        return matchesEquality(column, condition, collationComparator);
       case IS_NULL:
         return column.equals(condition.getColumn());
       case NE:
```
feeblefakie, line 432: Does a NULL column still pass `NE` and `LT` here? With `col` NULL, `col != 'x'` returns true, and `col < 'x'` also returns true, since `columnComparator` wraps `Comparator.nullsFirst`.

JDBC sends the same predicates to SQL as `<>?` and `<?` (`QueryUtils.getConditionString`), which exclude NULL rows, and Cassandra/Cosmos/DynamoDB never push conjunctions down -- `getConjunctions()` is only read to decide whether to wrap in `FilterableScanner`. So the same scan returns different rows depending on the backend.

`MutationConditionsValidator` guards its range operators with `hasNullValue()`, and the comment there says it does so to "mimic the behavior as if the condition was executed by the underlying storage", which suggests NULL isn't meant to satisfy a range operator.

This is pre-existing and unchanged by this PR, so not a blocker -- just wondering whether it's worth aligning while this switch is being touched.

feeblefakie (commented):
Overall, looking good. Left some comments. PTAL!

PR #3844, opened 2026-09-02
Collation 3/4:  follow the collation in Consensus Commit layer

Torch3333:
## Description

Consensus Commit repeats part of the storage's work in memory: keying its read, write, and delete sets by record key, checking whether a buffered write overlaps a later scan, and validating conditional mutations. The previous PR in this stack made the storage-side in-memory comparisons follow `scalar.db.collation`; this PR does the same here.

Under ICU, collate-equal spellings name the same row, so the snapshot treats them as one key and equality and range checks use the collation. Mutation grouping stays byte-exact, since every storage that groups per record or per partition rejects ICU at creation. Under BINARY, key identity stays byte-exact; the one default-path change is that TEXT range ordering in the scan-after-write check and in conditional-mutation validation moves from Java String order to unsigned UTF-8 byte order. Pattern matching stays byte-exact and would disagree with an ICU backend, so transactions reject LIKE and NOT LIKE under ICU.

## Related issues and/or PRs

This PR is the third of a stack of four PRs based on `feature/collation` that together replace #3791. It is based on the second PR's branch, `collation-pr2-utf8-text-comparison`.

## Changes made

- `Snapshot.Key` identity follows the collation: under ICU, TEXT key columns are identified by their collation key (`CollationComparator.canonicalTextFormOf`), so collate-equal keys hit one entry in the read, write, and delete sets and compare as equal; under BINARY, equality, hashing, and ordering are unchanged. The fields and `toString` keep the original bytes.
- The scan-after-write overlap check in `Snapshot` compares partition keys and tests clustering-key range membership with the collation's key comparator; inclusive boundaries are an ordering test, so a written key that collates equal to a boundary is in range. Conjunction re-evaluation on merged results and the scan-with-index column match use the collation's equality.
- `MutationConditionsValidator` decides EQ and NE with the collation's equality and GT, GTE, LT, LTE with the collation's column comparator; IS_NULL and IS_NOT_NULL are unchanged.
- `MergedResult` reports a key column with the stored record's spelling, from `getText` and from the deprecated `getPartitionKey` and `getClusteringKey`, and uses the Put's spelling only when there is no stored record.
- `CrudHandler` keeps a row returned by a Get with index whose stored index value collates equal to the queried value instead of filtering it out byte-exactly.
- `ConsensusCommitOperationChecker` takes the configured `Collation` and, under ICU, rejects a Get or Scan carrying a LIKE or NOT LIKE condition with `CoreError.COLLATION_ICU_LIKE_CONDITION_NOT_SUPPORTED` at every isolation level. It also throws `AssertionError` for a mutation on a storage whose atomicity unit is RECORD or PARTITION under ICU, with a comment recording why that branch is unreachable and that `MutationsGrouper` compares keys byte-exactly.
- `ConditionChecker` rejects a LIKE or NOT LIKE expression in a conditional mutation, since no storage evaluates pattern matching in a conditional write.
- `ConsensusCommitManager`, `TwoPhaseConsensusCommitManager`, and `ConsensusCommitParticipant` build one `CollationComparator` from `DatabaseConfig` and pass it to `Snapshot` and `CrudHandler`; the constructors of those classes, of `Snapshot.Key`, and of `MutationConditionsValidator` take it as a parameter.
- Deleted the two-argument `ScalarDbUtils.columnsMatchAnyOfConjunctions` overload; its remaining callers, `Snapshot` and `CrudHandler`, pass their own comparator. Range conditions (GT, GTE, LT, LTE) in `ScalarDbUtils` evaluate to false on a null column instead of comparing it.
- Tests: `SnapshotKeyTest` covers key identity under BINARY and case-insensitive ICU (case variants, composite keys, null text, keys built with different comparators); `MutationsGrouperTest` covers that case-differing keys stay in separate batches; `SnapshotTest` covers read-your-own-write, write-set merging, scan-after-write validation, and serializable validation with collate-equal keys, range boundaries, index values, and conjunctions under both collations; `MutationConditionsValidatorTest` covers EQ, NE, and GT on text and non-text under both collations; `CrudHandlerTest` covers the Get-with-index match; `ConsensusCommitOperationCheckerTest` and `ConsensusCommitManagerTest` cover the LIKE and NOT LIKE rejection for Get, Scan, and ScanAll and the atomicity-unit check; `MergedResultTest` covers the stored key spelling; `OperationCheckerTest` and `ScalarDbUtilsTest` cover the conditional LIKE rejection and null range filtering.
- Updated the remaining `consensuscommit` unit tests and the three `ConsensusCommit*IntegrationTestBase` classes in the `integration-test` module to the new constructor arities.

## Checklist

This PR adds no integration tests; the `integration-test` module changes only follow the new constructor arities. The next PR in this stack carries the integration tests and CI.

## Release notes

Fixed an issue where a conditional mutation carrying a LIKE or NOT LIKE condition failed with an `AssertionError` at execution. Such conditions are rejected up front with an `IllegalArgumentException`.


core/src/main/java/com/scalar/db/common/checker/ConditionChecker.java
```diff
@@ -69,6 +69,12 @@ public void visit(DeleteIfExists condition) {
 
   private void checkExpressions(List<ConditionalExpression> expressions) {
     for (ConditionalExpression expression : expressions) {
+      if (expression.getOperator() == Operator.LIKE
+          || expression.getOperator() == Operator.NOT_LIKE) {
+        // No storage evaluates pattern matching in a conditional write.
+        isValid = false;
+        break;
+      }
```
brfrn169, line 77: The `LIKE`/`NOT_LIKE` rejection added here doesn't take effect in Consensus Commit. `ConsensusCommitOperationChecker` calls `ConditionChecker.check()` for `Put` and `Delete` but discards the result:

- https://github.com/scalar-labs/scalardb/blob/cf6b7be7bfd64bec50ea718e6b5b6d55b83bbbe0/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java#L262
- https://github.com/scalar-labs/scalardb/blob/cf6b7be7bfd64bec50ea718e6b5b6d55b83bbbe0/core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java#L281

Since `ConditionChecker` has no side effects, these calls currently do nothing. A `LIKE` condition that bypasses the builder check (e.g., via `ConditionBuilder.putIf(List)` or `deleteIf(List)`) passes the checker and, when the target record exists, hits `default: throw new AssertionError()` in `MutationConditionsValidator.shouldMutate()`. Other conditions that the storage-side `OperationChecker` rejects, such as a nonexistent column or a type mismatch, aren't rejected up front in Consensus Commit either.

This bug isn't introduced by this PR. The result has been discarded since the call was added in #899, and the same code is on `3` and `3.16` through `3.19`. Could you fix it in a separate PR against `master`? The fix should be backported to all the supported release branches, which is easier as a standalone change than as part of this collation stack.

Some notes for that PR:

- Throwing `IllegalArgumentException` when `check()` returns `false`, as `OperationChecker.checkCondition()` does with `CoreError.OPERATION_CHECK_ERROR_CONDITION`, would align Consensus Commit with the storage API and JDBC transactions.
- This is a behavior change. Conditions on primary-key columns and comparisons against a null value (e.g., `isEqualToText(null)`) are currently accepted and evaluated in memory; they would start to be rejected.
- `ConsensusCommitOperationCheckerTest` uses an unstubbed `ConditionChecker` mock, which returns `false`, so the existing tests that assert no exception would need `check()` stubbed to return `true`.

core/src/test/java/com/scalar/db/common/checker/OperationCheckerTest.java
```diff
@@ -973,6 +974,27 @@ public void whenCheckingPutOperationWithDeleteIfCondition_shouldThrowIllegalArgu
         .isInstanceOf(IllegalArgumentException.class);
   }
 
+  @Test
+  public void whenCheckingPutOperationWithLikeCondition_shouldThrowIllegalArgumentException() {
+    // Arrange
+    Key partitionKey = Key.of(PKEY1, 1, PKEY2, "val1");
+    Key clusteringKey = Key.of(CKEY1, 2, CKEY2, "val1");
+    MutationCondition condition = new PutIf(ConditionBuilder.column(PKEY2).isLikeText("val%"));
```
brfrn169, line 982: This test passes without the new `LIKE` branch in `ConditionChecker`. `PKEY2` is a partition key column, and `ColumnChecker` is constructed with `requireNotPrimaryKey = true`, so it already returns `false` for this column and `OperationChecker` throws `IllegalArgumentException` regardless of the operator. The test table has no non-key TEXT column, so one would be needed for this test to exercise the `LIKE` branch.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -36,16 +38,35 @@ public class ConsensusCommitOperationChecker {
...
+      boolean isIncludeMetadataEnabled,
+      Collation collation) {
     this.transactionTableMetadataManager = transactionTableMetadataManager;
     this.virtualTableInfoManager = virtualTableInfoManager;
     this.storageInfoProvider = storageInfoProvider;
     this.isIncludeMetadataEnabled = isIncludeMetadataEnabled;
+    this.collation = collation;
+  }
+
+  private void throwIfLikeConditionUnderIcuCollation(Selection selection) {
```
brfrn169, line 56: nit: The other private helpers in this class, including `throwIfKeyedMutationAtomicityUnitUnderIcuCollation` added in this PR, are placed after the public `check` methods. Could you move this one there as well?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -36,16 +38,35 @@ public class ConsensusCommitOperationChecker {
...
+    if (collation != Collation.ICU) {
+      return;
+    }
+    for (Selection.Conjunction conjunction : selection.getConjunctions()) {
+      for (ConditionalExpression condition : conjunction.getConditions()) {
+        Operator operator = condition.getOperator();
+        if (operator == Operator.LIKE || operator == Operator.NOT_LIKE) {
+          throw new IllegalArgumentException(
+              CoreError.COLLATION_ICU_LIKE_CONDITION_NOT_SUPPORTED.buildMessage(
+                  operator, selection.forFullTableName().get(), condition.getColumn().getName()));
```
brfrn169, line 66: Question: Could you explain why `LIKE`/`NOT_LIKE` can't be supported under ICU?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -274,6 +298,28 @@ ConditionChecker createConditionChecker(TableMetadata tableMetadata) {
...
+    }
+    assert mutation.forNamespace().isPresent();
+    StorageInfo storageInfo = storageInfoProvider.getStorageInfo(mutation.forNamespace().get());
+    StorageInfo.MutationAtomicityUnit unit = storageInfo.getMutationAtomicityUnit();
+    if (unit == StorageInfo.MutationAtomicityUnit.RECORD
+        || unit == StorageInfo.MutationAtomicityUnit.PARTITION) {
+      // No storage with a record or partition atomicity supports ICU collation
+      // If we support such storage in the future, we need to make the MutationsGrouper collation
+      // aware
+      throw new UnsupportedOperationException(
```
brfrn169, line 314: This branch is effectively dead code: every storage whose mutation atomicity unit is `RECORD` or `PARTITION` rejects the ICU collation when it is created, and multi-storage doesn't allow overriding the collation per storage. How about throwing `AssertionError` instead and writing down why it's unreachable?

```suggestion
      // Unreachable: every storage whose mutation atomicity unit is RECORD or PARTITION rejects
      // the ICU collation when it is created (CoreError.COLLATION_ICU_NOT_SUPPORTED_BY_STORAGE),
      // and multi-storage doesn't allow overriding the collation per storage. If such a storage
      // supports the ICU collation in the future, revisit MutationsGrouper, which compares keys
      // byte-exactly.
      throw new AssertionError(
```

`checkForMutation_WithKeyedMutationAtomicityUnitUnderIcuCollation_ShouldThrowUnsupportedOperationException` would need to be updated as well.

core/src/test/java/com/scalar/db/transaction/consensuscommit/MergedResultTest.java
```diff
@@ -123,6 +123,25 @@ public void getPartitionKey_ResultAndPutGiven_ShouldReturnCorrectKey() {
...
   }
 
+  @Test
+  public void getText_ResultAndPutWithDifferentKeySpellingGiven_ShouldReturnStoredKeySpelling() {
+    // Arrange
+    Put put =
+        Put.newBuilder()
+            .table("test")
+            .partitionKey(Key.ofText(ANY_NAME_1, ANY_TEXT_1.toLowerCase()))
+            .clusteringKey(Key.ofText(ANY_NAME_2, ANY_TEXT_2.toLowerCase()))
```
brfrn169, line 133: nit: `ANY_TEXT_1` and `ANY_TEXT_2` are already lowercase (`"text1"` and `"text2"`), so `toLowerCase()` doesn't change the spelling and this test passes without the change in `MergedResult`. Using a spelling that differs from the stored one (e.g., `toUpperCase()`) would make the test cover the change.

Torch3333: Fixed in d9003ec89: the Put now uses the uppercase spelling, and the test fails without the MergedResult change.

core/src/test/java/com/scalar/db/common/checker/OperationCheckerTest.java
Torch3333: Fixed in c1ac1a8d7: the test table gains a non-key TEXT column (v4) and the LIKE condition targets it. Verified the test fails with the LIKE branch removed. The tests that used "v4" as an unknown column now use "v5".

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
Torch3333: Applied in a0d13f226 with your comment, and renamed the test to expect AssertionError.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
Torch3333: Moved in 34d21366d.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
```diff
@@ -274,6 +298,28 @@ ConditionChecker createConditionChecker(TableMetadata tableMetadata) {
...
+      return;
+    }
+    assert mutation.forNamespace().isPresent();
+    StorageInfo storageInfo = storageInfoProvider.getStorageInfo(mutation.forNamespace().get());
+    StorageInfo.MutationAtomicityUnit unit = storageInfo.getMutationAtomicityUnit();
+    if (unit == StorageInfo.MutationAtomicityUnit.RECORD
+        || unit == StorageInfo.MutationAtomicityUnit.PARTITION) {
+      // No storage with a record or partition atomicity supports ICU collation
+      // If we support such storage in the future, we need to make the MutationsGrouper collation
+      // aware
```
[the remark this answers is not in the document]
Torch3333: The description was wrong. MutationsGrouper is unchanged and compares keys byte-exactly. The description now says so, and the checker branch for RECORD/PARTITION storages is documented as unreachable in a0d13f226.

core/src/main/java/com/scalar/db/transaction/consensuscommit/MergedResult.java
```diff
@@ -32,9 +32,13 @@ public MergedResult(Optional<TransactionResult> result, Put put, TableMetadata m
...
     putColumns = new HashMap<>();
     putColumns.putAll(put.getColumns());
-    put.getPartitionKey().getColumns().forEach(c -> putColumns.put(c.getName(), c));
-    put.getClusteringKey()
-        .ifPresent(k -> k.getColumns().forEach(c -> putColumns.put(c.getName(), c)));
+    if (!result.isPresent()) {
+      // A Put cannot rewrite a stored key, so a stored record keeps its own key spelling even when
+      // the Put's key only collates equal to it.
+      put.getPartitionKey().getColumns().forEach(c -> putColumns.put(c.getName(), c));
+      put.getClusteringKey()
```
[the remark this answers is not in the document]
Torch3333: Fixed in 6c75087d3: getPartitionKey and getClusteringKey return the stored result keys when present and fall back to the Put only for an insert.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -1007,22 +1137,29 @@ public boolean equals(Object o) {
...
-          && this.table.equals(another.table)
-          && this.partitionKey.equals(another.partitionKey)
-          && this.clusteringKey.equals(another.clusteringKey);
+      // Byte-equal keys are collate-equal under every collation, so the common case of a re-read
+      // row skips collation-key generation.
+      if (namespace.equals(another.namespace)
+          && table.equals(another.table)
+          && partitionKey.equals(another.partitionKey)
+          && clusteringKey.equals(another.clusteringKey)) {
+        return true;
```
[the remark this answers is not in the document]
Torch3333: Keys built with different comparators never meet: every Snapshot.Key construction site passes the single CollationComparator owned by the transaction (Snapshot, CrudHandler, TransactionContext), so the byte-equal fast path and the identity-based path agree with hashCode within a snapshot. The column-name difference between equals and compareTo is also moot: compareTo orders namespace and table first, and two keys of the same table have the same key column names in the same positions. No change made.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -99,7 +105,7 @@ public Optional<Result> get(Get get, TransactionContext context) throws CrudExce
       // In case of a Get with index, we don't know the key until we read the record
       key = null;
     } else {
-      key = new Snapshot.Key(get);
+      key = new Snapshot.Key(get, collationComparator);
```
feeblefakie, line 108: `key` is collation-aware here, but `containsKeyInGetSet` just below still matches on the raw `Get`, and `getSet`/`scanSet` are keyed by the raw operation, whose `equals` is byte-exact. Under a case-insensitive ICU collation, `get(pk="ABC")` followed by `get(pk="abc")` misses the get set and re-reads storage, and `getResult(key, get)` takes its base value from `getSet`, so the second get returns the newer value while the read set still holds the first one. Since mutations are composed from the read set, prepare then uses the older before-image. I reproduced this at SNAPSHOT isolation: the two keys compare equal, the read set keeps the first value, and the second `get` returns the second one. Should `getSet` and `scanSet` be keyed on something collation-aware too?

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -937,44 +983,128 @@ private void throwExceptionDueToAntiDependency() throws ValidationConflictExcept
...
+
+    /**
+     * Memoized: a key is built for every result row of a scan, but only the ones actually looked up
+     * or compared need their identity.
+     */
+    private final Supplier<List<Object>> identity;
+
+    private final Supplier<Integer> hash;
+
+    public Key(Get get, CollationComparator collationComparator) {
```
feeblefakie, line 1011: The single-argument `Snapshot.Key(Get/Put/Delete)` constructors are removed here, and scalardb-cluster still calls them in `LogWriterBeforePreparationHookTest` at 14 sites. `scalardb-cluster/build.gradle` is back to `scalarDbVersion = projectVersion` and its CI runs the root `check` task, so that test source set will stop compiling once a snapshot publishes. Should we keep the one-argument forms as deprecated overloads, or prepare a paired cluster PR?

core/src/main/java/com/scalar/db/util/ScalarDbUtils.java
```diff
@@ -433,14 +404,19 @@ private static <T> boolean columnMatchesCondition(
         return !matchesEquality(column, condition, collationComparator);
       case IS_NOT_NULL:
         return !column.equals(condition.getColumn());
+        // A NULL satisfies no range condition, as in the storages' own evaluation.
```
feeblefakie, line 407: This one changes behavior outside Consensus Commit. `FilterableScanner` calls this for Cassandra, Cosmos, Dynamo and object storage, and a NULL column used to sort first and therefore match `LT`/`LTE`, so `DistributedStorage.scan` on those storages will stop returning rows it returns today, under BINARY as well as ICU. It looks like the right semantics, but should it be a separate PR, or at least a release note instead of N/A?

core/src/test/java/com/scalar/db/transaction/consensuscommit/MutationsGrouperTest.java
```diff
@@ -610,6 +610,84 @@ public void canBeGroupedAltogether_WithOverMaxCount_ShouldReturnFalse()
...
+    StorageInfo storageInfo = mock(StorageInfo.class);
+    when(storageInfo.getMutationAtomicityUnit()).thenReturn(unit);
+    when(storageInfo.getMaxAtomicMutationsCount()).thenReturn(100);
+    when(storageInfo.getStorageName()).thenReturn("storage1");
+    return storageInfo;
+  }
+
+  @Test
+  public void
+      groupMutations_WithPartitionAtomicityAndCaseDifferingPartitionKeys_ShouldGroupIntoTwoBatches()
```
feeblefakie, line 623: `MutationsGrouper` has no `CollationComparator` and there is nowhere to configure a collation in this test, so `"Apple"` vs `"apple"` here is just two different keys. These three tests pass with the whole PR reverted. Should we drop the collation framing from the names, or remove them given `ConsensusCommitOperationCheckerTest` already covers the guard?

core/src/test/java/com/scalar/db/transaction/consensuscommit/MutationConditionsValidatorTest.java
```diff
@@ -326,4 +331,156 @@ private void validateConditionIsSatisfied(Mutation mutation, TransactionResult e
...
+    // Act Assert
+    Assertions.assertThatCode(
+            () ->
+                collationValidator.checkIfConditionIsSatisfied(put, existingRecord, TRANSACTION_ID))
+        .doesNotThrowAnyException();
+  }
+
+  @Test
+  public void
+      validateConditionIsSatisfied_WithBinaryCollationAndEqOnText_ShouldThrowUnsatisfiedConditionException() {
```
feeblefakie, line 418: This looks byte-for-byte identical to `validateConditionIsSatisfied_WithBinaryCollationAndEqOnText_ShouldStayByteExactAndThrow` above, only the local variable name differs. Should one of them cover GTE/LTE at an ICU collate-equal boundary instead? That edge of `rangeCompare` has no test today.

core/src/test/java/com/scalar/db/transaction/consensuscommit/SnapshotTest.java
```diff
@@ -2734,4 +2779,946 @@ public void requiresBeforeIndexValidation_ScanAllWithSecondaryIndexCondition_Sho
...
+
+    Map<Snapshot.Key, TransactionResult> results = new HashMap<>();
+    results.put(keyUpper, prepareResult(ANY_ID, ANY_TEXT_1, "Apple"));
+    assertThat(results.containsKey(keyUpper)).isTrue();
+    assertThat(results.containsKey(keyLower)).isFalse();
+  }
+
+  @Test
+  public void
+      verifyNoOverlap_ScanWithRangeAndBinaryCollationGivenAndCaseDifferingWrittenKey_ShouldReproduceByteExactBehavior()
```
feeblefakie, line 2891: This is byte-for-byte the same as `..._ShouldNotThrowException` above. Should one of them become the ICU exclusive-boundary case instead? Every ICU boundary test here uses inclusive bounds, and the exclusive one is where the old and new code actually disagree.

core/src/test/java/com/scalar/db/transaction/consensuscommit/SnapshotTest.java
```diff
@@ -2734,4 +2779,946 @@ public void requiresBeforeIndexValidation_ScanAllWithSecondaryIndexCondition_Sho
...
+    // Act
+    Throwable thrown = catchThrowable(() -> snapshot.verifyNoOverlap(scan, Collections.emptyMap()));
+
+    // Assert
+    assertThat(thrown).doesNotThrowAnyException();
+  }
+
+  @Test
+  public void
+      verifyNoOverlap_PlainScanOfDistinctUnpairedSurrogatePartitionKeyUnderBinary_ShouldNotThrowException()
```
feeblefakie, line 3426: Does this pin anything? `compareByCodePoint` also separates U+D800 from U+DC00, so it still passes with the BINARY fast path in `partitionKeyEquals` removed. More broadly, no ordering assertion in this PR uses a character above U+007F, so a regression from `compareByCodePoint` back to `String::compareTo` would go unnoticed. Should we add one ordering case with a supplementary character against a U+E000..U+FFFF one?

core/src/test/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationCheckerTest.java
```diff
@@ -928,4 +942,224 @@ public void checkForDelete_ForVirtualTableAndConsistentReadsGuaranteed_ShouldNot
...
+                storageInfoProvider,
+                false,
+                Collation.ICU));
+    when(icuChecker.createConditionChecker(any())).thenReturn(conditionChecker);
+    return icuChecker;
+  }
+
+  @ParameterizedTest
+  @EnumSource(Isolation.class)
+  public void checkForScan_WithLikeConditionUnderIcuCollation_ShouldThrowIllegalArgumentException(
```
feeblefakie, line 1007: `throwIfLikeConditionUnderIcuCollation` is the second statement of `check`, before anything isolation-dependent runs, so the isolation parameter doesn't change the path being tested. Should these be plain `@Test`s, or parameterized over `Scan`/`ScanAll`/`ScanWithIndex` instead?

feeblefakie (commented):
Overall, looking good. Left some comments. PTAL!

core/src/main/java/com/scalar/db/common/checker/ConditionChecker.java
feeblefakie: One more thing on this hunk: the rejection is not ICU-only, it applies to every storage at every collation, so a `putIf`/`deleteIf` carrying LIKE that is accepted today starts being rejected. That looks correct given JDBC would otherwise hit the `AssertionError` in `QueryUtils.getConditionString`, but should it be called out in the release notes rather than N/A?

Torch3333: @brfrn169 Confirmed: both call sites drop the boolean, so `ConditionChecker` has had no effect in Consensus Commit since the call was added. The standalone fix against `master` is #3871. It throws `IllegalArgumentException` with `CoreError.OPERATION_CHECK_ERROR_CONDITION` when `check()` returns `false`, stubs the mocked `ConditionChecker` to return `true` in `ConsensusCommitOperationCheckerTest`, and adds one rejection test each for `Put` and `Delete`. The PR notes the behavior change for conditions on primary-key columns and comparisons against a null value.

@feeblefakie Also confirmed. `ConditionBuilder.putIf(List)` and `deleteIf(List)` skip the builder's operator check, and every storage's condition translator (`QueryUtils.getConditionString` for JDBC, `ConditionSetter` for Cassandra, `ConditionalQueryBuilder` for Cosmos, `ConditionExpressionBuilder` for DynamoDB) hits `default: throw new AssertionError()` on LIKE. On the storage API this hunk turns that `AssertionError` into an `IllegalArgumentException` from `OperationChecker`, at any collation. In Consensus Commit it takes effect once #3871 lands. The release notes of this PR say so.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitOperationChecker.java
Torch3333: The storage evaluates a LIKE on the pushed-down read and Consensus Commit evaluates it again in memory against the snapshot. The in-memory evaluation matches exact characters, while an ICU `Collator` only defines ordering and equality of whole strings and has no pattern-matching primitive, so under ICU the two evaluations can return different rows. Rejecting the operator was preferred to letting them silently diverge. The full reasoning is in the Javadoc of `throwIfLikeConditionUnderIcuCollation`,  see https://github.com/scalar-labs/scalardb/pull/3844/commits/67cb75f6b5794919c8c6861f8bf17b8467c7e814

PR #3850, opened 2026-09-07
Allow writes to records a scanner has already returned

brfrn169:
## Description

Consensus Commit rejects Update and Delete on a record while a `Scanner` is still iterating. At `Scanner.close()`, `verifyNoOverlap` detects any overlap between the write/delete set and the scan results and throws `DB-CORE-USER-0106`.

The check only asks *whether* the write set and the scan results overlap, never *when* the write happened relative to the read. So it rejects the safe case — a write issued after the scan had already returned the record — along with the unsafe one. The same restriction does not apply to `scan()`, which returns every row at once and verifies once at completion, so scan-then-write already works there.

The practical cost is that mutating while iterating is simply not expressible: callers have to fall back to `scan()` and process the result in two passes.

This PR adds the missing ordering signal.

**Why it is safe.** Writes are not applied to storage until commit, so a write issued mid-iteration cannot affect any row the scanner reads afterwards. For the rows a scanner has already delivered, the values the caller saw are identical to those it would have seen under "finish the whole scan, then apply all writes" — which Consensus Commit already permits through `scan()`.

The claim is per-scanner and does not extend to later reads. Any other scan or scanner whose range covers a written key is still rejected, because that read would return a stale row.

**What is not changed.** Every rejection that happens today still happens, at the same call site (`close()` / scan completion). Inserts into the scanned range stay rejected. The `scan()` batch path is untouched.

## Related issues and/or PRs

N/A

## Changes made

Four commits, each of which leaves the tree green on its own.

**1. Make the delete-set-to-write-set transition in `Snapshot` atomic**

`putIntoWriteSet` removed the key from the delete set before building the merged `Put`, so a throw from `getTableMetadata` left the key in neither set. That silently lost the delete, and it breaks the invariant the new ordering rule depends on. Build the merged `Put` first and perform both map mutations last. The invariant is now documented on both mutators.

**2. Decide scan overlap by when the write happened**

Scanners build a **filtered exempt set** as they produce rows: a key is recorded only when it is still absent from `writeSet ∪ deleteSet` at that moment, which means any later write to it happened after the caller had already seen the record. `close()` passes the set to a new `verifyNoOverlap(Scan, Map, Set)` overload that skips those keys in all three overlap checks, including the conjunction check — which would otherwise fire on an exempt record whose updated columns still match the predicate.

A key written *before* the scanner produced it never enters the set, so it is rejected exactly as before. The set is recorded only for rows that are actually delivered: rows dropped by the conjunction filter or by lazy recovery never reach the caller, and exempting them would let a later write escape the range and conjunction checks. The set is allocated only when overlap verification is required, so read-only and one-operation transactions track nothing.

`verifyNoOverlap` takes the exempt keys as a third parameter rather than gaining an overload. The batch scan path passes an empty set, which is exactly the old behaviour of rejecting every overlap, so `scan()` is unchanged. Keeping a two-argument overload was considered and dropped: it would have had no caller, and it would not have protected a downstream override either, since the scanners call the three-argument form and would silently bypass it. Removing it turns that into a compile error instead. The overlap checks now return the offending key instead of a boolean, so `0106` can name the record.

**3. Integration coverage**

The positive pattern goes in `DistributedTransactionIntegrationTestBase`, so iterating a scanner and mutating each returned record is an API-level guarantee. The Consensus Commit specific rejection and isolation cases stay in `ConsensusCommitSpecificIntegrationTestBase` and run under all three isolation levels.

**4. Pin the exempt keys passed to `verifyNoOverlap` in the scan tests**

The `ScanType`-parameterized tests now assert the exact exempt set instead of `any()`: empty for `scan()`, and the delivered keys for the scanners.

## Checklist

### User-facing rule

Within a transaction you may write to a record once a scan or scanner has returned it. A scan or scanner is rejected at `close()` / completion if it returns a record the transaction had already written or deleted, or if the transaction put a record that may fall within its range or conditions — judged from the key and the columns the put sets — without the scan having returned it first.

| Action | Result |
|---|---|
| Write a record the scan or scanner already returned | Allowed |
| Write or delete a record, then have a scan or scanner return it — including another scanner still open, or a later scan | Rejected at that scan's `close()` / completion |
| Put a record the scan has not returned but that may fall within its range or conditions — including an insert into the range | Rejected at `close()` / scan completion, even if a scanner never reaches the record |
| Delete a record a scanner covers but closes before returning | Not rejected; the caller never saw the record |
| `insert()` a record the scan already returned | Not an overlap; fails at commit because the record exists |
| Insert into the range after the scanner is closed | Allowed (pre-existing `scan()` behaviour) |

The exemption is per scanner; the rejection is transaction-wide. The sharp edge worth documenting: whether a write is allowed depends on whether the scanner has already reached that row, which for a scan with a `limit` or a partially consumed scanner is data-dependent.

### Behaviour changes reviewers should weigh

1. **`insert()` on a record a scanner already returned now fails later, and retriably.** Today it is
   caught at `close()` with `0106`. After this change the key is exempt — correctly, because the
   write cannot alter which rows the scan should have returned — so the invalid insert reaches
   prepare and surfaces as a retriable `CommitConflictException`. For a record that genuinely exists,
   retrying never converges. This matches the long-standing `get()`-then-`insert()` behaviour that
   `putAndCommit_PutWithInsertModeGivenForExistingAfterRead_ShouldThrowCommitConflictException`
   pins, but it does mean one path loses its fail-fast. Fixing it means reclassifying the
   prepare-time `CONSENSUS_COMMIT_PREPARING_RECORD_EXISTS` result, which affects the `get()` path too
   and belongs in its own change.

   An earlier revision of this PR added a CRUD-time rejection for it. It was withdrawn: under
   READ_COMMITTED a read returns the committed before-image of a record whose delete has already
   committed, so "present in the read set" does not imply the record exists, and the following insert
   is correct and must succeed — as
   `insert_InsertGivenForDeletedWhenCoordinatorStateCommitted_ShouldBehaveCorrectly` requires.

2. **A write can now be rejected at a different scanner's `close()`.** If a second scanner is still
   open, the write is accepted where it was made and rejected at the second scanner's `close()` when
   that scanner goes on to return the record or, for a put, when the put may fall within its range or
   conditions. Whether that happens depends on how far the second scanner gets and when it is closed
   relative to the write.

3. **The relaxation is unconditional.** No configuration restores the previous strict behaviour.
   Opt-in was considered and rejected: an opt-in caller carries the same data-dependent acceptance,
   and there is no precedent in this layer for a switch that toggles transaction semantics.

4. **`0106` message text changed** — it now names the record.

### Things checked, not assumed

- `JdbcTransaction` supports mutating while its scanner is open — verified on SQLite, PostgreSQL and MySQL. It has no overlap check at all, so nothing is relaxed there; the shared-base tests pin behaviour that had never been asserted. Oracle, SQL Server and YugabyteDB are left to CI.
- SERIALIZABLE validation tolerates exempted keys with a `limit`, with a partially consumed scanner, and with a conjunction that the update stops matching.

### Still open

- **Documentation**: scalardb-docs needs the user-facing rule above and the reworded `0106`. Not yet written.
- **Downstream check before merge**: confirm nothing in scalardb-cluster or scalardb-sql calls or overrides the public `Snapshot.verifyNoOverlap`. Its signature changed, so any such use is a compile error rather than a silent behaviour change or string-matches the `0106` message.
- **Separate task**: `close()` is not an unconditional verification point. `one()`'s catch blocks call `closeScanner()`, which sets `closed`, and `close()` returns early when it is set, so verification is skipped after a `one()` failure; both scanners also set `closed` before verifying, so a caller that swallows the rejection can still reach `commit()`. Pre-existing and untouched here, but this change makes `close()` the sole enforcement point for the new rule.

### Post-deploy monitoring and validation

No runtime rollout concerns — this is a library change with no background jobs, migrations, or external calls. What to watch after the first release that includes it:

- **Expected healthy signal**: `DB-CORE-USER-0106` occurrences drop, since the scan-then-write cases that used to trip it now pass.
- **Expected new signal**: a rise in `DB-CORE-CONCURRENCY-0013` (`The record being prepared already exists`). Inserts on records a scanner already returned now reach prepare instead of being stopped at `close()`. A large rise would mean applications are hitting the non-converging retry described above.
- **Rollback trigger**: unexpected `ValidationConflictException` under SERIALIZABLE on scan-heavy workloads, which would point at the exempt-key interaction with `validateScanResults`.

## Release notes

Allowed updating and deleting records that a scanner has already returned, so applications can mutate rows while iterating a scanner instead of buffering the whole result with `scan()`. Inserting into the scanned range remains disallowed.

### DIFF
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
@@ -29,11 +29,14 @@ import com.scalar.db.util.ScalarDbUtils;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
@@ -407,7 +410,9 @@ public class CrudHandler {
     TableMetadata metadata = txMetadata.getTableMetadata();
     LinkedHashMap<Snapshot.Key, TransactionResult> results =
         scanInternal(scan, context, txMetadata);
-    verifyNoOverlap(scan, results, context);
+    // The batch scan path exempts nothing: no write can be interleaved with it, so a key found in
+    // the write set here was necessarily written before the scan ran.
+    verifyNoOverlap(scan, results, Collections.emptySet(), context);
     return results.values().stream()
         .map(r -> new FilteredResult(r, scan.getProjections(), metadata, isIncludeMetadataEnabled))
         .collect(Collectors.toList());
@@ -561,10 +566,35 @@ public class CrudHandler {
   }
 
   private void verifyNoOverlap(
-      Scan scan, Map<Snapshot.Key, TransactionResult> results, TransactionContext context) {
+      Scan scan,
+      Map<Snapshot.Key, TransactionResult> results,
+      @Nullable Set<Snapshot.Key> exemptKeys,
+      TransactionContext context) {
     if (isOverlapVerificationRequired(context)) {
-      context.snapshot.verifyNoOverlap(scan, results);
+      // Reaching here guarantees exemptKeys was allocated: both scanners allocate it under exactly
+      // this condition, and the flags it reads are final on the context, so the condition cannot
+      // have differed at construction time. A scanner that skips the allocation therefore never
+      // gets here. The batch scan path passes an empty set explicitly.
+      assert exemptKeys != null;
+
+      context.snapshot.verifyNoOverlap(scan, results, exemptKeys);
     }
+  }
+
+  /**
+   * Returns whether this transaction has not yet written or deleted the given key.
+   *
+   * <p>A scanner calls this at the moment it produces a record. If the key is still absent from
+   * both the write set and the delete set, any later write to it necessarily happened after the
+   * caller had already seen the record, which is safe and makes the key exempt from the overlap
+   * check at close time. A key that is already written or deleted here is never exempt, so the scan
+   * is still rejected for it. Union membership is monotonic (see {@link Snapshot#putIntoWriteSet}),
+   * which is what makes this membership test equivalent to comparing the order of the two
+   * operations.
+   */
+  private boolean isNotYetWrittenOrDeleted(Snapshot.Key key, TransactionContext context) {
+    return !context.snapshot.containsKeyInWriteSet(key)
+        && !context.snapshot.containsKeyInDeleteSet(key);
   }
 
   private boolean isOverlapVerificationRequired(TransactionContext context) {
@@ -936,6 +966,12 @@ public class CrudHandler {
     private final Scanner scanner;
 
     @Nullable private final LinkedHashMap<Snapshot.Key, TransactionResult> results;
+
+    // The keys this scanner produced while they were still absent from the write set and the
+    // delete set. Writes to them happened after the caller saw the record, so they are exempt from
+    // the overlap check at close time.
+    @Nullable private final Set<Snapshot.Key> exemptKeys;
+
     private final AtomicInteger scanCount = new AtomicInteger();
     private final AtomicBoolean fullyScanned = new AtomicBoolean();
     private final AtomicBoolean closed = new AtomicBoolean();
@@ -957,6 +993,11 @@ public class CrudHandler {
         // into the scan set
         results = null;
       }
+
+      // This gate is narrower than the one for `results` above, which also fires for validation and
+      // snapshot reads. Read-only and one-operation transactions never verify overlap, so they
+      // track nothing here.
+      exemptKeys = isOverlapVerificationRequired(context) ? new HashSet<>() : null;
     }
 
     @Override
@@ -981,6 +1022,13 @@ public class CrudHandler {
               processScanResult(key, scan, result, context, metadata);
           if (!processedScanResult.isPresent()) {
             continue;
+          }
+
+          // Recorded only for rows that are actually delivered. Rows dropped by the conjunction
+          // filter or by lazy recovery above never reach the caller, so exempting them would let a
+          // later write to such a key escape the overlap check.
+          if (exemptKeys != null && isNotYetWrittenOrDeleted(key, context)) {
+            exemptKeys.add(key);
           }
 
           if (results != null) {
@@ -1057,7 +1105,7 @@ public class CrudHandler {
         putIntoScannerSetInSnapshot(scan, results, context);
       }
 
-      verifyNoOverlap(scan, results, context);
+      verifyNoOverlap(scan, results, exemptKeys, context);
     }
 
     @Override
@@ -1085,6 +1133,10 @@ public class CrudHandler {
     private final Iterator<Map.Entry<Snapshot.Key, TransactionResult>> resultsIterator;
 
     private final LinkedHashMap<Snapshot.Key, TransactionResult> results = new LinkedHashMap<>();
+
+    // See ConsensusCommitStorageScanner#exemptKeys
+    @Nullable private final Set<Snapshot.Key> exemptKeys;
+
     private boolean closed;
 
     public ConsensusCommitSnapshotScanner(
@@ -1096,6 +1148,7 @@ public class CrudHandler {
       this.context = context;
       this.metadata = metadata;
       resultsIterator = resultsInSnapshot.entrySet().iterator();
+      exemptKeys = isOverlapVerificationRequired(context) ? new HashSet<>() : null;
     }
 
     @Override
@@ -1105,6 +1158,11 @@ public class CrudHandler {
       }
 
       Map.Entry<Snapshot.Key, TransactionResult> entry = resultsIterator.next();
+
+      if (exemptKeys != null && isNotYetWrittenOrDeleted(entry.getKey(), context)) {
+        exemptKeys.add(entry.getKey());
+      }
+
       results.put(entry.getKey(), entry.getValue());
 
       return Optional.of(
@@ -1130,7 +1188,7 @@ public class CrudHandler {
     @Override
     public void close() {
       closed = true;
-      verifyNoOverlap(scan, results, context);
+      verifyNoOverlap(scan, results, exemptKeys, context);
     }
 
     @Override
--- a/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
+++ b/core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
@@ -74,9 +74,17 @@ public class Snapshot {
   private final List<ScannerInfo> scannerSet;
 
   // The write set stores information about writes in this transaction.
+  //
+  // Invariant: a key must never leave writeSet ∪ deleteSet, not even transiently on an exception
+  // path. Scanners test membership of this union at the moment they produce a record to decide
+  // whether a write happened before or after the read, so a key that briefly belongs to neither set
+  // would be misread as never written. putIntoWriteSet and putIntoDeleteSet are currently the only
+  // mutators of these two maps and both preserve it; any new mutator must preserve it too. See
+  // putIntoWriteSet for the details.
   private final Map<Key, Put> writeSet;
 
-  // The delete set stores information about deletes in this transaction.
+  // The delete set stores information about deletes in this transaction. See the invariant
+  // documented on writeSet above.
   private final Map<Key, Delete> deleteSet;
 
   public Snapshot(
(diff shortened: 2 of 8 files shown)


.claude/CLAUDE.md
```diff
@@ -0,0 +1,153 @@
+# CLAUDE.md
```
Torch3333: Several docs related to AI-assisted development were committed. Is that allright?

brfrn169: No, my AI mistakenly included them. I’ve fixed it. Thanks!

Torch3333 (approved):
LGTM, thank you!

yu2scalar (approved):
LGTM, Thank you!!

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -1057,7 +1105,7 @@ && checkAndRecoverBeforeIndexRecords(scan, context, txMetadata)) {
         putIntoScannerSetInSnapshot(scan, results, context);
       }
 
-      verifyNoOverlap(scan, results, context);
+      verifyNoOverlap(scan, results, exemptKeys, context);
```
feeblefakie, line 1108: `rollback()` calls `closeScanners()`, so this can throw from inside `rollback()`. It's an `IllegalArgumentException` and `ConsensusCommit.rollback()` only catches `CrudException`, so it escapes and can replace the exception the caller was already handling.

Does the check need to run on the rollback path at all? Nothing is committed there, so the result is never used. WDYT about skipping validation when closing for rollback, or catching `RuntimeException` there?

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
```diff
@@ -11309,6 +11309,277 @@ private void populateRecord(ConsensusCommitManager manager, String namespace, St
...
+    // stale row. The exemption is per-scanner while the rejection is transaction-wide.
+    assertThatThrownBy(scanner2::close).isInstanceOf(IllegalArgumentException.class);
+
+    transaction.rollback();
+  }
+
+  @ParameterizedTest
+  @EnumSource(Isolation.class)
+  public void
+      getScanner_WhenInsertingRecordAlreadyReturnedByScanner_ShouldThrowCommitConflictException(
```
feeblefakie, line 11517: This is retriable but won't converge if the record actually exists, right? The reasoning for fixing it separately makes sense to me.

Could we link a tracking issue for it? With `Related issues: N/A` it seems easy to lose.

core/src/test/java/com/scalar/db/transaction/consensuscommit/CrudHandlerTest.java
```diff
@@ -917,7 +918,7 @@ void scanOrGetScanner_ResultGivenFromStorage_ShouldUpdateSnapshotAndReturn(ScanT
     verify(scanner).close();
     verify(snapshot).putIntoReadSet(key, Optional.of(expected));
     verify(snapshot).putIntoScanSet(scan, Maps.newLinkedHashMap(ImmutableMap.of(key, expected)));
-    verify(snapshot).verifyNoOverlap(scan, ImmutableMap.of(key, expected));
+    verify(snapshot).verifyNoOverlap(eq(scan), eq(ImmutableMap.of(key, expected)), any());
```
feeblefakie: With `any()` here, is anything still pinning that the batch `scan()` path passes an empty exempt set? That's the "`scan()` is unchanged" guarantee, and I think a regression there would keep these tests green.

Would something like `eq(scanType == ScanType.SCAN ? ImmutableSet.of() : ImmutableSet.of(key))` work?

integration-test/src/main/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
```diff
@@ -1388,6 +1388,74 @@ public void scan_DefaultNamespaceGiven_ShouldWorkProperly() throws TransactionEx
     }
   }
 
+  @Test
+  public void getScanner_WhenUpdatingEachReturnedRecord_ShouldUpdateAllOfThem()
```
feeblefakie, line 1392: Should these also go in `TwoPhaseCommitTransactionIntegrationTestBase`? It mirrors the other `getScanner_*` tests and runs the same `CrudHandler`, so the new behavior applies there too but isn't covered.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
```diff
@@ -372,46 +408,82 @@ public void to(MutationComposer composer) throws ExecutionException {
...
+      return getKeyOverlappedWithWriteSet((ScanAll) scan, results, exemptKeys);
     } else {
-      return isWriteSetOverlappedWith(scan, results);
+      return getKeyOverlappedWithWriteSet(scan, results, exemptKeys);
     }
   }
 
-  private boolean isDeleteSetOverlappedWith(Map<Snapshot.Key, TransactionResult> results) {
+  @Nullable
+  private Key getKeyOverlappedWithDeleteSet(
```
feeblefakie, line 462: This one only checks `results.containsKey`, without the range/conjunction checks the write-set versions do. So for `delete()`, a key another still-open scanner covers but hasn't returned isn't caught, which doesn't quite match the "Rejected at that scanner's `close()`" row in the description's table.

Is the table just worded a bit too strongly here, or should this check the range too?

feeblefakie (approved):
LGTM! Thank you! Left some suggestions/comments, so PTAL!

integration-test/src/main/java/com/scalar/db/api/DistributedTransactionIntegrationTestBase.java
brfrn169: Thanks for pointing this out. `TwoPhaseCommitTransaction` is deprecated and scheduled for removal in 3.20.0, so I won't add these tests to `TwoPhaseCommitTransactionIntegrationTestBase` this time.

integration-test/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitSpecificIntegrationTestBase.java
brfrn169: Right, a retry won't converge when the record really exists. But this isn't something this PR introduces; it's how `insert()` already behaves. An insert on a key that already exists is reported as a retriable conflict at commit, whether or not the transaction read the record first, and `CrudOperable.insert` documents it that way. `putAndCommit_PutWithInsertModeGivenForExistingAfterRead_ShouldThrowCommitConflictException` has long pinned the get-then-insert case, and this PR only routes the scanner variant onto that same path.

The underlying question is how `insert()` should report a key that is already taken by a committed record, which calls for a different key or a duplicate error rather than a retry, while still treating the genuinely transient cases at prepare as retriable. That's an existing `insert()` issue rather than part of this change, so I'll handle it separately.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Good catch. This isn't introduced here, though: `close()` already threw the same `IllegalArgumentException` from `verifyNoOverlap` before this change, and `rollback()` has the same shape on the release branches. Exempt keys can only skip overlaps, so this PR narrows when it fires rather than widening it.

I agree the check shouldn't run on the rollback path — nothing is committed, so the result is never used, and today it also escapes before `groupCommitter.remove()` and before the remaining scanners are closed. Since it's pre-existing, I'll fix it in a separate PR so it can be backported, by closing scanners without verification on the cleanup path rather than catching `RuntimeException`.

core/src/main/java/com/scalar/db/transaction/consensuscommit/Snapshot.java
brfrn169: Good catch — the table is worded too strongly. The code is intended, and the delete-set check is unchanged from master apart from the exemption.

A delete can't bring a row into a scan, so the only way it can make a scanner's output wrong is by the scanner returning the deleted record. Writes aren't applied to storage until commit, so if the other scanner goes on to reach that record, it returns it while the key is already in the delete set; the key isn't exempt, and that scanner's `close()` rejects it. If the scanner is closed before reaching it, the caller never saw a stale row, so there's nothing to reject. The range/conjunction checks on the write set exist for a different reason: a put can move a record into the scan (case 3 in the `ScanAll` comment), which the results can't show.

Puts are similar in one respect: for a scan with conditions or an index scan, whether a put the scan hasn't returned is rejected is judged from the columns the put sets. I've reworded the rule and the table in terms of what the scan returned and what a put sets.

core/src/test/java/com/scalar/db/transaction/consensuscommit/CrudHandlerTest.java
brfrn169: Good point. `scanOrGetScanner_CalledAfterDeleteUnderRealSnapshot_ShouldThrowIllegalArgumentException` does catch it behaviourally, since it runs a real `Snapshot`, but none of the mock-based verifications pinned the argument. I've replaced `any()` with the expected exempt set in all seven `ScanType`-parameterized verifications: empty for `SCAN`, and the delivered keys for the scanner variants. With `scan()` changed to pass its result keys, six of them now fail alongside that test.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: Opened #3867 for this. The cleanup path now discards open scanners without running the before-index check or the overlap verification, instead of calling `close()`.

PR #3851, opened 2026-09-08
Fix get operation in Cloud Storage adapter to handle concurrent writes

KodaiD:
## Description

This PR fixes an issue reported by @jnmt where the get operation does not handle concurrent writes. 

The payload was downloaded with the _generation_ pinned in the blob ID, so a read that raced with a concurrent deletion or update failed with a 404 error instead of reporting the object as absent, because a non-current generation does not exist in a bucket without object versioning. The S3 and Blob Storage adapters don't have this window because they read the payload and its version in a single request.

## Related issues and/or PRs

N/A

## Changes made

- Update Cloud Storage adapter's get operation:
  - Download the payload with the precondition instead of pinning the generation in the blob ID, which distinguishes a concurrent deletion (`404`) from a concurrent update (`412`).
  - Return an empty optional when the object is deleted during the read.
  - Retry from the metadata retrieval when the object is updated during the read, up to `GET_MAX_RETRY_COUNT` times.
  - Throw `ConflictOccurredException` once the retries are exhausted.
- Update unit test for the get operation.

## Checklist

N/A

## Release notes

Fixed the get operation in the Cloud Storage adapter to handle concurrent writes.

### DIFF
--- a/core/src/main/java/com/scalar/db/storage/objectstorage/cloudstorage/CloudStorageWrapper.java
+++ b/core/src/main/java/com/scalar/db/storage/objectstorage/cloudstorage/CloudStorageWrapper.java
@@ -9,6 +9,7 @@ import com.google.cloud.storage.StorageBatch;
 import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
 import com.google.common.annotations.VisibleForTesting;
+import com.scalar.db.storage.objectstorage.ConflictOccurredException;
 import com.scalar.db.storage.objectstorage.ObjectStorageWrapper;
 import com.scalar.db.storage.objectstorage.ObjectStorageWrapperException;
 import com.scalar.db.storage.objectstorage.ObjectStorageWrapperResponse;
@@ -28,6 +29,10 @@ import javax.annotation.concurrent.ThreadSafe;
 public class CloudStorageWrapper implements ObjectStorageWrapper {
   // Batch API has a limit of 100 operations per request
   public static final int BATCH_DELETE_SIZE_LIMIT = 100;
+
+  // The maximum number of retries for the get operation, which is performed when the object is
+  // updated between the metadata retrieval and the payload download
+  @VisibleForTesting static final int GET_MAX_RETRY_COUNT = 3;
 
   private final Storage storage;
   private final String bucket;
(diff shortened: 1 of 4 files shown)


KodaiD:
Cloud Storage integration test result:
- https://github.com/scalar-labs/scalardb/actions/runs/34292433962
- https://github.com/scalar-labs/scalardb/actions/runs/34441615402

core/src/main/java/com/scalar/db/storage/objectstorage/cloudstorage/CloudStorageWrapper.java
```diff
@@ -55,18 +60,48 @@ public CloudStorageWrapper(CloudStorageConfig config, Storage storage) {
...
+            String.format("Failed to get the object with key '%s'", key), e);
       }
-      String payload = new String(blob.getContent(), StandardCharsets.UTF_8);
-      String generation = String.valueOf(blob.getGeneration());
-      return Optional.of(new ObjectStorageWrapperResponse(payload, generation));
-    } catch (Exception e) {
-      throw new ObjectStorageWrapperException(
-          String.format("Failed to get the object with key '%s'", key), e);
     }
+    throw new ConflictOccurredException(
```
Torch3333, line 100: The new `ConflictOccurredException` thrown is caught upstream but never translated into a retriable error.

`MutateStatementHandler:83` is the only place in the package that catches it, and it covers
the write path only. All three callers of `wrapper.get()` catch
`ObjectStorageWrapperException`:

- `MutateStatementHandler.getPartition` (L111) -> `ExecutionException` / `OBJECT_STORAGE_ERROR_OCCURRED_IN_MUTATION`
- `SelectStatementHandler.getPartition` (L135) -> `ExecutionException` / `OBJECT_STORAGE_ERROR_OCCURRED_IN_SELECTION`
- `StreamingRecordIterator.getPartition` (L70) -> same

Should it be mapped to `RetriableExecutionException` in those methods, the way `writePartition`
already does at MutateStatementHandler:83?

KodaiD: Good catch. Thank you! Fixed in https://github.com/scalar-labs/scalardb/pull/3851/commits/024e29bcc297ee9421fadaeb885d6eaeeda95c8f.

I left `SelectStatementHandler.getPartition` and `StreamingRecordIterator.getPartition` as they are,
for the following reasons:

- No other adapter throws `RetriableExecutionException` from a read path.
- Throwing it from a read path would also require changes in the Consensus Commit layer.
- I expect exhausting all the retries to be unlikely (Cloud Storage allows an item to be updated only once per second.).

Torch3333 (approved):
LGTM, thank you!

jnmt (approved):
LGTM, thank you!

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #3853, opened 2026-09-13
Retry conflict errors on recovery mutations

brfrn169:
## Description

A transient conflict raised while recovering a record fails the read or the commit that triggered the lazy recovery, instead of being retried.

`RecoveryExecutor` submits one recovery task per record to a cached thread pool (`RecoveryExecutor.java:68-71`, with `submit` at `:135`, `:203`, and `:306`), so the recovery of several records runs concurrently. Two recovery tasks can therefore write the same partition at the same time, and the storage can report a conflict. `RecoveryHandler` did not retry such a conflict: `rollbackRecord` and `rollforwardRecord` absorb only `NoMutationException`, so a conflict propagated out of the recovery future, and `CrudHandler#waitForRecoveryCompletion` reported it to the caller.

The Consensus Commit integration tests hit this on every branch. `ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation#get_GetWithIndexForTransientDeletedAndPreparedRecordsWhenWriterAborted_ShouldResolveToNoRecord` recovers two records of the same partition and fails with `ORA-08177: can't serialize access for this transaction` on Oracle and with `could not serialize access due to read/write dependencies among transactions` (SQLSTATE 40001) on PostgreSQL and AlloyDB. Some of those failures arrive through the production commit path (`ConsensusCommit#commit`), not only through the test helper, so this is not a test-only problem: a user transaction can fail to commit because a background recovery task lost a race that a retry would have won.

Separately, when such a conflict does reach the caller it is not reported as a conflict. `CrudHandler#waitForRecoveryCompletion` wraps it in a plain `CrudException`, even though `waitForRecoveryCompletionIfNecessary` already documents `@throws CrudConflictException if any recovery task fails due to a conflict` and `ConsensusCommit#commit` already maps `CrudConflictException` to `CommitConflictException`. A conflict raised by the storage never reached that branch, so the caller was not told that retrying the transaction resolves it.

## Related issues and/or PRs

Related to #3852. That PR makes `JdbcDatabase#put(Put)` and `#delete(Delete)` classify conflicts as `RetriableExecutionException`. Until it is merged, the retry added here does not fire for JDBC, because a single-mutation conflict on JDBC still arrives as a plain `ExecutionException`. The retry does take effect on DynamoDB, Cosmos DB, and object storage, whose single-mutation paths already throw `RetriableExecutionException` (`dynamo/PutStatementHandler.java:57-58`, `cosmos/MutateStatementHandler.java:69-70`). The two PRs touch disjoint files and can be merged in either order.

## Changes made

- `RecoveryHandler#mutate` now retries while the storage reports a `RetriableExecutionException`, bounded by `MAX_CONFLICT_RETRY_COUNT` (10) and with no backoff. Some conflicts are reported only after the competing work has ended, for which waiting does not improve the odds (the reasoning recorded for `JdbcAdmin#executeWithConflictRetry`). Others are reported while the competing operation is still in progress, such as a DynamoDB `TransactionConflictException` raised against an in-flight `TransactWriteItems`; the retries span several round trips, which is expected to outlast such an operation, and if they do not, the conflict reaches the caller as a conflict. This is the single funnel for both `rollbackRecord` and `rollforwardRecord`, so it covers lazy recovery and the `recoverRecord` API (`RecoveryExecutor#executeSynchronously`) alike. Non-conflict `ExecutionException`s are not retried.
- `CrudHandler#waitForRecoveryCompletion` now maps a `RetriableExecutionException` cause to `CrudConflictException`, so a conflict that outlives the retries reaches the caller as a conflict: `ConsensusCommit#commit` turns it into `CommitConflictException`, and the two-phase commit prepare paths turn it into `PreparationConflictException`. It carries a new `CONCURRENCY_ERROR` error code, `CONSENSUS_COMMIT_CONFLICT_OCCURRED_WHEN_RECOVERING_RECORDS`, because the existing `CONSENSUS_COMMIT_RECOVERING_RECORDS_FAILED` is in the `INTERNAL_ERROR` category and would label a retriable conflict as an internal error.
- Added unit tests for retry-then-succeed, giving up after the bound, not retrying a non-conflict error, absorbing a `NoMutationException` raised on a retry, and both exception mappings in `waitForRecoveryCompletion`, including the error code of the conflict.

Retrying is safe because recovery mutations are conditional on the record still being the one the recovery observed: `RollbackMutationComposer` builds `PutIf(tx_id AND state)` / `DeleteIf(tx_id AND state)` (`:98-101`, `:134-137`) and `CommitMutationComposer` does the same (`:86-89`, `:118-121`). Re-applying them cannot overwrite a newer state; if a concurrent actor resolved the record while we retried, the retry raises `NoMutationException`, which the callers already absorb as the record having been rolled back or committed by another transaction. `RetriableExecutionException` also means the mutation definitely did not apply, so a failed attempt leaves no partial effect behind.

## Checklist

The exact mechanism behind these conflicts has not been measured. Two concurrent recovery tasks writing the same partition is the leading explanation, and the PostgreSQL message (`Canceled on identification as a pivot`) does require two or more concurrent transactions with a read/write dependency, but a conflict between a recovery task and the reading transaction itself is also possible, and on Oracle `ORA-08177` can arise without concurrency at all. The retry added here does not depend on which of these it is.

That uncertainty does matter for two alternatives that were considered and not taken:

- Batching the recovery mutations of one atomicity unit into a single `storage.mutate` call. This would remove a self-conflict by construction, but it only helps if a self-conflict is the mechanism, it leaves conflicts with other transactions in place, and it needs the read path reworked, because `CrudHandler#executeRecovery` is called per record and returns a future per record.
- Serializing recovery per partition. Same dependence on the mechanism, plus a cost in recovery throughput.

Either could still be worth doing as an improvement once the mechanism is measured, but the retry is needed regardless, since neither removes conflicts with other transactions.

`ParticipantCommitHandler#commitRecords` and `#rollbackRecords` do not retry conflicts either, but their failures are deliberately left to lazy recovery, so they are out of scope here.

## Release notes

Fixed the Consensus Commit record recovery to retry conflict errors reported by the underlying storage, and to report a conflict that persists as a conflict so that the transaction can be retried.


Torch3333 (approved):
LGTM, thank you!

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
```diff
@@ -364,11 +370,58 @@ private boolean abortIfExpired(
...
       return;
     }
-    storage.mutate(mutations);
+
+    int attempt = 0;
+    while (true) {
+      try {
+        storage.mutate(mutations);
+        return;
+      } catch (RetriableExecutionException e) {
```
feeblefakie, line 408: Does this catch ever fire on JDBC?

`composer.get()` is always size 0 or 1 here, and `JdbcDatabase.mutate()` short-circuits a size-1 list to `put(Put)`/`delete(Delete)`. Those wrap every `SQLException` into a plain `ExecutionException` without consulting `rdbEngine.isConflict(e)`. The only JDBC site that maps a conflict to `RetriableExecutionException` is `mutateInternal`, the multi-mutation path.

If that is right, the ORA-08177 and 40001 failures in the commit message are not covered, and the new `CrudConflictException` mapping does not fire on JDBC either, so effective coverage would be DynamoDB, Cosmos and ObjectStorage only. `JdbcDatabaseTest.mutate_withConflictError_shouldThrowRetriableExecutionException` passes a two element list, so the single mutation path is untested for this.

Should `put(Put)`/`delete(Delete)` get the same `isConflict` mapping, or should the commit message be narrowed?

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
```diff
@@ -364,11 +370,58 @@ private boolean abortIfExpired(
...
+   * that triggered the recovery, even though it is transient.
+   *
+   * <p>Retrying is safe. Recovery mutations are conditional on the record still being the one this
+   * recovery observed (its {@code tx_id} and {@code state}), so re-applying them cannot overwrite a
+   * newer state: if a concurrent actor resolved the record while we retried, the retry raises
+   * {@link NoMutationException}, which the callers already absorb as the record having been rolled
+   * back or committed by another transaction. A {@link RetriableExecutionException} also means the
+   * mutation definitely did not apply, so the failed attempt leaves no partial effect behind.
+   *
+   * <p>No backoff is applied. Some conflicts are reported only after the competing work has ended,
```
feeblefakie, line 388: Is the no backoff rationale safe to apply to every `RetriableExecutionException`?

It holds for conflicts reported after the competing work ended (40001, ORA-08177, MySQL 1213). It seems not to hold for DynamoDB `TransactionConflictException`, Cosmos 449 RETRY_WITH and ObjectStorage precondition failures, where the competitor is still in flight, and those are the storages where this loop actually fires today.

At a few ms RTT the whole loop spans roughly 20ms. Is that enough to outlast a `TransactWriteItems` or a stored procedure that takes 5ms to 50ms?

Two related questions:

1. Should there be an elapsed time bound next to the count? `RdbEngineMysql#isConflict` also matches 1205 lock wait timeout, so if the JDBC gap above is closed, each attempt could block for `innodb_lock_wait_timeout` (50s by default) while a caller waits on an untimed `recoveryFuture.get()`.
2. Throttling is not mapped to `RetriableExecutionException` (DynamoDB `ProvisionedThroughputExceededException`, Cosmos 429). Could 11 immediate retries on a contended item push the table into throttling and turn a retriable conflict into a non-retriable error?

`CoordinatorStateAccessor#exponentialBackoff` already does backoff in this package, if it seems worth adding here.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
```diff
@@ -364,11 +370,58 @@ private boolean abortIfExpired(
...
+        if (attempt >= MAX_CONFLICT_RETRY_COUNT) {
+          logger.warn(
+              "Giving up on conflicting recovery mutations after {} retries. Transaction ID that wrote the record: {}",
+              attempt,
+              transactionId,
+              e);
+          throw e;
+        }
+        attempt++;
+        logger.warn(
```
feeblefakie, line 418: Should the per-attempt log be `debug`, and without the throwable?

Passing `e` as the last vararg logs a full stack trace on every attempt, and `RecoveryExecutor` submits one recovery task per record, so a contended partition can produce 10 stack traces per record.

The other six logs in this file are `info`, including the `NoMutationException` absorption just above. `JdbcAdmin#executeWithConflictRetry` logs `e.getErrorCode()` rather than the throwable, and its per-attempt warning is justified there by catalog contention indicating a bug, which does not seem to apply to this path.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
```diff
@@ -712,6 +713,16 @@ private void waitForRecoveryCompletion(
       if (cause instanceof CrudException) {
         throw (CrudException) cause;
       }
+      if (cause instanceof RetriableExecutionException) {
```
feeblefakie, line 746: Should the synchronous recovery paths get this mapping too?

`RecoveryExecutor#executeSynchronously` propagates the same exception to `ConsensusCommitManager#recoverRecord` and the finish transaction path. Both catch the broad `ExecutionException` (`RetriableExecutionException` extends it) and wrap it into `TransactionException` with an INTERNAL_ERROR code, so the reasoning in this comment seems to apply there too.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
brfrn169: You're right for the code as it stood. `JdbcDatabase#put(Put)` and `#delete(Delete)` now classify conflicts as `RetriableExecutionException` as of #3852, which was merged earlier today, so this catch does fire on JDBC for single-mutation recoveries. #3852 also covers the single-mutation path in `JdbcDatabaseTest`: `put_WithConflictError_ShouldThrowRetriableExecutionException`, `delete_WithConflictError_ShouldThrowRetriableExecutionException`, and `mutate_WithSingleMutationAndConflictError_ShouldThrowRetriableExecutionException`.

The PR description already noted the dependency ("Until it is merged, the retry added here does not fire for JDBC"), but the commit message does not, which is why it reads as overclaiming on its own. I'll rebase this branch on master before merging, so that the ORA-08177 and 40001 coverage the commit message describes holds as written.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
brfrn169: The two kinds are already separated in the code comment, and the retry is deliberate for both. For a conflict reported while the competitor is still in flight, the window is not a fixed 20ms: each attempt costs one round trip of a single-item write, so the window scales with the RTT of the environment rather than staying constant. Eleven attempts are expected to outlast an in-flight `TransactWriteItems`, and a single-item write losing to another single-item write does not surface as `TransactionConflictException` at all (the loser gets a conditional check failure, which we absorb as `NoMutationException`). If a conflict does outlive the retries, it reaches the caller as a conflict, so the transaction is retried one level up rather than failing as an internal error.

I'd rather not add backoff without a measurement showing the retries actually run out, since it would also lengthen the window during which the record stays unresolved for every other reader.

On the elapsed-time bound: the MySQL 1205 case is real now that #3852 maps it, but it needs an external long-lived holder of the same row lock — ScalarDB's own single-mutation writes are one statement each, so they do not hold a row lock for `innodb_lock_wait_timeout`. In that case the caller trades a fast internal-error failure for a slow conflict, which is the more accurate outcome, and a backoff would not shorten it: each attempt is already blocked inside the driver. A bound on total elapsed time is the remedy, and I'd rather add it when we have a case where the retries actually run out, so the value is chosen against something measured.

On throttling: each attempt is a single-item write and the loop is bounded at eleven per record, so the added volume is small, though I agree the retries concentrate on one item. I have not seen a throttled recovery, and if we do, the fix belongs in the adapters — classifying `ProvisionedThroughputExceededException` and Cosmos 429 as retriable — rather than in this loop, which would otherwise give up on a condition that is by definition transient.

core/src/main/java/com/scalar/db/transaction/consensuscommit/RecoveryHandler.java
brfrn169: I'd like to keep this one as `warn` with the throwable. In practice it is one line per conflicted recovery, not ten: the conflict is transient, so the first retry usually succeeds, and ten lines only appear when the retries run out — which is also the case the give-up `warn` above reports. Since a retry that succeeds produces no other record of the conflict, this line is the only evidence that contention is happening and self-healing, and the exception text is what identified ORA-08177 and 40001 when we chased these failures. `CoordinatorStateAccessor` logs its retries at `warn` in the same package for the same reason.

If the volume turns out to be a problem in a deployment, dropping the throwable and keeping the message is a one-line change, since the adapters already embed the cause's message.

core/src/main/java/com/scalar/db/transaction/consensuscommit/CrudHandler.java
brfrn169: You're right that both paths wrap it into `TransactionException` with an internal-error code. I left them out deliberately: `recoverRecord` and the finish-transaction path are low-level, idempotent maintenance APIs whose caller can simply call again, whereas `waitForRecoveryCompletion` sits underneath a user transaction's read or commit, where the distinction decides whether the caller should retry the transaction. Reporting a conflict from those two would need a conflict-carrying subtype of `TransactionException`, which is a public API addition and larger than this fix. Happy to do it separately if you'd like the parity.

feeblefakie (approved):
LGTM! Thank you!

PR #3854, opened 2026-09-13
Wait for the recovery to complete in the two-phase commit recovery tests

brfrn169:
## Description

The two-phase commit recovery tests leave their asynchronous recovery running after they finish, which makes the next test fail in `setUp`.

`TwoPhaseConsensusCommitSpecificIntegrationTestBase` has six helpers that trigger lazy recovery, and only the two `CoordinatorStateNotExistAndExpired` ones wait for it. The four others do not: the `CoordinatorStateCommitted` rollforward and the `CoordinatorStateAborted` rollback helpers, for both a `PREPARED` and a `DELETED` record. Each is driven by three tests (`get`, `scan`, and `scanAll`), so twelve tests can return while a recovery mutation is still in flight.

The transactions in those helpers are read-only, so the production wait does not cover them. `TwoPhaseConsensusCommit#prepare` calls `crud.waitForRecoveryCompletionIfNecessary(context)`, but that only waits when the recovered key is in the write or delete set, or when validation may be required. Neither holds here.

The next test's `setUp` calls `truncateTables()`. On Oracle, `TRUNCATE` is a DDL statement that requests an exclusive lock with `NOWAIT`, so it fails immediately with `ORA-00054: resource busy and acquire with NOWAIT specified or timeout` while a recovery write is in flight. That surfaced as `DB-CORE-30004: Truncating the table failed` thrown from `setUp`, failing whichever test happened to run next — which is why the victim differed from run to run.

This gap is specific to this test base. The three other integration test bases that write uncommitted records — `ConsensusCommitSpecificIntegrationTestBase`, `ConsensusCommitNullMetadataIntegrationTestBase`, and `ConsensusCommitImportTableIntegrationTestBase` — already wait in their `CoordinatorStateCommitted` and `CoordinatorStateAborted` helpers.

## Related issues and/or PRs

Related to #3852 and #3853, which address a different cause of failure in the same integration test runs (a transient conflict raised while recovering a record). This PR is independent of both: different root cause, different files, and no ordering requirement.

## Changes made

- Added `((TwoPhaseConsensusCommit) transaction).waitForRecoveryCompletion()` at the end of the four helpers that trigger lazy recovery without waiting for it, matching what the `CoordinatorStateNotExistAndExpired` helpers already do.

The `CoordinatorStateNotExistAndNotExpired` helpers are deliberately left alone. When the coordinator state is absent and the writer has not expired, recovery is intentionally skipped, so no mutation is in flight and there is nothing to wait for.

## Checklist

Verified locally by running the suite against SQLite, which needs no server: `TwoPhaseConsensusCommitSpecificIntegrationTestWithJdbcDatabase` passes with 74 tests, 0 failures, and all twelve affected tests ran.

`ORA-00054` is Oracle-specific, and the CI for a pull request does not run the Oracle jobs unless a storage adapter changes, so the JDBC integration tests were dispatched separately ([run](https://github.com/scalar-labs/scalardb/actions/runs/34745459460)). On Oracle 19, 21, and 23, the same test class passes with 74 tests and 0 failures, all twelve affected tests pass, and neither `DB-CORE-30004` nor `ORA-00054` appears in any Oracle report. The failure is intermittent, so a single clean run supports the fix rather than proving it. The two failing jobs in that run (Oracle 21 default and Oracle 23 group commit) fail in `ConsensusCommitSpecificIntegrationTestWithJdbcDatabaseInHighestIsolation#get_GetWithIndexForTransientDeletedAndPreparedRecordsWhenWriterAborted_ShouldResolveToNoRecord` with `ORA-08177`, which is the failure addressed by #3852 and #3853.

Retrying the `TRUNCATE` on `ORA-00054` was considered and not taken. `RdbEngineOracle#isConflict` covers 8177, 60, and 8176 but not 54, so adding 54 would put `truncateTable` on the existing `JdbcAdmin#executeWithConflictRetry` path. That helper deliberately applies no backoff, on the grounds that the conflicts it exists for are reported without consuming a wait. `ORA-00054` is raised immediately by `NOWAIT`, so ten immediate retries would be spent in well under a millisecond, while the race is against an asynchronous recovery write that takes longer than that. Removing the race is sounder than running it faster.

One 2PC failure seen in CI is not addressed here: an assertion failure inside the `CoordinatorStateNotExistAndExpired` helper, which already waits. It did not reproduce locally or in the dispatched run above, so it is left for a later CI run to characterize.

## Release notes

N/A


Torch3333 (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #3867, opened 2026-09-15
Fix rollback throwing IllegalArgumentException when a scanner is left open

brfrn169:
## Description

Rolling back a Consensus Commit transaction that still has an open scanner can fail with an `IllegalArgumentException` instead of completing the rollback.

`ConsensusCommit#rollback`, `TwoPhaseConsensusCommit#rollback`, and `ConsensusCommitParticipant#rollbackRecords` / `#releaseTransactionContext` close the transaction's open scanners through `TransactionContext#closeScanners`, which called the same `close()` that users call. That `close()` runs the before-index check, puts the results into the scan set or the scanner set, and finally calls `Snapshot#verifyNoOverlap`, which throws `IllegalArgumentException` (`DB-CORE-10106`) when the transaction has written a key in the scan range. The callers catch only `CrudException`, so the exception escapes the cleanup path. As a result:

- It replaces the exception the caller was handling when it decided to roll back.
- The loop over the scanners stops, so the remaining scanners are never closed. A JDBC storage scanner keeps its connection until it is closed.
- `ConsensusCommit#rollback` never reaches `groupCommitter.remove()`, so the group commit slot reserved at begin is left behind.

The same path runs when active transaction management expires or evicts a transaction (`ActiveTransactionManagedDistributedTransactionManager`, `ActiveTransactionManagedTwoPhaseCommitTransactionManager`, and `ActiveTransactionManagedTwoPhaseCommitParticipant`). There, `ActiveTransactionRegistry` only logs the exception, so the only symptoms are the unclosed scanners and the leftover slot.

This happens only when a scanner is still open at rollback time and its `close()` would reject an overlapping write, for example after writing a key in the scan range that the scanner has not returned yet. Using the scanner in try-with-resources and rolling back outside the block avoids it, because `close()` runs before the rollback. The issue predates #3850 and has the same shape on the release branches.

None of the work that `close()` does is needed on these paths. The transaction is not committed afterwards, so the scan results are never validated and the overlap check has nothing to protect. A before-index recovery there has no caller to ask for a retry, and lazy recovery handles those records when they are read.

## Related issues and/or PRs

Addresses https://github.com/scalar-labs/scalardb/pull/3850#discussion_r4012243546. The issue is pre-existing, so it is fixed separately from #3850 so that it can be backported.

## Changes made

- Added `ConsensusCommitScanner#discard()`, which closes a scanner without the before-index check, without putting the results into the scan set or the scanner set, and without the overlap verification. The storage scanner only closes the underlying storage scanner, using the same `closeScanner()` that its `one()` already uses on errors. The snapshot scanner only marks itself closed. The Javadoc states that it must only be used when the transaction will not be committed or prepared afterwards.
- `TransactionContext#closeScanners` now calls `discard()` instead of `close()` and no longer declares `CrudException`. The four callers drop their now-unreachable `catch (CrudException)` blocks, and `TwoPhaseConsensusCommit` and `ConsensusCommitParticipant` drop the loggers that only those blocks used.
- `ConsensusCommitSnapshotScanner#close()` now does nothing when the scanner is already closed, as `ConsensusCommitStorageScanner#close()` already does. Without this, rolling back inside a try-with-resources block would discard the scanner, and the implicit `close()` at the end of the block would then re-run the overlap verification and throw the same exception for a snapshot scanner, while a storage scanner stays silent.
- Tests:
  - `CrudHandlerTest`: `discard()` on both scanners, including a scan that requires the before-index check, touches neither the snapshot nor the storage. Closing after discarding and discarding after closing do nothing, and a snapshot scanner's second `close()` does nothing.
  - `TransactionContextTest`, `ConsensusCommitTest`, `TwoPhaseConsensusCommitTest`, and `ConsensusCommitParticipantTest`: each cleanup path discards open scanners without calling `close()`, even when `close()` would throw, and `ConsensusCommit#rollback` still releases the group commit slot.
  - `ConsensusCommitSpecificIntegrationTestBase`: opens a partition scanner, reads one record, inserts a key in the scan range that the scanner has not returned, and rolls back without closing the scanner. Before this change, the rollback fails with `DB-CORE-10106` for every isolation level.

The new tests do not depend on the argument list of `Snapshot#verifyNoOverlap` or on the `prepareInsert` helper, and they are placed next to tests that exist on all release branches, so they can be backported without changes.

## Checklist

- The one-operation scanners returned by `ConsensusCommitManager#getScanner` and `TwoPhaseConsensusCommitManager#getScanner` still call `close()` in their error paths. That `close()` is always a no-op there, because the storage scanner's `one()` closes itself before throwing.
- Catching `RuntimeException` around `closeScanners()` was considered and not taken. It would keep the before-index storage reads and recovery running on the cleanup path, and it would also hide unrelated bugs.

## Release notes

Fixed an issue where rolling back a Consensus Commit transaction with an open scanner, including the automatic rollback of an expired or evicted transaction, could fail with an IllegalArgumentException and leave the remaining scanners open and the group commit slot unreleased.


feeblefakie (approved):
LGTM! Thank you!

PR #3868, opened 2026-09-16
Recover uncommitted records blocking writes that create a record

brfrn169:
## Description

A write that creates an initial record is prepared with a `PutIfNotExists` condition, which fails when any record is physically present on the key. That is the case for an insert, and also for a write to a key that the transaction never read. Such a write never triggers the lazy recovery that the read path performs, so a record left behind by a transaction that did not finish blocks every subsequent attempt for that key until some other operation reads the key. For a key that is only ever written this way, that never happens, so every attempt keeps failing, even though no committed record exists for the key.

This was confirmed with a JDBC (SQLite) test: after placing a PREPARED record with no before image, an expired `tx_prepared_at`, and no coordinator state record, five consecutive inserts without a read all failed with `CommitConflictException`. A single read of the key triggered recovery, after which the insert succeeded.

The error message for the failed conditional write is also misleading. The same message is used when a record being inserted already exists or is being written by another transaction, and when a record being updated or deleted was modified by another transaction, but it only described the first case, which makes an update conflict look like a duplicate key.

## Related issues and/or PRs

- Based on #3888, which this PR targets until that one is merged.

## Changes made

- Recover the records of other transactions that are not committed yet and that sit on the keys a transaction writes, when the transaction rolls its records back.
  - The rollback reads the latest state of every record the transaction writes to decide what to restore, and those records are exactly where such a record shows up, so this costs no extra read and it runs after the rollback rather than before it.
  - A record this transaction prepared itself is rolled back there, and a committed record is a duplicate key that no recovery can resolve; neither is recovered.
  - A one-phase commit is not rolled back by the caller, since a failed one applies nothing, so it rolls back to recover the records blocking it. That rollback writes nothing.
  - The recovery goes through `RecoveryExecutor`, so the guard that leaves a record of a transaction that may still be in flight alone still applies. It is best-effort and does not change how the failure is reported: the attempt still fails with a conflict, and the next one is what benefits from the recovery.
- Thread `RecoveryExecutor` through `CommitHandler`, `CommitHandlerWithGroupCommit`, and `ParticipantCommitHandler`.
- Log a failed recovery in `RecoveryExecutor`. It runs on a background thread whose future the callers on the lazy-recovery path do not wait for, so such a failure was lost entirely.
- Reword `CONSENSUS_COMMIT_PREPARING_RECORD_EXISTS` (`DB-CORE-20013`) to cover both causes of the failed conditional write. The error code is unchanged.
- Note in the Javadoc of `CrudOperable.insert` that a conflict error caused by an already existing entry is not resolved by retrying the transaction with the same key.
- Add unit tests for the new path and an integration test that fixes the "fails once, then succeeds on retry" behavior.

## Checklist

The constructors of `CommitHandler` and `CommitHandlerWithGroupCommit` now take a `RecoveryExecutor`, so code outside this repository that constructs them needs a matching change.

The error message text is documented in the error code reference, which needs a matching update.

## Release notes

Fixed an issue where a write operation that creates a record kept failing for a key that holds a record of a transaction that did not finish, by recovering such a record when the transaction rolls back.


core/src/main/java/com/scalar/db/transaction/consensuscommit/ParticipantCommitHandler.java
```diff
@@ -229,4 +284,74 @@ void onePhaseCommitRecords(TransactionContext context)
...
+   * <p>This is best-effort and does not change what the caller reports: the conditional write may
+   * have failed for a record this method does not recover, or for a committed record that no
+   * recovery can resolve. A failure while reading a record is logged and ignored. The recovery
+   * itself runs on a background thread that nothing here waits for, so a failure inside it is not
+   * surfaced here either; the caller throws a conflict exception in either case and the next
+   * attempt retries the whole thing.
+   */
+  private void tryRecoverRecordsBlockingWrites(
+      TransactionContext context, List<Mutation> notAppliedMutations) {
+    for (Mutation mutation : notAppliedMutations) {
```
feeblefakie: This probes every mutation in the failed group, not the row that conflicted. With
`MutationAtomicityUnit.STORAGE` (JDBC, DynamoDB) the group is the whole transaction, so one
duplicate key in an N-row insert costs N sequential LINEARIZABLE reads before we throw.

Measured with 50 insert-mode puts and one conflicting row:

- STORAGE (JDBC): 50 reads
- STORAGE, cap 100 (DynamoDB): 50 reads
- PARTITION (Cassandra, Cosmos): 1 read

In the plain duplicate-key case all of them find a COMMITTED record and no recovery starts, so
it is pure overhead on a path we now tell applications to retry.

The failing index is already known and dropped today: `JdbcCrudService.executeBatch` has `i`
when it returns false, and Dynamo's `BatchHandler` finds the `ConditionalCheckFailed`
cancellation reason. Propagating it through `NoMutationException` would make this a single
read, but that is an SPI change across adapters.

For this PR, can we run the reads through the `parallelExecutor` that is already injected here,
and/or cap how many we probe? Recovering one blocking record per attempt is enough for retries
to converge.

feeblefakie, line 312: This probes every mutation in the failed group, not the row that conflicted. With
`MutationAtomicityUnit.STORAGE` (JDBC, DynamoDB) the group is the whole transaction, so one
duplicate key in an N-row insert costs N sequential LINEARIZABLE reads before we throw.

Measured with 50 insert-mode puts and one conflicting row:

- STORAGE (JDBC): 50 reads
- STORAGE, cap 100 (DynamoDB): 50 reads
- PARTITION (Cassandra, Cosmos): 1 read

In the plain duplicate-key case all of them find a COMMITTED record and no recovery starts, so
it is pure overhead on a path we now tell applications to retry.

The failing index is already known and dropped today: `JdbcCrudService.executeBatch` has `i`
when it returns false, and Dynamo's `BatchHandler` finds the `ConditionalCheckFailed`
cancellation reason. Propagating it through `NoMutationException` would make this a single
read, but that is an SPI change across adapters.

For this PR, can we run the reads through the `parallelExecutor` that is already injected here,
and/or cap how many we probe? Recovering one blocking record per attempt is enough for retries
to converge.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParticipantCommitHandler.java
```diff
@@ -49,38 +59,51 @@ class ParticipantCommitHandler {
...
       PrepareMutationComposer composer =
           new PrepareMutationComposer(context.transactionId, preparedAt, tableMetadataManager);
       context.snapshot.to(composer);
       List<List<Mutation>> groupedMutations = mutationsGrouper.groupMutations(composer.get());
 
-      List<ParallelExecutorTask> tasks = toTasks(groupedMutations);
+      List<ParallelExecutorTask> tasks = toTasks(groupedMutations, notAppliedMutations);
       parallelExecutor.prepareRecords(tasks, context.transactionId);
     } catch (NoMutationException e) {
+      tryRecoverRecordsBlockingWrites(context, notAppliedMutations);
```
feeblefakie, line 93: This runs before the transaction aborts itself. `abortState` and `rollbackRecords` only run
after `prepareRecords` throws (`CommitHandler.java:177-181`), so these reads delay our own
ABORTED write, and any records we did prepare successfully keep blocking other transactions
until then.

It does not apply on JDBC, where one failed group rolls back the whole transaction and nothing
stays prepared. It does apply where a transaction spans several groups: on DynamoDB one failed
batch can leave the other batches' records prepared while we do up to 100 reads.

Since nothing waits for the recovery result anyway, could we trigger this after the rollback
instead of before it?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParticipantCommitHandler.java
```diff
@@ -229,4 +284,74 @@ void onePhaseCommitRecords(TransactionContext context)
...
+          // The record already exists, which is a duplicate key that recovery cannot resolve
+          continue;
+        }
+        if (context.transactionId.equals(txResult.getId())) {
+          // The record was prepared by this transaction, and the caller rolls it back
+          continue;
+        }
+
+        RecoveryExecutor.Result recoveryResult =
+            recoveryExecutor.execute(
```
feeblefakie, line 337: `RecoveryExecutor` uses `newCachedThreadPool`, so it is unbounded, and there is no dedup by
record key or writer transaction ID. M transactions conflicting on the same key submit M
recovery tasks for that key.

That is pre-existing, but the read path only reaches it when someone actually reads an
uncommitted record. This fires on every insert conflict, including the common case where the
writer is still live and the not-expired guard in `RecoveryExecutor` means nothing is recovered
at all. On JDBC these threads also draw from the same connection pool as foreground traffic
(default 200).

Should we bound the pool or coalesce per key before adding a second producer to it?

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParticipantCommitHandler.java
```diff
@@ -229,4 +284,74 @@ void onePhaseCommitRecords(TransactionContext context)
...
+                key,
+                get,
+                txResult,
+                context.transactionId,
+                RecoveryExecutor.RecoveryType.RETURN_COMMITTED_RESULT_AND_RECOVER);
+
+        // Register the recovery task so that callers that wait for recovery completion (currently
+        // only tests) can observe it. The transaction itself does not wait for it: it is about to
+        // fail with a conflict, and the record is recovered for the next attempt.
+        context.recoveryResults.add(recoveryResult);
```
feeblefakie, line 347: `TransactionContext.recoveryResults` is documented as existing so recoveries can be awaited
before committing, but `waitForRecoveryCompletionIfNecessary` runs at `ConsensusCommit.java:211`,
before `commit.commit` at `:219`. So nothing added here is ever awaited in production, which the
comment above acknowledges.

Two side effects: it is a `CopyOnWriteArrayList`, so K adds cost O(K^2) in array copies, and each
`Result` carries a before image that is built synchronously on this thread and then discarded.

Could we return these from the method and expose them with `@VisibleForTesting` instead, so the
field keeps its stated meaning?

core/src/main/java/com/scalar/db/common/CoreError.java
```diff
@@ -1210,7 +1210,7 @@ public enum CoreError implements ScalarDbError {
   CONSENSUS_COMMIT_PREPARING_RECORD_EXISTS(
       Category.CONCURRENCY_ERROR,
       "0013",
-      "The record being prepared already exists. Details: %s",
+      "A conflict occurred. Either a record being inserted already exists or is being written by another transaction, or a record being updated or deleted has been modified by another transaction. Details: %s",
```
feeblefakie, line 1213: The message now covers three different situations, but the constant is still called
`CONSENSUS_COMMIT_PREPARING_RECORD_EXISTS`. Should we rename it so the two agree?

Also, does the public error code reference need the same update for `DB-CORE-20013`? There is
nothing to change in this repo, so it would be a docs repo follow-up.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
```diff
@@ -483,6 +483,24 @@ static Optional<TransactionResult> rereadRecord(
...
 
+  /**
+   * Creates a {@link Get} that reads the record identified by the specified snapshot key. All
+   * columns are read (no projections) with linearizable consistency so the before image and the
+   * transaction metadata are available.
+   *
+   * @param key the snapshot key identifying the record
+   * @return a {@code Get} for the record
+   */
+  static Get createGet(Snapshot.Key key) {
```
feeblefakie, line 494: `buildReReadGet` is about ten lines below this and builds the same `Get` with the same
consistency and nearly the same comment, and `ConsensusCommitManager.buildRecordGet` is a third
copy. Now that this one is package private, could those two delegate to it?

One thing to keep in mind if we also change the consistency or projections for the new caller:
that would want to stay separate, since `CrudHandler` does need LINEARIZABLE with no projections
here.

feeblefakie (commented):
Overall, looking good.
Left some comments/suggestions. PTAL!

brfrn169:
Thank you for the review. I am converting this to a draft while I restructure it, since the review changed where the recovery belongs.

On the read amplification: the rollback that the caller performs right after a failed preparation already reads the latest state of every record the transaction writes, through `RollbackMutationComposer.getLatestResult`. So the failing path already performs one read per record, and this PR was adding a second, separate round of reads for the same records.

Rather than capping or parallelizing that second round, the records blocking the writes can be taken from the rollback composition itself, which removes the extra reads entirely. Doing it there also runs the recovery after the rollback, so it no longer delays the ABORTED state and the rollback of the records this transaction prepared.

While looking at that, the reads in the rollback composition turn out to run on the calling thread, one after another, outside the parallel executor, so `parallel_rollback.enabled` only parallelizes the mutations. Fixing that first makes the change here simpler as well: with the latest results read up front, the blocking records fall out of them and the composer does not need to collect anything.

So I will split this into two:

1. A preceding PR that reads the latest records of a rollback up front, in parallel, and composes from those results.
2. This PR, rebased on it, which recovers the blocking records found there.

The preceding PR is #3888. I will rebase this on it and mark it ready for review again once that one is merged.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParticipantCommitHandler.java
brfrn169: The recovery no longer runs there. It is now triggered from the rollback of the records, after
the coordinator state is aborted and the records this transaction prepared are rolled back, so it
no longer delays either of them.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParticipantCommitHandler.java
brfrn169: The probe is gone. The rollback already reads the latest state of every record the transaction
writes, to tell the records it prepared from the ones another transaction holds, so the recovery
now takes its records from there: a record of another transaction that is not committed yet is
exactly what that read finds. Recovering them costs no extra read, whatever the atomicity unit of
the storage is, and #3888 also makes those reads run in parallel.

Thank you for the measurements; they are what moved this.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParticipantCommitHandler.java
brfrn169: You are right that nothing awaits these in production, and with the asynchronous rollback of #3888
they are now added on another thread after the transaction has failed, which makes the description
of the field even less accurate.

I corrected the description rather than moving the list (1b78ea7d5). The list holds the recovery
tasks the transaction started; the ones started while reading a record are awaited before the
records are committed, and the ones started while rolling back are not. That is a coherent thing
for the field to hold, and a caller that wanted to wait for the latter could do so.

On the two costs: the array copies and the before image are unchanged by where the list lives. The
before image comes from the recovery type that also recovers the record, so avoiding it would mean
a new type that recovers without returning anything. Both are proportional to the number of records
of other transactions that block the writes of one transaction, which is normally zero or one.

I did not take the `@VisibleForTesting` accessor because it moves the test-only affordance rather
than removing it, and the handler is shared by every transaction, so what it exposes is "the last
recovery", which is racy to describe even if a single-threaded test can use it.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ParticipantCommitHandler.java
brfrn169: This producer only fires on the failure path: a transaction that failed to commit, for the keys it
writes that another transaction holds uncommitted. That path is rare by design, and the tasks run
on a background thread that the transaction does not wait for, so they do not add latency to
anything the application sees.

I tried skipping a task whose record has not expired, since the recovery of such a record does
nothing when the writer has no coordinator state. I dropped it again: it delays the recovery of a
record whose writer wrote its coordinator state and then died by up to the transaction lifetime,
which works against what this PR is for, and it puts the expiration policy in a second place.

Bounding the pool or coalescing per key is worth doing on its own merits, since the read path is
the main producer there, but I would rather not change the shared executor as part of this PR.

core/src/main/java/com/scalar/db/common/CoreError.java
brfrn169: Renamed to `CONSENSUS_COMMIT_CONDITIONAL_MUTATION_NOT_APPLIED` (28bc672ef). The name now covers the
three situations, and it is also reported for a one-phase commit, which is not a preparation. The
error code is unchanged, so what an application matches on stays the same.

The error code reference does need the same update; that is a docs repository follow-up, noted in
the description.

core/src/main/java/com/scalar/db/transaction/consensuscommit/ConsensusCommitUtils.java
brfrn169: Done in #3888 (8f2fa73fd), since that is the PR that moved `createGet` next to `buildReReadGet`.
`createGet` now has an overload that takes the keys themselves, and the other two delegate to it;
`buildRecordGet` is gone. The consistency and the projections are the same in all three today, so
nothing is lost by sharing them; a caller that needs different ones would want its own builder, as
you say.

brfrn169:
I reworked the recovery since the last round, in 63e6d4109. The description is updated too.

**What the recovery covers now.** Only the records blocking the writes for the keys the transaction never read. When the recovery moved into the rollback, it covered every record of another transaction that is not committed yet on the keys the transaction writes, which is wider than it needs to be: the read path recovers such a record when the key is read, so a key the transaction reads is left to that read, or to the read a retry of the transaction performs. Only a key that nothing ever reads can stay blocked. This also limits the second producer of `RecoveryExecutor` you raised: an ordinary update conflict no longer submits anything.

The deletes are left out for the same reason. The record of a delete is always read before it is prepared, since the pre-read of the delete set is unconditional, unlike the one of the writes. I noted that in `CrudHandler` where it reads them, so the dependency is visible from that side as well.

**One-phase commit.** It no longer calls the rollback. It did so only to get the recovery, and that read the records of every write to compose rollback mutations that were always empty. It now reads the records of the keys that were never read, if any, and recovers the ones blocking them. A one-phase commit that read every key it writes reads nothing.

**`TransactionContext.recoveryResults`.** My earlier reply said that the recovery tasks started while reading a record are awaited before the records are committed. That holds only when the transaction writes or deletes the record, or when its reads may be validated, in SERIALIZABLE. The description in the code now says so.

I also corrected the description of `ParticipantCommitHandler`: it never writes the state of the transaction itself to the Coordinator table, but the recovery it triggers reads the state of other transactions there and may abort them.

PR #3871, opened 2026-09-17
Reject invalid mutation conditions in Consensus Commit

Torch3333:
## Description

`ConsensusCommitOperationChecker` calls `ConditionChecker.check()` for a conditional `Put` or `Delete` but ignores the boolean it returns, so a condition on a primary-key column, a comparison against a null value, an unknown column, or a type mismatch passes the checker. This PR makes Consensus Commit reject such conditions with an `IllegalArgumentException` whose message names the table and the conditional expressions.

The storage-side `OperationChecker` reports the same failure with `CoreError.OPERATION_CHECK_ERROR_CONDITION`, whose message prints the whole operation and the storage-level table metadata. Through the transaction API that would expose the transaction metadata columns (`tx_id`, `tx_state`, `before_*`), and for an `Update` it would describe the `Put` that Consensus Commit builds internally. A new error code is added instead of rewording that one because `OperationChecker` also throws it for `Scan` where-clauses, and its text is part of the storage API that users already match on.

## Related issues and/or PRs

Addresses https://github.com/scalar-labs/scalardb/pull/3844#discussion_r4006186144.

## Changes made

- `ConsensusCommitOperationChecker` throws `IllegalArgumentException` with the new `CoreError.CONSENSUS_COMMIT_CONDITION_NOT_PROPERLY_SPECIFIED` (`DB-CORE-10304`) when `ConditionChecker` rejects a `Put` or `Delete` condition. The `Put` and `Delete` paths share one `checkCondition` helper.
- `ConditionChecker` sets its validity flag explicitly for a `PutIf` or `DeleteIf` without expressions, so such a condition is rejected even on a reused checker instance.
- `ConsensusCommitOperationCheckerTest` checks the primary-key column, null comparison value, unknown column, and type mismatch cases against a real `ConditionChecker` with the exact message, plus the accepted cases. `ConditionCheckerTest` covers the empty-condition rule.
- `TwoPhaseCommitTransactionIntegrationTestBase.put_withPutIfWhenRecordDoesNotExist_shouldThrowUnsatisfiedConditionException` uses an `INT` null check on the `INT` balance column. The `TEXT` null check it used is a type mismatch that the checker rejects.

## Checklist

This is a behavior change for transaction API clients: a conditional mutation with a condition on a primary-key column, a null comparison value, an unknown column, a type mismatch, or no expressions was accepted by Consensus Commit and is rejected after this change. No documentation changes are included.

## Release notes

Fixed an issue where Consensus Commit accepted invalid conditions on conditional mutations instead of rejecting them.


brfrn169 (approved):
LGTM! Thank you!

